PR #14, opened 2021-11-11
Add pre-commit conf

### LINKED ISSUE #12: Integrate pre-commit hooks
based on tensorflow/tensorflow#48371

bhack:
Introduce [`pre-commit`](https://pre-commit.com/) and uses it in the Github Action.

resolves  #12


.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
+repos:
+  - repo: https://github.com/pycqa/isort
```
bhack, line 2: Are we asking to developers to run these two commands to setup their local env?:

```
pip install tensorflow
pip install -e ".[tests]" --progress-bar off --upgrade
```

In that case we could set `repo: local` and remove the extra bazel cache

bhack: One of the advantage to handling the lint tool versioning in the `pre-commit` config  Is that we could use the autoupdate PR if we want to use https://pre-commit.ci/.

There is also an alternative Github Action for managing the update:
https://github.com/marketplace/actions/pre-commit-autoupdate

For more context, also if it is a little bit outdated now, see:
https://browniebroke.com/blog/gh-action-pre-commit-autoupdate/

bhack:
If you are interested about handling a central conf for all the Keras repos please evaluate the author reply in:
https://stackoverflow.com/questions/59047996/how-can-i-configure-pre-commit-tool-to-rely-on-a-single-location-for-all-hook-co

bhack:
I've resolved the conflicts on master for your merged PR.

I hope that we could also discuss the open topics we have here.

.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
```
haifeng-jin, line 1: We prefer to keep all the configs in "setup.cfg" as we did in keras-tuner.

haifeng-jin (commented):
Thanks for the PR! we are preparing a doc for the best practices for setting up the repos in the Keras ecosystem. It will cover linting/formatting, github actions, testing and coverage, releasing and so on for all the repos. They should all follow the same standard. We also aims to have a centralized place to host the issue & PR templates & contributing guide, and distribute them to the different repos automatically at every update.

.flake8
bhack: Do you have commented the right line?

haifeng-jin: I think so. [This](https://github.com/keras-team/keras-tuner/blob/master/setup.cfg#L30-L43) is where we keep the flake8 configs.

bhack: Ok done. 
When https://github.com/PyCQA/flake8/issues/234 it will be ready probably we could unify `flake` config on the same file.

bhack: Do you want to copy the whole Keras `setup.cfg` here? As I've just mapped the current `args` that we already had  in this repo to config files.

haifeng-jin: I think it is good for now. We can move it later.

haifeng-jin:
The doc I mentioned is in the PR now.
https://github.com/keras-team/governance/pull/31

LukeWood:
@haifeng-jin do we have any update on what to do regarding our pre-commit workflows?

haifeng-jin:
@LukeWood I am not familiar with pre-commit. What we currently have for AutoKeras is a precommit to check the lint in a docker container. It requires the developer to setup mannually in the pre-commit file.

If the configuration in this PR can automatically setup the precommit for all developers, and check everything without much concerns on the local settings differences, it sounds good to me.

bhack:
You need `pip install pre-commit` and `pre-commit install` the first time you checkout the repo.

These two commands could be added also in the devel Dockerfile if we offer a devel ready image.

.pre-commit-config.yaml
bhack: Please check again this comment.

bhack:
P.s. honestly without speculating too much you could just try to run this PR on a sandboxed local system and give some feedback on your experience. 

I have already some self-done review comments in the PR for the tecno-political pending decisions.

LukeWood:
Hey @bhack - we discussed this in our keras ecosystem maintainers meetings and the consensus was generally against pre-commit hooks.  Initially I imagined symlinking a `hooks` dir to .git/hooks, but collectively we decided against that too.  Generally it seems the pre-commit package solves a problem we don't yet have.

Apologies for any inconvenience here due to miscommunication on this front.

bhack:
If we check all the failure results often we have a failing linting check: 
https://github.com/keras-team/keras-cv/actions?query=is%3Afailure

I still think that a `pre-commit` is useful also for the internal team.

bhack:
@haifeng-jin Following-up today meeting let me know if the team will re-evaluate this for the Keras-* repos.

haifeng-jin:
This is now mainly regarding the `keras-team/keras` repo.

The main concern for pre-commit hook was the inconsistency between internal and external dev infra. If the precommit hook checks something that internal cannot check, the internal developers will break the external developer's precommit check. If we can have a pre-commit hook just running `shell/lint.sh`, that should be OK.

However, I don't think we need or able to use the `pre-commit` package to do more checks.

LukeWood:
I don't think there is anything we can't get (that we currently need) from shell scripts by using the pre-commit package.

bhack:
Ok take a look a this so we don't have any conflict:

https://github.com/keras-team/keras-cv/pull/462

I will add an entry in the Markdown for who is not using the devcontainer/codespace

LukeWood:
Are .githooks automatically symlinked by the devcontainer?

PR #46, opened 2022-01-17
Fix test name for MixUp layer

chjort:
Fixed name of test for MixUp layer.


LukeWood:
cheers - thanks mate!

PR #50, opened 2022-01-21
Fix GIthub Actions Format Error

sampathweb:
Currently Github Actions is not running.  This PR aims to fix that by formatting the Yaml and adds pytest as requirement in setup.py

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -6,54 +6,64 @@ on:
   release:
     types: [created]
 jobs:
-  test:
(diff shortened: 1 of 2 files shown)


LukeWood:
Hey @sampathweb thank you for this contribution.  I did not realize that the actions had been broken - it seems a few lines were accidentally removed in a recent commit.  I fixed it @HEAD in a slightly different fashion than done here to match the actions of https://github.com/keras-team/keras_tuner.

Cheers and thanks again for pointing this out

PR #51, opened 2022-01-22
Update the bounding box utility to be more readable

qlzh727:
1. Added docstring header to contain more details about whats the formats we usually use for bounding box, and what's the math conversion between them.

2. Slight change the implementation of conversion function to be more readable. The current index slice based approach is very hard to understand.

3. Added unit test for bbox util.

### DIFF
--- a/keras_cv/utils/bbox.py
+++ b/keras_cv/utils/bbox.py
@@ -14,11 +14,11 @@
 
 """Shared utility functions for working with bounding boxes.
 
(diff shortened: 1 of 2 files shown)


LukeWood:
Thanks Scott!  Can look this over a bit later this eve

LukeWood (changes requested):
Looks good - looks like you need to locally run isort and it should be good to go.

qlzh727:
> Looks good - looks like you need to locally run isort and it should be good to go.

Done.

PR #56, opened 2022-01-26
Update the pad_bbox_batch_to_shape

qlzh727:
1. Fix the current implementation which has the typo in it (tartarget_shapeget)
2. Add docstring about the sample usage.
3. Add validation check for shape mismatch.
4. Add unit test.


keras_cv/utils/bbox_test.py
```diff
@@ -74,3 +74,24 @@ def test_xywh_to_corner(self):
...
+        # Make sure to raise error if the rank is different between bbox and target
+        # shape
+        with self.assertRaisesRegexp(
+                ValueError, 'Target shape should have same rank'):
+            bbox.pad_bbox_batch_to_shape(bboxes, [1, 2, 3])
+
+        # Make sure to not pad the dim if the target shape is smaller
+        target_shape = [3, 2]
+        result = bbox.pad_bbox_batch_to_shape(bboxes, target_shape)
+        self.assertAllClose(result, [[1, 2, 3, 4], [5, 6, 7, 8], [-1, -1, -1, -1]])
```
bhack: Is this the expected result `[[1, 2, 3, 4], [5, 6, 7, 8], [-1, -1, -1, -1]]`?

qlzh727: yes, the original shape is [2, 4] and since we couldn't shrink the second dim [4], the result shape is [3, 4].

qlzh727: what was your expected result in this case? I could raise an error if the target shape has smaller dim when comparing to input bboxs.

bhack: Yes probably it could be useful I think someone could go crazy when he is debugging a large code base. 
Do we need to support negative pad?

bhack: E.g. `tf.pad` has 
`InvalidArgumentError: must be non-negative` for negative inputs

qlzh727: Ack. Let me update the code/docstring/unittest

qlzh727: Done.

bhack (approved):

LukeWood:
Thanks Scott!

PR #61, opened 2022-01-27
Initial commit for the keras-cv roadmap/community guide line

qlzh727:
Mostly copy/pasted from internal roadmap doc (with some modifications).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,128 @@
 # KerasCV
-KerasCV is a repository of modular building blocks (layers, metrics, losses, data-augmentation) that applied computer vision engineers can leverage to quickly assemble production-grade, state-of-the-art training and inference pipelines for common use cases such as image classification, object detection, image segmentation, image data augmentation, etc. 
+KerasCV is a repository of modular building blocks (layers, metrics, losses, data-augmentation) that
+applied computer vision engineers can leverage to quickly assemble production-grade, state-of-the-art
+training and inference pipelines for common use cases such as image classification, object detection, 
+image segmentation, image data augmentation, etc. 
 
-KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras objects (layers, metrics, etc) that are too specialized to be added to core Keras, but that receive the same level of polish and backwards compatibility guarantees as the rest of the Keras API and that are maintained by the Keras team itself (unlike TFAddons).
+KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras 
+objects (layers, metrics, etc) that are too specialized to be added to core Keras, but that receive 
+the same level of polish and backwards compatibility guarantees as the rest of the Keras API and that 
+are maintained by the Keras team itself (unlike TFAddons).
 
-## Under Construction
-KerasCV is a relatively new project.  Expect regular updates and changes in the near future.
+# No Goals
+- **KerasCV is not a repository of blackbox end-to-end solutions (like TF Hub or ModelGarden).**
+
+    For the time being, it is focused on modular and reusable building blocks. We expect models in 
+    ModelGarden or TF Hub to eventually be built on top of the building blocks provided by KerasCV.
+
+    In the process of developing these building blocks, we will by necessity implement end-to-end 
+    workflows, but they're intended purely for demonstration and grounding purposes, they're not 
+    our main deliverable.
+
+
+- **KerasCV is not a repository of low-level image-processing ops, like tf.vision.** 
+
+    KerasCV is fundamentally an extension of the Keras API: it hosts Keras objects, like layers, 
+    metrics, or callbacks. Low-level C++ ops should go elsewhere.
+
+
+- **KerasCV is not a research library.**
+
+    Researchers may use it, but we do not consider researchers to be our target audience. Our target 
+    audience is applied CV engineers with experimentation and production needs. KerasCV should make 
+    it possible to quickly reimplement industry-strength versions of the latest generation of 
+    architectures produced by researchers, but we don't expect the research effort itself to be built
+    on top of KerasCV. This enables us to focus on usability and API standardization, and produce 
+    objects that have a longer lifespan than the average research project.
+
+## Areas of focus
+At the start of the project, we would like to take a task-oriented approach to outline our project, 
+and the first 2 tasks we would like to focus are:
+
+1. **Image classification** (Given an image and a list of labels, return the label and confidence for 
+   the image).
+2. **Object detection** (Given an image and a list of labels, return a list of bounding box, 
+   corresponding labels and confidence for the objects in the image).
+
+For each of the tasks, we are targeting to provide user model architectures, layers, preprocessing
+functions, losses, metrics and other necessary components to achieve state of the art accuracy.
+
+There will be common shareable components between those 2 tasks, like model building blocks, etc. We
+will have them in the repository as public API as well.
+
+### Common components
+- Model building block (served as backbone for the application)
+  - Resnet 
+  - Resnet RS
+  - ResNeXT
+  - Efficient
+- Preprocessing layers
+  - CutMix
+  - MixUp
+  - CutOut
+  - RandAugment
+  - RandomErasing
+
+### Image Classification
+- Dataset to evaluate
+  - [Imagenet](https://www.tensorflow.org/datasets/catalog/imagenet2012)
+  - [Imagenet v2](https://www.tensorflow.org/datasets/catalog/imagenet_v2) (for testing)
+- Target Metric 
+  - Top1 Accuracy > 80% (Note that some existing SotA models might not achieve this, like 
+    original ResNet, but this serve as a guideline for newly added models)
+
+### Object Detection
+- Dataset to evaluate
+  - [Coco 2017](https://www.tensorflow.org/datasets/catalog/coco#coco2017)
+  - [Open image v4](https://www.tensorflow.org/datasets/catalog/open_images_v4)
+  - [PASCAL voc](https://www.tensorflow.org/datasets/catalog/voc)
+- Target Metric
+  - COCO metric average precision (AP)
+- Detection model
+  - MaskRCNN
+  - YOLO (v3 or v4)
+  - RetinaNet
+- Other related utilities
+  - **ROI** (region of interest) layer (faster RCNN and Mask RCNN)
+  - Bounding box utils
+  - **RPN** (Region Proposal Network) used by faster RCNN and mask RCNN for bounding box 
+    generation.
+
+## Overlapping between keras.application and Keras-CV
+As you might notice, some area we would like to focus has overlapping with existing keras.application,
+and we would like to clarify the reason of the potential code duplication here.
+
+keras.application has a strong API contract that we can't easily make any backward incompatible
+change. On the other hand, there are a few of behavior we actually would like to change within the keras-cv
+  - Fetch imagenet weights by default, which might not want to do for keras-cv.
+  - Have the image preprocessing logic outside the model. We want to include all the preprocessing
+    within the model, so that user won't need to call the extra preprocessing methods.
+  - Only end-to-end model are exposed as public API, and Keras CV would like to include more building 
+    blocks as well.
+  - Some legacy model architectures are no longer commonly used and can't reach SotA performance, which
+    we might want to deprecate.
+
+Once the keras-cv applications are mature enough, we would add a deprecation warning in the 
+keras.applications code.
+
+
+## Community contribution guideline
+We would like to leverage/outsource the Keras community not only for bug reporting or fixes, 
+but also for active development for feature delivery. To achieve this, we will have the predefined
+process for how to contribute to this repository.
+
+The ideal workflow will be:
+
+- Keras defines the API interface for a certain task, eg a layer, model, dataset, metrics, etc. 
+- Post it to Github as a feature request. 
+- Have a hotlist/tag for those items and advocate them to the community.
+- Let community members claim the task they would like to work on.
+
+In the early stage, Keras team members will just claim those tasks by themselves. When there are 
+enough active community members, we should spend more time as reviewers rather than 
+authors. This should apply to bug fixes as well.
+
+User requested features will be carefully evaluated by keras team member before acceptance. 
+Based on the existing experience from tf-addons, even with a guideline of paper ref count > 50, 
+the image/cv related contributions are still quite scattered. We will make sure the contributions
+are coherent with the tasks/themes we have in the repository, or widely used for CV tasks. 
--- a/README.md
+++ b/README.md
@@ -4,12 +4,12 @@ applied computer vision engineers can leverage to quickly assemble production-gr
 training and inference pipelines for common use cases such as image classification, object detection, 
 image segmentation, image data augmentation, etc. 
 
-KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras 
-objects (layers, metrics, etc) that are too specialized to be added to core Keras, but that receive 
+KerasCV can be understood as a horizontal extension of the Keras API: the components are new first-party 
+Keras objects (layers, metrics, etc) that are too specialized to be added to core Keras, but that receive 
 the same level of polish and backwards compatibility guarantees as the rest of the Keras API and that 
 are maintained by the Keras team itself (unlike TFAddons).
 
-# No Goals
+# Non-Goals
 - **KerasCV is not a repository of blackbox end-to-end solutions (like TF Hub or ModelGarden).**
 
     For the time being, it is focused on modular and reusable building blocks. We expect models in 
@@ -50,19 +50,6 @@ functions, losses, metrics and other necessary components to achieve state of th
 There will be common shareable components between those 2 tasks, like model building blocks, etc. We
 will have them in the repository as public API as well.
 
-### Common components
-- Model building block (served as backbone for the application)
-  - Resnet 
-  - Resnet RS
-  - ResNeXT
-  - Efficient
-- Preprocessing layers
-  - CutMix
-  - MixUp
-  - CutOut
-  - RandAugment
-  - RandomErasing
-
 ### Image Classification
 - Dataset to evaluate
   - [Imagenet](https://www.tensorflow.org/datasets/catalog/imagenet2012)
@@ -77,16 +64,7 @@ will have them in the repository as public API as well.
   - [Open image v4](https://www.tensorflow.org/datasets/catalog/open_images_v4)
   - [PASCAL voc](https://www.tensorflow.org/datasets/catalog/voc)
 - Target Metric
-  - COCO metric average precision (AP)
-- Detection model
-  - MaskRCNN
-  - YOLO (v3 or v4)
-  - RetinaNet
-- Other related utilities
-  - **ROI** (region of interest) layer (faster RCNN and Mask RCNN)
-  - Bounding box utils
-  - **RPN** (Region Proposal Network) used by faster RCNN and mask RCNN for bounding box 
-    generation.
+  - COCO metric mean average precision (mAP)
 
 ## Overlapping between keras.application and Keras-CV
 As you might notice, some area we would like to focus has overlapping with existing keras.application,
@@ -117,10 +95,6 @@ The ideal workflow will be:
 - Post it to Github as a feature request. 
 - Have a hotlist/tag for those items and advocate them to the community.
 - Let community members claim the task they would like to work on.
-
-In the early stage, Keras team members will just claim those tasks by themselves. When there are 
-enough active community members, we should spend more time as reviewers rather than 
-authors. This should apply to bug fixes as well.
 
 User requested features will be carefully evaluated by keras team member before acceptance. 
 Based on the existing experience from tf-addons, even with a guideline of paper ref count > 50, 


README.md
```diff
@@ -1,7 +1,128 @@
...
 
-KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras objects (layers, metrics, etc) that are too specialized to be added to core Keras, but that receive the same level of polish and backwards compatibility guarantees as the rest of the Keras API and that are maintained by the Keras team itself (unlike TFAddons).
+KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras 
+objects (layers, metrics, etc) that are too specialized to be added to core Keras, but that receive 
+the same level of polish and backwards compatibility guarantees as the rest of the Keras API and that 
+are maintained by the Keras team itself (unlike TFAddons).
 
-## Under Construction
-KerasCV is a relatively new project.  Expect regular updates and changes in the near future.
+# No Goals
```
LukeWood, line 12: Non Goals

README.md
```diff
@@ -1,7 +1,128 @@
 # KerasCV
-KerasCV is a repository of modular building blocks (layers, metrics, losses, data-augmentation) that applied computer vision engineers can leverage to quickly assemble production-grade, state-of-the-art training and inference pipelines for common use cases such as image classification, object detection, image segmentation, image data augmentation, etc. 
+KerasCV is a repository of modular building blocks (layers, metrics, losses, data-augmentation) that
+applied computer vision engineers can leverage to quickly assemble production-grade, state-of-the-art
+training and inference pipelines for common use cases such as image classification, object detection, 
+image segmentation, image data augmentation, etc. 
 
-KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras objects (layers, metrics, etc) that are too specialized to be added to core Keras, but that receive the same level of polish and backwards compatibility guarantees as the rest of the Keras API and that are maintained by the Keras team itself (unlike TFAddons).
+KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras
```
LukeWood, line 7: > KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras 
they're -> the components are

README.md
```diff
@@ -1,7 +1,128 @@
...
+   corresponding labels and confidence for the objects in the image).
+
+For each of the tasks, we are targeting to provide user model architectures, layers, preprocessing
+functions, losses, metrics and other necessary components to achieve state of the art accuracy.
+
+There will be common shareable components between those 2 tasks, like model building blocks, etc. We
+will have them in the repository as public API as well.
+
+### Common components
+- Model building block (served as backbone for the application)
```
LukeWood, line 54: Let's remove the Common components section for now.  I think we should offer this through an API doc autogenerated from docstrings so it is accurate and up to date

README.md
```diff
@@ -1,7 +1,128 @@
...
+process for how to contribute to this repository.
+
+The ideal workflow will be:
+
+- Keras defines the API interface for a certain task, eg a layer, model, dataset, metrics, etc. 
+- Post it to Github as a feature request. 
+- Have a hotlist/tag for those items and advocate them to the community.
+- Let community members claim the task they would like to work on.
+
+In the early stage, Keras team members will just claim those tasks by themselves. When there are
```
LukeWood, line 121: I wonder if we need to include this information in the README.  Perhaps we can remove the information that is not relevant to community members for now

README.md
```diff
@@ -1,7 +1,128 @@
...
+  - Top1 Accuracy > 80% (Note that some existing SotA models might not achieve this, like 
+    original ResNet, but this serve as a guideline for newly added models)
+
+### Object Detection
+- Dataset to evaluate
+  - [Coco 2017](https://www.tensorflow.org/datasets/catalog/coco#coco2017)
+  - [Open image v4](https://www.tensorflow.org/datasets/catalog/open_images_v4)
+  - [PASCAL voc](https://www.tensorflow.org/datasets/catalog/voc)
+- Target Metric
+  - COCO metric average precision (AP)
```
LukeWood, line 80: mean average precision MaP.  I believe this is distinct from AP

README.md
```diff
@@ -1,7 +1,128 @@
...
+- Detection model
```
LukeWood, line 81: If you want we can leave the components in that we intend to add - but I'd personally opt to keep the README up to date iwth current info, and have a ROADMAP.md file with future info.

README.md
qlzh727: Done.

README.md
qlzh727: Done.

README.md
qlzh727: Ack. Moved some of them to the classification model section.

README.md
qlzh727: Removed.

README.md
qlzh727: Done.

README.md
qlzh727: sure. Remove the roadmap related items for now.

LukeWood:
Thanks Scott!

README.md
```diff
@@ -1,7 +1,128 @@
...
+KerasCV is a repository of modular building blocks (layers, metrics, losses, data-augmentation) that
+applied computer vision engineers can leverage to quickly assemble production-grade, state-of-the-art
+training and inference pipelines for common use cases such as image classification, object detection, 
+image segmentation, image data augmentation, etc. 
 
-KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras objects (layers, metrics, etc) that are too specialized to be added to core Keras, but that receive the same level of polish and backwards compatibility guarantees as the rest of the Keras API and that are maintained by the Keras team itself (unlike TFAddons).
+KerasCV can be understood as a horizontal extension of the Keras API: they're new first-party Keras 
+objects (layers, metrics, etc) that are too specialized to be added to core Keras, but that receive 
+the same level of polish and backwards compatibility guarantees as the rest of the Keras API and that 
+are maintained by the Keras team itself (unlike TFAddons).
```
bhack, line 10: 1.  What it means `unlike TFAddons`? 
I don't understand what this means by a community point of view. Is only a "political claim" that the Keras (internal) Team select what PR to merge in the Keras-cv repository as it is maintained only by the Keras (internal) Team members without a community codeowenership policy?

2. Also TFAddons doesn't collect e2e networks in the repository and it has almost the same, more or less formal threshold on paper citations as keras-cv when it is going evaluate a contribution. How we explain or route a contributor to keras-cv instead of TFAddons? I think that the old rationale, when also Keras itself was embedded in the TF repo, is not valid anymore: "this is too experimental/fresh try to propose it to TFAddons". With the current limited community resources we have we are not accepting this kind of contribution anymore. I think that not fragmenting more our community could be helpful to not waste the limited resources we have.

/cc @tensorflow/sig-addons-maintainers

README.md
```diff
@@ -1,7 +1,128 @@
...
 
-## Under Construction
-KerasCV is a relatively new project.  Expect regular updates and changes in the near future.
+# No Goals
+- **KerasCV is not a repository of blackbox end-to-end solutions (like TF Hub or ModelGarden).**
```
bhack, line 13: II think that this is a little bit too bold claim as models garden rely also on TFAddons (reusable?) components:

https://github.com/tensorflow/models/blob/master/official/requirements.txt#L13

But you are also aware that we historically had many reusable components duplication over time between TFAddons and Model Gardens without reusing TFAddons reusable components (we have many public tickets/threads on this topic).

/cc @seanpmorgan @fsx950223

README.md
```diff
@@ -1,7 +1,128 @@
...
+
+    For the time being, it is focused on modular and reusable building blocks. We expect models in 
+    ModelGarden or TF Hub to eventually be built on top of the building blocks provided by KerasCV.
+
+    In the process of developing these building blocks, we will by necessity implement end-to-end 
+    workflows, but they're intended purely for demonstration and grounding purposes, they're not 
+    our main deliverable.
+
```
bhack, line 21: In this block i think it is not clear where where you will store the e2d network, its training scripts or they pre-trained weights?
I think it is really important to understand by a community perspective where the trainable model and its weights will be stored.

Cause, in the current form, it seems that model gardens/TFHUB at some point will need to duplicate again all the e2e networks already implemented somewhere to test Keras-cv model driven components.

So I suppose that this is going to clarify where to contribute reusable components but not the e2e network.

README.md
```diff
@@ -1,7 +1,128 @@
...
+
+- **KerasCV is not a repository of low-level image-processing ops, like tf.vision.**
```
bhack, line 23: It is a little bit hard to me what are low level image processing ops. Is this only about performance of TF compiler stack?

E.g. how much operations on IOU or bbox utilites purely implemented in python in this repo are so different from   `tf.image.non_max_suppression` or ` tf.image.extract_patches` that are native TF ops/API.

I think here the contribution path is not clear. Is it really just about speed?

README.md
```diff
@@ -1,7 +1,128 @@
...
+    our main deliverable.
+
+
+- **KerasCV is not a repository of low-level image-processing ops, like tf.vision.** 
+
+    KerasCV is fundamentally an extension of the Keras API: it hosts Keras objects, like layers, 
+    metrics, or callbacks. Low-level C++ ops should go elsewhere.
+
+
+- **KerasCV is not a research library.**
```
bhack, line 29: I think partially we are trying to defining what is this more concretely at:

https://github.com/keras-team/keras-cv/discussions/52#discussioncomment-2058663

In a field like AI/Deep learning it is a little bit hard to talk about  the boundaries between `industry-strength versions` and research in many cases.

Where to put this boundaries is partially opinionated of course but I still think that we could find a baseline to share with the community to not arbitrarily discriminate too much when we need to reject a contribution proposal.
Using only the classic feedback "we have limited" resources it is generally a little bit too weak especially if plan to scale over the community co-ownership.

README.md
```diff
@@ -1,7 +1,128 @@
...
+2. **Object detection** (Given an image and a list of labels, return a list of bounding box, 
+   corresponding labels and confidence for the objects in the image).
+
+For each of the tasks, we are targeting to provide user model architectures, layers, preprocessing
+functions, losses, metrics and other necessary components to achieve state of the art accuracy.
+
+There will be common shareable components between those 2 tasks, like model building blocks, etc. We
+will have them in the repository as public API as well.
+
+### Common components
```
bhack: I suppose this is derived by the original proposal right?

README.md
```diff
@@ -1,7 +1,128 @@
...
+  - Efficient
+- Preprocessing layers
+  - CutMix
+  - MixUp
+  - CutOut
+  - RandAugment
+  - RandomErasing
+
+### Image Classification
+- Dataset to evaluate
```
bhack, line 54: For this point and the next one are we going to integrate with https://github.com/tensorflow/datasets ?

As historically in Keras we have inherited something the was create before TF datasets in the ecosystem:
https://keras.io/api/datasets/

I know that the two are currently  not strictly competing  as `api/dataset` has a very small coverage but as we plan to expand on multiple datasets it could be interesting to clarify this to the community.

README.md
```diff
@@ -1,7 +1,128 @@
...
+change. On the other hand, there are a few of behavior we actually would like to change within the keras-cv
+  - Fetch imagenet weights by default, which might not want to do for keras-cv.
+  - Have the image preprocessing logic outside the model. We want to include all the preprocessing
+    within the model, so that user won't need to call the extra preprocessing methods.
+  - Only end-to-end model are exposed as public API, and Keras CV would like to include more building 
+    blocks as well.
+  - Some legacy model architectures are no longer commonly used and can't reach SotA performance, which
+    we might want to deprecate.
+
+Once the keras-cv applications are mature enough, we would add a deprecation warning in the
```
bhack, line 83: It  think we need to expand this a little bit more this when we clarify my previous comment at the section:
> For the time being, it is focused...

README.md
```diff
@@ -1,7 +1,128 @@
...
+Once the keras-cv applications are mature enough, we would add a deprecation warning in the 
+keras.applications code.
+
+
+## Community contribution guideline
+We would like to leverage/outsource the Keras community not only for bug reporting or fixes, 
+but also for active development for feature delivery. To achieve this, we will have the predefined
+process for how to contribute to this repository.
+
+The ideal workflow will be:
```
bhack, line 92: It would be really nice if we could standardize this section (I suppose it will be something like CONTRIBUTING.md) with  a baseline template:
https://github.com/tensorflow/community/pull/384

I suppose that having a few common section it could help our contributors to navigate quickly our repositories.

README.md
```diff
@@ -1,7 +1,128 @@
...
+- Post it to Github as a feature request. 
+- Have a hotlist/tag for those items and advocate them to the community.
+- Let community members claim the task they would like to work on.
+
+In the early stage, Keras team members will just claim those tasks by themselves. When there are 
+enough active community members, we should spend more time as reviewers rather than 
+authors. This should apply to bug fixes as well.
+
+User requested features will be carefully evaluated by keras team member before acceptance. 
+Based on the existing experience from tf-addons, even with a guideline of paper ref count > 50,
```
bhack, line 100: I think that Keras previously had also its own history other then tf-addons:

https://github.com/keras-team/keras-contrib

But again it is not clear where we want to route the limited contributors resources.

bhack (commented):
I just made a very quick first pass. I hope you could find something useful.

README.md
LukeWood: > In this block i think it is not clear where where you will store the e2d network, its training scripts or they pre-trained weights?

In short, both.  we intend to ensure reproducibility of the weights published alongside models in the package.   We will have `keras_cv.applications.*`, which will host applications similar to `keras_cv.applications`.   Certain applications (i.e. ResNet50) could also considered critical building blocks for many model architectures people build. 

> So I suppose that this is going to clarify where to contribute reusable components but not the e2e network.

We are happy to host a widely reusable network architecture (i.e. ResNet50 if it were not in core Keras) in keras_cv.  I don't think that having a duplicated architecture between us and model garden will dissuade us from hosting a reusable architecture.

README.md
LukeWood: the bbox ops we have are closer to utility functions than low level ops.  They exist because users will need them to interact with the COCO metric implementation.

README.md
bhack: Ok but I still don't understand why we need to compete with models garden on the new model. Can we just figure out a shared way to add this models to Tensorflow garden?

I don't think that having duplicated training scripts, duplicated weights storages and hash  and duplicated models requests handling its is going to optimize what we always claim:
limited internal and community resoruces.

README.md
qlzh727: Sorry for the confusion.

1. Keras team is the dedicated group of engineer to maintain code. We would like to avoid any situation that the code contributor/owner become inactive/leave the community, and the code itself becomes unmaintained. By accepting the contribution from the community, Keras team will co-own the code and maintain it even the original owner no longer maintains it. I think that's the main difference we want to state here.

2. This comes down to whether TF/Keras/Keras-cv/Addon are holding the same acceptance criteria and API compatiblity level. From my expectation, keras-cv is more targeted for CV specify, well recognized common building blocks for keras components only. There are major difference/coverage between keras cv and addon, eg addon still host cv related ops (c++), and anything that is not CV related. @fchollet for more inputs.

LukeWood (approved):
I think there are some open questions here but I still think we are ironing out details, and that is ok.

README.md
bhack: 1. Ok the formulation in is clear now it would be nice if you could expose in some way in this section.
Also TFAddons lead and maintainers co-own the TFA repository code and maintain it as many other SIGs. The only difference I could see is that your "internal" non-volutary team will maintain the API when voluntary contributors or codeowners will disapper. This is similar to other SIGs with TF member leads and maintainer + voluatary contributor.
TF Addons instead is almostfull maintained by voluntary contributors so it is best effort without any SLA. I suppose other pure voluntary SIGS are in the same situation like SIG JVM etc..

2. This is true but if you [add also Keras-nlp](https://github.com/keras-team/keras-nlp/issues/11) (/cc @mattdangerw)  we are quite going to cover almost the same space as TF-Addons. It is true that historically we have c++ infra for tf.contrib migration but we are trying to avoid c++ ops as they have partial hw coverage, too much code overheadl. Also upstream TF custom ops infra is not actively maintained anymore and with the current codeowners retention rate the cost/value of accepting custom ops is very low.
I suppose as with Keras-* if we cannot go compositional or [interact with the compiler team](https://discuss.tensorflow.org/t/deformable-convolution-and-other-custom-ops/1951) for performance we will need to contribute some lowlevel c++ API in the TFcore repository with a PR.

README.md
LukeWood: The benchmarks serve to ensure we have the components in the repository required to achieve these scores.   There's not a specific component for datasets in Keras-CV.  

Basically, we should be able to score state of the art on imagenet, regardless of source, using KerasCV components.

README.md
bhack: In the current use case of a fresh repository but if we are abstracting a little bit bboxes manipulations operations are not so different from a NMS (`tf.image.non_max_suppression`) op.

README.md
bhack: But you still need to feed a dataset to produce and store the weighs and reproduce the scores (implementation reproducibility check). I suppose that in the `keras.application` this process was totally private if I remember correctly. 
Probably here we are going to store score reproducibility protocols/scripts/markdowns in this repo and just using private training job with internal resources to produce the trained weights.
But at the same time if we need to expand over https://keras.io/api/datasets/ I think this will be another source of duplication with the Tensorflow datasets repo that is trying to call the community to collect many research datasets adapters also in the Vision domain.

README.md
qlzh727: model garden was using some of the TFA optimizer and utils. 

For code sharing and duplication, I think if certain methods and useful but not worth a public API, it is reasonable to fork/copy or keep a slight different version in own code base for simplicity.

LukeWood: I agree, I don't see duplicated code as an issue.  I see it as a necessity in providing Keras users with a coherent API for completing CV tasks.

README.md
LukeWood: > Ok but I still don't understand why we need to compete with models garden on the new models.  Can we just figure out a shared way to add this models to Tensorflow garden?

I think we want to provide something unique here.  In particular, we want to really focus on providing a coherent Keras ecosystem.

README.md
LukeWood: I'm not sure I understand the question here

README.md
LukeWood: Yeah, this is a pretty nuanced point.

In particular, I believe we mean the focus is on practical components for applied ML.  This doesn't mean researchers won't use our preprocessing implementations or modelling implementations, they are just not the people we specifically consider when we are authoring the library.

README.md
qlzh727: I think we only want to exclude the c++ part in the keras project and keep it python only (for simplicity reason).

README.md
bhack: I think that it is ok to duplicate or copy private utils but I am not talking about this.
I think that all we agree that Model garden is not exactly a library but my point was not really about private utils but often about components like loss, activation etc.. that is hard to claim as `private`.
I have many example tickets/threads on github accumulated over time but if you want to have just a sample you could check:
https://github.com/tensorflow/models/pull/8962#pullrequestreview-457058720

I am ok that nobody want to depend on anything else external but IMHO by a community point of view there is a difference between an ecosystem and a collection of repos.

README.md
qlzh727: We will move this to the roadmap and not included in the readme.

README.md
LukeWood: > But you still need to feed a dataset to produce and store the weighs and reproduce the scores (implementation reproducibility check). I suppose that in the keras.application this process was totally private if I remember correctly.
Probably here we are going to store score reproducibility protocols/scripts/markdowns in this repo and just using private training job with internal resources to produce the trained weights.
But at the same time if we need to expand over https://keras.io/api/datasets/ I think this will be another source of duplication with the Tensorflow datasets repo that is trying to call the community to collect many research datasets adapters also in the Vision domain.

Sure, from that perspective our training scripts will probably use tensorflow datasets.  As of now, we will not be hosting datasets in this repo.

qlzh727: I think we will prefer the tf.datasets as the main integration point for consuming data, and not speed much effort on expand the support for keras.datasets. The new application model in keras_cv will have tf.datasets in mind, eg it will expect the tf.dataset instance as the preferred input (whereas the keras.datasets usually produce raw numpy array)

As for the training logic and scripts, I think we should make it available for end user as well (if they want to train the model by themselves). It will not be a public API, and we will use that to produce weights/validate performance.

README.md
bhack: I am totally ok but also with this claim I see a little bit of competitive forking between what we call now "Keras ecosystem" and something that was called as whole "TF ecosystem".
If the goal is to create this new entity "Keras ecosystem" I will not push more comments for a more holistic approach to coordinate with what we already have in the TF ecosystem and SIGs.

README.md
bhack: > I think we only want to exclude the c++ part in the keras project and keep it python only (for simplicity reason).

I seems that you confirm my guess about performances. If we cannot achieve this in python with the current compositional ops and compiler performances we could claim this a a low level op that need to be built in c++ and contributed it in the TF core repo or to contribute some extra transformation in the compiler stack.

Other then that this I think that it is a little bit hard to generally claim what it is low level or not.

README.md
qlzh727: > Ok but I still don't understand why we need to compete with models garden on the new models. Can we just figure out a shared way to add this models to Tensorflow garden?
> 
> I don't think that having duplicated training scripts, duplicated weights storages/hashes and duplicated models requests handling its is going to optimize what we always claim: limited internal and community resoruces.

Model garden has their opinionated way for implementing models/layers/training logic (Orbit), which is not necessary same as the way keras team would like to take. Forcing them to adopt the keras team process might be a big ask for them. We would like to provide our solution with keras best practice, and let model garden adopt it when we are ready. In this process, these components can be used by end user as well.

README.md
qlzh727: Ack, will put a contribution.md in a follow up PR.

README.md
bhack: Yes probably this is a little bit better. I think that we need to explain this in term o balance other then bold claims as I suppose we still want that researchers could use this also in the worst case scenario when it will be only used in paper only for comparing over strong and well known reproducible baselines.

README.md
qlzh727: I don't think we will put resource to keras-contrib anymore (I am not even sure why it exists in the first place, with tf.contrib deprecated/removed in tf2.0).

I will check with @fchollet  and see if we can just remove that repository.

README.md
bhack: I want to reinforce the claim that I am one of the promoter or reusable components. But if we need to create a competitive solution I will not going to push for more coordination.

README.md
bhack: Yes it is better. Also pointing on a github Wiki page it could be ok or easier to update.

README.md
bhack: Don't worry. 
After some of the last comments I think that we are more oriented to build a "Keras ecosystem" here so we don't need to enforce things that are proposed under the Tensorflow/community github organization for the Tensorflow Ecoystem.

But if you like or you want to give some inline review it will be appreciated.

I have also a `readme.md` template under review in the same repo:
https://github.com/tensorflow/community/pull/376

README.md
LukeWood: > model garden was using some of the TFA optimizer and utils.   For code sharing and duplication, I think if certain methods and useful but not worth a public API, it is reasonable to fork/copy or keep a slight different version in own code base for simplicity.

Personally, I also see no issue replicating components that may exist in model garden.  

I.E. cutmix & mixup,  both exist in model garden or TF Hub, but we still want to offer them as a first class preprocessing layer the Keras users

README.md
bhack: Yes I know we already discussed this on July:
https://github.com/tensorflow/community/pull/376
My point here was more about what the keras-contrib experience was and so that you can really rely also on the keras-contrib experience/history. 

Then we know some of the components of that repos was migrated in TFAddons like `tf.contribt`.

I still think that we could strongly evaluate if we could find a new role for TFAddons or we could just scheduling a migration of the community and the interesting components we currently have in Keras-* before we are going to create to much duplicates over the time in these two repo (I suppose that keras-nlp  would have a similar rationale as the one under review here).

README.md
bhack: Yes I am totally ok with this and also with copy adapting these from model garden.  I have posted many pointers to model gardes in many tickets on this repo so I am totally ok.
The only thing is that I want to avoid to create confusion on where contribute what for new layers. 
But if the goal is to build a competitive community or a brand new ecosystem I am ok.

README.md
bhack: Yes I agree with your point. 
Now that it is clear I think that we could add something in this section.

README.md
bhack: I think that we could resolve this as now we have a de-facto competitive call for contribution in the communiy and we could not reduce this spending in extra coordination.

bhack: I still think that we could still expand this section with the clarification we had in this thread.

README.md
qlzh727: Sure. It will make a lot of sense to discuss with TFA for any of the existing components and see if they are good fit for keras-cv/nlp so that we can avoid any duplication.

README.md
qlzh727: I will add @martin-gorner and @fchollet in this thread to comment on the interaction between keras-team and model garden as well.

bhack: Thanks, I am ok also if the final outcome will be to compete over concurrent philosophies, at that point we could just describe here our different point of view and then users will select what they want to use and the contributors will choice on which repo the will want to spent their time.

README.md
LukeWood: Yeah, this is probably too strongly worded.  Perhaps we can get the point across by moving this to goals and instead stating that our target is applied ML.

bhack:
More in general I will move this in a `Manifesto.md` or something like that with a link from a `Readme.md` section.

If you like and you want to give some online feedback I have a draft template Readme for the ecosystem at:

https://github.com/tensorflow/community/pull/376

It could help to lower the cognitive overhead for our new community members when they navigate across our ecosystem repositories.

README.md
bhack: A small additional note. Are we going to collect here higher level API in the style of [TF Object detection API](https://tensorflow-object-detection-api-tutorial.readthedocs.io/en/latest/)?

This kind of API were quite popular on stackoverflow, our forum and on github.

README.md
bhack: Just to have a quick pointer to a recent overlap between models grader contribution proposal and keras-cv. I have tried to manually route the contributor with a comment:
https://github.com/keras-team/keras/issues/15545#issuecomment-1025728605

If we are in a competitive setup on model collection between keras-* and models-garden I don't think that we could route users with comment like the one I've done. 
If we disambiguate a little bit what we want to do with the new models it will be easier to a more clean feedback to users on github and on the forum.

qlzh727:
> 

I am not sure what's the normal content for MANIFEST.md (not sure if MANIFEST should be a structured file that to be consumed by other app). I will leave the current content as is, and move to different files if applicable (eg contribution.md).

bhack:
Are we going to lost all the thread that we had?  I see many review points still open and not outdated by changes.

PR #70, opened 2022-01-30
Removes label smoothing from CutMix and MixUp

LukeWood:
Currently, we can accomplish label smoothing in two ways: in CutMix/MixUp, or in the loss function.  Each approach to label smoothing yields a slightly different result.  

To me, this could be a source of confusion to users.  I foresee no functional difference to the learning process between smoothing the labels before the label update is applied by cutmix, and applying it after.  

Besides, if users really want, they can smooth labels before passing them to CutMix and MixUp.  This is pretty trivial, and I think removing the functionality from these layers greatly reduces the cognitive complexity of the code.

https://colab.research.google.com/drive/1ECNMsLHTn6_LQefE2fosQVnD-QWjtZIi?usp=sharing


bhack:
The colab is private.

fchollet (approved):
LGTM, good call IMO.

PR #65, opened 2022-01-30
Batch fill rectangle

chjort:
Vectorized/batched implementation of `fill_rectangle` for significant performance improvements.

This implementation has been benchmarked against the non-vectorized implementation of `fill_rectangle` that uses `tf.map_fn`.

The following table summarizes the time (in milliseconds) it takes to fill a batch of rectangles into a batch of 32x32 images. These are the numbers I get on my own machine using GPU.

| Batch sizes | Batch fill time [milliseconds]| Map fill time [milliseconds] |
|-------------|------------|-----------|
| 1           | 0.5903     | 0.8888    |
| 2           | 0.6151     | 1.1497    |
| 4           | 0.5357     | 1.5249    |
| 8           | 0.5238     | 2.8148    |
| 16          | 0.5359     | 5.3349    |
| 32          | 0.5364     | 9.8739    |
| 64          | 0.5424     | 19.0392   |
| 128         | 0.5553     | 46.8254   |
| 256         | 0.5877     | 89.102    |
| 512         | 0.6801     | 136.8117  |
| 1024        | 0.6406     | 279.6941  |
| 2048        | 0.6969     | 588.2947  |
| 4096        | 0.7975     | 1190.1174 |

And plot:
fill_benchmark

PR #92, opened 2022-02-02
add grid_mask preprocessing layers

innat:
@LukeWood @kartik4949

I used the grid mask-code from the following implementation by kartik https://github.com/google/automl/blob/master/efficientdet/aug/gridmask.py.  Below are some sample outputs for random-grid-mask (`rate=0.5`). 

download

---

**[Concern]:**

The [`fill_value`](https://github.com/google/automl/blob/0b0ba5ebd0860edd939465fc4152da4ff9f79b44/efficientdet/aug/gridmask.py#L30), in my understanding, would change the spatial view of grid-block. For example, in the above example, the gird-block shows black. So, based on the `fill_value`,  it should update those spatial regions. 

But, with `fill_value = 0`, it makes the whole image black, and also for example, `fill_value = 100` makes the whole image noisy. I'm not sure if these are expected or not. In the [test file](https://github.com/google/automl/blob/0b0ba5ebd0860edd939465fc4152da4ff9f79b44/efficientdet/aug/gridmask_test.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L22), I didn't find any check regarding this.

### DIFF
--- a/examples/layers/preprocessing/grid_mask_demo.py
+++ b/examples/layers/preprocessing/grid_mask_demo.py
@@ -33,7 +33,7 @@ def main():
     )
 
     gridmask = preprocessing.GridMask(
-        ratio=0.6, gridmask_size_ratio=0.8, fill_value=1, rate=0.8
+        ratio=0.6, gridmask_size_ratio=0.8, rate=0.8
     )
     train_ds = train_ds.map(
         lambda x, y: (gridmask(x, training=True), y),
--- a/examples/layers/preprocessing/grid_mask_demo.py
+++ b/examples/layers/preprocessing/grid_mask_demo.py
@@ -1,0 +1,53 @@
+"""gridmask_demo.py shows how to use the GridMask preprocessing layer.
+
+Operates on the oxford_flowers102 dataset.  In this script the flowers
+are loaded, then are passed through the preprocessing layers.
+Finally, they are shown using matplotlib.
+"""
+
+import matplotlib.pyplot as plt
+import tensorflow as tf
+import tensorflow_datasets as tfds
+from keras_cv.layers import preprocessing
+
+IMG_SIZE = (224, 224)
+BATCH_SIZE = 64
+
+
+def resize(image, label, num_classes=10):
+    image = tf.image.resize(image, IMG_SIZE)
+    label = tf.one_hot(label, num_classes)
+    return image, label
+
+
+def main():
+    data, ds_info = tfds.load("oxford_flowers102", with_info=True, as_supervised=True)
+    train_ds = data["train"]
+
+    num_classes = ds_info.features["label"].num_classes
+
+    train_ds = (
+        train_ds.map(lambda x, y: resize(x, y, num_classes=num_classes))
+        .shuffle(10 * BATCH_SIZE)
+        .batch(BATCH_SIZE)
+    )
+
+    gridmask = preprocessing.GridMask(
+        ratio=0.6, gridmask_size_ratio=0.8, fill_value=1, rate=0.8
+    )
+    train_ds = train_ds.map(
+        lambda x, y: (gridmask(x, training=True), y),
+        num_parallel_calls=tf.data.AUTOTUNE
+    )
+
+    for images, labels in train_ds.take(1):
+        plt.figure(figsize=(8, 8))
+        for i in range(9):
+            plt.subplot(3, 3, i + 1)
+            plt.imshow(images[i].numpy().astype("uint8"))
+            plt.axis("off")
+        plt.show()
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/layers/preprocessing/grid_mask_demo.py
+++ b/examples/layers/preprocessing/grid_mask_demo.py
@@ -33,7 +33,7 @@ def main():
     )
 
     gridmask = preprocessing.GridMask(
-        ratio=0.6, gridmask_rotation_factor=0.5
+        ratio=0.6, rotation_factor=0.5
     )
     train_ds = train_ds.map(
         lambda x, y: (gridmask(x, training=True), y),
--- a/examples/layers/preprocessing/grid_mask_demo.py
+++ b/examples/layers/preprocessing/grid_mask_demo.py
@@ -33,7 +33,7 @@ def main():
     )
 
     gridmask = preprocessing.GridMask(
-        ratio=0.6, gridmask_size_ratio=0.8, rate=0.8
+        ratio=0.6, rate=0.8
     )
     train_ds = train_ds.map(
         lambda x, y: (gridmask(x, training=True), y),
--- a/examples/layers/preprocessing/grid_mask_demo.py
+++ b/examples/layers/preprocessing/grid_mask_demo.py
@@ -8,6 +8,7 @@ Finally, they are shown using matplotlib.
 import matplotlib.pyplot as plt
 import tensorflow as tf
 import tensorflow_datasets as tfds
+
 from keras_cv.layers import preprocessing
 
 IMG_SIZE = (224, 224)
@@ -33,11 +34,11 @@ def main():
     )
 
     gridmask = preprocessing.GridMask(
-        ratio=0.6, rotation_factor=0.5
+        ratio="random", rotation_factor=0.5, fill_mode="gaussian_noise"
     )
     train_ds = train_ds.map(
         lambda x, y: (gridmask(x, training=True), y),
-        num_parallel_calls=tf.data.AUTOTUNE
+        num_parallel_calls=tf.data.AUTOTUNE,
     )
 
     for images, labels in train_ds.take(1):
--- a/examples/layers/preprocessing/grid_mask_demo.py
+++ b/examples/layers/preprocessing/grid_mask_demo.py
@@ -33,7 +33,7 @@ def main():
     )
 
     gridmask = preprocessing.GridMask(
-        ratio=0.6, rate=0.8
+        ratio=0.6, gridmask_rotation_factor=0.5
     )
     train_ds = train_ds.map(
         lambda x, y: (gridmask(x, training=True), y),
--- a/keras_cv/layers/preprocessing/grid_mask.py
+++ b/keras_cv/layers/preprocessing/grid_mask.py
@@ -1,0 +1,148 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow.keras import layers, backend
+
+
+class GridMask(layers.Layer):
+    """GridMask class for grid-mask augmentation.
+
+    Args:
+        ratio: Grid mask ratio i.e if 0.5 grid and spacing will be equal.
+            Defaults to 0.5.
+        gridmask_size_ratio: Grid mask size, grid to image size ratio.
+            Defaults to 0.5.
+        fill_value: an integer represents the value to be filled inside the grid.
+            Defaults to 1.
+        rate: Float between 0 and 1. The probability of augmenting an input.
+            Defaults to 0.5.
+
+    Sample usage:
+    ```python
+    (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    random_gridmask = keras_cv.layers.preprocessing.GridMask(0.5, 0.5, rate=1.0)
+    augmented_images = random_gridmask(images)
+    ```
+    """
+    
+    def __init__(
+        self,
+        ratio=0.6,
+        gridmask_size_ratio=0.5,
+        fill_value=1,
+        rate=0.5,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.ratio = ratio
+        self.gridmask_size_ratio = gridmask_size_ratio
+        self.fill_value = fill_value
+        self.rate = rate
+
+    @staticmethod
+    def crop(mask, image_height, image_width):
+        '''crops in middle of mask and image corners.'''
+        ww = hh = tf.shape(mask)[0]
+        mask = mask[
+            (hh - image_height) // 2 : (hh - image_height) // 2 + image_height,
+            (ww - image_width) // 2 : (ww - image_width) // 2 + image_width,
+        ]
+        return mask
+
+    @tf.function
+    def mask(self, image_height, image_width):
+        """mask helper function for initializing grid mask of required size."""
+        image_height = tf.cast(image_height, tf.float32)
+        image_width = tf.cast(image_width, tf.float32)
+
+        mask_w = mask_h = tf.cast(
+            tf.cast((self.gridmask_size_ratio + 1), tf.float32)
+            * tf.math.maximum(image_height, image_width),
+            tf.int32,
+        )
+
+        mask = tf.zeros(shape=[mask_h, mask_w], dtype=tf.int32)
+        gridblock = tf.random.uniform(
+            shape=[],
+            minval=int(tf.math.minimum(image_height * 0.5, image_width * 0.3)),
+            maxval=int(tf.math.maximum(image_height * 0.5, image_width * 0.3)) + 1,
+            dtype=tf.int32
+        )
+
+        if self.ratio == 1:
+            length = tf.random.uniform(
+                shape=[], minval=1, maxval=gridblock + 1, dtype=tf.int32
+            )
+        else:
+            length = tf.cast(
+                tf.math.minimum(
+                    tf.math.maximum(
+                        int(tf.cast(gridblock, tf.float32) * self.ratio + 0.5), 1
+                    ),
+                    gridblock - 1,
+                ),
+                tf.int32,
+            )
+
+        for _ in range(2):
+            start_w = tf.random.uniform(
+                shape=[], minval=0, maxval=gridblock + 1, dtype=tf.int32
+            )
+
+            for i in range(mask_w // gridblock):
+                start = gridblock * i + start_w
+                end = tf.math.minimum(start + length, mask_w)
+                indices = tf.reshape(tf.range(start, end), [end - start, 1])
+                updates = (
+                    tf.ones(shape=[end - start, mask_w], dtype=tf.int32)
+                    * self.fill_value
+                )
+                mask = tf.tensor_scatter_nd_update(mask, indices, updates)
+            mask = tf.transpose(mask)
+        return mask
+
+    @tf.function
+    def _grid_mask(self, image, training):
+        image_height = tf.shape(image)[0]
+        image_width = tf.shape(image)[1]
+        grid = self.mask(image_height, image_width)
+
+        mask = tf.reshape(
+            tf.cast(self.crop(grid, image_height, image_width), image.dtype),
+            (image_height, image_width),
+        )
+        mask = tf.expand_dims(mask, -1) if image._rank() != mask._rank() else mask
+
+        rate_cond = tf.less(
+            tf.random.uniform(shape=[], minval=0, maxval=1.0), self.rate
+        )
+        augment_cond = tf.logical_and(rate_cond, training)
+        return tf.cond(augment_cond, lambda: image * mask, lambda: image)
+
+    def call(self, images, training=True):
+        """Masks input image tensor with random grid mask."""
+        if training is None:
+            training = backend.learning_phase()
+        return tf.map_fn(lambda image: self._grid_mask(image, training), images)
+
+    def get_config(self):
+        config = {
+            "ratio": self.ratio,
+            "rate": self.rate,
+            "gridmask_size_ratio": self.gridmask_size_ratio,
+            "fill_value": self.fill_value,
+        }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
--- a/keras_cv/layers/preprocessing/grid_mask.py
+++ b/keras_cv/layers/preprocessing/grid_mask.py
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 import tensorflow as tf
-from tensorflow.keras import layers, backend
+from tensorflow.keras import layers
+from tensorflow.python.keras.utils import layer_utils
 
 
 class GridMask(layers.Layer):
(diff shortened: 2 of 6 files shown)


bhack:
Can we reuse vectorized `fill_rectangle` in https://github.com/keras-team/keras-cv/pull/65?

innat:
@bhack (cc. @chjort )

It would be nice. But I'm not sure if it's a generic solution. Also like, as mentioned [here](https://stackoverflow.com/a/38922451/9215780), the [tf.map_fn](https://www.tensorflow.org/api_docs/python/tf/map_fn) supposed to behave efficiently for larger batch sizes but it looks like the benchmark shown [here](https://github.com/keras-team/keras-cv/pull/65) didn't find its validity.

kartik4949:
@innat Hi,
you can restrict fill value to 1, as I haven't tested it other values, also it should be 0-1 as per I remember its been 1.5 years I wrote the code :(
Thanks

innat:
@kartik4949 thanks for confirming. I understand. I also think that for the original grid mask, it's better to have the fill value fixed. ( There're some variants of grid-mask which probably use adaptive value for filling the gridblock. ). Apart from that, do you have any suggestions regarding the batch operation, i.e vectorization/tf._map_fn?

LukeWood:
> It would be nice. But I'm not sure if it's a generic solution. Also like, as mentioned here, the tf.map_fn supposed to behave efficiently for larger batch sizes but it looks like the benchmark shown here didn't find its validity.

Unfortunately, map_fn still gets heavily outperformed by map.   I bet we could reuse fill_rectangle.

kartik4949:
@innat 
When i wrote the code, i didnt had the fill_rectangle utility,
So i think we can reuse it, it make sense :)
@LukeWood

chjort:
Hey!

I also believe we can reuse the fill utils. I have not fully looked into this, but these are my first thoughts:
If we first compute coordinates of all rectangles (for all images in a batch) in a vectorized manner, we can use `rectangle_masks` from `keras_cv.utils.fill_utils` to generate masks for all rectangles in a vectorized manner. We then have to gather and apply masks to their corresponding image, but this could be a bit tricky since each image has a different number of rectangles (masks).

I would be happy to contribute this, but I also think this PR should be reviewed as is. The vectorization can be included in a new PR so it doesnt have to block this.

LukeWood:
> Hey!
> 
> I also believe we can reuse the fill utils. I have not fully looked into this, but these are my first thoughts: If we first compute coordinates of all rectangles (for all images in a batch) in a vectorized manner, we can use `rectangle_masks` from `keras_cv.utils.fill_utils` to generate masks for all rectangles in a vectorized manner. We then have to gather and apply masks to their corresponding image, but this could be a bit tricky since each image has a different number of rectangles (masks).
> 
> I would be happy to contribute this, but I also think this PR should be reviewed as is. The vectorization can be included in a new PR so it doesnt have to block this.

I think this reflects the general policy I am landing on for new features.  A slow implementation is better than no implementation.  After contribution, I will open a GitHub issue to vectorize each unvectorized component.

bhack:
I think that it is healthy to annotate #TODO sections in the code (or any other string convention). 
So we could quickly find these section also if a vectorized PR will not land just after the merge.

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,148 @@
...
+
+class GridMask(layers.Layer):
+    """GridMask class for grid-mask augmentation. The expected images should be [0-255] pixel ranges.
+    
+    Args:
+        ratio: Grid mask ratio i.e if 0.5, grid and spacing will be equal.
+            Defaults to 0.5. It ranges from 0.0 to 1.0.
+        gridmask_size_ratio: Grid mask size, grid to image size ratio.
+            Defaults to 0.5. It ranges from 0.0 to 1.0.
+        rate: Float between 0 and 1. The probability of augmenting an input.
```
LukeWood, line 27: We don't need to support a rate parameter for this layer.  Thanks

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,148 @@
...
+
+import tensorflow as tf
+from tensorflow.keras import layers, backend
+
+
+class GridMask(layers.Layer):
+    """GridMask class for grid-mask augmentation. The expected images should be [0-255] pixel ranges.
+    
+    Args:
+        ratio: Grid mask ratio i.e if 0.5, grid and spacing will be equal.
```
LukeWood, line 23: "The ratio from grit masks to spacings.  Float in range [0, 1].  Defaults to 0.5, which indicates that grid and spacing will be equal."

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,148 @@
...
+            Defaults to 0.5. It ranges from 0.0 to 1.0.
+        gridmask_size_ratio: Grid mask size, grid to image size ratio.
```
LukeWood, line 25: Similar edits here, how does this differ from the previous parameter?  Expand in the docstring a bit.

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,148 @@
...
+            Defaults to 0.5. It ranges from 0.0 to 1.0.
+        rate: Float between 0 and 1. The probability of augmenting an input.
+            Defaults to 0.5.
+
+    Sample usage:
+    ```python
```
LukeWood, line 31: indent python blurb

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,148 @@
...
+    (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    random_gridmask = keras_cv.layers.preprocessing.GridMask(0.5, 0.5, rate=1.0)
```
LukeWood, line 33: you can remove defaults.

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,148 @@
...
+        augment_cond = tf.logical_and(rate_cond, training)
+        return tf.cond(augment_cond, lambda: image * mask, lambda: image)
+
+    def call(self, images, training=True):
+        """Masks input image tensor with random grid mask."""
+        if training is None:
+            training = backend.learning_phase()
+        
+        # TODO: Make the batch operation vectorize.
+        return tf.map_fn(lambda image: self._grid_mask(image, training), images)
```
LukeWood, line 139: if not training:
    return images

then you can drop the training param

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Sample usage:
+    ```python
+    (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    random_gridmask = keras_cv.layers.preprocessing.GridMask(0.5, 0.5, rate=1.0)
+    augmented_images = random_gridmask(images)
+    ```
+    """
+    
+    def __init__(
```
LukeWood, line 38: In order to match the original implementation, you also will need to implement rotation for the gridmask - right?

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,148 @@
...
+        self,
+        ratio=0.6,
+        gridmask_size_ratio=0.5,
+        rate=0.5,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.ratio = ratio
+        self.gridmask_size_ratio = gridmask_size_ratio
+        self.fill_value = 1 # TODO: make it adaptive, i.e. 'constant' or 'gaussian_noise'
```
LukeWood, line 48: let's support fill value.  it can default to `constant`, with `fill_value=0`.

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,148 @@
...
+        self.rate = rate
+        # TODO: set seed for deterministic result
```
LukeWood, line 50: you can copy the seed code from CutMix

keras_cv/layers/preprocessing/grid_mask.py
LukeWood: You can check the Keras `RandomRotation` layer's implementation in order to see how you can rotate an image randomly.

keras_cv/layers/preprocessing/grid_mask.py
innat: https://github.com/keras-team/keras-cv/issues/100#issuecomment-1029965632

keras_cv/layers/preprocessing/grid_mask.py
innat: How about like this?

```python
 def call(self, images, training=None):
     """Masks input image tensor with random grid mask."""
     if training:
        return tf.map_fn(lambda image: self._grid_mask(image, training), images)
     return images
```

keras_cv/layers/preprocessing/grid_mask.py
innat: @LukeWood I think there is no reference of having a rotation on grid-mask in the paper, [source-paper](https://arxiv.org/pdf/2001.04086.pdf). But it can be an option.

keras_cv/layers/preprocessing/grid_mask.py
innat: I've just observed that in grid-mask, unlike cutout-random_erasing, the grid-block only appears to fill with black value only. So, a param like `fill_value` shouldn't necessarily be an option here for grid-mask augmentation. WDYT?

innat:
download 
download2
download3
download4

keras_cv/layers/preprocessing/grid_mask.py
bhack: I think that more generally many mask/mix fill ops could have an applicabile transform

keras_cv/layers/preprocessing/grid_mask.py
LukeWood: Typically guard clauses scale better.  The default case should be the final return.

so let's just switch the if training and the image return statements.

Additionally, include the None check in training (see RandomCutout).  

Thanks!

keras_cv/layers/preprocessing/grid_mask.py
LukeWood: Gotcha, thank you for adding it.  Cheers

keras_cv/layers/preprocessing/grid_mask.py
LukeWood: Eventually we plan to do some experimentation regarding mirror fill modes for all of these sorts of operations.  For now, this is fine.

LukeWood:
Thank you @innat - this is great.  Keep it up, a few open comments.

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -176,7 +176,19 @@ def _grid_mask(self, image):
             return mask * image
 
     def call(self, images, training=True):
```
LukeWood, line 178: `training=None`

LukeWood:
@innat go ahead and resolve any comments that are addressed - then I can do another review to get this ready for a merge :)

keras_cv/layers/preprocessing/grid_mask.py
innat: @LukeWood Is this ok now?

keras_cv/layers/preprocessing/grid_mask.py
innat: Can you please elaborate? I followed this https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/preprocessing/random_cutout.py#L52-L55

innat:
@LukeWood I think most of the feedback is addressed. Could you please re-check?

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,221 @@
...
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+    Output shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        ratio: The ratio from grid masks to spacings.
+            Float in range [0, 1]. Defaults to 0.5, which indicates that grid and spacing will be equal.
+            In orther word, higher value makes grid size smaller and equally spaced, and opposite.
```
LukeWood, line 33: typo in orther - thanks

keras_cv/layers/preprocessing/grid_mask.py
LukeWood: Indent the python blurb so it is one level in from `Sample Usage:` - also capitolize `Sample Usage`

examples/layers/preprocessing/grid_mask_demo.py
```diff
@@ -0,0 +1,53 @@
...
+    train_ds = train_ds.map(
+        lambda x, y: (gridmask(x, training=True), y),
+        num_parallel_calls=tf.data.AUTOTUNE
+    )
+
+    for images, labels in train_ds.take(1):
+        plt.figure(figsize=(8, 8))
+        for i in range(9):
+            plt.subplot(3, 3, i + 1)
+            plt.imshow(images[i].numpy().astype("uint8"))
```
LukeWood, line 47: Try running this again after fixing the multiply by zero bug I pointed out

LukeWood (changes requested):

LukeWood:
Thanks @innat for the Pr, left an updated code review

keras_cv/layers/preprocessing/grid_mask.py
LukeWood: I was wrong on this one - the indentation is fine here.

Rename `Sample usage` => `Usage` though and add an extra newline after it.

keras_cv/layers/preprocessing/grid_mask.py
LukeWood: Sure, in `RandomCutout` `height_factor` and` width_factor` are required arguments.  In this one, `(0.5, 0.5`, rate=1.0`) are all optional arguments.  They can be removed in the sample usage.

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,221 @@
...
+
+        gridblock = tf.random.uniform(
+            shape=[],
+            minval=int(tf.math.minimum(image_height * 0.5, image_width * 0.3)),
+            maxval=int(tf.math.maximum(image_height * 0.5, image_width * 0.3)) + 1,
+            dtype=tf.int32,
+            seed=self.seed,
+        )
+
+        if self.ratio == 1:
```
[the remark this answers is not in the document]
innat: @LukeWood the `ratio` ranges from `0.0, - 1.0`. It can be `0.4`, `0.9` but giving `0.999` is not considered as a valid pass.

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,221 @@
...
+    augmented_images = random_gridmask(images)
+    ```
+    """
+
+    def __init__(
+        self,
+        ratio=0.6,
+        gridmask_rotation_factor=0.1,
+        fill_mode="constant",
+        fill_value=1,
```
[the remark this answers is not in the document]
innat: The args `fill_value` suppose to be integer, valid range `[1,2,...255]`, where 1 for full black, 100 can be gray, 255 is white. Now, passing fill value `(255 + 100)` will be gray, and iteratively with other values.

keras_cv/layers/preprocessing/grid_mask.py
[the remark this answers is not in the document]
innat: Also, the ratio of 1.0 is a kinda special case, I updated the doc with some comments about it. With a ratio of 1.0, the size and spatial distribution of the grid mask will be applied randomly.

innat:
@LukeWood 
A [colab](https://colab.research.google.com/drive/1ZpId3sDcJUXCkztX4XZTUTQztJe83_3A?usp=sharing) link for a quick test.

LukeWood:
Thanks !  the last thing I am thinking of is if `ratio` is the best name for the parameter.  I am not able to think of anything better off the top of my head, but it isn't super clear to me what it does until I look at the implementation. 

Maybe `fill_ratio` is better.  Also, I'd like to drop the special case for 1.0 and give it a name like `random`.

Can you also provide a demo of what it looks like when the `ratio=1.0`?

Thanks!  @innat almost there, last few changes!

innat:
@LukeWood 
The naming `ratio` is also used in the original implementation, [link](https://github.com/dvlab-research/GridMask/blob/b2439a7ddec41d6b0917a9b91c3ec8687c871239/imagenet_grid/utils/grid.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L9). 

For the demo, I actually gave a colab file for that, link above. It's a small file. In the file, in the last cell, just change the  `ratio = 0.6` to `ratio=1.0`.  

It's special in the sense that, it gives different levels of spatial gird distribution in one call, which I think is a nice option to keep. In colab, you can try it yourself. Again, sooner or later when there will be a vectorized function instead of `tf.map`, many things may need to change accordingly. 

Ps. I'm currently working on [rgb-shift](https://github.com/keras-team/keras-cv/issues/117) and channel-shuffle, I'll send a pr asap.

chjort:
> Again, sooner or later when there will be a vectorized function instead of tf.map, many things may need to change accordingly.

I am pretty far with a vectorized implementation of this layer. I will make a PR not long after this one is merged.

innat:
@chjort Thanks, looking forward to it. I also think there should be some built-in generic method for parallel mapping, for example [tf.vectorized_map](https://www.tensorflow.org/api_docs/python/tf/vectorized_map).

LukeWood:
@innat I was doing some final copyedits and came across some issues with the layer.  I will perform another review to address these issues.

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -67,26 +67,26 @@ def __init__(
...
-            factor=rotation_factor, seed=seed
-        )
+        self.random_rotate = layers.RandomRotation(factor=rotation_factor, seed=seed)
         self.seed = seed
 
     def _check_input_range(self, ratio, fill_mode, fill_value):
-        if ratio not in [ x / pow(0.1, -1) for x in range(0, 10) ]:
-            raise ValueError(f"ratio should be within [0.0, 0.9 ] with 0.1 interval Got {ratio}")
-        
+        if ratio < 0 or ratio > 1:
```
innat, line 80: @LukeWood if `ratio = 0.999` is given, which isn't valid, then how to check beforehand?

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,232 @@
...
+    References:
+        - https://arxiv.org/abs/2001.04086
+    """
+
+    def __init__(
+        self,
+        ratio=0.6,
+        rotation_factor=0.1,
+        fill_mode="constant",
+        fill_value=0,
```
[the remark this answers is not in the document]
innat: Yes, I agree, it's somewhat not intuitive to exclude 0. Let's see what can be done. It may need some changes in the implementation elsewhere.

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -0,0 +1,232 @@
...
+
+    Input shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+    Output shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        ratio: The ratio from grid masks to spacings.  Higher values make the grid size smaller, and
```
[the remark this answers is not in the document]
innat: @LukeWood note this [comment](https://github.com/keras-team/keras-cv/pull/92#issuecomment-1034660222).

keras_cv/layers/preprocessing/grid_mask.py
```diff
@@ -123,8 +123,12 @@ def _compute_mask(self, image_height, image_width):
 
         if self.fill_mode == "constant":
             mask = tf.fill([mask_h, mask_w], self.fill_value)
-        else:
+        elif self.fill_mode == "gaussian_noise":
             mask = tf.cast(tf.random.normal([mask_h, mask_w]), dtype=tf.int32)
+        else:
+            raise ValueError(
+                "Unsupported fill_mode.  `fill_mode` should be 'constant' or 'gaussian_noise'."
+            )
```
innat, line 131: @LukeWood the else check here may not be needed as we already check the input params in the `init` method, within 

```python 
super().__init__(**kwargs)
self._check_input_range(ratio, fill_mode, fill_value)

if fill_mode.lower() not in ["constant", "gaussian_noise"]:
     raise ValueError(
          f"fill_mode should be either 'constant', or 'gaussian_noise' ]. Got {fill_mode}"
)
````

File: keras_cv/layers/preprocessing/grid_mask.py (hunk 15 above)
Comment (LukeWood) [reply keras_cv/layers/preprocessing/grid_mask.py:80]: Sorry, I may misunderstand.  Why is 0.999 not valid?  can't it be any value in [0, 1]?

Comment (innat) [reply keras_cv/layers/preprocessing/grid_mask.py:80]: In the previous implementation, it wasn't working, but I will send an update on your last review. It's fine now.

File: keras_cv/layers/preprocessing/grid_mask.py [hunk 19]
```diff
@@ -0,0 +1,232 @@
...
+        mask = tf.reshape(
+            tf.cast(self._crop(grid, image_height, image_width), dtype=image.dtype),
+            (image_height, image_width),
+        )
+        mask = tf.expand_dims(mask, -1) if image._rank() != mask._rank() else mask
+
+        if self.fill_mode == "constant":
+            return tf.where(mask < self.fill_value, image, mask)
+        else:
+            return mask * image
```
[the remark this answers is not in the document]
innat: not sure, because I can use tf where and fill value but not in gaussian cases.

innat:
@LukeWood The issue you pointed out should be resolved by now. Could you please recheck and give some feedback?

LukeWood:
Great work @innat !  I performed a final round of edits.  Thank you for your contribution!

LukeWood:
@chjort this is now done, feel free to prepare your PR with the vectorized version.

LukeWood:
Great job @innat!  Thanks a lot!

chjort:
@LukeWood  I have now made a vectorized implementation of this layer, but it uses a bit much memory I think. So I'm looking to reduce the memory consumption before submitting a PR.

bhack:
@chjort Do you have already a public branch?

bhack:
Oh I think that it is this one https://github.com/chjort/keras-cv/blob/vectorize_gridmask/keras_cv/layers/preprocessing/grid_mask.py

chjort:
> Oh I think that it is this one https://github.com/chjort/keras-cv/blob/vectorize_gridmask/keras_cv/layers/preprocessing/grid_mask.py

Yes, that is the branch I'm working on

PR #98, opened 2022-02-03
refactor arguments of fill_utils.rectangle_masks

chjort:
Minor change to arguments as per https://github.com/keras-team/keras-cv/pull/65#discussion_r798096578 from @qlzh727

### DIFF
--- a/keras_cv/utils/fill_utils.py
+++ b/keras_cv/utils/fill_utils.py
@@ -74,7 +74,6 @@ def fill_rectangle(images, centers_x, centers_y, widths, heights, fill_values):
         images with filled rectangles.
     """
     images_shape = tf.shape(images)
-    batch_size = images_shape[0]
     images_height = images_shape[1]
     images_width = images_shape[2]
 
--- a/keras_cv/utils/fill_utils.py
+++ b/keras_cv/utils/fill_utils.py
@@ -16,13 +16,14 @@ import tensorflow as tf
 from keras_cv.utils import bbox
 
 
-def rectangle_masks(corners, width, height):
+def rectangle_masks(corners, mask_shape):
     """Computes masks of rectangles
 
     Args:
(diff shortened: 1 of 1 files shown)


keras_cv/utils/fill_utils.py
```diff
@@ -16,15 +16,16 @@
...
 
 
-def rectangle_masks(mask_shape, corners):
-    """Computes positional masks of rectangles in images
+def rectangle_masks(corners, width, height):
+    """Computes masks of rectangles
 
     Args:
-        mask_shape: shape of the masks as [batch_size, height, width].
         corners: rectangle coordinates in corners format.
```
qlzh727, line 23: I think its fine to just use mask_shape as the param (which is in shape [batch, width, height]). The width and height usually should be combined together as one param.

One a side note, please add shape/rank infrmation for the docstring.

chjort: The `batch_size` of the masks, is determined by the number of bboxes in `corners` argument, so I think `mask_shape` should only be [width, height] then.

keras_cv/utils/fill_utils.py
```diff
@@ -16,15 +16,17 @@
...
-        mask_shape: shape of the masks as [batch_size, height, width].
-        corners: rectangle coordinates in corners format.
+        corners: tensor of rectangle coordinates with shape (batch_size, 4) in
+            corners format (x0, y0, x1, y1).
+        mask_shape: a shape tuple as (width, height) indicating the output
+            width and height of masks.
 
     Returns:
-        boolean masks with True at rectangle positions.
+        boolean masks with True at positions within rectangle coordinates.
```
qlzh727, line 29: The output shape should be [batch, width, height].

qlzh727 (approved):

Issue #83, opened 2022-02-02
Invert preprocessing layer


LukeWood:
Reference implementation: https://github.com/tensorflow/similarity/blob/906d141b75acf00d0e1d37dd0d88c432bbb30f57/tensorflow_similarity/augmenters/img_augments.py#L679

LukeWood:
https://github.com/keras-team/keras-cv/pull/91

PR #106, opened 2022-02-07
Added Solarize preprocessing layer.

sebastian-sz:
Introduces Invert operation as a KLP.  Invert operation is one of the operations used by AutoAugment.

[Linked Issue](https://github.com/keras-team/keras-cv/issues/83)

This is mostly a wrapper, around the [original implementation](https://github.com/tensorflow/similarity/blob/906d141b75acf00d0e1d37dd0d88c432bbb30f57/tensorflow_similarity/augmenters/img_augments.py#L679), except I removed the `tf.convert_to_tensor` call, as I was unsure if it was really necessary.

Regarding input discussion: The layer should work with both float and integer inputs, as long as the values are in range 0-255.

Things to consider:
* Is `tf.convert_to_tensor` necessary?
* Perhaps there is a better way to parametrize the test's under `tf.test.TestCase`. I know that `absltest` has [parametrized](https://github.com/abseil/abseil-py/blob/3e4749af698201224165a98930c6845d8c102e0d/absl/testing/parameterized.py#L22) option (`absltest` comes preinstalled with `tensorflow`). I could not find any information for `tf.test.TestCase` equivalent.

### DIFF
--- a/keras_cv/layers/preprocessing/invert.py
+++ b/keras_cv/layers/preprocessing/invert.py
@@ -42,3 +42,11 @@ class Invert(tf.keras.layers.Layer):
 
     def call(self, images):
         return self.max_value - images + self.min_value
+
+    def get_config(self):
+        config = {
+            "min_value": self.min_value,
+            "max_value": self.max_value,
+        }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
--- a/keras_cv/layers/preprocessing/invert.py
+++ b/keras_cv/layers/preprocessing/invert.py
@@ -29,5 +29,16 @@ class Invert(tf.keras.layers.Layer):
         images: Tensor of type int or float, pixels in range [0, 255].
     """
 
+    def __init__(self, min_value=0, max_value=255):
+        super().__init__()
+
+        assert min_value < max_value, (
+            "`min_value` should be smaller than `max_value`. "
+            f"Received: min_value: {min_value}, max_value: {max_value}"
+        )
+
+        self.min_value = min_value
+        self.max_value = max_value
+
     def call(self, images):
-        return 255 - images
+        return self.max_value - images + self.min_value
--- a/keras_cv/layers/preprocessing/invert.py
+++ b/keras_cv/layers/preprocessing/invert.py
@@ -15,18 +15,27 @@ import tensorflow as tf
 
 
 class Invert(tf.keras.layers.Layer):
-    """Inverts the image pixels.
+    """Applies (max_value - pixel + min_value) for each pixel in the image.
+
+    Args:
+        min_value: int or float. Lower bound of input pixel values.
+        max_value: int or float. Upper bound of input pixel values.
 
     Usage:
-        ```
-        invert = Invert()
-        (images, labels), _ = tf.keras.dataset.cifar10.load_data()
+
+    ```python
+        (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+        print(images[0, 0, 0])  # [59 62 63]
         # Note that images are Tensor with values in the range [0, 255]
+        invert = Invert(min_value=0, max_value=255)
         images = invert(images)
-        ```
+        print(images[0, 0, 0])  # [196, 193, 192]
+    ```
 
     Call arguments:
-        images: Tensor of type int or float, pixels in range [0, 255].
+        images: Tensor of type int or float, with pixels in
+            range [`min_value`, `max_value`] and shape [batch, height, width, channels]
+            or [height, width, channels].
     """
 
     def __init__(self, min_value=0, max_value=255):
--- a/keras_cv/layers/preprocessing/invert.py
+++ b/keras_cv/layers/preprocessing/invert.py
@@ -11,8 +11,23 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import tensorflow as tf
 
-from keras_cv.layers.preprocessing.cut_mix import CutMix
-from keras_cv.layers.preprocessing.invert import Invert
-from keras_cv.layers.preprocessing.mix_up import MixUp
-from keras_cv.layers.preprocessing.random_cutout import RandomCutout
+
+class Invert(tf.keras.layers.Layer):
+    """Inverts the image pixels.
+
+    Usage:
+        ```
+        invert = Invert()
+        (images, labels), _ = tf.keras.dataset.cifar10.load_data()
+        # Note that images are Tensor with values in the range [0, 255]
+        images = invert(images)
+        ```
+
+    Call arguments:
+        images: Tensor of type int or float, pixels in range [0, 255].
+    """
+
+    def call(self, images):
+        return 255 - images
--- a/keras_cv/layers/preprocessing/invert_test.py
+++ b/keras_cv/layers/preprocessing/invert_test.py
@@ -17,28 +17,91 @@ from keras_cv.layers.preprocessing.invert import Invert
 
 
 class InvertTest(tf.test.TestCase):
-    invert = Invert()
+    def test_raises_error_on_invalid_min_max_values(self):
+        with self.assertRaises(AssertionError):
+            Invert(min_value=255, max_value=0)  # min must be smaller than max
 
-    def test_return_shape_unchanged(self):
-        dummy_input = tf.ones(shape=(2, 224, 224, 3))
+    def test_range_0_to_255(self):
+        invert = Invert(min_value=0, max_value=255)
 
-        output = self.invert(dummy_input)
-
-        self.assertEqual(dummy_input.shape, output.shape)
-
-    def test_return_values(self):
         test_parameters = [
-            {"input_value": 0, "expected_value": 255},
-            {"input_value": 255, "expected_value": 0},
-            {"input_value": 127.5, "expected_value": 127.5},
+            {"input_value": 0, "expected_output_value": 255},
+            {"input_value": 64, "expected_output_value": 191},
+            {"input_value": 127, "expected_output_value": 128},
+            {"input_value": 191, "expected_output_value": 64},
+            {"input_value": 255, "expected_output_value": 0},
         ]
 
         for parameters in test_parameters:
-            self._test_input_output(**parameters)
+            self._test_input_output(
+                layer=invert,
+                input_value=parameters["input_value"],
+                expected_value=parameters["expected_output_value"],
+                dtype=tf.uint8,
+            )
 
-    def _test_input_output(self, input_value, expected_value):
-        dummy_inputs = tf.ones(shape=(1, 224, 224, 3)) * input_value
+    def test_range_0_to_1(self):
+        invert = Invert(min_value=0.0, max_value=1.0)
 
-        outputs = self.invert(dummy_inputs)
+        test_parameters = [
+            {"input_value": 1.0, "expected_output_value": 0.0},
+            {"input_value": 0.75, "expected_output_value": 0.25},
+            {"input_value": 0.5, "expected_output_value": 0.5},
+            {"input_value": 0.25, "expected_output_value": 0.75},
+            {"input_value": 0.0, "expected_output_value": 1.0},
+        ]
 
-        self.assertTrue((outputs == expected_value).numpy().all())
+        for parameters in test_parameters:
+            self._test_input_output(
+                layer=invert,
+                input_value=parameters["input_value"],
+                expected_value=parameters["expected_output_value"],
+                dtype=tf.float32,
+            )
+
+    def test_range_minus_1_to_plus_1(self):
+        invert = Invert(min_value=-1.0, max_value=1.0)
+
+        test_parameters = [
+            {"input_value": -1.0, "expected_output_value": 1.0},
+            {"input_value": -0.5, "expected_output_value": 0.5},
+            {"input_value": 0.0, "expected_output_value": 0.0},
+            {"input_value": 0.5, "expected_output_value": -0.5},
+            {"input_value": 1.0, "expected_output_value": -1.0},
+        ]
+
+        for parameters in test_parameters:
+            self._test_input_output(
+                layer=invert,
+                input_value=parameters["input_value"],
+                expected_value=parameters["expected_output_value"],
+                dtype=tf.float32,
+            )
+
+    def test_custom_range_minus_8_to_plus_16(self):
+        invert = Invert(min_value=-8, max_value=16)
+
+        test_parameters = [
+            {"input_value": -8, "expected_output_value": 16},
+            {"input_value": -2, "expected_output_value": 10},
+            {"input_value": 4, "expected_output_value": 4},
+            {"input_value": 10, "expected_output_value": -2},
+            {"input_value": 16, "expected_output_value": -8},
+        ]
+
+        for parameters in test_parameters:
+            self._test_input_output(
+                layer=invert,
+                input_value=parameters["input_value"],
+                expected_value=parameters["expected_output_value"],
+                dtype=tf.int32,
+            )
+
+    @staticmethod
+    def _test_input_output(layer, input_value, expected_value, dtype):
+        dummy_input = tf.ones(shape=(2, 224, 224, 3), dtype=dtype) * input_value
+        expected_output = tf.ones(shape=(2, 224, 224, 3), dtype=dtype) * expected_value
+
+        output = layer(dummy_input)
+
+        tf.debugging.assert_equal(output, expected_output)
--- a/keras_cv/layers/preprocessing/invert_test.py
+++ b/keras_cv/layers/preprocessing/invert_test.py
@@ -1,0 +1,44 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


LukeWood:
Thanks for the PR Sebastian!

Let's add some constructor parameters:
- `max_value=255`
- `min_value=0`

This way users can specify custom ranges if needed:
```
invert = Invert(max_value=1, min_value=-1)
```

An alternative API would be:
```
invert = Invert(value_range=[0, 1])
```

LukeWood:
> Thanks for the PR Sebastian!
> 
> Let's add some constructor parameters:
> 
> * `max_value=255`
> * `min_value=0`
> 
> This way users can specify custom ranges if needed:
> 
> ```
> invert = Invert(max_value=1, min_value=-1)
> ```
> 
> An alternative API would be:
> 
> ```
> invert = Invert(value_range=[0, 1])
> ```

I'd be curious to hear @fchollet's opinion on this.

keras_cv/layers/preprocessing/invert.py
```diff
@@ -0,0 +1,33 @@
...
+        # Note that images are Tensor with values in the range [0, 255]
+        images = invert(images)
+        ```
+
+    Call arguments:
+        images: Tensor of type int or float, pixels in range [0, 255].
+    """
+
+    def call(self, images):
+        return 255 - images
```
sampathweb, line 33: What's the expected behavior if the values are outside of this range.  Should we Clip the Values to 0-255?
Also, can the same layer be used when its scaled to 0-1 instead of 0-255?

This layer makes an assumption that all values are 0-255.  Can this be parameterized?

LukeWood: We should raise a `ValueError` when a value is outside of the range - as an `Invert` function should only have a domain of [minpixel, maxpixel]

keras_cv/layers/preprocessing/invert_test.py
```diff
@@ -0,0 +1,44 @@
...
+            {"input_value": 0, "expected_value": 255},
+            {"input_value": 255, "expected_value": 0},
+            {"input_value": 127.5, "expected_value": 127.5},
+        ]
+
+        for parameters in test_parameters:
+            self._test_input_output(**parameters)
+
+    def _test_input_output(self, input_value, expected_value):
+        dummy_inputs = tf.ones(shape=(1, 224, 224, 3)) * input_value
```
qlzh727, line 40: Could u also test with int dtype? (tf.ones default to float32 I think).

keras_cv/layers/preprocessing/invert.py
qlzh727: 1. I think we should have a init param to control what's the range of the value we should accept (eg, if its on the scale of 0~1 or 0~255). Based on imagenet and most of the other datasets, I think we should default the value to 255
2. Raise ValueError means we have to check against the tensor value of the input, which might not be a cheap computation and we probably would like to not pay this every single time.

keras_cv/layers/preprocessing/invert.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Invert(tf.keras.layers.Layer):
+    """Inverts the image pixels.
+
+    Usage:
+        ```
+        invert = Invert()
+        (images, labels), _ = tf.keras.dataset.cifar10.load_data()
+        # Note that images are Tensor with values in the range [0, 255]
+        images = invert(images)
```
qlzh727, line 25: we might want to add some numerical value here, eg [100, 150, 200] is invert to [155, 105, 55]

qlzh727 (changes requested):
Thanks for the PR.

keras_cv/layers/preprocessing/invert.py
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+
+class Invert(tf.keras.layers.Layer):
+    """Inverts the image pixels.
+
+    Usage:
+        ```
```
fchollet, line 21: Add line break above, unindent code block, and add `python` tag

keras_cv/layers/preprocessing/invert.py
```diff
@@ -0,0 +1,33 @@
...
+        invert = Invert()
+        (images, labels), _ = tf.keras.dataset.cifar10.load_data()
+        # Note that images are Tensor with values in the range [0, 255]
+        images = invert(images)
+        ```
+
+    Call arguments:
+        images: Tensor of type int or float, pixels in range [0, 255].
```
fchollet, line 29: Specify shape information

keras_cv/layers/preprocessing/invert.py
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+
+class Invert(tf.keras.layers.Layer):
+    """Inverts the image pixels.
```
fchollet, line 18: This is a somewhat ambiguous description, please provide the equation below. In particular, "invert" taking literally means "1 / x", but this is `255 - x`

fchollet (commented):
Thanks for the PR. Is "invert" really what this operation is canonically called? "Invert" seems to suggest `1 / x`. I believe the operation `255 - x` is actually called "solarization" (though solarization is usually performed on a single channel, not all channels at once).

keras_cv/layers/preprocessing/invert.py
LukeWood: Very good point on 2.) - thanks for pointing that out.

keras_cv/layers/preprocessing/invert.py
LukeWood: For what its worth, in the RandAugment linked below they distinguish it by applying a threshold to `Solarize`.  

https://github.com/tensorflow/similarity/blob/906d141b75acf00d0e1d37dd0d88c432bbb30f57/tensorflow_similarity/augmenters/img_augments.py#L449

That being said... the original `RandAugment` specification (https://arxiv.org/pdf/1909.13719.pdf) doesn't even include `Invert`.  Only `solarize`. 

AutoAugment does include a mention of `invert`, and defines the two operations as seen below:

Screen Shot 2022-02-07 at 9 43 43 PM

sebastian-sz: Apart from what @LukeWood mentioned, `Invert` name is also used in most other implementations. It might not be the best, but changing it now could introduce confusion.

[Link 1](https://github.com/tensorflow/models/blob/dd02069717128186b88afa8d857ce57d17957f03/official/legacy/image_classification/augment.py#L520)
[Link 2](https://github.com/tensorflow/similarity/blob/906d141b75acf00d0e1d37dd0d88c432bbb30f57/tensorflow_similarity/augmenters/img_augments.py#L679)
[Link 3](https://imgaug.readthedocs.io/en/latest/source/overview/arithmetic.html#invert)

bhack: In the official reference impl of the paper It Is just:
https://github.com/tensorflow/tpu/blob/master/models/official/efficientnet/autoaugment.py#L398-L401

In other Google papers is PIL/PILLOW `invert`

https://github.com/google-research/google-research/blob/master/ieg/dataset_utils/randaugment.py#L239:L241

keras_cv/layers/preprocessing/invert.py
sebastian-sz: 1) Agreed on `min_value` and `max_value` parameters. I will be adding this.

2) Is it decided not to include input values validation in `call` method? It is not only costly, but also problematic in graph mode. 

Personally, I think it could be user's responsibility to test if they are passing values of correct range. To highlight this we could not provide default values of 0 and 255 to `min_value` and `max_value` and require the user to explicitly provide them. 

Please, let me know what you think.

keras_cv/layers/preprocessing/invert.py
fchollet: Precedent is one thing to look at, but correctness matters too. Deep learning researchers are memoryless and tend to be ignorant about established nomenclature, which leads to a lot of confusion for everyone down the road. A few examples from our recent past:

- What people call "normalization" in contexts such as "batch normalization" or "layer normalization" (which is an even more meaningless name...) is actually standardization. "Normalization" refers to processing data so that it has a certain norm (usually unit L2 norm). That's where the name "normalization" comes from. Seeking a fixed standard deviation is "standardization" (the name "standardization" comes "standard deviation"). There are no norms involved in batch normalization / etc, which should be called batch standardization.

- A majority of researchers are calling "dilated convolution" "atrous convolution". There are several layers of confusion here. "Atrous" is an English word which means "black". That's entirely unrelated to dilated convolution. The word seems to come from the French term "à trous" which literally means "with holes" and is used in French math textbooks to mean "sparse". But "sparse convolution" is convolution with sparse tensors, not dilated convolution (and if you really wanted to talk about sparse convolution, you'd call it sparse convolution, not "convolution à trous", because you're writing in English, not French). The fact that the Keras API went with the correct term "dilated convolution" played a meaningful role in fixing the nomenclature confusion here.

- Once upon a time most people were calling "transposed convolution" "deconvolution". Deconvolution is actually a different operation. Unfortunately the Keras API originally made the mistake of going with the mainsteam, and we had to change our layer names back to "Conv2DTranspose" later down the road.

- Etc.

In the current situation it does appear that the proper and specific term is solarization, not inversion. "Inversion" is super vague anyway. The first thing comes to mind when reading "inversion" is "taking the inverse" (1 / x). The second that comes to mind is "reversing", as in reversing the order of an input sequence. Etc.

keras_cv/layers/preprocessing/invert.py
sebastian-sz: Added shape information.

keras_cv/layers/preprocessing/invert.py
sebastian-sz: Added line break and `python` tag unindented code block.

keras_cv/layers/preprocessing/invert.py
sebastian-sz: Added printing of the three first values across channels.

keras_cv/layers/preprocessing/invert_test.py
sebastian-sz: I refactored the tests to include:
* 0-255 uint8
* 0-1 float32
* -1 +1 float32
* custom range -8 +16 int32

sebastian-sz:
I updated the PR. Summary:
* Updated docstring.
* Added construction params. Values modification now happens based on user provided `min_value` and `max_value`.
* Refactored tests to cover most common value ranges.

Things to discuss:
@fchollet mentioned that perhaps `Invert` should be renamed to `Solarization`. I can rename the layer, but `Solarize` is already a name of other operation in AutoAugment (it is basically a Solarize operation with a threshold). One would have to remember and change the other name too.

keras_cv/layers/preprocessing/invert.py
sebastian-sz: @sampathweb I don't think that clipping values is going to help much. I would not add it here.

Example scenario 1: Invert with values 0 and 1 but user passes 0-255.
The values would be clipped to 0 and 1. These would be the only 2 unique values in the image.

Example scenario 2: Invert with values 0-255 but user passes 0-1
Many values close to 0 and 255, clipping does not help here and image is almost unchanged.

EIther way the layer does not do what it's supposed to. I would still transfer the responsibility of validating input to the end user.

LukeWood:
> I updated the PR. Summary:
> 
> * Updated docstring.
> * Added construction params. Values modification now happens based on user provided `min_value` and `max_value`.
> * Refactored tests to cover most common value ranges.
> 
> Things to discuss: @fchollet mentioned that perhaps `Invert` should be renamed to `Solarization`. I can rename the layer, but `Solarize` is already a name of other operation in AutoAugment (it is basically a Solarize operation with a threshold). One would have to remember and change the other name too.

If this is the case, we can simply support a `threshold` parameter and use our Solarization layer for both.  I agree with Francois and would prefer to go with the proper established terminology here.

keras_cv/layers/preprocessing/invert.py
```diff
@@ -0,0 +1,61 @@
...
+        print(images[0, 0, 0])  # [196, 193, 192]
+    ```
+
+    Call arguments:
+        images: Tensor of type int or float, with pixels in
+            range [`min_value`, `max_value`] and shape [batch, height, width, channels]
+            or [height, width, channels].
+    """
+
+    def __init__(self, min_value=0, max_value=255):
```
LukeWood, line 41: IMO min_value and max_value should be required parameters - this will prevent people from passing floats, oblivious to the internal behavior of the layer.

keras_cv/layers/preprocessing/invert.py
```diff
@@ -0,0 +1,61 @@
...
+
+    Usage:
+
+    ```python
+        (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+        print(images[0, 0, 0])  # [59 62 63]
+        # Note that images are Tensor with values in the range [0, 255]
+        invert = Invert(min_value=0, max_value=255)
+        images = invert(images)
+        print(images[0, 0, 0])  # [196, 193, 192]
```
LukeWood, line 32: nit: let's place these # [...] comments on newlines - slightly better readability

keras_cv/layers/preprocessing/invert.py
LukeWood: opinon @qlzh727 @fchollet ?

LukeWood: Sorry - I mean't passing differently ranged values.

LukeWood: Perfect, thanks!

fchollet: The layer would then start by clipping to the value range, then apply `max_value - x` ?

LukeWood: if you're in the range [-1, 1] a solarize operation is actually:

```
(2 - (v + 1)) - 1
```

Right?  If I understand correctly you can't just subtract from max value if negatives are included in the value range.

v=0.5
maxval=1
minval=-1

then if we just do:
1-0.5 => 0.5

This is not solarizing that value, which should actually become -0.5 if I understand the solarize operation correctl.

Would:

(maxval - minval) - (v - minval) - minval =>
1 - - 1 - 0.5 + 1 - -1 => 
2 - 0.5 + 2 => -0.5

work?

keras_cv/layers/preprocessing/invert.py
LukeWood: Well, one thing is clear to me:

Even if we wanted to match their terminology, we can still fit this need by just using a solarization layer with a threshold=None. Then it solarizes everything, executing the "Invert" operation.  So in this case we should go with a thresholded solarization layer.

sebastian-sz: Agree with @LukeWood . Adding threshold here, and setting it to None sounds like a good compromise. I will be adding this.

keras_cv/layers/preprocessing/invert.py
sebastian-sz: > if you're in the range [-1, 1] a solarize operation is actually:
> 
> ```
> (2 - (v + 1)) - 1
> ```
> 
> Right? If I understand correctly you can't just subtract from max value if negatives are included in the value range.
> 
> v=0.5 maxval=1 minval=-1
> 
> then if we just do: 1-0.5 => 0.5
> 
> This is not solarizing that value, which should actually become -0.5 if I understand the solarize operation correctly.

It still works for negative ranges, like -1 and +1:
```python
min_value = -1
max_value = 1
# the equation is currently: max_value - x + min_value
# for value 0.5 the equation essentially becomes:
1 - 0.5 + (-1) => 1 - 0.5 - 1 => 0.5 - 1 = -0.5
```
Similary to other values, like 0.25 resulting in -0.25, (or reverse: -0.25 resulting in 0.25) this should be ok.
I added tests to test few example values.

sebastian-sz: > The layer would then start by clipping to the value range, then apply `max_value - x` ?

@fchollet I am not sure about clipping input values. If it is decided that it's good to have I will add it, but personally I'm not sure if it's necessary. Please, let me know.
(This is probably what @sampathweb had in mind before, but I misunderstood him)

After that, yes, the value would be calculated by getting `max_value - x + min_value`.

fchollet: > @fchollet I am not sure about clipping input values. If it is decided that it's good to have I will add it, but personally I'm not sure if it's necessary. Please, let me know.

The problem with ignoring the value range is that output might silently be garbage if it's outside the expected range. There are possible alternatives to clipping:

- Scaling to the target range, by subtracting the min and dividing by the max. An issue here is that the scaling function may be different for different inputs, which messes with the input distribution.
- Erroring out if the input values fall outside the target range. This is the safest option.

> If this is the case, we can simply support a threshold parameter and use our Solarization layer for both

I think that's a great idea, it also avoids the logic redundancy with having two layers for similar use cases!

bhack: We could put some reference in the docstring about  the inversion naming. E.g. `some paper etc..`

keras_cv/layers/preprocessing/invert.py
sebastian-sz: Moved comments to newlines

keras_cv/layers/preprocessing/invert.py
sebastian-sz: Input values clipping has been added.

@LukeWood I removed the default constructor values and added optional threshold parameter. 

@bhack I mentioned AutoAugment in the docstring and which AutoAugment transform will be applied based on whether`threshold` parameter has been specified.

sebastian-sz:
@LukeWood thanks! Cheers!

Issue #123, opened 2022-02-10
ChannelShuffling preprocessing layer


innat:
ref. https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.ChannelShuffle

ref https://pytorch.org/docs/stable/generated/torch.nn.ChannelShuffle.html

LukeWood:
Any numbers showing the usefulness of this in training?

bhack:
In Pytorch it was not strictly a [vision preprocessing layer](https://pytorch.org/docs/stable/generated/torch.nn.ChannelShuffle.html)

If I remeber correctly it was intrdouced implementing Shufflenets:
https://arxiv.org/abs/1707.01083

LukeWood:
That's fine.  Users can use it however they like.

qlzh727:
https://github.com/keras-team/keras-cv/pull/124

PR #124, opened 2022-02-10
Add ChannelShuffle preprocessing layers

innat:
makeup https://github.com/keras-team/keras-cv/issues/123

download


innat:
demo 1

innat:
demo 2

LukeWood:
Thanks for the PR @innat .  Let's discuss some more references in https://github.com/keras-team/keras-cv/issues/123 before merging.

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,89 @@
...
+        super().__init__(**kwargs)
+        self.seed = seed
+
+    @tf.function
+    def _channel_shuffling(self, image):
+        x = tf.transpose(image)
+        x = tf.random.shuffle(x, seed=self.seed)
+        return tf.transpose(x)
+
+    def call(self, images, training=None):
```
fchollet, line 54: Use `training=True` for all augmentation layers (default behavior should be to apply the augmentation)

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,89 @@
...
+        """call method for the ChannelShuffle layer.
+
+        Args:
+            images: Tensor representing images of shape
+                [batch_size, width, height, channels], with dtype tf.float32 / tf.uint8, or,
+                [width, height, channels], with dtype tf.float32 / tf.uint8
+        Returns:
+            images: augmented images, same shape as input.
+        """
+        if training is None:
```
fchollet, line 64: We can just assume `training` is a Python bool.

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_config(self):
+        config = {
+            "seed": self.seed,
+        }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
+
+    def compute_output_shape(self, input_shape):
+        return input_shape
```
fchollet, line 89: Add line break below

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,89 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf 
+from tensorflow.keras import layers, backend
+
+class ChannelShuffle(layers.Layer):
+    """ChannelShuffle class for shuffling the channel of RGB image. The expected images
```
fchollet, line 19: Please format docstring to follow the same format as other docstrings

fchollet (commented):
Thanks for the PR!

keras_cv/layers/preprocessing/channel_shuffle.py
LukeWood: Do we need `compute_output_shape` methods?  What utility do they serve?  Are shapes not simply inferred from the graph?

fchollet: They are. Currently they serve a unit testing purpose: in standard layer tests, we compare the actual (computed) with the expected shape (from `compute_output_shape()`).

Other than that, the only situation where the method is required is for non-graphable layers (built with `dynamic=True`). This is a use case so niche we might potentially get rid of it in the future.

We could get rid of the method here, but then we should think of an alternative for standardized unit testing.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: Done.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: Followed this style https://github.com/keras-team/keras/blob/v2.8.0/keras/layers/preprocessing/image_preprocessing.py#L253-L255

keras_cv/layers/preprocessing/channel_shuffle.py
innat: I made some changes, could you please check now?

keras_cv/layers/preprocessing/channel_shuffle.py
bhack: Can we standardize the call signature? sometimes we have images in other case we have inputs

bhack:
As we have discussed in https://github.com/keras-team/keras-cv/issues/123#issuecomment-1039747156 this probably is not a strictly preprocessing (only) layer.

innat:
@bhack 
Yes, I was thinking about it. But not sure what to decide. 

Like, if we consider channel-shuffle from shuffle-net, then the question would be whether we want channel-shuffle as a layer or not to build the model, i.e shuffle-net. For example, layer like, [permute](https://keras.io/api/layers/reshaping_layers/permute/), [repeat_vector](https://keras.io/api/layers/reshaping_layers/repeat_vector/), [reshape](https://keras.io/api/layers/reshaping_layers/reshape/) etc . Or, just a [preprocessing](https://keras.io/api/layers/preprocessing_layers/) layer.

Also, the `torch` has a shuffle-net in its application, unlike keras. I also wonder how many models do have such a channel-shuffling strategy.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: Until now, in most cases `images` were chosen in [keras-cv](https://github.com/keras-team/keras-cv/tree/master/keras_cv/layers/preprocessing), which makes it reasonable to pick for the CV layers. But I think layers in `keras` have the `inputs` signature.

innat:
@fchollet what do you think about the above discussion? @bhack pointed out some valid points. Referring to the layer name as **ChannelShuffle** is practiced by the [albumentation](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.ChannelShuffle) for image augmentation. On the other hand, it's referred to as the **same name** but as a **standard layer** in the [torch](https://pytorch.org/docs/stable/generated/torch.nn.ChannelShuffle.html).  Apart from that, I am not sure what minimum criteria an image processing method needs to fulfill to be part of KPL! 

Below are some option, I think that can be picked:

- [Option 1]: Keep it as an RGB image processing technique but rename it to something else reasonable, i.e `RGBShuffle`.
- [Option 2]: Consider converting it as a standard layer, like torch does: `torch.nn.ChannelShuffle`

innat:
> It subclasses `Layer` and it will be accessed via `keras_cv.layers.ChannelShuffle`, so it doesn't seem particularly important whether it's classified as a "preprocessing layer" or not. It could be used for any use case. I think it's better to keep it more general.

@fchollet It's not generalized. To match up with the existing channel-shuffle [layer](https://pytorch.org/docs/stable/_modules/torch/nn/modules/channelshuffle.html#ChannelShuffle), a new param `group` is introduced. I set 3 by default to work on RGB normally. Please take a look. 

Sample usages. 

```python
a = tf.constant(
    [[
      [[ 1,  2,  3,  4],
       [ 5,  6,  7,  8]],
      
      [[ 9, 10, 11, 12],
       [13, 14, 15, 16]]
     ]]
)

a.shape : (1, 2, 2, 4))
```

```python
aug_fn = ChannelShuffle(group=2)
aug_fn(a, training=True)

<tf.Tensor: shape=(1, 2, 2, 4), dtype=int32, numpy=
array([[[[ 3,  1,  4,  2],
         [ 7,  5,  8,  6]],

        [[11,  9, 12, 10],
         [15, 13, 16, 14]]]], dtype=int32)>
```

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -39,17 +41,34 @@ class ChannelShuffle(layers.Layer):
...
-        x = tf.random.shuffle(x, seed=self.seed)
-        return tf.transpose(x)
+    def _channel_shuffling(self, images):
+        unbatched = images.shape.rank == 3
+        
+        if unbatched:
+            images = tf.expand_dims(images, axis=0)
+
+        batch_size, height, width, num_channels = images.get_shape().as_list()
+        assert num_channels % self.groups == 0 , ('input channels should be divisible by the number of group')
```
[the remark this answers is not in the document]
innat: Fixed.

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+        groups=3,
+        seed=None,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.groups=groups
+        self.seed = seed
+
+    def _channel_shuffling(self, images):
+        unbatched = images.shape.rank == 3
```
LukeWood, line 53: we can't use `.shape` in graph mode.  I believe we need to use `tf.shape`.

keras_cv/layers/preprocessing/channel_shuffle.py
LukeWood: > Followed this style https://github.com/keras-team/keras/blob/v2.8.0/keras/layers/preprocessing/image_preprocessing.py#L253-L255

Going forwards we can assume `training` is a Python bool.

keras_cv/layers/preprocessing/channel_shuffle_test.py
```diff
@@ -0,0 +1,88 @@
+import tensorflow as tf
+from keras_cv.layers.preprocessing.channel_shuffle import ChannelShuffle
+
+
+class ChannelShuffleTest(tf.test.TestCase):
```
LukeWood, line 5: Let's add a test case to test it in a `@tf.function` or a `keras.Model`

keras_cv/layers/preprocessing/channel_shuffle.py
innat: I think it's safe. Check here https://github.com/keras-team/keras/blob/v2.8.0/keras/layers/preprocessing/image_preprocessing.py#L861-L872

keras_cv/layers/preprocessing/channel_shuffle_test.py
innat: There is a function inside this class called `test_in_tf_function`, is that what you're asking?

keras_cv/layers/preprocessing/channel_shuffle.py
innat: @LukeWood 
Can you elaborate on your quote? Did you mean `if training:` instead of  `if training is None:`?

keras_cv/layers/preprocessing/channel_shuffle.py
fchollet: Right, `shape.rank` will work, however e.g. `shape[3]` might be None.

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf 
+from tensorflow.keras import layers, backend
+
+class ChannelShuffle(layers.Layer):
+    """ChannelShuffle performs shuffling the channel of inputs.
```
fchollet, line 19: "Shuffle channels of an input image."

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+        The expected images should be [0-255] pixel ranges.
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+    Output shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        groups:
+            Number of groups to divide the input channels. Default 3.
```
fchollet, line 31: Remove line break, e.g. "groups: Number of..."

Same below.

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+from tensorflow.keras import layers, backend
+
+class ChannelShuffle(layers.Layer):
+    """ChannelShuffle performs shuffling the channel of inputs.  
+
+    Input shape:
+        The expected images should be [0-255] pixel ranges.
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+    Output shape:
```
fchollet, line 25: Add line break above

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+    Args:
+        groups:
+            Number of groups to divide the input channels. Default 3. 
+        seed:
+            Integer. Used to create a random seed.
+
+    Usage:
+    ```python
+    (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    channelshuffle = keras_cv.layers.preprocessing.ChannelShuffle()
```
fchollet, line 38: Just `keras_cv.layers. ChannelShuffle()`. Also make sure to import the symbol in the `__init__` of the layers module

fchollet, line 38: `channel_shuffle =`

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+    ```
+    """
+    def __init__(
+        self,
+        groups=3,
+        seed=None,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.groups=groups
```
fchollet, line 49: Make sure to use spaces around `=`. Try running a linter on your PR.

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+        self.seed = seed
+
+    def _channel_shuffling(self, images):
+        unbatched = images.shape.rank == 3
+        
+        if unbatched:
+            images = tf.expand_dims(images, axis=0)
+
+        batch_size, height, width, num_channels = images.get_shape().as_list()
```
fchollet, line 58: This is incorrect, it should use `tf.shape`

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if unbatched:
+            images = tf.squeeze(images, axis=0)
+
+        return images
+
+    def call(self, images, training=True):
+        """call method for the ChannelShuffle layer.
+
+        Args:
```
fchollet, line 78: Remove this docstring and add a `Call arguments:` section in the layer docstring.

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+            images: Tensor representing images of shape
+                [batch_size, width, height, channels], with dtype tf.float32 / tf.uint8, or,
+                [width, height, channels], with dtype tf.float32 / tf.uint8
+        Returns:
+            images: augmented images, same shape as input.
+        """
+        if training is None:
+            training = backend.learning_phase()
```
fchollet, line 86: Remove this block, `training` should be a Python bool

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+
+        return tf.cond(
```
fchollet, line 88: This should just be a Python `if`.

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+            lambda: images,
+        )
+
+    def get_config(self):
+        config = {
+            "groups": self.groups, 
+            "seed": self.seed,
+        }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
```
fchollet, line 100: We should be consistent in how we do dictionary update across our layers. In other layers in the project, we normally do `config = super().get_config()` followed by `config.update(...)`

fchollet (commented):
Thanks for the updates!

keras_cv/layers/preprocessing/channel_shuffle.py
LukeWood: Yep!  we don't need the if training is None block anymore.

you can just use training as if it is a python bool.  it will never be None.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: @fchollet 

With this change, one test case failed. 

```python
batch_size = tf.shape(images)[0]
height = tf.shape(images)[1]
width = tf.shape(images)[2]
num_channels = tf.shape(images)[3]

if not num_channels % self.groups == 0:
    raise ValueError("The number of input channels should be divisible by the number of groups."
                             f"Received: channels={num_channels}, groups={self.groups}")
```

```python
def test_in_tf_function(self):
   xs = tf.cast(
       tf.stack([2 * tf.ones((100, 100, 1)), tf.ones((100, 100, 1))], axis=0),
       dtype=tf.float32,
     )

layer = ChannelShuffle(groups=1)

@tf.function
def augment(x):
   return layer(x, training=True)

xs = augment(xs)
```

> ValueError: The number of input channels should be divisible by the number of groups.Received: channels=Tensor("channel_shuffle/strided_slice_3:0", shape=(), dtype=int32), groups=1

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _channel_shuffling(self, images):
+        unbatched = images.shape.rank == 3
+        
+        if unbatched:
+            images = tf.expand_dims(images, axis=0)
+
+        batch_size, height, width, num_channels = images.get_shape().as_list()
+        if not num_channels % self.groups == 0:
+            raise ValueError("input channels should be divisible by the number of group. "
```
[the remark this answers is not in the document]
innat: Done.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: Done.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: Done.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: Done.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: Done.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: Done.

keras_cv/layers/preprocessing/channel_shuffle.py
innat: Done.

keras_cv/layers/preprocessing/channel_shuffle.py
fchollet: For the channel number you can actually use `images.shape[3]` since the channel dimension will always be static. However it's important to use `tf.shape` for image dimensions, since they will usually be None for width/height.

innat: Fixed.

fchollet (approved):
LGTM, thanks

Issue #117, opened 2022-02-09
RGBShift preprocessing layer 


innat:
ref: https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.RGBShift

PR #128, opened 2022-02-10
add RGBShift preprocessing layers

innat:
Make-up https://github.com/keras-team/keras-cv/issues/117

demo 1: 

```
RGBShift(r_shift_limit=(40, 60), g_shift_limit=[40, 50], b_shift_limit=50)
```


innat:
demo 2:

download

innat:
[Concern]
- Between `tf.unstack` and `tf.split` (etc), are all these computationally same about their time complexity? 
- No support for `tf.uint8` yet (to-do). Suggestion welcomed. 
- How to check if a loaded image is an RGB or bgr format in tf efficiently! In [python](https://stackoverflow.com/a/30802605/9215780), like `image.mode` in PIL.

bhack:
> Between tf.unstack and tf.split (etc), are all these computationally same about their time complexity?

`unstack` is  `unpack`

>   // Except for shape, unpack is a special case of split, so we reuse the
    // same computational kernels.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/unpack_op.cc#L97-L98

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,144 @@
...
+        g_shift = tf.random.uniform([], 
+                                    minval=self._g_shift_limit[0], 
+                                    maxval=self._g_shift_limit[1], 
+                                    dtype=tf.float32, seed=self.seed)
+        b_shift = tf.random.uniform([], 
+                                    minval=self._b_shift_limit[0],
+                                    maxval=self._b_shift_limit[1], 
+                                    dtype=tf.float32, seed=self.seed)
+        
+        unstack_rgb = tf.unstack(image, axis=-1)
```
qlzh727, line 94: I don't think you need to unstack the image and apply the diff, as long as the diff is broadcast-able to the image, you can directly add the diff to the image (which will be faster).

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,144 @@
...
+        shifted_rgb = tf.stack([unstack_rgb[0] + r_shift,
```
qlzh727, line 95: Could u add a test for int input here? I guess the + will fail if the input is int and the shift is float.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,144 @@
...
+            images: Tensor representing images of shape
+                [batch_size, width, height, channels], with dtype tf.float32 / tf.uint8, or,
+                [width, height, channels], with dtype tf.float32 / tf.uint8
+        Returns:
+            images: augmented images, same shape as input.
+        """
+        if training is None:
+            training = backend.learning_phase()
+
+        if not training:
```
qlzh727, line 114: Please use the tf.cond here since the `training` value could be a tensor, which will fail with the if check.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,144 @@
...
+            # The transform op only accepts rank 4 inputs, so if we have an unbatched
+            # image, we need to temporarily expand dims to a batch.
+            if unbatched:
+                images = tf.expand_dims(images, axis=0)
+
+            # TODO: Check if input tensor is RGB or not. 
+            # TODO: Support tf.uint8
+
+            # TODO: Make the batch operation vectorize.
+            output = tf.map_fn(lambda image: self._rgb_shifting(image), images)
```
qlzh727, line 127: I think we should fix the batch verctorization within this PR. Eg you can generate the shift based on teh input rank. Use tf.map will have a performance implication.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,144 @@
...
+        if not training:
+            return images
+        else:
+            unbatched = images.shape.rank == 3
+            # The transform op only accepts rank 4 inputs, so if we have an unbatched
+            # image, we need to temporarily expand dims to a batch.
+            if unbatched:
+                images = tf.expand_dims(images, axis=0)
+
+            # TODO: Check if input tensor is RGB or not.
```
qlzh727, line 123: I think we will always assume the input is RGB with [0, 255].

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,144 @@
...
+                    bound number. Got {shift_limit}')
+            return sorted(shift_limit)
+        elif isinstance(shift_limit, (int, float)):
+            shift_limit = abs(shift_limit)
+            return [-shift_limit, shift_limit]
+        else:
+            raise ValueError(f'The {channel} should be scalar, tuple or list of two upper and lower bound\
+                 number. Got {shift_limit}')
+
+    @tf.function
```
qlzh727, line 79: I don't think tf.function is necessary here, unless this is strictly needed.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,144 @@
...
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+    Output shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        r_shift_limit: A scalar or tuple or list of two upper and lower bound number. It will change r-channel by
+            adding with the r-channel. If r_shift_limit is a single integer, the range will be (-r_shift_limit, r_shift_limit).
+            Default: (-10, 10).
```
qlzh727, line 35: we might want to just call it r_shift.

qlzh727, line 35: Btw, in https://github.com/keras-team/keras-cv/pull/122, I was using a scale value [-1.0, 1.0] as the input, which hint the limits here. It is bit unclear in this PR that what's the allowed range for those limits (I guess its -255 ~ 255)?

keras_cv/layers/preprocessing/rgb_shift_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+        layer = RGBShift(
+            r_shift_limit=40,
+            g_shift_limit=30,
+            b_shift_limit=20
+        )
+
+        xs = layer(xs, training=True)
+        self.assertFalse(tf.math.reduce_any(xs == 1.0))
+
```
qlzh727, line 86: Could u add a test for batched inputs?

qlzh727 (changes requested):
Thanks for the PR.

qlzh727:
Thinking about this more. This PR is producing similar math result as https://github.com/keras-team/keras-cv/pull/122. The only difference here is that your PR allow user to control the range for each of the channel, whereas #122 only allow user to control 3 channel in one range.

We might want to just share most of the implementation details, to avoid the duplications.

innat:
@qlzh727 The RGBShift as an augmentation layer is supported in [albumentation](https://github.com/albumentations-team/albumentations#albumentations), which is still mostly used in practice. And I think the RGBShift and Brightness, both are mutual approaches. 

Now, after seeing the demo result here https://github.com/keras-team/keras-cv/pull/122, some output looks like almost RGBshited rather than Brightness preprocessing, for example, first two examples row-wise. Between them, the first example indeed looks like RGBshifted whereas the second example looks like the result of Brightness adjusted (darker). 

image

keras_cv/layers/preprocessing/rgb_shift_test.py
innat: I think this is the only single input test function. Besides this, all tests take batched input.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Yes, I agree. But I wanted to set a checker if the input is RGB or not.  I tried something as follows but for some reason, it wasn't working. I'll check later. 

```python

def is_rgb_image(image):
    return tf.equal(len(image.shape), 3) and tf.equal(image.shape[-1], 3)

if not is_rgb_image(image):
      raise TypeError(....)
```

keras_cv/layers/preprocessing/rgb_shift.py
innat: Yes, as long as the diff are same (`r_shift_limit == g_shift_limit == b_shift_limit`), we don't need to unpack/split the channel-wise. But if they aren't the same value, don't we need to unpack here?

qlzh727:
> 

Thanks for pointing this out. I am fixing it in https://github.com/keras-team/keras-cv/pull/122/commits/e385c3fc3ae082d4bf939a12a695f10debbd061c. This actually means that those two layer is just one line different in the implementation detail.

innat:
> This actually means that those two layer is just one line different in the implementation detail.

Agree. So, WDYT? If it's ok to have it, I'll send an update according to the above your review.

qlzh727:
> > This actually means that those two layer is just one line different in the implementation detail.
> 
> Agree. So, WDYT? If it's ok to have it, I'll send an update according to the above your review.

There are more discussion on #122 about where should it live. Since this PR and #122 will probably live together, let's wait for #122 to be finalized, and then move forward.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,144 @@
...
+   
+    Input shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+    Output shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        r_shift_limit: A scalar or tuple or list of two upper and lower bound number. It will change r-channel by
```
fchollet, line 33: - Is it really worth splitting the shift argument into 3? Most use cases will want to use the same value for all channels, and having to respecify it 3 times add significant UX friction.
- This type of argument is normally called something like `factor` is other preprocessing layers. We should align on existing naming covnentions. https://keras.io/api/layers/preprocessing_layers/image_augmentation/
- Likewise we should align on the same semantics as with other preprocessing layers: an integer is used as a level number, a float is used as a fraction of the total (e.g. `0.1` is interpreted as "shift within a +-10% range")

I'd recommend a single argument `factor` with the semantics above. To cover the use case of per-channel factors (which is rare), we could simply add a `target_channel=None` argument so you can specify which channel(s) the shift should apply to (default to all). What do you think?

fchollet (commented):
Thanks for the PR!

keras_cv/layers/preprocessing/rgb_shift.py
innat: @fchollet 
Regarding the shift value to the RGB channel, I don't have empirical evidence to use shift values independently to all channels, but I think it's nice to have an option.  Also, independent scaling to all channels looks quite different or randomized than the original channel whereas single scaling to all channels more and more looks like a brightness adjusted. That's why I think it's fine to have an option for independent channel shifting.  So, using `target_channel`, we can prove an option to the end-user to pick which one suits their goals. Like, 

```python
(pseudo)

if target_channel == 0:  # R channel 
    # only R channel will be shifted and concat with others 

if target_channel == 1:  # G channel 
    # only G and concat with others 

if target_channel == 2:  # B channel 
    # only B and concat with others 

if target_channel == "RGB":
    # shit independently to all channel and concat all 
```

WDYT?

innat: @fchollet 

Regarding the `factor` param, are you suggesting the following setup? 

```python

import tensorflow as tf

factor = (-0.6, 0.6)
shift_limit = tf.random.uniform([], 
                                minval=factor[0],
                                maxval=factor[1], 
                                dtype=tf.float32)*255.0
shift_limit
<tf.Tensor: shape=(), dtype=float32, numpy=-7.2357845>
```

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

innat: fchollet suggested using the `factor` term, like in other augmentation, [here](https://github.com/keras-team/keras-cv/pull/128/files#r806313141). So, for r g b input, should it be 

```python
r_factor, g_factor, b_factor
```

keras_cv/layers/preprocessing/rgb_shift.py
fchollet: Upon reflection I'd recommend dropping the argument altogether -- just keep `factor` -- and apply the random to all channels (a different random shift value for each channel).

innat:
> Upon reflection I'd recommend dropping the argument altogether -- just keep factor -- and apply the random to all channels (a different random shift value for each channel).

@fchollet ok. Before pushing an update, here is some sample look of this change, compared to the [above](https://github.com/keras-team/keras-cv/pull/128#issuecomment-1042858855) output. 

```
RGBShift(factor=(10, 90))
```
r

---

```
RGBShift(factor=50)
```
g

---

```
RGBShift(factor=20)
```
b

innat:
@fchollet Eventually I've realized that if we pick a single `factor` to rescale all channels (R G B), it would be more like the [RandomBrightness, line: 125-128](https://github.com/keras-team/keras-cv/pull/122/files) implementation. @qlzh727 pointed this out earlier, [here](https://github.com/keras-team/keras-cv/pull/128#issuecomment-1035578763).

innat:
@fchollet @qlzh727 Could you please give some feedback on the above queries/issues?

fchollet:
> Eventually I've realized that if we pick a single factor to rescale all channels (R G B), it would be more like the [RandomBrightness](https://github.com/keras-team/keras-cv/pull/122/files)

See my earlier comment:

> Upon reflection I'd recommend dropping the argument altogether -- just keep `factor` -- and apply the random to all channels (a different random shift value for each channel).

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Fixed.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        factor: A scalar or tuple or list of two upper and lower bound number. If factor is a single value, 
+                the range will be (-factor, factor). The factor value can be float or integer; for float the 
+                valid limits are (-1.0, 1.0) and for integer the valid limits are (-255, 255).
+        seed: Integer. Used to create a random seed. Default: None.
+
+    Call arguments: call method for the RGBShift layer.
```
fchollet, line 37: remove "call method for the RGBShift layer."

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+    Input shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+    Output shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        factor: A scalar or tuple or list of two upper and lower bound number. If factor is a single value, 
+                the range will be (-factor, factor). The factor value can be float or integer; for float the
```
fchollet, line 33: Use backticks around code keywords (e.g. tuple above)

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+    Args:
+        factor: A scalar or tuple or list of two upper and lower bound number. If factor is a single value, 
+                the range will be (-factor, factor). The factor value can be float or integer; for float the 
+                valid limits are (-1.0, 1.0) and for integer the valid limits are (-255, 255).
+        seed: Integer. Used to create a random seed. Default: None.
+
+    Call arguments: call method for the RGBShift layer.
+        Args:
+            images: Tensor representing images of shape
+                [batch_size, width, height, channels], with dtype tf.float32 / tf.uint8, or,
```
fchollet, line 40: For shapes, use tuples, and use backticks around them

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+                [width, height, channels], with dtype tf.float32 / tf.uint8
+        Returns:
+            images: augmented images, same shape as input.
+   
+    Usage:
+    ```python
+    (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    rgbshift = keras_cv.layers.RGBShift(factor=(-2, 2))
```
fchollet, line 48: rgb_shift

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+    augmented_images = rgbshift(images)
```
fchollet, line 49: rgb_shift

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.factor = self._set_shift_limit(factor)
+        self.seed = seed
+
+    def _set_shift_limit(self, factor):
+        if isinstance(factor, (tuple, list)):
+            if len(factor) != 2: 
+                raise ValueError(f'The factor should be scalar, tuple or list of two upper and lower \
```
fchollet, line 66: Don't use `\`. Instead, use string concatenation, e.g. 
```
raise ValueError(
    'The factor should be a scalar, '
    'a tuple or a list of two upper and lower '
    f'bound values. Received: factor={factor}')
```

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+                    bound number. Got {factor}')
+            return self._check_factor_range(sorted(factor))
+        elif isinstance(factor, (int, float)):
+            factor = abs(factor)
+            return self._check_factor_range([-factor, factor])
+        else:
+            raise ValueError(f'The factor should be scalar, tuple or list of two upper and lower bound\
```
fchollet, line 73: Also fix this following the example in my comment above

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+                 number. Got {factor}')
+        
+    @staticmethod
+    def _check_factor_range(factor):
+        if all(isinstance(each_elem, float) for each_elem in factor):
+            if factor[0] < -1.0 or factor[1] > 1.0:
+                raise ValueError(f"Got {factor}")
```
fchollet, line 80: Properly format the error message

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+            return factor
+        elif all(isinstance(each_elem, int) for each_elem in factor):
+            if factor[0] < -255 or factor[1] > 255:
+                raise ValueError(f"Got {factor}")
```
fchollet, line 84: Properly format the error message

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,155 @@
...
+            return factor
+        else:
+            raise ValueError(f'Both bound must be same dtype. Got {factor}')
```
fchollet, line 87:
```
raise ValueError(
    'Both lower/upper bound values must have the same dtype. '
    f'Received: factor={factor} where type(factor[0]) is {type(factor[0])} '
    f'and type(factor[1]) is {type(factor[1])}')
```

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,158 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np 
+import tensorflow as tf
+from tensorflow.keras import layers, backend
+
+
+class RGBShift(layers.Layer):
+    """RGBShift class randomly shift values for each channel of the input RGB image. The expected images
```
fchollet, line 21: Docstring should start with a single line shorter than 80 char, ending with a period. You can add more info in a paragraph below.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Input shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+    Output shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        factor: A scalar or tuple or list of two upper and lower bound number. If factor is a single value,
```
fchollet, line 32: Please shorten lines to be <80 chars

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,158 @@
...
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        factor: A scalar or tuple or list of two upper and lower bound number. If factor is a single value, 
+                the range will be `(-factor, factor)`. The factor value can be float or integer; for float the 
+                valid limits are `(-1.0, 1.0)` and for integer the valid limits are `(-255, 255)`.
+        seed: Integer. Used to create a random seed. Default: None.
+
+    Call arguments: 
+        Args:
```
fchollet, line 38: No need for "Args" here, nor for "Returns" section below. Just list the arguments after "Call arguments"

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,158 @@
...
+            images: Tensor representing images of shape
```
fchollet, line 39: Missing the `training` arg

fchollet (approved):
Thanks for the update!

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,160 @@
...
+class RGBShift(layers.Layer):
+    """RGBShift class randomly shift values for each channel of the 
+    input RGB image. The expected images should be `(0-255)` pixel ranges.
+
+    Input shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `channels_last` format
+    Output shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `channels_last` format
```
innat, line 29: @fchollet Should I remove lines 24-29 in a favor of `Call argument` docstring in lines 38-44.

keras_cv/layers/preprocessing/rgb_shift.py
bhack: Pre-commit hooks helps

keras_cv/layers/preprocessing/rgb_shift.py
bhack: Ditto

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

innat:
@fchollet a gentle reminder on this and https://github.com/keras-team/keras-cv/pull/124 PR.

keras_cv/layers/preprocessing/rgb_shift.py
fchollet: These lines are fine, we can keep them.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class RGBShift(layers.Layer):
+    """RGBShift class randomly shift values for each channel of the 
+    input RGB image. The expected images should be `(0-255)` pixel ranges.
+
+    Input shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `channels_last` format
+    Output shape:
```
fchollet, line 26: Add line break above

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,164 @@
...
+    def __init__(self, factor, seed=None, **kwargs):
+        super().__init__(**kwargs)
+        self.factor = self._set_factor_limit(factor)
+        self.seed = seed
+
+    def _set_factor_limit(self, factor):
+        if isinstance(factor, (tuple, list)):
+            if len(factor) != 2:
+                raise ValueError(
+                    self._FACTOR_VALIDATION_ERROR + f"Received: factor={factor}"
```
fchollet, line 68: You're missing a space after "integer." and before "Received:"

This applies to all locations of this error message.

fchollet (approved):
LGTM, thank you! Some minor comments.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Didn't get the integer part. Is it:

```
if len(factor) != 2 : spacing after int ??
```

fchollet: Replace
```
self._FACTOR_VALIDATION_ERROR + f"Received: factor={factor}"
```
with
```
self._FACTOR_VALIDATION_ERROR + f" Received: factor={factor}"
```

At every location this line appears.

innat: Uh, I've already did that. I was confused a bit when you said ' after  "integer" ' .  Placing a space before 'Received' was done in line 69, 77, 84, 90. And in line 96, a space was set in previous line. So, I think it's ok.

innat:
@fchollet I think most of the last reviews are addressed. Could you please check?

LukeWood:
> @fchollet I think most of the last reviews are addressed. Could you please check?

Hey @innat can you "resolve" the conversations for commments that are resolved

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,165 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+class RGBShift(layers.Layer):
+    """RGBShift class randomly shift values for each channel of the
```
fchollet, line 20: The header description should fit in one line.

```
"""Randomly shift values for each channel of the input image(s).

The input images should have values in the `[0-255]` range.
```

fchollet (approved):
LGTM, sorry for the review delay

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

innat:
@fchollet 
Let me know if there's anything to modify. Thanks.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,165 @@
...
+        else:
+            _rand_uniform = tf.random.uniform(
+                rgb_delta_shape,
+                minval=factor_limit[0],
+                maxval=factor_limit[1],
+                dtype=tf.float32,
+            )
+
+        if all(isinstance(each_elem, float) for each_elem in factor_limit):
+            _rand_uniform = _rand_uniform * 85.0
```
fchollet, line 120: I hadn't caught that initially, but why 85? Why not 255?

innat: When the input factor is a **float** of ranges `[-1.0, 1.0]`, each `_rand_uniform` of r, g, and b channel get scaled up by **1/3th** of `255.0` (or, `85.0`). 

I think it's needed as the expected input ranges are `[0-255]`. For **integer** factor of `[0, 255]` ranges, the `_rand_uniform` is not required to scale up.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,168 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+class RGBShift(layers.Layer):
```
LukeWood, line 19: Can we subclass `BaseImageAugmentationLayer`?

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,168 @@
...
+    """Randomly shift values for each channel of the input image(s).
+
+    The input images should have values in the `[0-255]` range.
```
LukeWood, line 22: lets add a value_range parameter like the other color based layers now have

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,168 @@
...
+    (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    rgb_shift = keras_cv.layers.RGBShift(factor=(-2, 2))
+    augmented_images = rgb_shift(images)
+    ```
+    """
+
+    _FACTOR_VALIDATION_ERROR = (
+        "The factor should be a scalar, "
+        "a tuple or a list of two upper and lower "
+        "bound values in the range `(-1.0, 1.0)` as float or `(-255, 255) as integer."
```
LukeWood, line 60: lets make factor scaled to the range: `[0, 1]` and perform a transformation to map to `[-255, 255]` internally.  Then, we can use `utils.preprocessing.transform_value_range` to always operate in the [0, 255] range.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+
+    def __init__(self, factor, seed=None, **kwargs):
+        super().__init__(**kwargs)
+        self.factor = self._set_factor_limit(factor)
```
LukeWood, line 65: Note that there are now utilities to parse factor ranges in keras_cv.utils.preprocessing

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,168 @@
...
+            return factor
+        else:
+            raise ValueError(
+                "Both lower/upper bound values must have the same dtype. "
+                f"Received: factor={factor} where type(factor[0]) is {type(factor[0])} "
+                f"and type(factor[1]) is {type(factor[1])}"
+            )
+
+    def _get_random_uniform(self, factor_limit, rgb_delta_shape):
+        if self.seed is not None:
```
LukeWood, line 104: Lets rely on `BaseImageAugmentationLayer.get_random_transformation` - check out RandomShear as a canonical example.

keras_cv/layers/preprocessing/rgb_shift.py
LukeWood: lets always assume [0, 255] as the value range, with an optional `value_range` parameter that performs a transformation on the images at the start of the `augment_image` call.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,168 @@
...
+        else:
+            return images
+
+    def get_config(self):
+        config = super().get_config()
+        config.update({"factor": self.factor, "seed": self.seed})
+        return config
+
+    def compute_output_shape(self, input_shape):
+        return input_shape
```
LukeWood, line 168: compute_output_shape() isn't needed anymore.

LukeWood:
Thanks for the PR @innat - this is a really useful layer.  Appreciate the contribution!  I left some minor comments.

innat:
@LukeWood Thank you for the feedback. Regarding your suggestions, is it possible to address those in new PR? I think the current PR was quite ready to get merged.

LukeWood:
> @LukeWood Thank you for the feedback. Regarding your suggestions, is it possible to address those in new PR? I think the current PR was quite ready to get merged.

The API will actually be different for the end layer - especially in the factor arguments.  I can help get it ready to merge if needed.   Sorry for the delays and shifting requirements - but I think it is best to get the api right the first time around.

LukeWood:
@innat let me know if you'll address the comments or if you want some help.  My time is sparing so I am focusing on landing v1 but I am happy to do it at some point if needed due to the shifting requirements.  Cheers.

innat:
@fchollet 
Some changes have been made to match up with the latest API design. Previously, it was supported that 


> - use can pass actual pixel value of (0, 255) ranges.
> - or, can pass factor values of [-1, 1] ranges. 

But with the new changes, 

> - user can pass a factor value of [0.0, 1.0].
> - and internally transform / convert to [-1, 1] ranges. 


demo o/p with new changes

```
RGBShift(factor=(0.3, 0.8))
```

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: I think it doesn't require that much here. It's assumed that all input would be in the range of `[0, 255]`.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
LukeWood: We can't assume all inputs in [0, 255].  We do need a value range to be consistent with the other ones.  Otherwise all other KPLs that rely on color values will support value_range=(0, 1) and this one will silently fail for users in that case.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    def __init__(self, factor, **kwargs):
+        super().__init__(**kwargs)
+        self.factor = preprocessing.parse_factor_value_range(
+            factor, min_value=0.0, max_value=1.0
+        )
+
+    def get_random_transformation(self, image=None, label=None, bounding_box=None):
+        r_shift_rand_uniform = self._random_generator.random_uniform(
```
LukeWood, line 57: These can just be named `r_shift`, `g_shift`, and `b_shift`

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,94 @@
...
+        b_shift_rand_uniform = self._random_generator.random_uniform(
+            shape=(), minval=self.factor[0], maxval=self.factor[1], dtype=tf.float32
+        )
+
+        return [r_shift_rand_uniform, g_shift_rand_uniform, b_shift_rand_uniform]
+
+    def augment_image(self, image, transformation=None):
+        # Convert sampled value from [0.0, 1.0] ranges to [-1.0, 1.0] ranges.
+        transformation = [
+            (each_transformation * 2.0 - 1.0) for each_transformation in transformation
```
LukeWood, line 74: can rename the `each_transformation` to `transform`

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,94 @@
...
+        ]
+
+        image = preprocessing.transform_value_range(image, [0, 255], [0.0, 1.0])
```
LukeWood, line 77: yeah, with this call we do need to support the user provided `value_range`

keras_cv/layers/preprocessing/rgb_shift_test.py
```diff
@@ -0,0 +1,84 @@
...
+                [2 * tf.ones((1024, 512, 3)), tf.ones((1024, 512, 3))],
+                axis=0,
+            ),
+            dtype=tf.float32,
+        )
+
+        layer = RGBShift(factor=[0.1, 0.3])
+
+        xs = layer(xs, training=True)
+        self.assertFalse(tf.math.reduce_any(xs[0] == 2.0))
```
LukeWood, line 42: Is this consistently passing?  If you're using reduce_any xs[0] and it jitters all 3 channels wont this fail?

keras_cv/layers/preprocessing/rgb_shift_test.py
```diff
@@ -0,0 +1,84 @@
...
+# limitations under the License.
+
+import numpy as np
+import tensorflow as tf
+
+from keras_cv.layers.preprocessing.rgb_shift import RGBShift
+
+
+class RGBShiftTest(tf.test.TestCase):
+    def test_return_shapes(self):
```
LukeWood, line 22: lets implement a value test that actually inspects the return values.

This can be done by passing `factor=(0.5, 0.5)` which will then make get_random_transformation always return 0.5 for the transforms, so you can know what it will do.  Try this for 0.5, maybe 0.7

keras_cv/layers/preprocessing/rgb_shift.py
LukeWood, line 74: I don't think we want to sample this way, because this means that 0.5=>0 which is not idiomatic to the meaning of `factor`.  Also, 0.0 => -1 which is the strongest transformation.

What you want to do here is also sample a -1 or 1 for all 3 channels in get_random_transformation.

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+
+        g_shift_rand_uniform = self._random_generator.random_uniform(
+            shape=(), minval=self.factor[0], maxval=self.factor[1], dtype=tf.float32
+        )
+
+        b_shift_rand_uniform = self._random_generator.random_uniform(
+            shape=(), minval=self.factor[0], maxval=self.factor[1], dtype=tf.float32
+        )
+
```
LukeWood, line 68: so, you will need to also sample a -1 or 1 here in addition to the raw value.

LukeWood (changes requested):

keras_cv/layers/preprocessing/rgb_shift.py
innat: Regarding this transformation, isn't it the same as [this](https://github.com/keras-team/keras-cv/blob/ca6eaa83f31ca55d8cab6b20310d408f19d0bd5f/keras_cv/layers/preprocessing/random_hue.py#L58-L60)?

keras_cv/layers/preprocessing/rgb_shift_test.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift_test.py
innat: Can you describe it a little more? I did run pytest, all ok though. It should not cause any issue:

```python

xs = 2 * tf.ones((1024, 1024, 3)) 

layer = RGBShift( .... )

xs = layer(xs) 

self.assertFalse(tf.math.reduce_any(xs == 2.0))

```

keras_cv/layers/preprocessing/rgb_shift.py
innat: Assuming that it's resolved.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Assuming that it's resolved.

keras_cv/layers/preprocessing/rgb_shift.py
LukeWood: I've opened a PR to fix RandomHue - that one was also incorrect.

LukeWood: Check out

https://github.com/keras-team/keras-cv/pull/269

You can use the utility in there to update it.

innat: okay, it seems one of an issue raised [here](https://github.com/keras-team/keras-cv/pull/158#issuecomment-1084443373), is also sovled. cool.

LukeWood: Definitely, yeah that's good to know.  Thanks, let me know when this is updated.

In general:

users should be able to logically reason about factors.  A factor range should be [0, 1] if possible, this means that 1.0 is the strongest augmentation and 0.0 is the weakest.

LukeWood: Thanks for your persistence on this issue @innat It will be a great contribution!

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,99 @@
...
+    (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    rgb_shift = keras_cv.layers.RGBShift(factor=(0.3, 0.8), value_range=(0, 255))
+    augmented_images = rgb_shift(images)
+    ```
+    """
+
+    def __init__(self, factor, value_range, **kwargs):
+        super().__init__(**kwargs)
+        self.value_range = value_range
+        self.factor = preprocessing.parse_factor_value_range(
```
LukeWood, line 57: if you rebase on origin/master you can use the `Factor` API.  Check the other classes like shear

keras_cv/layers/preprocessing/rgb_shift.py
```diff
@@ -0,0 +1,99 @@
...
+            factor, min_value=0.0, max_value=1.0
+        )
+
+    def get_random_transformation(self, image=None, label=None, bounding_box=None):
+        r_shift = self._random_generator.random_uniform(
```
LukeWood, line 62: so these will be replaced with self.factor.sample()

LukeWood (changes requested):

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

keras_cv/layers/preprocessing/rgb_shift.py
innat: Done.

innat:
@LukeWood I've updated the rgb-shitf layers based on the previous review. But `format.sh` causes issues on `lint.sh` and thus affected elsewhere.

LukeWood:
> @LukeWood I've updated the rgb-shitf layers based on the previous review. But `format.sh` causes issues on `lint.sh` and thus affected elsewhere.

I will do a pass on Monday!  Thank you for your patience as the API has evolved.  I appreciate your understanding and effort!

cheers!

innat:
> @LukeWood I've updated the rgb-shitf layers based on the previous review. But `format.sh` causes issues on `lint.sh` and thus affected elsewhere.

It was due to `black == 19.10b0`. Upgrading to `22.3.0` solves the issue.

LukeWood:
@innat I have made some edits! Largest one is supporting an arbitrary # of channels. Thank you for your wonderful contribution.

keras_cv/layers/preprocessing/random_channel_shift.py
```diff
@@ -0,0 +1,102 @@
...
+        invert = preprocessing.random_inversion(self._random_generator)
+        return invert * self.factor() * 0.5
+
+    def augment_image(self, image, transformation=None):
+        image = preprocessing.transform_value_range(image, self.value_range, (0, 1))
+        unstack_rgb = tf.unstack(image, axis=-1)
+
+        result = []
+        for c_i in range(self.channels):
+            result.append(unstack_rgb[c_i] + transformation[c_i])
```
innat, line 79: @LukeWood not sure, but isn't it safe to use `tf.add` instead of `+`? I also used `+` at the beginning but qlzh727 discouraged to use it.

LukeWood: it will only fail when image is an int, shift is float.  BaseImageAugmentationLayer casts images to _compute_dtype so this is fine.

LukeWood:
congratulations @innat!  Thanks for the PR.

And thank you for your continued patience and perseverence on this.

Issue #84, opened 2022-02-02
Posterize preprocessing layer


LukeWood:
https://github.com/tensorflow/similarity/blob/906d141b75acf00d0e1d37dd0d88c432bbb30f57/tensorflow_similarity/augmenters/img_augments.py#L494

sebastian-sz:
I think PIL implementation assumes uint8 dtype, and TF's `bitwise` ops only support integers. 

Is it safe to assume that this layer's input will only accept integer values in range 0-255? Or should it accept any dtype, cast internally to uint8, posterize and cast back to original dtype?

LukeWood:
Thanks so much @sebastian-sz

PR #136, opened 2022-02-15
Introduces Posterization preprocessing layer.

sebastian-sz:
Posterization is one of the operations used by AutoAugment and RandAugment.

[Linked Issue](https://github.com/keras-team/keras-cv/issues/84)

### DIFF
--- a/keras_cv/layers/preprocessing/posterization.py
+++ b/keras_cv/layers/preprocessing/posterization.py
@@ -49,7 +49,10 @@ class Posterization(tf.keras.layers.Layer):
 
     def __init__(self, bits: int):
         super().__init__()
-        assert 0 < bits < 9, f"Bits value must be between 1-8. Received bits: {bits}."
+
+        if not (0 < bits < 9):
+            raise ValueError(f"Bits value must be between 1-8. Received bits: {bits}.")
+
         self.shift = 8 - bits
 
     def call(self, images):
--- a/keras_cv/layers/preprocessing/posterization.py
+++ b/keras_cv/layers/preprocessing/posterization.py
@@ -12,9 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import tensorflow as tf
+from tensorflow.keras.__internal__.layers import BaseImageAugmentationLayer
+
+from keras_cv.utils.preprocessing import transform_value_range
 
 
-class Posterization(tf.keras.layers.Layer):
+class Posterization(BaseImageAugmentationLayer):
     """Reduces the number of bits for each color channel.
 
     References:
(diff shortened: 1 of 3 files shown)


keras_cv/layers/preprocessing/posterization.py
```diff
@@ -0,0 +1,66 @@
...
+    ```
+
+     Call arguments:
+        images: Tensor with pixels in range [0, 255] and shape
+            [batch, height, width, channels] or [height, width, channels].
+    """
+
+    def __init__(self, bits: int):
+        super().__init__()
+        assert 0 < bits < 9, f"Bits value must be between 1-8. Received bits: {bits}."
```
bhack, line 52: Generally we `raise`:
https://github.com/keras-team/keras-cv/blob/58b298c411c1eb67b3bd1b244b789908673f1b49/keras_cv/layers/preprocessing/grid_mask.py#L100-L101

keras_cv/layers/preprocessing/posterization_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class PosterizationTest(tf.test.TestCase):
+    rng = tf.random.Generator.from_seed(1234)
+
+    def test_raises_error_on_invalid_bits_parameter(self):
+        invalid_values = [-1, 0, 9, 24]
+
+        for value in invalid_values:
+            with self.assertRaises(AssertionError):
```
bhack, line 26: Ditto

keras_cv/layers/preprocessing/posterization_test.py
```diff
@@ -0,0 +1,45 @@
...
+                Posterization(bits=value)
+
+    def test_output_shape_unchanged(self):
```
bhack, line 29: @LukeWood We are writing this quite the same in every class. Is there any refactor solution?

keras_cv/layers/preprocessing/posterization_test.py
```diff
@@ -0,0 +1,45 @@
...
+        dummy_input = self.rng.uniform(shape=(2, 224, 224, 3), minval=0, maxval=256)
+        layer = Posterization(bits=4)
+
+        output = layer(dummy_input)
+
+        self.assertEqual(output.shape, dummy_input.shape)
+
+    def test_output_dtype_unchanged(self):
```
bhack, line 37: Probably the same also on this one.

keras_cv/layers/preprocessing/posterization_test.py
LukeWood: we are working on exporting a parameterized `layer_test` method that extends `tf.TestCase` from keras core.  This will cover most of the generic behaviors.

keras_cv/layers/preprocessing/posterization.py
sebastian-sz: Yes, you are right. Changed to `ValueError`.

keras_cv/layers/preprocessing/posterization.py
```diff
@@ -0,0 +1,69 @@
...
+        print(images[0, 0, 0])
+        # [48, 48, 48]
+    ```
+
+     Call arguments:
+        images: Tensor with pixels in range [0, 255] and shape
+            [batch, height, width, channels] or [height, width, channels].
+    """
+
+    def __init__(self, bits: int):
```
LukeWood, line 50: Hey there - please include `value_range` as an argument and scale images up to [0, 255] according to value_range, then scale back down to `value_range` before returning.  Thank.

LukeWood:
Thank you for the PR. Left a starting comment, will do a thorough review.  Please be aware we are doing some major design work around KPLs in KerasCV and the API may shift as a result.  Apologies if this delays reviews.

sebastian-sz:
> Thank you for the PR. Left a starting comment, will do a thorough review. Please be aware we are doing some major design work around KPLs in KerasCV and the API may shift as a result. Apologies if this delays reviews.

Hi, no worries. Let me know if you're done with refactoring and design changes, I will update this PR accordingly.

sebastian-sz:
@LukeWood I wanted to update the layer to subclass `BaseImageAugmentationLayer`, but it's not available in TF 2.8. Should I wait for 2.9?

bhack:
> @LukeWood I wanted to update the layer to subclass `BaseImageAugmentationLayer`, but it's not available in TF 2.8. Should I wait for 2.9?

It seems that we rely on Tensorflow stable in this repo but in Keras repo on master we track TF Nightly:
https://github.com/keras-team/keras-cv/blob/master/setup.py#L27

LukeWood:
> > @LukeWood I wanted to update the layer to subclass `BaseImageAugmentationLayer`, but it's not available in TF 2.8. Should I wait for 2.9?
> 
> It seems that we rely on Tensorflow stable in this repo but in Keras repo on master we track TF Nightly: https://github.com/keras-team/keras-cv/blob/master/setup.py#L27

I'll update the repo to rely on nightly until 2.9.  We're in pre-release so this is fine for now.

LukeWood:
> BaseImageAugmentationLayer

Yes, sorry for the delayed response.

sebastian-sz:
@LukeWood I'm starting to use `BaseImageAugmentationLayer`. I noticed it force-casts all inputs to `tf.float32` - is this intended?

```python
import tensorflow as tf
from tensorflow.keras.__internal__.layers import BaseImageAugmentationLayer


class DummyLayer(BaseImageAugmentationLayer):
    def augment_image(self, image, transformation=None):
        return image 

layer = DummyLayer()
image = tf.random.uniform((1, 224, 224, 3),dtype=tf.int32)
output = layer(image)
print(output.dtype)  # tf.float32, but input was tf.int32
```

LukeWood:
The layer will cast your inputs to `self.compute_dtype`.  This is pretty standard for keras.

LukeWood:
Basically @sebastian-sz to ensure consistency across dtypes we will just always assume that the inputs are float, and make no assumptions about their value range.

LukeWood:
Hey @sebastian-sz a sample usage of BaseImageAugmentationLayer is available in https://github.com/keras-team/keras-cv/pull/180.

Feel free to base yours on this.  Upload when a copy is ready and I will review :)

sebastian-sz:
Thanks @LukeWood I'm almost done with updating the code but I'm still unhappy with how the tests look. 

This is in progress and will probably be done by the end of the week (probably sooner).

LukeWood:
awesome - yeah feel free to upload as you go.  I may begin lending a hand in some of these components as the list of components required for RandAugment shrinks.

sebastian-sz:
@LukeWood I updated the PR. Two questions:
1. What is the `tf.function` policy? Should the `augment_image` or any other function be decorated?
2. Is it ok to override `_batch_augment` as well if the implementation is already vectorized and does not need to rely on `vectorized_map / map_fn` ? See code for more details, if not I will remove this part.

keras_cv/layers/preprocessing/posterization.py
```diff
@@ -0,0 +1,101 @@
...
+            target_range=[0, 255],
+        )
+        image = tf.cast(image, tf.uint8)
+
+        image = self._posterize(image)
+
+        return transform_value_range(
+            images=image,
+            original_range=[0, 255],
+            target_range=self._value_range,
```
LukeWood, line 85: lets cast back to compute_dtype before `transform_value_range`.  That way we are certain it is the correct dtype.  transform_value_range can be skipped if value_range is [0, 255]

LukeWood:
> @LukeWood I updated the PR. Two questions:
> 
> 1. What is the `tf.function` policy? Should the `augment_image` or any other function be decorated?
> 2. Is it ok to override `_batch_augment` as well if the implementation is already vectorized and does not need to rely on `vectorized_map / map_fn` ? See code for more details, if not I will remove this part.

1.) no need for tf.function
2.) I think it is actually fine.  I will check with @qlzh727 too.

LukeWood (changes requested):

bhack:
> 1.) no need for tf.function

The model logic is wrapped by default with a `tf.function` until you will go to explicitly enforce the eager mode [like in model compilation](https://www.tensorflow.org/api_docs/python/tf/keras/Model#compile)

`jit_compile` could be different instead as we could need to limit the performance region to a specific function. See more at https://github.com/keras-team/keras-cv/issues/165

LukeWood:
Thanks a ton @sebastian-sz - great contribution!  Looks good to go!

Issue #26, opened 2022-01-03
GaussianBlur preprocessing layer


LukeWood:

bhack:
Se have both 2d and 3d:

https://github.com/tensorflow/models/blob/master/official/vision/beta/projects/video_ssl/ops/video_ssl_preprocess_ops.py#L248


And:
https://github.com/tensorflow/models/blob/master/official/vision/beta/projects/simclr/dataloaders/preprocess_ops.py#L233

LukeWood:
We can also use: https://github.com/google-research/simclr/blob/master/tf2/data_util.py#L326 if we want

bhack:
But If you need to support a 4d tensor with batch size you need to look at addons:

https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/filters.py#L232

artu1999:
Happy to look into this if no one has started working on it

LukeWood:
@artu1999 thanks for offering to take this on.  Please feel free.

we may also want to support:

- [MedianBlur](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.MedianBlur)
- [MotionBlur](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.MotionBlur)

but I consider these to be lower priority.  So focus on Gaussian first is recommended.

artu1999:
Awesome! After implementing GaussianBlur it should be fairly easy to implement the others as well. In terms of input args, I guess we want to keep it similar to this?

https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.GaussianBlur

LukeWood:
I'd like to match the mathematical definition of GaussianBlur.  If theirs matches that, then that is a great reference.  Thanks

bhack:
Gaussian Blur Is already available in TFA. If you want you could port/adapt from there.

PR #143, opened 2022-02-20
Gaussian Blur layer implementation

artu1999:
implemented Gaussian Blur preprocessing layer using the gaussian blur implemented in tensorflow additions ([Reference](https://www.tensorflow.org/addons/api_docs/python/tfa/image/gaussian_filter2d))

[Linked Issue](https://github.com/keras-team/keras-cv/issues/26)

`layer = GaussianBlur(kernel_size=(10,7), sigma=1)`

Figure_1

`layer = GaussianBlur(kernel_size=(10,7), sigma=2)`

Figure_2

### DIFF
--- a/keras_cv/layers/preprocessing/gaussian_blur.py
+++ b/keras_cv/layers/preprocessing/gaussian_blur.py
@@ -1,0 +1,81 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+import tensorflow_addons as tfa
+from tensorflow.keras import backend
+from tensorflow.keras import layers
+
+
+class GaussianBlur(layers.Layer):
+    """GaussianBlur is a preprocessing layer that applies Gaussian Blur to RGB and
+    greyscale images. Input images should have values in the range of [0,255]
+
+        Args:
+            kernel_size: An integer or tuple/list of 2 integers, specifying
+                height and weight of Gaussian kernel. If integer, this represents
+                both dimensions of a square kernel.
+            sigma: Float or tuple/list of 2 floats representing the standard deviation
+                used to calculate kernel. tuple/list can be used to apply different
+                standard deviations for height and weight.
+
+        Usage:
+        ```python
+        blur = GaussianBlur()
+
+        (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+        # Note that images are an int8 Tensor with values in the range [0, 255]
+        images = equalize(images)
+        ```
+
+        Call arguments:
+            images: Tensor of pixels in range [0, 255], in RGB or greyscale format.
+                Can be of type float or int.  Should be in NHWC format.
+    """
+
+    def __init__(self, kernel_size=5, sigma=1, **kwargs):
+        super().__init__(**kwargs)
+        self.kernel_size = kernel_size
+        self.sigma = sigma
+
+    def call(self, inputs, training=True):
+        """call method for the GaussianBlur layer.
+        Args:
+            images: Tensor representing images of shape
+                [batch_size, width, height, channels] or
+                [width, height, channels] with type float or int.
+                Pixel values should be in the range [0, 255]
+        Returns:
+            images: Blurred input images, same as input.
+        """
+        if training is None:
+            training = backend.learning_phase()
+
+        def _blur(image, kernel_size, sigma):
+            image = tfa.image.gaussian_filter2d(
+                image=image,
+                filter_shape=kernel_size,
+                sigma=sigma,
+                padding="CONSTANT",
+                constant_values=1,
+            )
+            return image
+
+        augment = lambda: _blur(inputs, self.kernel_size, self.sigma)
+        no_augment = lambda: inputs
+        return tf.cond(tf.cast(training, tf.bool), augment, no_augment)
+
+    def get_config(self):
+        config = {"kernel_size": self.kernel_size, "sigma": self.sigma}
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
--- a/keras_cv/layers/__init__.py
+++ b/keras_cv/layers/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from keras_cv.layers.preprocessing.cut_mix import CutMix
+from keras_cv.layers.preprocessing.gaussian_blur import GaussianBlur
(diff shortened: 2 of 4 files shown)


artu1999:
I have just realised the email I have for author wasn't matching the email on the CLA. Can I proceed with reverting the PR and resubmit it?

bhack:
@qlzh727 Do you know if we have an update from the TF team on https://github.com/tensorflow/tensorflow/issues/39050?

LukeWood:
FYI @artu1999 we'll need you to sign the CLA to contribute code.  Thanks

keras_cv/layers/preprocessing/gaussian_blur.py
```diff
@@ -0,0 +1,81 @@
...
+                [width, height, channels] with type float or int.
+                Pixel values should be in the range [0, 255]
+        Returns:
+            images: Blurred input images, same as input.
+        """
+        if training is None:
+            training = backend.learning_phase()
+
+        def _blur(image, kernel_size, sigma):
+            image = tfa.image.gaussian_filter2d(
```
LukeWood, line 65: we don't rely on TFA in this repo.  We will need to ship our own implementation.  Gaussian blurs can actually be implemented through convolution (if tf.image doesn't have anything we can use).   Perhaps check the SimCLR repo for a reference implementation

keras_cv/layers/preprocessing/gaussian_blur.py
```diff
@@ -0,0 +1,81 @@
...
+        """call method for the GaussianBlur layer.
+        Args:
+            images: Tensor representing images of shape
+                [batch_size, width, height, channels] or
+                [width, height, channels] with type float or int.
+                Pixel values should be in the range [0, 255]
+        Returns:
+            images: Blurred input images, same as input.
+        """
+        if training is None:
```
LukeWood, line 61: assume training is a python bool.

LukeWood, line 61: assume training is a python bool.

LukeWood:
Thanks for the PR . I've left some early comments.

artu1999:
@LukeWood I have realised I used the wrong author name, I will make sure to use the correct one for the next commit. In terms of the next steps:

- Is there any desire of making the Blur random? the implementation in simCLR already does that ([Reference](https://github.com/tensorflow/models/blob/master/official/vision/beta/projects/simclr/dataloaders/preprocess_ops.py#L233)). Examples where Random Blur is used:

https://arxiv.org/pdf/2009.00104.pdf

https://arxiv.org/pdf/2006.10955.pdf

I think it might be something very useful as other libraries don't seem to support this. 

- Will start readapting the code so it won't rely on tfa. 
- Include training in docstring

LukeWood:
> @LukeWood I have realised I used the wrong author name, I will make sure to use the correct one for the next commit. In terms of the next steps:
> 
> * Is there any desire of making the Blur random? the implementation in simCLR already does that ([Reference](https://github.com/tensorflow/models/blob/master/official/vision/beta/projects/simclr/dataloaders/preprocess_ops.py#L233)). Examples where Random Blur is used:
> 
> https://arxiv.org/pdf/2009.00104.pdf
> 
> https://arxiv.org/pdf/2006.10955.pdf
> 
> I think it might be something very useful as other libraries don't seem to support this.
> 
> * Will start readapting the code so it won't rely on tfa.
> * Include training in docstring

I'd be in favor of support for random behavior.    For example, we could allow `filter_size` to support ranges as well as sigma to support ranges, then randomly sample within the range.

bhack:
>I'd be in favor of support for random behavior. For example, we could allow filter_size to support ranges as well as sigma to support ranges, then randomly sample within the range. 

@LukeWood Are we going to always be random inside the batch? Cause I think that this sometimes is going already to complicate the vectorizzation a bit. Do we have any reference about the convergence improvement with in-batch randomization?
If we could randomize only between the batches instead I think that a Random could be more a policy class then something that we need to be supported with every args at preprocessing layer level.

LukeWood:
> > I'd be in favor of support for random behavior. For example, we could allow filter_size to support ranges as well as sigma to support ranges, then randomly sample within the range.
> 
> @LukeWood Are we going to always be random inside the batch? Cause I think that this sometimes is going already to complicate the vectorizzation a bit. Do we have any reference about the convergence improvement with in-batch randomization? If we could randomize only between the batches instead I think that a Random could be more a policy class then something that we need to be supported with every args at preprocessing layer level.

We are leaning towards guiding users towards implementing:
```augment_image```

Then relying on XLA to perform optimization.  This may be slightly less performant but the entire JAX ecosystem relies on this, so it should actually be fine.  We want to minimize dependence between samples so this is optimal.

bhack:
Yes but also if you want to use XLA or `vectorize_map` doesn't it make sense to have the preprocessing layer itself Random? Isn't better to have a Random policy class that handle the randomization of the args for the underline preprocessing layer?
In that case this could enable more complex policies still controlling the preprocessing args.

artu1999:
Having a higher level Random Blur layer might also be useful in case we want to give users the possibility of applying random MotionBlur and MedianBlur as well. Possibly, we could use ab argument to specify which type of random blur the user wants to apply or even apply a random type of Blur for each picture in the batch.

qlzh727:
> @qlzh727 Do you know if we have an update from the TF team on [tensorflow/tensorflow#39050](https://github.com/tensorflow/tensorflow/issues/39050)?

I don't think there is any active development on that issue at the moment.

LukeWood:
> Yes but also if you want to use XLA or `vectorized_map` does it make sense to have the preprocessing layer itself Random? Isn't better to have a Random policy class that handle the randomization of the args for the underline preprocessing layer? In that case this could enable more complex policies still controlling the preprocessing args.

This sounds really heavy handed.  I think we just want the layers to handle randomization themselves

bhack:
> This sounds really heavy handed. I think we just want the layers to handle randomization themselves

I still believe that I will be harder to remove the implicit randomization over the (hype)parameters space especially as we don't expose the non randomized ops in the internal API.

LukeWood:
@artu1999 any update here?

artu1999:
> @artu1999 any update here?

Sorry for the inactivity, got quite busy with my final university project. I am going to make a commit either tomorrow or the day after with the new code, I just need to fix a couple of things here and there.

artu1999:
Also, quick question: would you be happy with using keras.engine.base_layer.BaseRandomLayer as the class to inherit from or would you prefer to stick with keras.layers.Layer?

LukeWood:
Hey @artu1999  - Please use `BaseImageAugmentationLayer`.  Please see RandomShear as an example.  We will be moving all of our KPL layers to use this:

https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/preprocessing/random_shear.py#L21

Basically, you write a function `augment_image` to augment a single image and it handles vectorization and input format processing for you :)

artu1999:
> Hey @artu1999 - Please use `BaseImageAugmentationLayer`. Please see RandomShear as an example. We will be moving all of our KPL layers to use this:
> 
> https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/preprocessing/random_shear.py#L21
> 
> Basically, you write a function `augment_image` to augment a single image and it handles vectorization and input format processing for you :)

Will use that layer then, thanks!

LukeWood:
Closing for your other PR.  Thanks @artu1999 for starting this process

PR #144, opened 2022-02-21
Add `value_range` utilities.

LukeWood:
This will help us implement the `value_range` parameter
for all of our KPLs.

We can normalize our inputs at the start of each `call` method, then assume [0, 255] range internally, then transform back to the users' dtype and value range at the end.

### DIFF
--- a/keras_cv/utils/preprocessing.py
+++ b/keras_cv/utils/preprocessing.py
@@ -1,3 +1,16 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 import tensorflow as tf
 
 
--- a/keras_cv/utils/preprocessing.py
+++ b/keras_cv/utils/preprocessing.py
@@ -14,17 +14,20 @@
 import tensorflow as tf
 
 
-def transform_value_range(images, original_range, target_range):
+def transform_value_range(images, original_range, target_range, dtype=None):
     """transforms values in input tensor from original_range to target_range.
 
     This function is intended to be used in preprocessing layers that
     rely upon color values.  This allows us to assume internally that
     the input tensor is always in the range [0, 255].
 
+    Note that
+
     Args:
         images: the set of images to transform to the [0, 255] range.
         original_range: the value range to transform from.
         target_range: the value range to transform to.
+        dtype: the dtype to compute the conversion with.  Defaults to tf.float32.
     Returns:
         a new Tensor with values in the target range.
 
@@ -45,9 +48,12 @@ def transform_value_range(images, original_range, target_range):
     )
     ```
     """
-    images = tf.cast(images, dtype=tf.float32)
-    original_min_value, original_max_value = _unwrap_value_range(original_range)
-    target_min_value, target_max_value = _unwrap_value_range(target_range)
+    dtype = dtype or tf.float32
+    images = tf.cast(images, dtype=dtype)
+    original_min_value, original_max_value = _unwrap_value_range(
+        original_range, dtype=dtype
+    )
+    target_min_value, target_max_value = _unwrap_value_range(target_range, dtype=dtype)
 
     # images in the [0, 1] scale
     images = (images - original_min_value) / (original_max_value - original_min_value)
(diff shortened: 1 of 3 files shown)


keras_cv/utils/preprocessing.py
```diff
@@ -0,0 +1,83 @@
...
+    the input tensor is always in the range [0, 255].  This function
+    should be used at the end of the function body before returning
+    the input tensor to the user.
+
+    Args:
+        images: the set of images to transform to the value range
+        value_range: the value range to transform into
+
+    Returns:
+        a new Tensor with values in the range [0, 255]
```
qlzh727, line 61: I think the return value should be in `value_range`, rather than [0, 255].

keras_cv/utils/preprocessing.py
```diff
@@ -0,0 +1,83 @@
...
+    """transforms input Tensor into value_range.
+
+    This function is intended to be used in preprocessing layers that
+    rely upon color values.  This allows us to assume internally that
+    the input tensor is always in the range [0, 255].  This function
+    should be used at the end of the function body before returning
+    the input tensor to the user.
+
+    Args:
+        images: the set of images to transform to the value range
```
qlzh727, line 57: The images should have the default value range, which is [0, 255].

keras_cv/utils/preprocessing.py
```diff
@@ -0,0 +1,83 @@
...
+        value_range
+    )
+    images = tf.math.minimum(images + 10, 255)
+    images = keras_cv.utils.transform_to_value_range(images, value_range)
+    ```
+    """
+    images = tf.cast(images, dtype=tf.float32)
+    min_value, max_value = _unwrap_value_range(value_range)
+    images = (images - min_value) / (max_value - min_value)
+    return images * 255.0
```
qlzh727, line 44: Did we change the dtype of the output? we should make it clear in the docstring.

keras_cv/utils/preprocessing.py
```diff
@@ -0,0 +1,83 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+
+def transform_to_standard_range(images, value_range):
```
qlzh727, line 17: I think this interface is somehow confusing, since the input param `value _range` is not the standard_range, and the default standard range [0, 255] is hidden/implied in the implementation.

My original thought was to have it like, where user need to explicitly specify the from/to value range.

```
def convert_image_plx_value(images, from_range, to_range)
```

keras_cv/utils/preprocessing.py
```diff
@@ -0,0 +1,83 @@
...
+    """transforms the input Tensor to range [0, 255].
```
qlzh727, line 18: Another random thought, since we always expect user to convert from and to its original image range (the convert function comes with pair), is it possible to do this in a context?

```
with utils.standard_image_value(images, original_value_range) as images:
   tf.image.ops(images)

# in the final block of the context, convert the image back to original value range.
tf.other_ops(image) # the image should be convert back to its original value range.


```

qlzh727 (changes requested):

keras_cv/utils/preprocessing.py
LukeWood: Let me see if I can make this work.  Really smart idea.

LukeWood: Just tested - not possible to do in a context unfortunately.

LukeWood:
Update to rely on a single function.  This is a much more explicit user API - good point.

keras_cv/utils/preprocessing.py
```diff
@@ -0,0 +1,70 @@
...
+    ```python
+    original_range = [0, 1]
+    target_range = [0, 255]
+    images = keras_cv.utils.preprocessing.transform_value_range(
+        images,
+        original_range,
+        target_range
+    )
+    images = tf.math.minimum(images + 10, 255)
+    images = keras_cv.utils.transform_value_range(
```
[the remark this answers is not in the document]
LukeWood: Fixed

LukeWood:
@qlzh727 addressed comments.  Thanks!

keras_cv/utils/preprocessing.py
```diff
@@ -0,0 +1,66 @@
...
+def transform_value_range(images, original_range, target_range, dtype=tf.float32):
+    """transforms values in input tensor from original_range to target_range.
+
+    This function is intended to be used in preprocessing layers that
+    rely upon color values.  This allows us to assume internally that
+    the input tensor is always in the range [0, 255].
+
+    Args:
+        images: the set of images to transform to the target range range.
+        original_range: the value range to transform from.
```
bhack, line 26: Do we need to have these ranges free?  Are we not going to practically support just 3 value ranges as in tests?

LukeWood: We actually do need them to be free.  `value_range` will likely be a user supplied value and we can't make assumptions about their dataset/formatting.

bhack: Ok I don't know what other ranges could be interesting other then `-1,1`, `0,1` and `0,255`. 
But If we want to have any free range is responsibility of the caller to validates the inputs as we could theoretically have `nan` in:

`images = (images - original_min_value) / (original_max_value - original_min_value)`

qlzh727 (approved):
Thanks for the PR.

PR #147, opened 2022-02-23
Add performance benchmarking scripts for COCO Metrics

LukeWood:


LukeWood:
This could be a useful example for our users if they want to benchmark other metrics/components.

bhack:
Do we have the CI ready to test notebooks?

bhack:
If understand correctly this will let us to have just to get the final texecution time but any insight like  Tensorflow profiler about the ops, timeline etc.
Can we add [profiler directives](https://www.tensorflow.org/api_docs/python/tf/profiler/experimental/start)?

If I remember correctly it requires a base model to fulfill stats.

bhack:
> If I remember correctly it requires a base model to fulfill stats.

P.s. probably the Keras model isn't strictly required if we use  https://www.tensorflow.org/api_docs/python/tf/profiler/experimental/Trace

bhack:
Some Tensorflow and Google repos are using:

https://www.reviewnb.com/

LukeWood:
I'd rather avoid using a non-github tool to review PRs.  I'll migrate these to scripts.  There's no real reason for these to not be scripts anyways - just an artifact of how I developed these.

bhack:
My proposal:

- I think that we could use just the plain python code as a regular test
- We could add a [performance custom markers](https://docs.pytest.org/en/6.2.x/example/markers.html#mark-examples) so that we could enable/filter this in the CI as we like
- We could use `tf.profiler` like I've done in this [dummy example](https://colab.research.google.com/gist/bhack/36b6900d49a62487778fbaab3e2082dc/trace_profiler.ipynb)

LukeWood:
@fchollet re-formatted as scripts.

LukeWood:
FYI @qlzh727 as well

PR #156, opened 2022-02-26
feat(utils): Added get four corners from bbox.

kartik4949:
solves #151


LukeWood:
what exactly does this do that is different from `xywh_to_corners`?

kartik4949:
This converts 
x1,y1 x2,y2 
To 
4 corners of box format.

LukeWood:
Ah I see what you are trying to do here now.  We already use `x1,y1 x2,y2` as corners format.  Encoding the information twice is lossy and not worth it.  thanks for the contribution - but yeah lets leave it as is!!!

Issue #152, opened 2022-02-26
Create makefile with existing shell/ cmd


kartik4949:

LukeWood:
I'm not sure we really want this.  The current shell setup is lightweight and easy enough.  makefiles get less and less readable as time goes on. 

The problem a makefile solves is managing a multi-step build process.  We don't have that right now.  If we do end up with that, we can adopt `make`.

PR #157, opened 2022-02-26
cli(keras-cv): Added makefile

kartik4949:
solves #152

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,0 +1,9 @@
+.PHONY: format clean
+
+clean:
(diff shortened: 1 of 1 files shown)


kartik4949:
will update it with lint as well so we can get rid of shell completely.

LukeWood:
See response on the issue - we don't have a multi step build right now, so using make is not currently idiomatic.  We can reconsider if we someday do have a multi step build.

Issue #141, opened 2022-02-17
XLA compile all possible for loops in COCOMeanAveragePrecision


LukeWood:
We may also be able to XLA compile some of the inner loops on COCORecall!

bhack:
I suppose that XLA has still some limits with `tf.math.bincount` op (at least on CPU):

> on XLA_CPU_JIT: Bincount (No registered 'Bincount' OpKernel for XLA_CPU_JIT devices compatible with node 

But the list is of CPU and GPU ops is outdated:
https://github.com/tensorflow/tensorflow/issues/14798#issuecomment-350814888

LukeWood:
Ahhh man.  Yeah we can’t get around bincount.  Maybe we can use it for result though?

bhack:
I've tried also to give a run with the "next-gen" MLIR bridge compiler `tf.config.experimental.enable_mlir_bridge()`. 
It seems that it fail to legalize it:

> OP_REQUIRES failed at xla_ops.cc:248 : INVALID_ARGUMENT: TF to XLA legalization failed: /usr/local/lib/python3.7/site-packages/tensorflow/python/framework/ops.py:2203:1: error: 'tf.TensorListReserve' op unknown tensor list element shape

bhack:
Just to reference the code block for visibility:

https://github.com/keras-team/keras-cv/blob/a25ee77ab8466163322332a6d707606794ddf259/keras_cv/metrics/coco/mean_average_precision.py#L135-L253

bhack:
>  Maybe we can use it for result though?


Instead for result XLA without MLIR bridge is working 

With MLIR:
> INVALID_ARGUMENT: TF to XLA legalization failed: /usr/local/lib/python3.7/site-packages/tensorflow/python/framework/ops.py:2203:1: error: The following operations cannot be legalized: tf.LowerBound (count: 10); tf.Where (count: 10). These legalization failure(s) may be due to missing TF to HLO lowerings and/or unsupported attributes, etc.


https://github.com/keras-team/keras-cv/blob/a25ee77ab8466163322332a6d707606794ddf259/keras_cv/metrics/coco/mean_average_precision.py#L255-L316

bhack:
For result:
HLO graph CPU before optimization

![visualized_output](https://user-images.githubusercontent.com/1710528/154601409-90c4406b-069f-4c00-a447-44f3324e8798.svg)

HLO graph CPU after optimization
[...]

PR #161, opened 2022-03-03
Add benchmarking script, update API_DESIGN.md to reflect the results.

LukeWood:


.github/API_DESIGN.md
```diff
@@ -4,15 +4,33 @@ In general, KerasCV abides to the  [API design guidelines of Keras](https://gith
...
 `segm` are often used.  In KerasCV, we will *not* be using these abbreviations.  This is done
 to ensure full consistency in our naming convention.  While the team is fond of the abbreviation
 `bbox`, we are loss fond of `segm`.  In order to ensure full consistency, we have decided to
 use the full names for label types in our code base.
 
-## Preprocessing Layers
-### Color Based Preprocessing Layers
+# Preprocessing Layers
+## BaseImageAugmentationLayer
+When implementing preprocessing, we encourage users to subclass the `tf.keras.__internal__.layers.BaseImageAugmentationLayer`.  This layer provides a common `call` method, auto vectorization, and more.
```
fchollet, line 16: Nit: use `()` after method/function names, and keep lines under ~80 char to help with version control & reviews

benchmarks/vectorization_strategy_benchmark.py
```diff
@@ -0,0 +1,700 @@
...
+plt.legend()
+plt.show()
+
+"""
+# Sanity check
+all of these should have comparable outputs
+"""
+
+images = []
+for aug in [VectorizedRandomCutout, VMapRandomCutout, MapFnRandomCutout]:
```
fchollet, line 680: Would it be useful to also have numbers with XLA compilation?

fchollet (approved):

.github/API_DESIGN.md
LukeWood: Thanks - done.

benchmarks/vectorization_strategy_benchmark.py
LukeWood: https://github.com/keras-team/keras-cv/issues/165 lets do in a follow up.

LukeWood: (but I do think so)

benchmarks/vectorization_strategy_benchmark.py
```diff
@@ -0,0 +1,700 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+# Setup/utils
+"""
+import time
+
+import matplotlib.pyplot as plt
```
bhack, line 17: Is the benchmark a test or an example?

https://github.com/keras-team/keras-cv/blob/master/setup.py#L30

.github/API_DESIGN.md
```diff
@@ -1,28 +1,59 @@
...
+- `augment_bounding_box()` allows updates to bounding boxes to be made
+
+When a canonical layer subclassing BaseImageAugmentationLayer is available, a 
+link to it will be added here.
+
+## Vectorization
+`BaseImageAugmentationLayer` requires you to implement augmentations in an 
+image-wise basis instead of using a vectorized approach.  This design choice 
+was based made on the  results found in the 
+[vectorization\_strategy\_benchmark.py](../benchmarks/vectorization_strategy_benchmark.py)
```
bhack, line 35: How we are going to run this from command-line?

benchmarks/vectorization_strategy_benchmark.py
```diff
@@ -0,0 +1,700 @@
...
+    images.append(layer(x_train[:3]))
+images = [y for x in images for y in x]
+
+
+plt.figure(figsize=(8, 8))
+for i in range(9):
+    plt.subplot(3, 3, i + 1)
+    plt.imshow(images[i].numpy().astype("uint8"))
+    plt.axis("off")
+plt.show()
```
bhack, line 691: This is not going to work well in containers like in our official [keras dev container](https://github.com/keras-team/keras/tree/master/.devcontainer). Isn't better to default plot on file?

benchmarks/vectorization_strategy_benchmark.py
bhack: In this case for Vectorized and VMap:

immagine

bhack: We also need to consider the general discussion we had at https://github.com/keras-team/keras-cv/pull/146#issuecomment-1049352918

As this one is a special case that is running fine with `vectorized_map/fallback_to_while_loop=False`.

Also,  this simple case doesn't have any specific not JITable ops like e.g. `ImageProjectiveTransformV3` (E.g. for rotation) or  `tf.bincount` (see https://github.com/keras-team/keras-cv/issues/141#issuecomment-1047817015)

Issue #134, opened 2022-02-12
Add `num_cutouts` parameter to CutOut layer


innat:
It's kinda an alternate of cutout augmentation but with more options.

tf-code reference. https://www.kaggle.com/cdeotte/tfrecord-experiments-upsample-and-coarse-dropout

Demo. 
![output_13_0](https://user-images.githubusercontent.com/17668390/153725511-6070a4f0-d99a-4b6a-b184-922b7914aae6.png)

innat:
cc. @chjort

chjort:
If we want to support this, I suggest we merge it into `RandomCutout` by giving it an additional argument like `num_cutouts` or something. @LukeWood wdyt?

LukeWood:
Yeah, I'd prefer a `num_cutouts` parameter... perhaps supporting a range to randomly sample from.

parikshit14:
hi, I would like to contribute to this task. 
So from the reference link provided above, does ```num_cutouts``` stand for **number of cutouts** that must be cut from the image provided?

LukeWood:
absolutely!  And I'd also like this to support a range.

So you can have

```
CutOut(num_cutouts=(1, 10))
```

Then it will randomly sample from 1, 10 and perform that many cutouts.

frostbyte012:
@LukeWood  what will those cutouts do?  From the link provided, I understood it's done to reduce overfitting . If I'm not wrong. And how will those have an impact on the accuracy though? And is this kinda data augmentation?

frostbyte012:
@LukeWood  what will those cutouts do?  From the link provided, I understood it's done to reduce overfitting . If I'm not wrong. And how will those have an impact on the accuracy though? And is this kind of data augmentation?
I want to try this one and contribute if this is still open.

quantumalaviya:
I think this can be closed now

innat:
@quantumalaviya Thanks for this PR. One query and concern on this 

```
for _ in tf.range(self._sample_num_cutouts())
```

Isn't loping through the num of cutouts causing performance issues?  It's the same if you use `tf.map_fn`, under the hood, it also does something like that, I think. @parikshit14 did some benchmark, [HERE](https://github.com/keras-team/keras-cv/pull/178#issuecomment-1075135062), is it reproduced? 

cc. @LukeWood @bhack

quantumalaviya:
Yeah, I mentioned it in the PR (#207). I was waiting on @parikshit14 for the changes.

[...]

PR #178, opened 2022-03-20
Adding num_cutouts to RandomCutout

quantumalaviya:
Added `num_cutouts` parameter to `RandomCutout` as mentioned in issue #134 

Sample usage:
```python
(images, labels), _ = tf.keras.datasets.cifar10.load_data()
random_cutout = keras_cv.layers.preprocessing.RandomCutout(0.5, 0.5, num_cutouts = (10, 30))
augmented_images = random_cutout(images)
```
This cutout will add random cutouts in the range [10, 30]

### DIFF
--- a/keras_cv/layers/preprocessing/random_cutout.py
+++ b/keras_cv/layers/preprocessing/random_cutout.py
@@ -172,7 +172,7 @@ class RandomCutout(layers.Layer):
 
     def _random_cutout(self, inputs):
         """Apply random cutout."""
-        for _ in range(self._sample_num_cutouts()):
+        for _ in tf.range(self._sample_num_cutouts()):
             center_x, center_y = self._compute_rectangle_position(inputs)
             rectangle_height, rectangle_width = self._compute_rectangle_size(inputs)
             rectangle_fill = self._compute_rectangle_fill(inputs)
--- a/keras_cv/layers/preprocessing/random_cutout.py
+++ b/keras_cv/layers/preprocessing/random_cutout.py
@@ -51,7 +51,7 @@ class RandomCutout(layers.Layer):
             - A tuple of size 2, representing the range for the number of cutouts.
             For example, `num_cutouts=10` results in 10 cutouts.
             `num_cutouts=(2,8)` results in num_cutouts between [2, 8]. Can be used
-            to implement coarse dropout.
+            to implement coarse dropout. Defaults to 1.
 
     Sample usage:
     ```python
@@ -172,7 +172,7 @@ class RandomCutout(layers.Layer):
 
     def _random_cutout(self, inputs):
         """Apply random cutout."""
-        for _ in range(self._compute_num_cutouts()):
+        for _ in range(self._sample_num_cutouts()):
             center_x, center_y = self._compute_rectangle_position(inputs)
             rectangle_height, rectangle_width = self._compute_rectangle_size(inputs)
             rectangle_fill = self._compute_rectangle_fill(inputs)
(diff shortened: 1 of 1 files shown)


keras_cv/layers/preprocessing/random_cutout.py
```diff
@@ -46,6 +46,12 @@ class RandomCutout(layers.Layer):
             - *gaussian_noise*: Pixels are filled with random gaussian noise.
         fill_value: a float represents the value to be filled inside the patches
             when `fill_mode="constant"`.
+        num_cutouts: One of:
```
LukeWood, line 49: add the default "Defaults to 1."

keras_cv/layers/preprocessing/random_cutout.py
```diff
@@ -172,6 +209,16 @@ def _compute_rectangle_position(self, inputs):
         )
         return center_x, center_y
 
+    def _compute_num_cutouts(self):
```
LukeWood, line 212: rename to `_sample_num_cutouts`

LukeWood (changes requested):
Nice work!  Thanks

LukeWood (approved):

LukeWood:
Thanks @quantumalaviya !!!

keras_cv/layers/preprocessing/random_cutout.py
```diff
@@ -136,17 +172,18 @@ def call(self, inputs, training=True):
...
-        rectangle_fill = self._compute_rectangle_fill(inputs)
-        inputs = fill_utils.fill_rectangle(
-            inputs,
-            center_x,
-            center_y,
-            rectangle_width,
-            rectangle_height,
-            rectangle_fill,
-        )
+        for _ in range(self._sample_num_cutouts()):
```
LukeWood, line 175: Does this work with `tf.range` instead of `range`?

LukeWood (changes requested):

parikshit14:
hi, but dont you think implementation should be in vectorized format

parikshit14:
hi, but dont you think implementation should be in vectorized format

parikshit14:
@LukeWood dont you think this will effect upon the overall performance as the techniques used are not in vectorized format and is simply using loops

parikshit14:
@LukeWood dont you think this will effect upon the overall performance as the techniques used are not in vectorized format and is simply using loops

LukeWood:
> @LukeWood dont you think this will effect upon the overall performance as the techniques used are not in vectorized format and is simply using loops

I think we can rely on `BaseImageAugmentationLayer` to vectorize this.

parikshit14:
I was able to complete the vectorized form so instead of just relying on `BaseImageAugmentationLayer` we can simply use the vectorized form at the very base, it will in one way or another give an edge, if you wanna have a look at the vectorized code I can create a PR for that!

quantumalaviya:
@parikshit14 I assume you're talking about vectorizing across the number of cutouts? Did you use `fill_utils` in your PR or drop the reference altogether? If you used it, you might have had to edit `fill_rectangle` as well?

parikshit14:
> @parikshit14 I assume you're talking about vectorizing across the number of cutouts? Did you use `fill_utils` in your PR or drop the reference altogether? If you used it, you might have had to edit `fill_rectangle` as well?

yes, I did some changes in `fill_rectangle` as well as in `rectangle_masks` that's the reason it took me so long to get to the vectorized code.

parikshit14:
@LukeWood I tried implementing the new benchmark(according to `num_cutouts` ) and looks like tf.range cannot be optimized by  `tf.vectorized_map` as it is giving error `OperatorNotAllowedInGraphError: Iterating over a symbolic 'tf.Tensor' is not allowed` .

Also did some benchmarks to compare vectorized vs non-vectorized approaches for `num_cutouts=5`. 
**Results :** 
vectorized_vs_unvectorized

_UnvRandomCutout_ -> unvectorized approch
_VRandomCutout_ -> vectorized approch

PR #180, opened 2022-03-21
Create RandomShear layer using the BaseImageAugmentationLayer

LukeWood:
random_shear


keras_cv/layers/preprocessing/random_shear.py
```diff
@@ -0,0 +1,119 @@
...
+        x: float, 2 element tuple, or `None`.  For each augmented image a value is
+            sampled from the provided range.  If a float is passed, the range is
+            interpreted as `(0, x)`.  If `None` is passed, no shear occurs on the X
+            axis.  Defaults to `None`.
+        y: float, 2 element tuple, or `None`.  For each augmented image a value is
+            sampled from the provided range.  If a float is passed, the range is
+            interpreted as `(0, y)`.  If `None` is passed, no shear occurs on the Y
+            axis.  Defaults to `None`.
+        interpolation: interpolation method used in the `ImageProjectiveTransformV3` op.
+             Supported values can be found in the `ImageProjectiveTransformV3`
```
qlzh727, line 34: I think it is better to explicitly list the options here.

keras_cv/layers/preprocessing/random_shear.py
```diff
@@ -0,0 +1,119 @@
...
+             documentation.  Defaults to `"bilinear"`.
+        fill_mode: fill_mode in the `ImageProjectiveTransformV3` op.
+             Supported values can be found in the `ImageProjectiveTransformV3`
```
qlzh727, line 37: Same here. Please explicitly list the options here.

keras_cv/layers/preprocessing/random_shear.py
```diff
@@ -0,0 +1,119 @@
...
+        fill_value=0.0,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        if isinstance(x, float):
+            x = (0, x)
+        if isinstance(y, float):
+            y = (0, y)
+        if x is None and y is None:
+            warnings.warn(
```
qlzh727, line 59: Should this be an error instead?

keras_cv/layers/preprocessing/random_shear.py
```diff
@@ -0,0 +1,119 @@
...
+        x=None,
+        y=None,
+        interpolation="bilinear",
+        fill_mode="reflect",
+        fill_value=0.0,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        if isinstance(x, float):
+            x = (0, x)
```
qlzh727, line 55: Do u expect x and y to be positive? or it can be any value? (I didn't see any restriction in docstring.)

keras_cv/layers/preprocessing/random_shear.py
```diff
@@ -0,0 +1,119 @@
...
+                fill_mode=self.fill_mode,
+                fill_value=self.fill_value,
+            )
+
+        return tf.squeeze(image, axis=0)
+
+    @staticmethod
+    def _format_transform(transform):
+        transform = tf.convert_to_tensor(transform, dtype=tf.float32)
+        return transform[None]
```
qlzh727, line 119: Humm, what is transform[None]?

keras_cv/utils/preprocessing.py
```diff
@@ -64,3 +65,96 @@ def _unwrap_value_range(value_range, dtype=tf.float32):
     min_value = tf.cast(min_value, dtype=dtype)
     max_value = tf.cast(max_value, dtype=dtype)
     return min_value, max_value
+
+
+def transform(
```
qlzh727, line 70: Assume this method is copied from core keras, should we consider expose this as __internal__ method?

keras_cv/layers/preprocessing/random_shear.py
LukeWood: it adds a new axis on [1,].  Its required to use `transform` as ImageProjectiveTransform expects a batch dim.

keras_cv/layers/preprocessing/random_shear.py
LukeWood: Good point - updated

keras_cv/layers/preprocessing/random_shear.py
LukeWood: No - because some pipelines may gridsearch over hyperparameters and not want to special case remove the layer from their codebase.  Or perhaps users will pass `x_shear` as a flag, and the code base will not know if it is `None` or a value.

I say we keep it as a warning in this case - the COCO metrics had a similar case with sample weights.

LukeWood:
> Thanks for the change. Can you paste the demo output of RandomShear to the the PR description?

Done - thanks!

keras_cv/utils/preprocessing.py
LukeWood: I discussed this with Matt and we decided to just copy paste it as we want the freedom to make changes to it going forward.  We could expose it in __internal__ - but I think copy pasting this is easy enough for now.

keras_cv/layers/preprocessing/random_shear.py
LukeWood: Done

keras_cv/layers/preprocessing/random_shear.py
LukeWood: Done

keras_cv/layers/preprocessing/random_shear.py
LukeWood: https://stackoverflow.com/questions/1408311/numpy-array-slice-using-none same is supported in TensorFlow - using `None` is equivalent to `tf.new_axis`

keras_cv/layers/preprocessing/random_shear.py
```diff
@@ -0,0 +1,119 @@
...
+import tensorflow as tf
+
+from keras_cv.utils import preprocessing
+
+
+class RandomShear(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Randomly shears an image.
+
+    Args:
+        x: float, 2 element tuple, or `None`.  For each augmented image a value is
```
qlzh727, line 25: Is this value based on PI? eg if I want the image to shear between 0 and 60 degree, should I put (0, 60/PI) here? (I think its better to have some sample value/range here)

keras_cv/layers/preprocessing/random_shear.py
LukeWood: Updated to `tf.newaxis` for readability.

keras_cv/layers/preprocessing/random_shear.py
```diff
@@ -0,0 +1,119 @@
...
+            sampled from the provided range.  If a float is passed, the range is
+            interpreted as `(0, x)`.  Provided values should be positive.  If `None` is
```
qlzh727, line 27: In the case of two element, should they all be positive? or the positive rule is only applied to single element input?

keras_cv/layers/preprocessing/random_shear_test.py
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+from keras_cv.layers import preprocessing
+
+NUM_CLASSES = 10
```
qlzh727, line 18: this value seems to be unused in this file.

qlzh727 (changes requested):

keras_cv/layers/preprocessing/random_shear.py
LukeWood: They should all be positive - updated to be clearer

keras_cv/layers/preprocessing/random_shear.py
LukeWood: Updated with value information - it is a percentage of the image.  Thanks

qlzh727 (approved):

PR #186, opened 2022-03-22
benchmarks for num_cutouts

parikshit14:
I tried implementing the new benchmark(according to num_cutouts ) and looks like tf.range cannot be optimized by tf.vectorized_map as it is giving error OperatorNotAllowedInGraphError: Iterating over a symbolic 'tf.Tensor' is not allowed.

### DIFF
--- a/benchmarks/vectorization_strategy_benchmark.py
+++ b/benchmarks/vectorization_strategy_benchmark.py
@@ -13,6 +13,7 @@
 # Setup/utils
 """
 import time
+import numba
 
 import matplotlib.pyplot as plt
 import tensorflow as tf
@@ -23,7 +24,7 @@ from tensorflow.keras import backend
 from keras_cv.utils import bounding_box
 from keras_cv.utils import fill_utils
 
-
+@numba.jit()
 def single_rectangle_mask(corners, mask_shape):
     """Computes masks of rectangles
 
@@ -41,24 +42,23 @@ def single_rectangle_mask(corners, mask_shape):
     corners = corners[..., tf.newaxis, tf.newaxis]
 
     # split coordinates
-    x0 = corners[:, 0]
-    y0 = corners[:, 1]
-    x1 = corners[:, 2]
-    y1 = corners[:, 3]
+    x0 = corners[0]
+    y0 = corners[1]
+    x1 = corners[2]
+    y1 = corners[3]
 
     # repeat height and width
     width, height = mask_shape
-    x0_rep = tf.repeat(x0, height, axis=1)
-    y0_rep = tf.repeat(y0, width, axis=2)
-    x1_rep = tf.repeat(x1, height, axis=1)
-    y1_rep = tf.repeat(y1, width, axis=2)
+    x0_rep = tf.repeat(x0, height, axis=0)
+    y0_rep = tf.repeat(y0, width, axis=1)
+    x1_rep = tf.repeat(x1, height, axis=0)
+    y1_rep = tf.repeat(y1, width, axis=1)
 
     # range grid
-    batch_size = tf.shape(corners)[0]
     range_row = tf.range(0, height, dtype=corners.dtype)
     range_col = tf.range(0, width, dtype=corners.dtype)
-    range_row = tf.repeat(range_row[tf.newaxis, :, tf.newaxis], batch_size, 0)
-    range_col = tf.repeat(range_col[tf.newaxis, tf.newaxis, :], batch_size, 0)
+    range_row = range_row[:, tf.newaxis]
+    range_col = range_col[tf.newaxis, :]
 
     # boolean masks
     mask_x0 = tf.less_equal(x0_rep, range_col)
@@ -70,8 +70,8 @@ def single_rectangle_mask(corners, mask_shape):
 
     return masks
 
-
-def fill_single_rectangle(images, centers_x, centers_y, widths, heights, fill_values):
+@numba.jit()
+def fill_single_rectangle(image, centers_x, centers_y, widths, heights, fill_values):
     """Fill rectangles with fill value into images.
 
     Args:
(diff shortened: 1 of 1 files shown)


parikshit14:
@LukeWood  attaching the colab notebook link [colab](https://colab.research.google.com/drive/1XEsxciVP5EF4K9NB4Buio66mv9LAanir?usp=sharing) . In this, you can observe the changes according to num_cutouts and the error generated due to looping over a tf.range.

benchmarks/vectorization_strategy_benchmark.py
```diff
@@ -356,64 +425,77 @@ def _parse_bounds(self, factor):
...
-        input = fill_single_rectangle(
-            input,
-            center_x,
-            center_y,
-            rectangle_width,
-            rectangle_height,
-            rectangle_fill,
-        )
-        return input
+    def _random_cutout(self, inputs):
```
bhack, line 435: Can you try to move the @tf.function from `call` to `_random_cutout`

parikshit14: > Can you try to move the @tf.function from `call` to `_random_cutout`?

done

parikshit14:
> the benchmark only existed to show the performance of map and vectorized map for the design purposes of BaseImageAugmentationLayer. It is ok if num_cutouts doesn't work for it - though this is not ideal for when it comes to migrating RandomCutout to BaseImageAugmentationLayer

So what do you suggest, should I create a PR for the vectorized approach for `num_cutouts`?

benchmarks/vectorization_strategy_benchmark.py
bhack: Are you sure? I don't see this. Then try to use again the `vectorized_map` and check if you have bypassed the error.

bhack: You can track this ticket https://github.com/tensorflow/tensorflow/issues/55340

quantumalaviya:
@parikshit14 if the `fill_utils` functions have been tested across all their references (mainly `CutMix` as far as I know), I think that vectorizing it is a better idea indeed since `BaseImageAugmentationLayer` relies on `tf.vectorize_map`.

parikshit14:
> @parikshit14 if the `fill_utils` functions have been tested across all their references (mainly `CutMix` as far as I know), I think that vectorizing it is a better idea indeed since `BaseImageAugmentationLayer` relies on `tf.vectorize_map`.

The changes to vectorize `num_cutouts` in `fill_rectangles` and `rectangle_mask` will make it incompatible for `CutMix`. So to counter that we can create **FLAG** sort of a variable to separate both of these(RandomCutout and CutMix) inside `fill_utils` or we can create a separate file for the vectorized `fill_utils`, whichever is the best fit.

benchmarks/vectorization_strategy_benchmark.py
parikshit14: ok, got what you were trying to say, but looks like there needs to be some changes to be done in `fill_single_rectangle` as well. Working on it.......

parikshit14:
Changes: 

- changed @tf.function position as mentioned by @bhack 
- updated `fill_single_rectangle` to accept `tf.map_fn` and `tf.vectorized_map` calls
- Hard coded `num_cutouts` value to 5 so comparison between all three calls is fair

Colab link to see the output benchmark data : [colab-benchmark-comparision](https://colab.research.google.com/drive/1XEsxciVP5EF4K9NB4Buio66mv9LAanir?usp=sharing)

output graph:
graph (1)

At number of images = 50,000 
time taken by `VMapRandomCutout` is more than three times as that of `VectorizedRandomCutout` . For GPU as well similar results were obtained.

bhack:
Can you check with `jit_compile=true`  (in the case all the involved ops are supported)?

parikshit14:
Results:

| No. of Images | vectorized (s) | vectorized JIT (s) | vectorized_map (s) | vectorized_map JIT (s) |
| --- | --- | --- | --- | --- |
| 1000 | 0.232527494430542 | 0.14783048629760742 | 1.7844336032867432 | 1.9857478141784668 |
| 2000 | 0.4784221649169922 | 0.29476451873779297 | 2.4432573318481445 | 2.3840723037719727 |
| 5000 | 1.1677978038787842 | 0.6889147758483887 | 3.5335452556610107 | 4.823124170303345 |
| 10000 | 2.1287717819213867 | 1.4040074348449707 | 5.350322008132935 | 5.415518760681152 |
| 25000 | 5.106718301773071 | 3.551025152206421 | 11.242588758468628 | 11.360033750534058 |
| 37500 | 6.948775053024292 | 5.144469976425171 | 16.840341806411743 | 16.344752550125122 |
| 50000 | 6.931020975112915 | 6.8766396045684814 | 21.10237979888916 | 21.155300617218018 |

bhack:
vectorized_map is slower?

parikshit14:
Thats the craziest part coz the so called `VectorizedRandomCutout` isn't even vectorized. And if we use `tf.vectorize_map` over it, it is not able to optimise loops.

LukeWood:
> Results:
> 
> No. of Images	vectorized (s)	vectorized JIT (s)	vectorized_map (s)	vectorized_map JIT (s)
> 1000	0.232527494430542	0.14783048629760742	1.7844336032867432	1.9857478141784668
> 2000	0.4784221649169922	0.29476451873779297	2.4432573318481445	2.3840723037719727
> 5000	1.1677978038787842	0.6889147758483887	3.5335452556610107	4.823124170303345
> 10000	2.1287717819213867	1.4040074348449707	5.350322008132935	5.415518760681152
> 25000	5.106718301773071	3.551025152206421	11.242588758468628	11.360033750534058
> 37500	6.948775053024292	5.144469976425171	16.840341806411743	16.344752550125122
> 50000	6.931020975112915	6.8766396045684814	21.10237979888916	21.155300617218018

Awesome.  is the jit_compile=True code available?  Can we add it to the benchmark?

LukeWood:
Thanks for the PR.  Once we fix the tests happy to merge

parikshit14:
I have done the changes in the colab notebook(for JIT compile): [benchmarks](https://colab.research.google.com/drive/1XEsxciVP5EF4K9NB4Buio66mv9LAanir?usp=sharing) . If you want I can update this PR with the changes @LukeWood but I think the benchmark for vectorized implementation will make more sense we are going to merge that.

LukeWood:
If you add the jit compilation I think this will add value!  I don't want to include the num_cutouts part because it adds complexity to the benchmark but doesn't add value to the benchmark with respect to it's original purpose.  Hope you understand!

LukeWood:
Ah - I didn't realize you were using numba.jit.  I thought we were discussing @tf.function(jit_compile=True).  Can we try that instead - or does it not work?

parikshit14:
@LukeWood updated according to the changes suggested, had to update the corners to **_coco_to_corners** as it has been replaced  in _bounding_box.py_

LukeWood:
Sorry I think there is a miscommunication occurring here.  We want to benchmark BOTH with and without jit_compile=True.  Not just with jit_compile=True.  Otherwise we are throwing out useful information.

parikshit14:
So in total, you want six classes, two of each i.e. with and without jit_compile for Fully Vectorized, map_fn, and vectorized_map.

LukeWood:
> So in total, you want six classes, two of each i.e. with and without jit_compile for Fully Vectorized, map_fn, and vectorized_map?

Basically yes, this is just for benchmark purposes so yes!

parikshit14:
@LukeWood updated with jit-compiled for each of the three classes

LukeWood:
Rebased for you to get ready for submission.  Thanks @parikshit14

PR #193, opened 2022-03-22
feat(utils): Initial setup mask to bboxes

### LINKED ISSUE #173: Repurposing Masks into Bounding Boxes.
**Convert the Segmentation masks into Object localization annotations for different tasks, e.g. Instance and Panoptic segmentation.**

Reference: [torchvision.ops.masks_to_boxes](https://pytorch.org/vision/stable/generated/torchvision.ops.masks_to_boxes.html#torchvision.ops.masks_to_boxes)
Details and Code: [Mask-to-BoundingBox](https://pytorch.org/vision/stable/auto_examples/plot_repurposing_annotations.html#sphx-glr-auto-examples-plot-repurposing-annotations-py)

Demo.

sphx_glr_plot_repurposing_annotations_001 sphx_glr_plot_repurposing_annotations_002

kartik4949:
closes #173

a simple function as of now, which only gets bbox from single mask, will be working to make it for (n, mask)

technique
get a first occurrence of positive pixel across rows and columns which would give top and left 
similar for Bottom and right


keras_cv/utils/bounding_box_test.py
```diff
@@ -114,3 +118,18 @@ def test_bounding_box_padding(self):
             ValueError, "Target shape should be larger than bounding box shape"
         ):
             bounding_box.pad_bounding_box_batch_to_shape(bounding_boxes, target_shape)
+
+    def test_mask_to_bbox(self):
```
LukeWood, line 122: use `bounding_box` everywhere over bbox

keras_cv/utils/bounding_box_test.py
```diff
@@ -114,3 +118,18 @@ def test_bounding_box_padding(self):
...
+        target_boundingbox = [1, 1, 1, 1]
```
LukeWood, line 123: can we try a bouding box where the coordinates are not all 1s?


[1, 1, 5, 5] maybe?

LukeWood, line 123: add _ between bounding and box

keras_cv/utils/bounding_box_test.py
```diff
@@ -114,3 +118,18 @@ def test_bounding_box_padding(self):
...
+        result = bounding_box.mask_to_bbox(self.center_box_mask_image)
+        self.assertAllClose(result, target_boundingbox)
+
+        self.center_box_mask_image_3d = tf.expand_dims(
```
LukeWood, line 127: this can just be a variable, doesn't need to be assigned to `self`

keras_cv/utils/bounding_box_test.py
```diff
@@ -27,6 +27,10 @@ def setUp(self):
             [[60, 60, 100, 100], [70, 70, 100, 100]], dtype=tf.float32
         )
 
+        self.center_box_mask_image = tf.constant(
```
LukeWood, line 30: use a local variable in the test case

keras_cv/utils/bounding_box.py
```diff
@@ -149,3 +149,33 @@ def pad_bounding_box_batch_to_shape(bounding_boxes, target_shape, padding_values
     return tf.pad(
         bounding_boxes, paddings, mode="CONSTANT", constant_values=padding_values
     )
+
+
+def mask_to_bbox(masks):
```
LukeWood, line 154: lets try to suppose batch dimension before we submit this!  Thank you!

keras_cv/utils/bounding_box.py
```diff
@@ -149,3 +149,33 @@ def pad_bounding_box_batch_to_shape(bounding_boxes, target_shape, padding_values
...
+    def _get_bbox(mask):
+        def _get_positive_pixel_index(group, occurence=0):
+            pos_group = tf.where(group)[:-1]
```
LukeWood, line 157: avoid abbreviations: position_group

LukeWood (changes requested):
Thanks for the PR!  This is a useful function.  Lets touch it up a bit per the review!

Cheers!

keras_cv/utils/bounding_box_test.py
kartik4949: yes, by mistake.

keras_cv/utils/bounding_box_test.py
kartik4949: @LukeWood make sense,
btw should be check top, left as well, like the sequence of them
using 

1,2, 3,4 may be?
in all test cases

keras_cv/utils/bounding_box.py
kartik4949: hmm, let me think how to apply mapping along a axis 
i.e of apply_along_axis in Numpy.

kartik4949:
@LukeWood 
pr seems ready,
I have covered following scenario
* empty masks (empty mean no bbox masks)
* mixture of empty and no empty masks
* (H, W) masks
* (N, H, W) masks 
* batch_dim support

kartik4949:
mask

boxed

kartik4949:
@LukeWood can u kick off ci :)

LukeWood:
Hey Kartik, I appreciate this PR but I'm not sure we want to include this in the library at this point.  We may revisit in the future, in which case we can reference back to this PR.

Issue #176, opened 2022-03-18
Support dynamic setting of DropBlock.dropout_rate


LukeWood:

LukeWood:
Currently the constant value is baked into the graph, I suggest a schedule parameter to support this.

kksinghal:
I want to look into this.

sebastian-sz:
@LukeWood What could this API look like? Do we want to support only linear schedule, mentioned in the paper?
`layer = DropBlock2D(schedule={"total_steps": 100, "starting_rate": 0.0}`

Or maybe accept a separate class that will handle scheduling in it's own call method?
```python
class DropoutRateScheduler:
    def __init__(self, total_steps: int, starting_rate = 0.0):
        self._current_step = 0

    def __call__(current_rate):
          self._current_step += 1
          # modify `current_rate` accordingly


layer = DropBlock2D(scheduler=DropoutRateScheduler(...))
```

Or maybe something completely different?

LukeWood:
I think we want to also support something like lambda functions:
```
DropBlock2D(schedule=lambda current_step: current_step / 1000)
```

or something

kksinghal:
@sebastian-sz, I don't think it's worth making a separate DropoutRateScheduler class since it is a very generic function, it returns a linear function of a value. Otherwise, we could also use any scheduler function, like tf.keras.optimizers.schedules.PolynomialDecay, with power=1.

Instead,
1. The user can pass the initial dropout rate, end dropout rate, total train steps as arguments, and we can handle the linear scheduler inside our DropBlock2D class so that user don't have to care about the scheduler. 
2. The user can also pass a custom-defined scheduler function, which takes the current step as input and returns the rate, as @LukeWood mentioned. Here, the user can also use the tf schedulers functions.
So, the user can go either way. What do you say?

kksinghal:
@LukeWood, Regarding #184, 
Do you want me to add arguments of linear scheduler to DropBlock2D? or do you want to pass the scheduler function only.
About #184, Do you want to create a new class of schedulers like tf.keras.optimizers.schedules?
Please mention if you are working on it. Otherwise, I would like to work on both of these.

LukeWood:
Lets create a schedule that works like tf.keras.optimizers.schedules.

Feel free to create a class in the PR and I will provide API feedback in the code review.  Thank you @kksinghal.

So...

[...]

PR #196, opened 2022-03-22
Support dynamic setting of DropBlock.dropout_rate #176

kksinghal:
Three options for users:
1. Pass dropout_rate for constant dropout
2. Pass initial_dropout_rate, end_dropout_rate, total_train_steps. We handle the linear scheduler through function _linear_scheduler implemented within DropBlock2D class
3. Pass a custom function, which takes current_train_step as input, and returns a float dropout_rate

### DIFF
--- a/keras_cv/core/__init__.py
+++ b/keras_cv/core/__init__.py
@@ -1,0 +1,14 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_cv.core.schedules.linear_scheduler import LinearScheduler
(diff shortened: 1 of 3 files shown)


LukeWood:
Lets do something like https://github.com/keras-team/keras-cv/pull/238 but for schedules

kksinghal:
@LukeWood Please have a look

LukeWood:
@kksinghal Thanks for the API.  (Sorry this got lost in my inbox).  

Due to the "core" nature of this API I am still figuring out some details as to how this will fit into the entire ecosystem.  I think your design is good, I just want to be sure it is "future proof" before I accept and merge.  But overall this PR looks pretty good.  This is something I will be looking at soon now that v1 is out.

lucasdavid:
Just a few thoughts on this, if it's of any help:

1. There's a leak of responsibility, where the Dropblock is now also expected to keep track of training status.
2. The paper clearly states that all Dropblock layers receive the same alpha, so having different schedulers and `training_step` for each layer is outside the scope of this technique and hasn't been validated by any scientific work, as far as I know.
1. The method `DropBlock2D#call` will not be invoked each feed-forward if its traced. Instead, this function is converted into a TF graph, which will bake the number `1` as a constant and always return the exact same value. You need to use a tf.Variable in here and update it each step/epoch. There's already a variable available for this (`model.optimizer.iterations <tf.Tensor[type: int, shape: ()]>`). Maybe there's a way to reuse it in here?

sebastian-sz:
I'm wondering if wouldn't it be better to write a separate, lightweight wrapper layer `ScheduledDropBlock2D`, (could even be in the same file) which accepts scheduler + all those other params and calls regular `DropBlock2D` internally.

This would simplify the constructor and maintain the separation of concerns.

LukeWood:
> I'm wondering if wouldn't it be better to write a separate, lightweight wrapper layer `ScheduledDropBlock2D`, (could even be in the same file) which accepts scheduler + all those other params and calls regular `DropBlock2D` internally.
> 
> This would simplify the constructor and maintain the separation of concerns.

We're also still figuring out what the best way to design this API is.  There are valid use cases in Keras core as well.

Issue #165, opened 2022-03-07
Augment benchmarks/vectorization_strategy_benchmark.py to include XLA compilation


LukeWood:

bhack:
Check:
https://github.com/keras-team/keras-cv/pull/161/files#r820872659

bhack:
It seems that in faster `jit_compiled` version all the `tf.cast` we have are still quite relevant

![immagine](https://user-images.githubusercontent.com/1710528/157131716-9bc840cc-5ff6-4ba5-a90e-6e2d6a7f5988.png)


![immagine](https://user-images.githubusercontent.com/1710528/157132032-467636b7-6fd6-47f2-9834-e8bc387d8102.png)

bhack:
@qlzh727 What do you think about adding an extra parameter to the base class for `jit_compile`?
https://github.com/keras-team/keras/blob/master/keras/layers/preprocessing/image_preprocessing.py#L413-L414

So that we could optionally use
```python 
f = def_function.function(self._augment, jit_compile=True)
self._map_fn(f, inputs)
```

LukeWood:
> @qlzh727 What do you think about adding an extra parameter to the base class for `jit_compile`? https://github.com/keras-team/keras/blob/master/keras/layers/preprocessing/image_preprocessing.py#L413-L414
> 
> So that we could optionally use
> 
> ```python
> f = def_function.function(self._augment, jit_compile=True)
> self._map_fn(f, inputs)
> ```

I personally would support a jit_compile attribute on the BaseImageAugmentationLayer.

so in a constructor you'd do:

jit_compile = True 

and it would work pass it to the tf.function wrapping `call`.

bhack:
We need to care about massively closing bug a we lost interesting threads. This is now at https://github.com/keras-team/keras/issues/16403

LukeWood:
> We need to care about massively closing bugs a we lose interesting threads. This is now at https://github.com/keras-team/keras/issues/16403

This is a distinct issue.  I do not think benchmarks do not need this as the benchmarks existed to guide our direction. 

you are right, BaseImageAugmentationLayer should support jit_compile, but the benchmarks do not need this so I closed this issue.

Closing bug is part of maintaining a clean project state.

bhack:
> Closing bug is part of maintaining a clean project state.

Yes but if there are still interesting parts emerged in the thread you need to spinoff that part on a new ticket or it will be lost forever.

PR #198, opened 2022-03-22
Implement check to skip value range transform

LukeWood:


qlzh727 (approved):
Probably add a simple test case, just for coverage.

LukeWood:
> Probably add a simple test case, just for coverage.

Yeah, I was wondering about that.  Any suggestion @qlzh727?  We’d have to inspect the graph, right?

bhack:
@qlzh727 How is this working between the preprocessing layers chain?

Are we going to have some important casting overhead as we have profiled at:
https://github.com/keras-team/keras-cv/issues/165#issuecomment-1061225089

qlzh727:
if the value range for all the image are (0, 255) for the input image, then it will benefit from this change (since it skip the conversion). When the input range are (0, 1), then any chain of the layer will keep converting them between (0, 1) and (0, 255). We might want to add some warning to user if this method has been hit several times in the pipeline, and suggest user to pre-convert their value to (0, 255). 

Anyway, this method is always good to have since it always skip the unnecessary conversion.

bhack:
> We might want to add some warning to user if this method has been hit several times in the pipeline, and suggest user to pre-convert their value to (0, 255).

Can we find a way to communicate/propagate between KLP layers that the range is converted?

bhack:
As I suppose that a chain of transformation is a quite common use case and it could be the O() of the casting overhead.

PR #201, opened 2022-03-23
Vectorize Equalization

### LINKED ISSUE #192: Migrate Equalization to use BaseImageAugmentationLayer
https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/preprocessing/equalization.py

quantumalaviya:
Port Equalization to BaseImageAugmentationLayer 
closes #192

### DIFF
--- a/keras_cv/layers/preprocessing/equalization.py
+++ b/keras_cv/layers/preprocessing/equalization.py
@@ -95,7 +95,9 @@ class Equalization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
         return result
 
     def augment_image(self, image, transformation=None):
-        image = preprocessing.transform_value_range(image, self.value_range, (0, 255), dtype=image.dtype)
+        image = preprocessing.transform_value_range(
+            image, self.value_range, (0, 255), dtype=image.dtype
+        )
         image = tf.cast(image, tf.int32)
         image = tf.vectorized_map(
             lambda channel: self.equalize_channel(image, channel),
--- a/keras_cv/layers/preprocessing/equalization.py
+++ b/keras_cv/layers/preprocessing/equalization.py
@@ -45,7 +45,10 @@ class Equalization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     def __init__(self, value_range, bins=256, **kwargs):
         super().__init__(**kwargs)
         self.bins = bins
+<<<<<<< HEAD
         self.value_range = value_range
+=======
+>>>>>>> Vectorizing Equalization
 
     @tf.function
     def equalize_channel(self, image, channel_index):
--- a/keras_cv/layers/preprocessing/equalization.py
+++ b/keras_cv/layers/preprocessing/equalization.py
@@ -19,7 +19,6 @@ from keras_cv.utils import preprocessing
 @tf.keras.utils.register_keras_serializable(package="keras_cv")
 class Equalization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     """Equalization performs histogram equalization on a channel-wise basis.
-
     Args:
         value_range: a tuple or a list of two elements. The first value represents
             the lower bound for values in passed images, the second represents the
@@ -27,16 +26,13 @@ class Equalization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
             `value_range`.
         bins: Integer indicating the number of bins to use in histogram equalization.
             Should be in the range [0, 256].
-
     Usage:
     ```python
     equalize = Equalization()
-
     (images, labels), _ = tf.keras.datasets.cifar10.load_data()
     # Note that images are an int8 Tensor with values in the range [0, 255]
     images = equalize(images)
     ```
-
     Call arguments:
         images: Tensor of pixels in range [0, 255], in RGB format.  Can be
             of type float or int.  Should be in NHWC format.
@@ -49,7 +45,6 @@ class Equalization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
 
     def equalize_channel(self, image, channel_index):
         """equalize_channel performs histogram equalization on a single channel.
-
         Args:
             image: int Tensor with pixels in range [0, 255], RGB format,
                 with channels last
@@ -64,7 +59,7 @@ class Equalization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
         # constant across input sizes for compatibility with vectorized_map
         histogram_without_zeroes = tf.where(
             tf.equal(histogram, 0),
-            tf.multiply(tf.ones((), tf.int32), 1410065408),
+            tf.ones((), tf.int32) * 1410065408,
             histogram,
         )
 
--- a/keras_cv/layers/preprocessing/equalization.py
+++ b/keras_cv/layers/preprocessing/equalization.py
@@ -45,10 +45,7 @@ class Equalization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     def __init__(self, value_range, bins=256, **kwargs):
         super().__init__(**kwargs)
         self.bins = bins
-<<<<<<< HEAD
         self.value_range = value_range
-=======
->>>>>>> Vectorizing Equalization
 
     @tf.function
     def equalize_channel(self, image, channel_index):
--- a/keras_cv/layers/preprocessing/equalization.py
+++ b/keras_cv/layers/preprocessing/equalization.py
@@ -47,6 +47,7 @@ class Equalization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
         self.bins = bins
         self.value_range = value_range
 
+    @tf.function
     def equalize_channel(self, image, channel_index):
         """equalize_channel performs histogram equalization on a single channel.
 
@@ -60,9 +61,15 @@ class Equalization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
         histogram = tf.histogram_fixed_width(image, [0, 255], nbins=self.bins)
 
         # For the purposes of computing the step, filter out the nonzeros.
-        nonzero = tf.where(tf.not_equal(histogram, 0))
-        nonzero_histogram = tf.reshape(tf.gather(histogram, nonzero), [-1])
-        step = (tf.reduce_sum(nonzero_histogram) - nonzero_histogram[-1]) // (
+        # Zeroes are replaced by 1410065408 while calculating min to keep shape
+        # constant across input sizes for compatibility with vectorized_map
+        histogram_without_zeroes = tf.where(
+            tf.equal(histogram, 0),
+            tf.multiply(tf.ones((), tf.int32), 1410065408),
+            histogram,
+        )
+
+        step = (tf.reduce_sum(histogram) - tf.reduce_min(histogram_without_zeroes)) // (
             self.bins - 1
         )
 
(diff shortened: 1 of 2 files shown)


kartik4949:
I can second this, the same error for tf.vectorized_map on my latest mask_to_boxes pr

qlzh727:
Thanks for the PR. I think the error might be similar as https://github.com/keras-team/keras-cv/pull/146#issuecomment-1061324744. Basically there are some intermidiate tensor has the dynamic shape that depends on the value of the input (eg the bucketing part). We should make it have the fixed shape, and then try the vectorized map, which should be way faster than the map_fn.

bhack:
> Thanks for the PR. I think the error might be similar as [#146 (comment)](https://github.com/keras-team/keras-cv/pull/146#issuecomment-1061324744). Basically there are some intermidiate tensor has the dynamic shape that depends on the value of the input (eg the bucketing part). We should make it have the fixed shape, and then try the vectorized map, which should be way faster than the map_fn.

/cc @wangpengmit

quantumalaviya:
It does give out these warnings while computing but I assumed it's not a big problem. Let me know if these need to be fixed, however.

```
WARNING:tensorflow:Using a while_loop for converting HistogramFixedWidth
WARNING:tensorflow:Using a while_loop for converting HistogramFixedWidth
WARNING:tensorflow:Using a while_loop for converting HistogramFixedWidth
```

bhack:
I suppose that `tf.histogram_fixed_width` Is not supported so we could have some performance drain.

quantumalaviya:
Makes sense, I wonder if designing a supported util for computing histograms will make a huge difference

bhack:
I think you can open a related ticket on tensorflow to track this with a very minimal example.

bhack:
We are also trying to give some hints on the root cause of the fallback with https://github.com/tensorflow/tensorflow/pull/55192

keras_cv/layers/preprocessing/equalization.py
```diff
@@ -53,9 +54,15 @@ def equalize_channel(self, image, channel_index):
         histogram = tf.histogram_fixed_width(image, [0, 255], nbins=self.bins)
 
         # For the purposes of computing the step, filter out the nonzeros.
-        nonzero = tf.where(tf.not_equal(histogram, 0))
-        nonzero_histogram = tf.reshape(tf.gather(histogram, nonzero), [-1])
-        step = (tf.reduce_sum(nonzero_histogram) - nonzero_histogram[-1]) // (
+        # Zeroes are replaced by 1410065408 while calculating min to keep shape
```
qlzh727, line 57: Can u add more comments about how we come up this magic number (1410065408)?

qlzh727 (changes requested):

keras_cv/layers/preprocessing/equalization.py
```diff
@@ -79,13 +86,15 @@ def build_mapping(histogram, step):
...
 
-    def call(self, images):
+    def augment_image(self, image, transformation=None):
         # Assumes RGB for now.  Scales each channel independently
         # and then stacks the result.
         # TODO(lukewood): ideally this would be vectorized.
-        r = tf.map_fn(lambda x: self.equalize_channel(x, 0), images)
-        g = tf.map_fn(lambda x: self.equalize_channel(x, 1), images)
-        b = tf.map_fn(lambda x: self.equalize_channel(x, 2), images)
+        dtype = image.dtype
```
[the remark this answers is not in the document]
LukeWood: Good point - yeah absolutely should do this.

keras_cv/layers/preprocessing/equalization.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, bins=256, **kwargs):
         super().__init__(**kwargs)
         self.bins = bins
 
+    @tf.function
```
LukeWood, line 42: Does this need tf.function?

LukeWood:
Thanks for the PR!

LukeWood (changes requested):

keras_cv/layers/preprocessing/equalization.py
```diff
@@ -79,13 +86,15 @@ def build_mapping(histogram, step):
...
+    def augment_image(self, image, transformation=None):
         # Assumes RGB for now.  Scales each channel independently
         # and then stacks the result.
         # TODO(lukewood): ideally this would be vectorized.
-        r = tf.map_fn(lambda x: self.equalize_channel(x, 0), images)
-        g = tf.map_fn(lambda x: self.equalize_channel(x, 1), images)
-        b = tf.map_fn(lambda x: self.equalize_channel(x, 2), images)
+        dtype = image.dtype
+        image = tf.cast(image, tf.int32)
+        r = self.equalize_channel(image, 0)
```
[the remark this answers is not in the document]
quantumalaviya: This is done. However, currently, I am using a nested `tf.vectorized_map`. The warnings are as follows now:

```
WARNING:tensorflow:Using a while_loop for converting StridedSlice
WARNING:tensorflow:Using a while_loop for converting HistogramFixedWidth
WARNING:tensorflow:Using a while_loop for converting HistogramFixedWidth
WARNING:tensorflow:Using a while_loop for converting DynamicPartition
WARNING:tensorflow:Using a while_loop for converting DynamicStitch
```
I hope this isn't a problem

keras_cv/layers/preprocessing/equalization.py
```diff
@@ -53,9 +53,15 @@ def equalize_channel(self, image, channel_index):
...
-        step = (tf.reduce_sum(nonzero_histogram) - nonzero_histogram[-1]) // (
+        # Zeroes are replaced by 1410065408 while calculating min to keep shape
+        # constant across input sizes for compatibility with vectorized_map
+        histogram_without_zeroes = tf.where(
+            tf.equal(histogram, 0),
+            tf.multiply(tf.ones((), tf.int32), 1410065408),
+            histogram,
+        )
+
+        step = (tf.reduce_sum(histogram) - tf.reduce_min(histogram_without_zeroes)) // (
```
qlzh727, line 64: I don't think this update is exactly the same as the original implementation. 

The original nonzero_histogram was a vector (which can have dynamic length based on the input value), we would like to get a fix length nonzero_histogram. The `nonzero_histogram[-1]` in original implementation is not exactly same as the `tf.reduce_min(histogram_without_zeroes)`

quantumalaviya: That makes sense. I referred to a different source that used the minimum non-zero value. The results seem to be similar.

Do you suggest I somehow find the last nonzero value from the histogram instead? I believe it can be done using a loop or something similar to ensure that an intermediate tensor doesn't interfere with `vectorized_map`.

qlzh727: @LukeWood  for why we choose to use the last nonzero element in histogram. I would like to make sure we are mathematically correct.

LukeWood: Yeah, we probably need to give this layer another look and add unit tests over the behavior.  This is a tricky layer.

LukeWood: I am not 100% confident that we have the right implementation.

quantumalaviya: The implementation I was aware of involves division of the histogram by the total number of pixels, followed by cumulative sum and multiplication with the highest possible value in the range (here, 255). Since this method seems to be different compared to what several open-source implementations follow, I have been completely referencing sources online. 

If there is any single source I should follow that explains the desired variation, I will code the approach in a follow-up commit. I have noticed all of these approaches correctly do the job of expanding the intensity range, however.

keras_cv/layers/preprocessing/equalization.py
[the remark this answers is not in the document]
innat: @qlzh727 @LukeWood 
Is the above warning message safe to ignore? Does the end-user need to worry about anything regarding this? If it's 100% okay to ignore, then I think the warning message should not appear in the first place. Otherwise, it would be a bad UX.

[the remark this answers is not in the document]
kartik4949: It means that particular ops is not supported yet hence falling to while loop 
This does not affect the result but drain some performance.

Well if there are too many fallbacks to while loop, i suspect we wont have any/substantial  vectorization gain, (I might be wrong :) )
May be because too many fallback latency.

May be a benchmark, needed. 
@innat

keras_cv/layers/preprocessing/equalization.py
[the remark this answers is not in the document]
quantumalaviya: #212

keras_cv/layers/preprocessing/equalization.py
quantumalaviya: @LukeWood are we keeping this PR on hold until #210 is resolved?

keras_cv/layers/preprocessing/equalization.py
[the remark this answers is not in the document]
bhack: We have talked about this also at https://github.com/keras-team/keras-cv/pull/146#issuecomment-1049352918 and we are also trying to improve the feedback on the root cause of the fallback with https://github.com/tensorflow/tensorflow/pull/55192

The main issue also is that `tf.vectorized_map` [is currently orphan](https://github.com/keras-team/keras-cv/pull/146#issuecomment-1068486128)

keras_cv/layers/preprocessing/equalization.py
LukeWood: Yeah, maybe I will be taking a stab at this tomorrow.   I think theres a good chance I drastically rewrite the layer.  It looks like the logic might actually be incorrect.

quantumalaviya: > The implementation I was aware of involves division of the histogram by the total number of pixels, followed by cumulative sum and multiplication with the highest possible value in the range (here, 255). 

Do you want me to implement this method? Might save you some time @LukeWood

LukeWood: https://github.com/keras-team/keras-cv/pull/234 will handle it - thanks for offering

LukeWood: Done - Can you rebase

LukeWood: (Rebase and retitle vectorize Equalization now that this is what this PR does)

LukeWood (changes requested):

keras_cv/layers/preprocessing/equalization.py
quantumalaviya: This is done! @LukeWood

keras_cv/layers/preprocessing/equalization.py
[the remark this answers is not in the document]
LukeWood: So, we should really mute this errors.  BUT I don't think it is blocking for now.  So resolving for now, but this is a tricky issue because it falls on the TensorFlow side, not our side.

keras_cv/layers/preprocessing/equalization.py
```diff
@@ -60,9 +60,15 @@ def equalize_channel(self, image, channel_index):
...
 
         # For the purposes of computing the step, filter out the nonzeros.
-        nonzero = tf.where(tf.not_equal(histogram, 0))
-        nonzero_histogram = tf.reshape(tf.gather(histogram, nonzero), [-1])
-        step = (tf.reduce_sum(nonzero_histogram) - nonzero_histogram[-1]) // (
+        # Zeroes are replaced by 1410065408 while calculating min to keep shape
+        # constant across input sizes for compatibility with vectorized_map
+        histogram_without_zeroes = tf.where(
+            tf.equal(histogram, 0),
+            tf.multiply(tf.ones((), tf.int32), 1410065408),
```
LukeWood, line 67: can you replace this with `tf.ones(()) * 1410065408`

LukeWood, line 67: Can we add a tiny test case asserting that 1410065408 shows up nowhere in the result?  Just pass a random uniform distribution with value_range=(0, 255) and then assert 1410065408 doesn't show up,

LukeWood (changes requested):

LukeWood:
Apologies for the delay @quantumalaviya I think this looks pretty good.  I'm a little concerned about that hard coded value in there but if we can add a test case to ensure it doesn't show up I'll feel pretty confident in it.  Can we also include a demo with a side by side image - one augmented one not?

keras_cv/layers/preprocessing/equalization.py
quantumalaviya: Done! I added a more general testcase that ensures that the resultant output range is the same. This will exclude 1410065408 as well.

However, 1410065408 is not really interfering with the range and simply changing the frequency.

quantumalaviya: Done. Added tf.int32 to ensure the same dtype, however.

quantumalaviya:
I agree, I am also not a fan of having the hardcoded value. However, it just serves the purpose of being a really big number to exclude zeroes.

On another note, I also added `tf.clip_by_value(lookup_table, self.value_range[0], self.value_range[1])` instead of `tf.clip_by_value(lookup_table, 0, 255)`  so that this function isn't hardcoded across one value range.

LukeWood:
> I agree, I am also not a fan of having the hardcoded value. However, it just serves the purpose of being a really big number to exclude zeroes.

Thanks for the explanation.

LukeWood:
Ok @quantumalaviya - I updated the PR with some copyedits and minor fixes around dtype handling.  Also added some new test cases to cover the correctness.  

Thanks for the great PR!  Good idea to add a big value to mask out 0s.  I appreciate the contribution!

LukeWood:
Thanks you @quantumalaviya !!!

PR #204, opened 2022-03-24
Fix overridden method in RandomShear and RandomSharpness

chjort:
Tests were failing due to these typos in overridden method names. Fixing these before any Pull requests stumbles into them.


LukeWood:
yep - yeah there was a commit to keras core causing this.  Thank

PR #207, opened 2022-03-24
Migrate RandomCutout to use BaseImageAugmentationLayer

### LINKED ISSUE #189: Migrate RandomCutout to use BaseImageAugmentationLayer
https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/preprocessing/random_cutout.py

Example in RandomShear

quantumalaviya:
closes #189

### DIFF
--- a/keras_cv/layers/preprocessing/random_cutout.py
+++ b/keras_cv/layers/preprocessing/random_cutout.py
@@ -12,13 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import tensorflow as tf
-import tensorflow.keras.layers as layers
-from tensorflow.keras import backend
 
 from keras_cv.utils import fill_utils
 
 
-class RandomCutout(layers.Layer):
+class RandomCutout(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     """Randomly cut out rectangles from images and fill them.
 
     Args:
@@ -162,17 +160,14 @@ class RandomCutout(layers.Layer):
         else:
             return type(factor)(0), factor
 
-    def call(self, inputs, training=True):
-        if training is None:
-            training = backend.learning_phase()
+    def get_random_transformation(self):
+        return self._sample_num_cutouts()
 
-        augment = lambda: self._random_cutout(inputs)
-        no_augment = lambda: inputs
-        return tf.cond(tf.cast(training, tf.bool), augment, no_augment)
-
-    def _random_cutout(self, inputs):
+    @tf.function
+    def augment_image(self, image, transformation=None):
         """Apply random cutout."""
-        for _ in tf.range(self._sample_num_cutouts()):
+        inputs = tf.expand_dims(image, 0)
+        for _ in tf.range(transformation):
             center_x, center_y = self._compute_rectangle_position(inputs)
             rectangle_height, rectangle_width = self._compute_rectangle_size(inputs)
             rectangle_fill = self._compute_rectangle_fill(inputs)
@@ -184,7 +179,13 @@ class RandomCutout(layers.Layer):
                 rectangle_height,
                 rectangle_fill,
             )
-        return inputs
+        return inputs[0]
+
+    def _sample_num_cutouts(self):
+        num_cutouts = self._random_generator.random_uniform(
+            (), self.num_cutouts_lower, self.num_cutouts_upper, dtype=tf.int32
+        )
+        return num_cutouts
 
     def _compute_rectangle_position(self, inputs):
         input_shape = tf.shape(inputs)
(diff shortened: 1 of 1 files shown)


quantumalaviya:
Vectorizing across `num_cutouts` might be useful here.

keras_cv/layers/preprocessing/random_cutout.py
```diff
@@ -162,17 +160,14 @@ def _parse_bounds(self, factor):
...
-            training = backend.learning_phase()
+    def get_random_transformation(self):
+        return self._sample_num_cutouts()
 
-        augment = lambda: self._random_cutout(inputs)
-        no_augment = lambda: inputs
-        return tf.cond(tf.cast(training, tf.bool), augment, no_augment)
-
-    def _random_cutout(self, inputs):
+    @tf.function
```
bhack, line 166: @qlzh727 I suppose that the model is already wrapped by default in a `tf.function` at build time (later we need to  identify how we want to control to activate `jit_compile` on a specific region/function)

bhack:
> Vectorizing across num_cutouts might be useful here.

Is there any limit to achieve this?

quantumalaviya:
> > Vectorizing across num_cutouts might be useful here.
> 
> Is there any limit to achieve this?

No limit, it will require changes inside fill_utils. I believe @parikshit14 has already worked on the implementation.

keras_cv/layers/preprocessing/random_cutout.py
qlzh727: yea, I don't think the tf.function is necessary here.

keras_cv/layers/preprocessing/random_cutout.py
```diff
@@ -162,17 +160,14 @@ def _parse_bounds(self, factor):
         else:
             return type(factor)(0), factor
 
-    def call(self, inputs, training=True):
-        if training is None:
-            training = backend.learning_phase()
+    def get_random_transformation(self):
```
qlzh727, line 163: I think this is probably not correct. This function should generate the transformation for one image, and in future bbox and label with it. So it should not only contain the number of images, but also the position/size of the cutout. When the augment_image got the transformation, it can augment the image without any random behavior involved.

qlzh727: it seems that this method might need the actual input (for the size of the image) to produce all the transformations. We probably will need to update the method signature in base class. @LukeWood

qlzh727 (changes requested):

keras_cv/layers/preprocessing/random_cutout.py
LukeWood: Ah I see - this is an interesting issue.  I guess we could `vectorized_map(images, get_random_transformation)`?

keras_cv/layers/preprocessing/random_cutout.py
LukeWood: yeah we only decorate components with tf.function if we intend for them to be used in isolation AND they are computationally expensive.

i.e. coco metrics - they will be used outside of models most likely due to the complexity of BBOX models and they are extremely expensive so best to compile them.  Other than rare cases no need to decorate code with tf.function

quantumalaviya: I am not sure why but if the function is not decorated with `tf.function`, the graph doesn't support the loop.
This problem persists when I create a one-layer model that applies CutOut:

```
OperatorNotAllowedInGraphError: Exception encountered when calling layer "random_cutout" (type RandomCutout).

Iterating over a symbolic `tf.Tensor` is not allowed in Graph execution. Use Eager execution or decorate this function with @tf.function.

Call arguments received by layer "random_cutout" (type RandomCutout):
  • inputs=tf.Tensor(shape=(None, 504, 672, 3), dtype=float32)
  • training=True
```

bhack: I am working on this at:
https://github.com/tensorflow/tensorflow/pull/55350

I hope It could cover also this case.

quantumalaviya: No point in merging this PR without the decorator as long as that isn't covered, right?

keras_cv/layers/preprocessing/random_cutout.py
qlzh727: Sending https://github.com/keras-team/keras/pull/16299.

LukeWood: https://github.com/keras-team/keras/pull/16299 is now merged and thus this issue can be solved.  @quantumalaviya

LukeWood: you will need to upgrade to the latest keras-nightly.

quantumalaviya: I'll work on this. I think I will wait for #217 to be completely merged before making the changes.

keras_cv/layers/preprocessing/random_cutout.py
LukeWood: no, we should add a test case to cover this as well.  Will you add a test case running cutout in a model?

quantumalaviya: Sure, I will add the one-layer model in the test case!

keras_cv/layers/preprocessing/random_cutout.py
LukeWood: it is as merged as it will be for awhile - thanks.  Just need to upgrade keras nightly!

LukeWood: I've updated this - we should remove num_cutouts for now.  It doesn't work as we want

LukeWood:
merged and rebased in https://github.com/keras-team/keras-cv/pull/236

LukeWood:
Thanks for the PR!

keras_cv/layers/preprocessing/random_cutout.py
quantumalaviya: @LukeWood This hasn't been done yet, has it? I can create a pull request for it. 
I wonder if these model testcases would be useful for all layers.

PR #213, opened 2022-03-25
Update solarization in preparation for use in RandAugment

### LINKED ISSUE #211: Solarization must support value_range

LukeWood:
fixes #211

### DIFF
--- a/keras_cv/layers/preprocessing/solarization.py
+++ b/keras_cv/layers/preprocessing/solarization.py
@@ -13,8 +13,10 @@
 # limitations under the License.
 import tensorflow as tf
 
+from keras_cv.utils import preprocessing
 
-class Solarization(tf.keras.layers.Layer):
+
+class Solarization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     """Applies (max_value - pixel + min_value) for each pixel in the image.
 
     When created without `threshold` parameter, the layer performs solarization to
@@ -25,22 +27,29 @@ class Solarization(tf.keras.layers.Layer):
     - [AutoAugment: Learning Augmentation Policies from Data](
         https://arxiv.org/abs/1805.09501
     )
+    - [RandAugment](https://arxiv.org/pdf/1909.13719.pdf)
 
     Args:
-        threshold: (Optionally) int or float. If specified, only pixel values above this
+        addition: (Optional) int or float.  If specified, this value is added to each
+            pixel before solarization and thresholding.  The addition value should be
+            scaled accoring to the value range (0, 255).  Defaults to 0.0.
+        threshold: (Optional) int or float. If specified, only pixel values above this
             threshold will be solarized.
+        value_range: a tuple or a list of two elements. The first value represents
+            the lower bound for values in passed images, the second represents the
+            upper bound. Images passed to the layer should have values within
+            `value_range`. Defaults to `(0, 255)`.
 
     Usage:
-
     ```python
-        (images, labels), _ = tf.keras.datasets.cifar10.load_data()
-        print(images[0, 0, 0])
-        # [59 62 63]
-        # Note that images are Tensor with values in the range [0, 255]
-        solarization = Solarization()
-        images = solarization(images)
-        print(images[0, 0, 0])
-        # [196, 193, 192]
+    (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    print(images[0, 0, 0])
+    # [59 62 63]
+    # Note that images are Tensor with values in the range [0, 255]
+    solarization = Solarization()
+    images = solarization(images)
+    print(images[0, 0, 0])
+    # [196, 193, 192]
     ```
 
     Call arguments:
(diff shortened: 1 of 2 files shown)


keras_cv/layers/preprocessing/solarization.py
```diff
@@ -49,16 +58,31 @@ class Solarization(tf.keras.layers.Layer):
             or [height, width, channels].
     """
 
-    def __init__(self, threshold=None):
+    def __init__(
+        self,
+        addition=0.0,
+        threshold=None,
```
qlzh727, line 64: if the value should be int or float, should we have default value to be 0.0?

keras_cv/layers/preprocessing/solarization.py
```diff
@@ -49,16 +58,31 @@ class Solarization(tf.keras.layers.Layer):
...
 
-    def call(self, images):
-        images = tf.clip_by_value(images, clip_value_min=0, clip_value_max=255)
+    def augment_image(self, image, transformation=None):
+        image = preprocessing.transform_value_range(
+            image, original_range=self.value_range, target_range=(0, 255)
+        )
+        result = image + self.addition
+        result = tf.clip_by_value(result, 0, 255)
         if self.threshold is None:
```
qlzh727, line 78: if we have the default value to be 0, then we can actually remove this condition and always use the _solarize_above_threshold.

keras_cv/layers/preprocessing/solarization_test.py
```diff
@@ -36,14 +36,21 @@ def test_range_0_to_255(self):
...
 
-    @staticmethod
-    def _test_input_output(layer, input_value, expected_value, dtype):
-        dummy_input = tf.ones(shape=(2, 224, 224, 3), dtype=dtype) * input_value
-        expected_output = tf.ones(shape=(2, 224, 224, 3), dtype=dtype) * expected_value
+    def test_solarization_with_addition(self):
+        solarization = Solarization(addition=10.0)
 
-        output = layer(dummy_input)
+        test_parameters = [
```
qlzh727, line 42: I think we should consider using absl parameterized to config the test case. It will generate one test case per param setting. In the current setting, the failing test in front will block the following test cases.

https://github.com/abseil/abseil-py/blob/3e4749af698201224165a98930c6845d8c102e0d/absl/testing/parameterized.py#L73

qlzh727 (changes requested):
Thanks for the PR.

keras_cv/layers/preprocessing/solarization_test.py
kartik4949: I will second for parametized test cases, useful in refactoring other test cases as well

keras_cv/layers/preprocessing/solarization.py
LukeWood: Good point, fixed

keras_cv/layers/preprocessing/solarization.py
LukeWood: Fixed

keras_cv/layers/preprocessing/solarization_test.py
LukeWood: done - thanks for the suggestion, this is great.

keras_cv/layers/preprocessing/solarization.py
```diff
@@ -49,16 +58,28 @@ class Solarization(tf.keras.layers.Layer):
...
+        )
+        result = image + self.addition
+        result = tf.clip_by_value(result, 0, 255)
+        result = self._solarize_above_threshold(result)
+        result = preprocessing.transform_value_range(
+            result, original_range=(0, 255), target_range=self.value_range
+        )
+        return result
 
     def _solarize(self, images):
```
qlzh727, line 84: Should we remove this method since this is one-liner and only called by _solarize_above_threshold?

keras_cv/layers/preprocessing/solarization.py
```diff
@@ -67,6 +88,10 @@ def _solarize_above_threshold(self, images):
         return tf.where(images < self.threshold, images, self._solarize(images))
```
qlzh727, line 88: Maybe rename this method to _solarize().

keras_cv/layers/preprocessing/solarization.py
LukeWood: Good point

keras_cv/layers/preprocessing/solarization.py
LukeWood: Done

keras_cv/layers/preprocessing/solarization_test.py
```diff
@@ -12,72 +12,71 @@
...
-                layer=solarization,
-                input_value=parameters["input_value"],
-                expected_value=parameters["expected_output_value"],
-                dtype=tf.uint8,
-            )
-
-    def test_input_values_outside_of_specified_range_are_clipped(self):
-        solarization = Solarization()
+    def _test_input_output(self, layer, input_value, expected_value, dtype):
+        dummy_input = tf.ones(shape=(2, 224, 224, 3), dtype=dtype) * input_value
```
qlzh727, line 70: Not sure why this is called dummy input, seems to be normal input.

keras_cv/layers/preprocessing/solarization_test.py
```diff
@@ -12,72 +12,71 @@
...
+from absl.testing import parameterized
 
 from keras_cv.layers.preprocessing.solarization import Solarization
 
 
-class SolarizationTest(tf.test.TestCase):
-    def test_range_0_to_255(self):
+class SolarizationTest(tf.test.TestCase, parameterized.TestCase):
+    @parameterized.named_parameters(
+        ("0_255", 0, 255),
```
qlzh727, line 22: The naming of the test case doesn't seems to carry any meaningful info.

qlzh727 (approved):

keras_cv/layers/preprocessing/solarization_test.py
LukeWood: yeah-  good point updated.

keras_cv/layers/preprocessing/solarization_test.py
LukeWood: how so?  They tell you which values fail.  They get postpended: test_only_values_above_threshold_are_solarized_0_0

PR #217, opened 2022-03-25
vectorized num_cutouts

parikshit14:
Vectorized implementation of `num_cutouts` solves issue #134 
Might help with issue #189 PR #207 , #186 

**Note**:
The changes from this PR disrupt `RandomCutMix` as that uses `rectangle_masks` and `fill_rectangle` from `fill_utils.py` which has now been vectorized for num_cutouts. Which I think can be solved using (non-exhaustive):
- we can create **FLAG** sort of a variable to separate both of these(RandomCutout and CutMix) inside `fill_utils` 

- we can create a separate file for the vectorized `fill_utils`.

### DIFF
--- a/keras_cv/layers/preprocessing/random_cutout.py
+++ b/keras_cv/layers/preprocessing/random_cutout.py
@@ -152,15 +152,17 @@ class RandomCutout(layers.Layer):
         self.fill_value = fill_value
         self.seed = seed
         self.num_cutouts = num_cutouts
-        if isinstance(num_cutouts, tuple):
-            self.num_cutouts = tf.random.uniform(
+
+    def get_random_transformation(self):
+        if isinstance(self.num_cutouts, tuple):
+            num_cutouts = tf.random.uniform(
                 shape=(1,),
                 minval=self.num_cutouts_lower,
                 maxval=self.num_cutouts_upper + 1,
                 dtype=tf.int32,
                 seed=self.seed,
             )
-            self.num_cutouts = self.num_cutouts[0]
+            self.num_cutouts = num_cutouts[0]
 
     def _parse_bounds(self, factor):
         if isinstance(factor, (tuple, list)):
@@ -172,6 +174,7 @@ class RandomCutout(layers.Layer):
         if training is None:
             training = backend.learning_phase()
 
+        self.get_random_transformation()
         augment = lambda: self._random_cutout(inputs)
         no_augment = lambda: inputs
         return tf.cond(tf.cast(training, tf.bool), augment, no_augment)
--- a/keras_cv/layers/preprocessing/random_cutout.py
+++ b/keras_cv/layers/preprocessing/random_cutout.py
@@ -76,10 +76,6 @@ class RandomCutout(layers.Layer):
         self.height_lower, self.height_upper = self._parse_bounds(height_factor)
         self.width_lower, self.width_upper = self._parse_bounds(width_factor)
         self.num_cutouts_lower, self.num_cutouts_upper = self._parse_bounds(num_cutouts)
-        self.num_cutouts_lower = (
-            self.num_cutouts_lower if self.num_cutouts_lower != 0 else 1
-        )
-        self.num_cutouts_upper += 1
 
         if fill_mode not in ["gaussian_noise", "constant"]:
             raise ValueError(
@@ -155,6 +151,16 @@ class RandomCutout(layers.Layer):
         self.fill_mode = fill_mode
         self.fill_value = fill_value
         self.seed = seed
+        self.num_cutouts = num_cutouts
+        if isinstance(num_cutouts, tuple):
+            self.num_cutouts = tf.random.uniform(
+                shape=(1,),
+                minval=self.num_cutouts_lower,
+                maxval=self.num_cutouts_upper + 1,
+                dtype=tf.int32,
+                seed=self.seed,
+            )
+            self.num_cutouts = self.num_cutouts[0]
 
     def _parse_bounds(self, factor):
         if isinstance(factor, (tuple, list)):
@@ -172,18 +178,17 @@ class RandomCutout(layers.Layer):
 
     def _random_cutout(self, inputs):
         """Apply random cutout."""
-        for _ in tf.range(self._sample_num_cutouts()):
-            center_x, center_y = self._compute_rectangle_position(inputs)
-            rectangle_height, rectangle_width = self._compute_rectangle_size(inputs)
-            rectangle_fill = self._compute_rectangle_fill(inputs)
-            inputs = fill_utils.fill_rectangle(
-                inputs,
-                center_x,
-                center_y,
-                rectangle_width,
-                rectangle_height,
-                rectangle_fill,
-            )
+        center_x, center_y = self._compute_rectangle_position(inputs)
+        rectangle_height, rectangle_width = self._compute_rectangle_size(inputs)
+        rectangle_fill = self._compute_rectangle_fill(inputs)
+        inputs = fill_utils.fill_rectangle(
+            inputs,
+            center_x,
+            center_y,
+            rectangle_width,
+            rectangle_height,
+            rectangle_fill,
+        )
         return inputs
 
     def _compute_rectangle_position(self, inputs):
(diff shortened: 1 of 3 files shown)


chjort:
I would actually suggest a couple of refactors to achieve this vectorization, and at the same time
also make rectangle filling more flexible.

First would be a refactor of layers using `fill_utils.fill_rectangle`. Right now this includes `CutMix` and
`RandomCutOut`. I suggest the layers are now responsible for computing the bounding boxes (rectangles) in corners
format, and then directly use `fill_utils.rectangle_masks` instead of `fill_utils.fill_rectangle`.

Specifically, in `CutMix._cutmix(self, images, labels)` replace the following piece of code on lines 112-119
```python
images = fill_utils.fill_rectangle(
    images,
    random_center_width,
    random_center_height,
    cut_width,
    cut_height,
    tf.gather(images, permutation_order),
)
```
with this piece of code:
```python
# convert to corners bounding boxes
xywh = tf.stack([random_center_width, random_center_height, cut_width, cut_height], axis=1)
xywh = tf.cast(xywh, tf.float32)
corners = bounding_box.xywh_to_corners(xywh)

# compute bounding box masks
input_shape = tf.shape(images)
batch_size, image_height, image_width = (
    input_shape[0],
    input_shape[1],
    input_shape[2],
)
mask_shape = (image_width, image_height)
is_rectangle = fill_utils.rectangle_masks(corners, mask_shape)
is_rectangle = tf.expand_dims(is_rectangle, axis=-1)

# fill bounding boxes
images = tf.where(is_rectangle, tf.gather(images, permutation_order), images)
```
essentially unpacking `fill_utils.fill_rectangle` directly into the layer. 

And specifically for the `RandomCutout._random_cutout(self, inputs)` in this PR, replace
```python
inputs = fill_utils.fill_rectangle(
    inputs,
    center_x,
    center_y,
    rectangle_width,
    rectangle_height,
    rectangle_fill,
)
```
with this
```python
# convert to corners bounding boxes
xywh = tf.stack([center_x, center_y, rectangle_width, rectangle_height], axis=1)
xywh = tf.cast(xywh, tf.float32)
corners = bounding_box.xywh_to_corners(xywh)

# compute bounding box masks
input_shape = tf.shape(inputs)
batch_size, image_height, image_width = (
    input_shape[0],
    input_shape[1],
    input_shape[2],
)
mask_shape = (image_width, image_height)
is_rectangle = fill_utils.rectangle_masks(corners, mask_shape)

# merge num_cutouts bounding boxes
is_rectangle = tf.reshape(is_rectangle, [batch_size, self.num_cutouts, image_width, image_height])
is_rectangle = tf.reduce_any(is_rectangle, axis=1)
is_rectangle = tf.expand_dims(is_rectangle, axis=-1)

# fill bounding boxes
inputs = tf.where(is_rectangle, rectangle_fill, inputs)
```
while also changing lines [L202](https://github.com/keras-team/keras-cv/blob/1e87cbe6871e311a96b3011cda5ed01db51c4065/keras_cv/layers/preprocessing/random_cutout.py#L202), [L209](https://github.com/keras-team/keras-cv/blob/1e87cbe6871e311a96b3011cda5ed01db51c4065/keras_cv/layers/preprocessing/random_cutout.py#L209), [L225](https://github.com/keras-team/keras-cv/blob/1e87cbe6871e311a96b3011cda5ed01db51c4065/keras_cv/layers/preprocessing/random_cutout.py#L225), and [L231](https://github.com/keras-team/keras-cv/blob/1e87cbe6871e311a96b3011cda5ed01db51c4065/keras_cv/layers/preprocessing/random_cutout.py#L231)
from `shape=[batch_size, self.num_cutouts]` to `shape=[batch_size * self.num_cutouts]`.

This way we simply compute masks for all cutouts in a flattened manner (batch_size * num_cutouts), and then reshape masks
so we can reduce them on num_cutout axis to have a single mask per image.

The second, but optional refactor, is to rename `keras_cv.utils.fill_utils.rectangle_masks` into `keras_cv.utils.bounding_box.corners_to_mask` as this seems to be where it belongs.

Any thoughts on this @LukeWood?

keras_cv/utils/fill_utils.py
```diff
@@ -33,24 +33,28 @@ def rectangle_masks(corners, mask_shape):
     corners = corners[..., tf.newaxis, tf.newaxis]
 
     # split coordinates
-    x0 = corners[:, 0]
-    y0 = corners[:, 1]
-    x1 = corners[:, 2]
-    y1 = corners[:, 3]
+    x0 = corners[:, :, 0]
```
LukeWood, line 36: can we instead do `[..., 0]`?  Does that work or cause issues?

keras_cv/utils/fill_utils.py
```diff
@@ -33,24 +33,28 @@ def rectangle_masks(corners, mask_shape):
...
+    x1 = corners[:, :, 2]
+    y1 = corners[:, :, 3]
 
     # repeat height and width
     width, height = mask_shape
-    x0_rep = tf.repeat(x0, height, axis=1)
-    y0_rep = tf.repeat(y0, width, axis=2)
-    x1_rep = tf.repeat(x1, height, axis=1)
-    y1_rep = tf.repeat(y1, width, axis=2)
+    x0_rep = tf.repeat(x0, height, axis=2)
```
LukeWood, line 43: while we are updating this lets rename these variables to `x0_repeat`.  In keras we avoid abbreviations most of the time.

keras_cv/utils/fill_utils.py
```diff
@@ -33,24 +33,28 @@ def rectangle_masks(corners, mask_shape):
...
+    y0_rep = tf.repeat(y0, width, axis=3)
```
LukeWood, line 44: can we maybe use negative indexes here so that the number of dimensions in `x0` and `y0` etc don't matter?

LukeWood (changes requested):
Thanks for the PR.  I've left some minor comments.

parikshit14:
@chjort thanks for the suggestions, my only concern though is if  `shape=[batch_size * self.num_cutouts]`  is better then `shape=[batch_size, self.num_cutouts]` or not.

keras_cv/utils/fill_utils.py
parikshit14: yes it causes some errors as 
`x0 = corners[...,0] `gives shape **(None, num_cutouts, 4, 1)** 
`x0 = corners[:,:,0]` gives shape **(None, num_cutouts, 1, 1)**

LukeWood:
> 

it is the same.

keras_cv/layers/preprocessing/random_cutout.py
```diff
@@ -155,6 +151,16 @@ def __init__(
...
+        self.num_cutouts = num_cutouts
+        if isinstance(num_cutouts, tuple):
+            self.num_cutouts = tf.random.uniform(
+                shape=(1,),
+                minval=self.num_cutouts_lower,
+                maxval=self.num_cutouts_upper + 1,
+                dtype=tf.int32,
+                seed=self.seed,
+            )
+            self.num_cutouts = self.num_cutouts[0]
```
LukeWood, line 163: This doesn't look right - it looks like you always assign to self.num_Countouts[0]

keras_cv/layers/preprocessing/random_cutout.py
```diff
@@ -194,31 +199,21 @@ def _compute_rectangle_position(self, inputs):
             input_shape[2],
         )
         center_x = tf.random.uniform(
-            shape=[batch_size],
+            shape=[batch_size, self.num_cutouts],
```
LukeWood, line 202: self.num_cutouts is baked into the graph here.  You'll need to leverage `get_random_transformation` to handle the case where ranges are supported.  You can see an example of how this works in `random_shear.py`

keras_cv/layers/preprocessing/random_cutout.py
```diff
@@ -155,6 +151,16 @@ def __init__(
         self.fill_mode = fill_mode
         self.fill_value = fill_value
         self.seed = seed
+        self.num_cutouts = num_cutouts
+        if isinstance(num_cutouts, tuple):
+            self.num_cutouts = tf.random.uniform(
```
LukeWood, line 156: This sampling should occur inside `get_random_transformation`.

parikshit14:
@LukeWood updated `cut_mix` by adding extra dimension to make it compatible with the vectorized fill_utils, but couldn't actually follow what you meant by 

> self.num_cutouts is baked into the graph here. You'll need to leverage `get_random_transformation` to handle the case where ranges are supported. You can see an example of how this works in `random_shear.py`

anyways did other changes suggested

innat:
@parikshit14 Great work. 
Is it possible to share some demo output of this addition in the cutout?

innat:
@parikshit14 Great work. 
Is it possible to share some demo output of this addition in the cutout?

LukeWood:
Instead of this API I think I'd rather just support `Repeat(CutOut(), 3)` instead.  High level abstractions are more reusable and maintainable IMO.

PR #218, opened 2022-03-26
Migrate channel_shuffle to use BaseImageAugmentationLayer

### LINKED ISSUE #187: Migrate channel_shuffle.py to use BaseImageAugmentationLayer
example available in:
https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/preprocessing/random_shear.py

quantumalaviya:
closes #187

### DIFF
--- a/keras_cv/layers/preprocessing/channel_shuffle.py
+++ b/keras_cv/layers/preprocessing/channel_shuffle.py
@@ -52,8 +52,8 @@ class ChannelShuffle(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
         self.seed = seed
 
     def augment_image(self, image, transformation=None):
-        height = tf.shape(image)[0]
-        width = tf.shape(image)[1]
+        shape = tf.shape(image)
+        height, width = shape[0], shape[1]
         num_channels = image.shape[2]
 
         if not num_channels % self.groups == 0:
--- a/keras_cv/layers/preprocessing/channel_shuffle.py
+++ b/keras_cv/layers/preprocessing/channel_shuffle.py
@@ -13,10 +13,9 @@
 # limitations under the License.
 
 import tensorflow as tf
-from tensorflow.keras import layers
 
 
-class ChannelShuffle(layers.Layer):
+class ChannelShuffle(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     """Shuffle channels of an input image.
 
     Input shape:
@@ -52,15 +51,10 @@ class ChannelShuffle(layers.Layer):
         self.groups = groups
         self.seed = seed
 
-    def _channel_shuffling(self, images):
-        unbatched = images.shape.rank == 3
-
-        if unbatched:
-            images = tf.expand_dims(images, axis=0)
-
-        height = tf.shape(images)[1]
-        width = tf.shape(images)[2]
-        num_channels = images.shape[3]
+    def augment_image(self, image, transformation=None):
+        height = tf.shape(image)[0]
+        width = tf.shape(image)[1]
+        num_channels = image.shape[2]
 
         if not num_channels % self.groups == 0:
             raise ValueError(
@@ -70,24 +64,13 @@ class ChannelShuffle(layers.Layer):
             )
 
         channels_per_group = num_channels // self.groups
-        images = tf.reshape(
-            images, [-1, height, width, self.groups, channels_per_group]
-        )
-        images = tf.transpose(images, perm=[3, 1, 2, 4, 0])
-        images = tf.random.shuffle(images, seed=self.seed)
-        images = tf.transpose(images, perm=[4, 1, 2, 3, 0])
-        images = tf.reshape(images, [-1, height, width, num_channels])
+        image = tf.reshape(image, [height, width, self.groups, channels_per_group])
+        image = tf.transpose(image, perm=[2, 0, 1, 3])
+        image = tf.random.shuffle(image, seed=self.seed)
+        image = tf.transpose(image, perm=[1, 2, 3, 0])
+        image = tf.reshape(image, [height, width, num_channels])
 
-        if unbatched:
-            images = tf.squeeze(images, axis=0)
-
-        return images
-
-    def call(self, images, training=True):
-        if training:
-            return self._channel_shuffling(images)
-        else:
-            return images
+        return image
 
     def get_config(self):
         config = super().get_config()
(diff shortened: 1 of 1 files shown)


keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -52,15 +51,10 @@ def __init__(self, groups=3, seed=None, **kwargs):
...
-        if unbatched:
-            images = tf.expand_dims(images, axis=0)
-
-        height = tf.shape(images)[1]
-        width = tf.shape(images)[2]
-        num_channels = images.shape[3]
+    def augment_image(self, image, transformation=None):
+        height = tf.shape(image)[0]
+        width = tf.shape(image)[1]
+        num_channels = image.shape[2]
```
LukeWood, line 57: you will need to use tf.shape(image)[3].  We can't use `.shape` in graph mode.

keras_cv/layers/preprocessing/channel_shuffle.py
```diff
@@ -52,15 +51,10 @@ def __init__(self, groups=3, seed=None, **kwargs):
...
-        unbatched = images.shape.rank == 3
-
-        if unbatched:
-            images = tf.expand_dims(images, axis=0)
-
-        height = tf.shape(images)[1]
-        width = tf.shape(images)[2]
-        num_channels = images.shape[3]
+    def augment_image(self, image, transformation=None):
+        height = tf.shape(image)[0]
```
LukeWood, line 55: reduce this to a single tf.shape call.

```
height, width, num_channels = tf.shape(image)
```

or 
```
shape = tf.shape(image)
height, width, num_channels = shape[0], shape[1], shape[2]

**Comment by LukeWood** [comment]
Thanks for the great PR.  Few minor comments.  @quantumalaviya

File: keras_cv/layers/preprocessing/channel_shuffle.py (hunk 1 above)
Comment (quantumalaviya) [reply keras_cv/layers/preprocessing/channel_shuffle.py:57]: @LukeWood This gives the following error:

```
Using a symbolic `tf.Tensor` as a Python `bool` is not allowed in Graph execution. Use Eager execution or decorate this function with @tf.function.
```

Using `tf.cond` seems like the only option. 
Also,  `.shape` seems to be working fine as part of a model too.

File: keras_cv/layers/preprocessing/channel_shuffle.py [hunk 3]
```diff
@@ -52,15 +51,10 @@ def __init__(self, groups=3, seed=None, **kwargs):
...
+        width = tf.shape(image)[1]
+        num_channels = image.shape[2]
 
         if not num_channels % self.groups == 0:
```
LukeWood, line 59: If you want to do a runtime check you must run `tf.debugging.Assert` instead of using an if statement.  This will fix the above issue too.

LukeWood: Although, this should be a solveable issue at trace time - I see.

keras_cv/layers/preprocessing/channel_shuffle.py
LukeWood: The first bug makes sense - you do need to use tf.cond there.  as for the .shape usage - there are cases where it will cause issues.  I'm fairly sure we can't use .shape inside of call methods - @qlzh727 can you confirm this?

keras_cv/layers/preprocessing/channel_shuffle.py
quantumalaviya: Exactly, when I use `tf.debugging.Assert` and  `tf.debugging.assert_equal`, the assertions get bypassed  at trace time and the function raises an error during reshaping.

When I use `tf.cond`  and `if`, the raise function just gets executed as part of tracing thus throwing an exception even when the code is completely fine.

quantumalaviya: @LukeWood Would love any inputs on how to fix this!^

keras_cv/layers/preprocessing/channel_shuffle.py
innat: You can actually use .shape here. Check the fchollet comment here. https://github.com/keras-team/keras-cv/pull/124#discussion_r815002419

quantumalaviya: This would solve everything actually.

quantumalaviya: @LukeWood This seems to solve all the other errors as well. I believe we can ensure that the channels dimension will always be populated. Do you think this can be merged as is?

Of course, I'll reduce the width and height to one single call, however.

innat: @qlzh727 @LukeWood (cc. @quantumalaviya )
I checked channel shuffle on actual training for augmenting the input image. It works without any issue. [HERE](https://www.kaggle.com/code/ipythonx/tf-hybrid-efficientnet-swin-transformer-gradcam). So, as a preprocessing layer or augmenting layer, it's completely safe with the current implementation. No code changes may be required. 

Now, if you followed a discussion over [HERE](https://github.com/keras-team/keras-cv/pull/124#issuecomment-1042876194), it later extended to work with intermediate layers (i.e `shuffle-net`). But it wasn't checked earlier by passing it as the intermediate layer. I'm checking it now. As the `num_channels` is known for input value and may not be known for intermediate layer output. So, it may cause issues. 

So, I tried to run the following model but it raised another issue for `tf.random_shuffle`. 

```python
input_shape = (28, 28, 3)
model = keras.Sequential(
    [
        keras.Input(shape=input_shape),
        ChannelShuffle(groups=3), # OK 
        layers.Conv2D(32, kernel_size=(3, 3), activation="relu"),
        ChannelShuffle(groups=4), # Not OK
        layers.MaxPooling2D(pool_size=(2, 2)),
        ChannelShuffle(groups=8), # Not OK
        layers.Conv2D(64, kernel_size=(3, 3), activation="relu"),
        layers.MaxPooling2D(pool_size=(2, 2)),
        layers.Flatten(),
        layers.Dropout(0.5),
        layers.Dense(num_classes, activation="softmax"),
    ]
)
````
```
StagingError: in user code:
....
 LookupError: gradient registry has no entry for RandomShuffle
```

---

- https://github.com/tensorflow/tensorflow/issues/6269#issuecomment-465850464.
- https://github.com/tensorflow/tensorflow/issues/22505
- https://stackoverflow.com/q/70875986/9215780 

What are the main causes of it?

LukeWood: Awesome - thank you for testing it in training.  Great to know!

LukeWood (approved):

LukeWood:
Thanks @quantumalaviya

PR #221, opened 2022-03-27
Add RandomSaturation KPL 

### LINKED ISSUE #119: HueSaturationValue preprocessing layer
ref https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.HueSaturationValue

qlzh727:
Fix #119 

Demo:
Original:
original_saturation

factor (0.0, 1.0)
augmented_saturation_0_1


keras_cv/layers/preprocessing/random_saturation.py
```diff
@@ -0,0 +1,67 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import numpy as np
+import tensorflow as tf
+
+from keras_cv.utils import preprocessing
+
+
+class RandomSaturation(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Randomly performs the saturation adjustment on given images.
```
LukeWood, line 21: """Randomly adjusts the saturation on given images.

keras_cv/layers/preprocessing/random_saturation.py
```diff
@@ -0,0 +1,67 @@
...
+
+    This layer will randomly increase/reduce the saturation for the input RGB
+    images. At inference time, the output will be identical to the input.
+    Call the layer with `training=True` to adjust the brightness of the input.
+
+    Note that different saturation adjustment factors will be apply to each the
```
LukeWood, line 27: I wonder if we need to include:

"    Note that different saturation adjustment factors will be apply to each the images in the batch" -> 

This is true for all of our layers so it can probably just be assumed.

keras_cv/layers/preprocessing/random_saturation.py
```diff
@@ -0,0 +1,67 @@
...
+    images in the batch.
+
+    Args:
+        factor: Either a tuple of two floats or a single float. `factor` controls the
+            extent to which the image saturation is impacted. `factor=1.0` makes
+            this layer perform a no-op operation. `factor=0.0` makes the image to be
+            fully grayscale. Any value larger than 1.0 will increase the saturation
+            of the image.
+
+            Values should be between `0.0` and +inf. If a tuple is used, a `factor`
```
LukeWood, line 37: what does +inf do?  Isn't 1.0 the maximum?

keras_cv/layers/preprocessing/random_saturation.py
```diff
@@ -0,0 +1,67 @@
...
+            two identical floats: `(0.5, 0.5)`.
+    """
+
+    def __init__(self, factor, **kwargs):
+        super().__init__(**kwargs)
+        self.factor = preprocessing.parse_factor_value_range(
+            factor, min_value=0.0, max_value=np.inf)
+
+    def get_random_transformation(self, image=None, label=None, bounding_box=None):
+        del image, label, bounding_box
```
LukeWood, line 50: does the del accomplish any performance optimization or is it just to satisfy the unused variable check?  Seems like noise to me.  Maybe we can configure out linter to not complain about unused variables for get_random_transformation.

LukeWood (changes requested):

LukeWood:
Might need to upgrade your `black` version.

keras_cv/layers/preprocessing/random_saturation.py
qlzh727: Ack. Removed. (This was noticed on the core keras KPLs.)

keras_cv/layers/preprocessing/random_saturation.py
qlzh727: Updated.

keras_cv/layers/preprocessing/random_saturation.py
qlzh727: Oh, I was trying to suppress the unused var warning.

keras_cv/layers/preprocessing/random_saturation.py
qlzh727: The saturation is bit tricky here. Think about the `new_saturation == factor * existing_saturation`, the existing_saturation could be 0 (eg a grey scale input). The valid range of factor is [0, inf]. 

One thing we could try is to map the [0, inf] range to [0, 1], eg by f(x) = x / (1 + x), this also works when x == 1 (no saturation update). 

I also feel the factor range here is a bit different from all other factors, but this is the direct math value we give to the backend image API. If we choose to do a mapping, then we need to provide some explanation to user for how to choose the value range. 

Let me know what do u think.

keras_cv/layers/preprocessing/random_saturation.py
LukeWood: fair enough, maybe I can disable the check in a followup somehow.  not a huge deal either way - just a nit.

keras_cv/layers/preprocessing/random_saturation.py
LukeWood: So one consideration is that in most of our layers a factor of [1.0] means you are applying the strongest possible augmentation the layer can produce, while a factor of 0.0 produces no augmentation.

Maybe we can adjust our design to map both from [0, 1] => [0, inf] and also invert the range so that it is consistent across all of our KPLs.  That way, users always know [0, 0.5] produces medium strength augmentations, and [0, 0.1] always produces weak augmentations.

LukeWood:
Looks good other than the value range.  I think we should make that match the [0, 1] with 1 being the strongest and 0 being the weakest.  I will update API design guidelines to mention this as I think it is important to be consistent here for a good UX

keras_cv/layers/preprocessing/random_saturation.py
qlzh727: Done updated with factor value mapping. Now the input factor is [0, 1]

keras_cv/layers/preprocessing/random_saturation.py
```diff
@@ -0,0 +1,78 @@
...
+
+class RandomSaturation(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Randomly adjusts the saturation on given images.
+
+    This layer will randomly increase/reduce the saturation for the input RGB
+    images. At inference time, the output will be identical to the input.
+    Call the layer with `training=True` to adjust the brightness of the input.
+
+    Args:
+        factor: Either a tuple of two floats or a single float. `factor` controls the
```
LukeWood, line 27: Is this docstring still accurate?

LukeWood (approved):

keras_cv/layers/preprocessing/random_saturation.py
qlzh727: yes. I have updated it to reflect the latest status.

PR #231, opened 2022-03-28
Add RandomHue KPL

qlzh727:
Update for https://github.com/keras-team/keras-cv/issues/119

Original image
original_saturation

Random hue (0, 1)
augmented_hue_0_1

### DIFF
--- a/keras_cv/layers/preprocessing/random_hue.py
+++ b/keras_cv/layers/preprocessing/random_hue.py
@@ -1,0 +1,66 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+from keras_cv.utils import preprocessing
+
+
+class RandomHue(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Randomly adjusts the hue on given images.
+
+    This layer will randomly increase/reduce the hue for the input RGB
+    images. At inference time, the output will be identical to the input.
+    Call the layer with `training=True` to adjust the brightness of the input.
+
+    The image hue is adjusted by converting the image(s) to HSV and rotating the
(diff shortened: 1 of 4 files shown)


keras_cv/layers/preprocessing/random_hue.py
```diff
@@ -0,0 +1,66 @@
...
+    images. At inference time, the output will be identical to the input.
+    Call the layer with `training=True` to adjust the brightness of the input.
+
+    The image hue is adjusted by converting the image(s) to HSV and rotating the
+    hue channel (H) by delta. The image is then converted back to RGB.
+
+    Args:
+        factor: Either a tuple of two floats or a single float. `factor` controls the
+            extent to which the image saturation is impacted. `factor` =
+            `0.0`, `0.5` or `1.0` makes this layer perform a no-op operation.
```
LukeWood, line 32: > 0.0`, `0.5` or `1.0` makes this layer perform a no-op operation.
is this true for all three values?  isn't that a little strange an unexpected?

LukeWood: ah - I see this is because of the way we transform the value range.

keras_cv/layers/preprocessing/random_hue.py
```diff
@@ -0,0 +1,66 @@
...
+            `factor=0.25` and `factor=0.75` makes the image to have fully opposite
+            hue value. Values should be between `0.0` and `1.0`.
+            If a tuple is used, a `factor` is sampled
+            between the two values for every image augmented.  If a single float is
+            used, a value between `0.0` and the passed float is sampled.
+            In order to ensure the value is always the same, please pass a tuple with
+            two identical floats: `(0.5, 0.5)`.
+    """
+
+    def __init__(self, factor, **kwargs):
```
LukeWood, line 42: Can we actually provide a sensible default for this layer's factor?  Mainly because the range is a little confusing

LukeWood (approved):

keras_cv/layers/preprocessing/random_hue.py
qlzh727: yea, the hue adjustment is like a cycle. When u adjust to the max or min, it comes back to the original state. I could change the mapping to [-0.5, 0.5] so that the the input [0, 1] could achieve the max opposite hue value.

Let me know what you prefer.

qlzh727: Note that even we change the value mapping, the 0 and 1 will achieve the max adjustment, but the output will be the same between 0 and 1, (since the adjustment is a cycle, 0 and 1 just go from different direction.)

keras_cv/layers/preprocessing/random_hue.py
qlzh727: It really depends on how user want to adjust the color. If we have to come up a default value, I would just give [0, 1]

Issue #233, opened 2022-03-28
All the KerasCV layers should register to tf.keras.utils.register_keras_serializable


qlzh727:
This will allow layers to be serialized and deserialized correctly by keras framework, and enable save model etc.

qlzh727:
See https://www.tensorflow.org/api_docs/python/tf/keras/utils/register_keras_serializable for more details.

LukeWood:
would this automatically occur?  Can we make it work in a first party manner where users don't have to provide layer implementations themselves when loading?

qlzh727:
It won't happen automatically, usually the layer owner has to register it, otherwise the model loading will fail since it doesn't know what layer to map to when it see the string version of the config.

qlzh727:
See #235 and all the related test failures discovered during unit test.

PR #235, opened 2022-03-29
Register all the keras cv layers with keras serializable.

qlzh727:
Also update all the layers that has a broken `get_config`

Fix https://github.com/keras-team/keras-cv/issues/233


LukeWood (approved):

LukeWood:
Thanks @qlzh727 - great idea.

PR #244, opened 2022-03-30
Grad-CAM

### LINKED ISSUE #171: GradCam implementation
https://keras.io/examples/vision/grad_cam/

The call signature should be something like:

overlaid_result_images = gradcam(model, images, classes)

lucasdavid:
Fix #171

Add a Grad-CAM implementation and a demo of its usage.

#### Additions
- `visualization` module (perhaps `explaining` is a better name?)
- `visualization.cam.gradcam` (tf function) containing an implementation of the Grad-CAM explaining method, expecting a batch of images and batch of output units to be explained
- `visualization.utils.normalize`: scales the activation intensities of a map between 0 and 1  -- perhaps this shouldn't exist inside keras-cv, just as an utility function in the demo script?
- `visualization.utils.collect_endpoints`: a function that collects intermediate symbolic tensors within the model being explained. This can also be added as a simple utility function in the demo script. However, it is best (in my experience) to have a well-defined and documented way to perform this, as more complex typologies can result in confusing errors, such as disconnected gradients and empty maps.

#### Limitations
- ~Currently, it only works for 2D images (tensors of shape `BHWC`). With a few more arguments or assumptions (e.g. features in last dim), we can generalize onto N-dims positional signal (3D, 4D, N-D), if it's interesting.~

#### Demo Output

gradcam


LukeWood:
Thanks for the PR @lucasdavid !  Can you include some demo outputs from the demo script in the PR description!

lucasdavid:
@LukeWood done

lucasdavid:
Ok, no problem. Let me know if I can help with anything else.

keras_cv/visualization/utils.py
```diff
@@ -0,0 +1,56 @@
...
+        endpoints_.append(
+            get_nested_layer(model, layer).get_input_at(bind)
+            if link == "input"
+            else get_nested_layer(model, layer).get_output_at(bind)
+        )
+
+    return endpoints_
+
+
+def normalize(x: tf.Tensor, axis: Tuple[int, int] = (-3, -2)) -> tf.Tensor:
```
LukeWood, line 51: Let's use `utils.transform_value_range()` instead

keras_cv/visualization/utils.py
```diff
@@ -0,0 +1,56 @@
+from typing import Dict
+from typing import List
+from typing import Tuple
+from typing import Union
+
+import tensorflow as tf
+
+E = Union[str, Dict[str, Union[str, int]]]
```
LukeWood, line 8: Let's just drop the type annotations.

examples/visualization/gradcam_demo.py
```diff
@@ -0,0 +1,83 @@
...
+    tf.keras.utils.get_file("wolf.jpg", "https://i.imgur.com/rRcsL2D.jpeg"),
+    tf.keras.utils.get_file("cat_and_dog.jpg", "https://i.imgur.com/6wxtMEp.jpeg"),
+]
+
+images = np.stack([load_image(path) for path in paths]).astype("uint8")
+x = tf.convert_to_tensor(images.astype("float32") / 127.5 - 1)
+
+# Load a model with pre-trained weights from ImageNet
+input_tensor = tf.keras.Input([None, None, 3], name="images")
+rn50 = tf.keras.applications.ResNet50V2(
```
LukeWood, line 27: expand names: rn50 -> resnet

examples/visualization/gradcam_demo.py
```diff
@@ -0,0 +1,83 @@
...
+probs = tf.nn.softmax(logits)
+decoded = tf.keras.applications.imagenet_utils.decode_predictions(probs.numpy())
+
+logits, maps = visualization.gradcam(rn50_w_acts, x, labels)
+
+# As we are only interested in pixels that positively contribute to the
+# classification of a label, we crunch pixels that negatively contribute to
+# its classification and unrelated pixels (zero contribution) together:
+maps = tf.nn.relu(maps)
+maps = normalize(maps)
```
LukeWood, line 67: Ah - I see how you use this.  Is there any way to encapsulate this information into the gradcam implementation?

keras_cv/visualization/cam.py
```diff
@@ -0,0 +1,41 @@
...
+    logits.
+
+    If `indices` is passed, the specific logits indexed by elements in this
+    tensor are selected before the gradients are computed, effectivelly
+    reducing the columns in the jacobian, and the size of the output
+    explaining map.
+
+    """
+    with tf.GradientTape(watch_accessed_variables=False) as t:
+        t.watch(inputs)
```
LukeWood, line 29: t -> tape

LukeWood (changes requested):

LukeWood:
> visualization module (perhaps explaining is a better name?)

I like your name explaining, I'm not fully decided on this but will begin drafting out names.

innat:
@lucasdavid 
> Currently, it only works for 2D images (tensors of shape BHWC). With a few more arguments or assumptions (e.g. features in last dim), we can generalize onto N-dims positional signal (3D, 4D, N-D), if it's interesting.

If possible, please consider supporting the 3D grad-cam too. It would be essential for medical imaging. Some code reference below

- https://github.com/fitushar/3D-Grad-CAM
- https://github.com/AdamLabISU/3DGradCAM

keras_cv/visualization/utils.py
lucasdavid: I don't think it's possible for two reasons:
1. `transform_value_range` requires the original range, which is unknown (the signal from the last conv. layer and from logits might be arbitrarily large)
2. I want to normalize the activation map for each class independently, thus using `axis=(-3, -2)=(height, width)`; but `transform_value_range` was clearly developed to normalize with respect to a single pair (min, max) of numbers.

keras_cv/visualization/utils.py
lucasdavid: ok :(

examples/visualization/gradcam_demo.py
lucasdavid: There is, but I wouldn't recommend adding it directly to the gradcam traced function.
Some other methods will use gradcam or saliency-grad-backprop internally (for example, Smooth-Grad or cameras).
The former consists of replicating the input image N times and adding noise to the data, computing the saliency map with grad-backprop and averaging the results.

Normalizing this could make it difficult for us to reuse these functions on the long run.

#### Some options:

Opt 2: we could have a wrapper:
```
def explain(model, endpoints, explaining_units, images, method='gradcam', normalize=True):
  method = deserialize(method)
  maps = method(..., images)
  if normalize:
    maps = normalize(maps)
  return maps
```

This simplifies everything into a single call, but introduces a performance issue in case the explanation method must be applied over multiple batches (endpoints are extracted and a new model is synthesized each step).

Opt 3: keras-vis opted for a object-oriented approach:
```
class Vis:
  def __init__(self, normalize=True):
    self.normalize = normalize

  def explain(images):
    images = self.perform(images)
    if self.normalize: images = self.normalize(images)
    return images

class GradCam(Vis):
  def perform(images):
    ...
```
Note: that's not the actual props/methods names, I'm just typing on top of my memory to explain it.

This opt takes care of the wrapping and leaves the actual implementation for the sub-classes. The prepared model (with extracted endpoints) can be cached as a property of the Vis' object and calling explain onto multiple batches should be fine. The issue with this approach is that we get a little further from functional programming and it might be a little confusing for some.

keras_cv/visualization/utils.py
LukeWood: Haha - sorry just trying to keep it consistent across the codebase for now!

LukeWood: Maybe we will be using them in the future, but I'd like to keep it consistent.

lucasdavid:
@LukeWood fixed all inconsistencies.
I left the `normalize` function untouched as it's not clear how to proceed.

@innat added the `positional_axis=(-3, -2)` argument. So you can now compute for 3D conv by passing `positional_axis=(-4, -3, -2)` or `(1, 2, 3)`.

examples/visualization/gradcam_demo.py
```diff
@@ -0,0 +1,83 @@
...
+from keras_cv import visualization
+from keras_cv.visualization.utils import normalize, collect_endpoints
+
+SIZES = (299, 299)
+LABELS = 5
+
+# Load images from the web
+def load_image(path):
+    image = tf.keras.preprocessing.image.load_img(path, target_size=SIZES)
+    return tf.keras.preprocessing.image.img_to_array(image)
```
innat, line 14: @lucasdavid The `tf.keras.preprocessing.image.*` will be depreciated, [source](https://github.com/keras-team/keras-cv/issues/320#issuecomment-1099518637)). But we can port that two function from `tf.keras.utils.*`.

keras_cv/visualization/cam.py
```diff
@@ -0,0 +1,40 @@
+import tensorflow as tf
+
+
+@tf.function(experimental_relax_shapes=True, experimental_follow_type_hints=True)
+def gradcam(
```
innat, line 5: Is the file name `cam.py` or should be `grad_cam.py`?

innat: or, we're going to have also variants of cam, ie. `grad_cam_pluss_plus`, and `guided_gram` func in the single file? That's fine then.

lucasdavid: It could go whichever way, just let me know. I usually place them in the same file in my research projects, as these methods aren't usually very long.

lucasdavid: Correction: I place all cam-based methods inside `explaining/cams.py` and all saliency-based into `explaining/saliency.py`, with `__init__.py` importing and exposing all members.

lucasdavid:
@LukeWood I implemented the wrap function and added the `simple_gradcam_demo.py` example. Hopefully it's easier to use now.

LukeWood:
> @LukeWood I implemented the wrap function and added the `simple_gradcam_demo.py` example. Hopefully it's easier to use now.

Thanks @lucasdavid!  I will look at this in depth tomorrow.

LukeWood:
@lucasdavid checking this PR out now locally to play around with it and look more deeply into the API.

LukeWood:
Hi @lucasdavid - we have decided `explain` is the package name we will use for explainability tools.  Can you migrate this to `explain`?  Also, I think we want the simplest API possible.  Perhaps we can expose more behavior behind keywords.  

after you move to `explain`, I will try my hand at API design.

lucasdavid:
@LukeWood done.

lucasdavid:
@LukeWood friendly ping.

From what I understood, I'm finished with all requested changes. (Or is anything missing?)
Let me know if I can help with those keywords, or in any way to speed up this PR.

semihcanturk:
Hi @lucasdavid, thanks for this PR! Your implementation is the most robust I've found that avoids the graph disconnected issues when using nested models (e.g. when transfer learning). I have two questions/suggestions:

1) Models that have "." in their nested model names names (e.g. `mobilenetv2_1.00_224`) will break the structure in `get_nested_layer`:
```
def get_nested_layer(model, name):
    for n in name.split("."):
        model = model.get_layer(n)
    return model
```
May I suggest wrapping with a try/except so that it uses the name itself without splitting as a fallback option?
```
def get_nested_layer(model, name):
    try:
        for n in name.split("."):
            model = model.get_layer(n)
    except:
        model = model.get_layer(name)
```

2) As far as I can tell, the pooling layer selection in `wrappings.py` through `pooling_layer = model.layers[-2].name` will not generalize to arbitrary models:
```
if logits_layer is None:
    logits_layer = model.layers[-1]
if pooling_layer is None:
    pooling_layer = model.layers[-2].name
if isinstance(pooling_layer, str):
    pooling_layer = {"name": pooling_layer, "link": "input"}
```
As an alternative, I tested with finding the _final_ layer with 4-dimensional output, and using its output `"link": "output"` instead, which worked well. What do you think -- is there a reason for selecting the pooling layer specifically?
```
if logits_layer is None:
    logits_layer = model.layers[-1]
if pooling_layer is None:
    for layer in reversed(model.layers):
        if len(layer.input_shape) == 4:
            pooling_layer = layer.name
            break
if isinstance(pooling_layer, str):
    pooling_layer = {"name": pooling_layer, "link": "output"}
```

Let me know, and I can add a commit with the suggested changes if you find them suitable.

lucasdavid:
Hello @semihcanturk. I'm happy to hear you found it useful. :)
About the suggestions:

1. I see the problem, but asking for any sub-layer of `mobilenetv2_1.00_224` would break with your proposal as well, wouldn't it? For example: `mobilenetv2_1.00_224.avg_pool` would not work because neither `mobilenetv2_1` or `mobilenetv2_1.00_224.avg_pool` are valid layer names.
    I break inner layers by periods because they are less frequently used in names than slashes. However, if it's impossible to represent them without ambiguity, then perhaps we should use other representations (e.g. `pooling_layer=('mobilenetv2_1.00_224', 'avg_pool')`)...
2. I guess you're right. `pooling_layer` is not a very good name because Grad-CAM can be computed with respect to any intermediate positional signal. Though there is an example where the last 4D signal is not the correct choice: when the logits are a `Conv2D` layer:
```py
model = Sequential([
  backbone,
  Conv2D(20, kernel_size=1, padding='same', name='logits'),
  GlobalAveragePooling2D(name='avg_pool'),
  Activation('sigmoid', name='predictions'),
])
```

Ideally, I think we want to default to the last 4d layer before the `logits` layer, though it might be difficult to determine it for all cases... Maybe your suggestion is indeed the best guess.

LukeWood:
hey @lucasdavid.  We're still considering whether we want to branch out into model explainability tools.  We might just leave this out of scope and rely on tf-explain and other packages for this functionality. 

Any thoughts regarding tf-explain?  do you find it fits your needs as a user?  Trying to avoid reinventing the wheel and target our resources.

lucasdavid:
I see, no worries. I do believe tf-explain fits most industry cases, specially if you are only interested in analyzing individual examples.

I haven't used in my research because I found it difficult to extend/modify (I research explainability and WSSS). Plus it's a little inefficient -- it doesn't work for multiple GPUs, for input batches, or n-d signals. It doesn't explain multiple units at once; it uses `gradients` instead of `batch_jacobian`; etc.

LukeWood:
> I see, no worries. I do believe tf-explain fits most industry cases, specially if you are only interested in analyzing individual examples.
> 
> I haven't used in my research because I found it difficult to extend/modify (I research explainability and WSSS). Plus it's a little inefficient -- it doesn't work for multiple GPUs, for input batches, or n-d signals. It doesn't explain multiple units at once; it uses `gradients` instead of `batch_jacobian`; etc.

Sorry for all of the confusion on this.  I do think this is awesome, just want to be aware of domain creep on KerasCV.

PR #267, opened 2022-04-05
Devcontainer

### LINKED ISSUE #242: Devcontainer
Can we have a `.devcontainer` like in Keras?

https://github.com/keras-team/keras/tree/master/.devcontainer

bhack:
resolve  #242

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,5 @@
 ARG VARIANT
 
 FROM tensorflow/tensorflow:nightly${VARIANT}
+
 RUN pip install flake8 isort black pytest
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,4 @@
+ARG VARIANT
+
+FROM tensorflow/tensorflow:nightly${VARIANT}
+RUN pip install flake8 isort black pytest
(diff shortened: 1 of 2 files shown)


LukeWood:
Thanks for the PR @bhack

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,5 @@
+ARG VARIANT
```
LukeWood, line 1: Any way I can test this?  I've never used .devcontainer - perhaps github codespaces is the easiest way?  Looks good from a birds-eye view.

bhack: You can test this "locally" with Vscode (CPU/GPU), locally with Codespace, or in the browser with Github with Codespace top right button in this page (CPU only instance)

LukeWood: I don't have VSCode - and I'm not in the codespaces beta.  For now this looks fine - if it works for you it looks good to me.

LukeWood:
Thanks @bhack

.devcontainer/Dockerfile
bhack: Ok but Codespace is not in beta.
Probably you have an Org account and you Org has not activated Codespaces.

With your personal Github account you will find the button (top right)

PR #268, opened 2022-04-05
[Mscoco] decrease 10x delta tollerance after bbox format aligment

bhack:
https://github.com/keras-team/keras-cv/pull/246#discussion_r842005074


bhack:
I close now it seems it was working yesterday. But there are still some relevant differences

PR #269, opened 2022-04-05
    Update RandomHue for v1.0

LukeWood:
Updates:

- RandomHue now respects factor ranges better, 1 is a strong
  augmentation and 0 is no-op
- RandomHue now respects value range, and respects tf.image's convention
  of using [0, 1] for floats.


LukeWood:
Note: this is rebased on factor.

qlzh727:
This is bit hard to see what has been changed by this PR. Could u update this to be self-contained.

keras_cv/layers/preprocessing/random_hue.py
```diff
@@ -28,40 +28,45 @@ class RandomHue(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     hue channel (H) by delta. The image is then converted back to RGB.
 
     Args:
-        factor: Either a tuple of two floats or a single float. `factor` controls the
-            extent to which the image saturation is impacted. `factor` =
-            `0.0`, `0.5` or `1.0` makes this layer perform a no-op operation.
+        factor: A tuple of two floats, a single float or `keras_cv.FactorSampler`.
+            factor` controls the extent to which the image saturation is impacted.
+            `factor`=`0.0`, `0.5` or `1.0` makes this layer perform a no-op operation.
```
qlzh727, line 33: I think this will need some update (since you mentioned in the PR description that the 0.0 is noop and 1.0 is the strongest augmentation.)

keras_cv/layers/preprocessing/random_hue.py
```diff
@@ -28,40 +28,45 @@ class RandomHue(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
...
             If a tuple is used, a `factor` is sampled
             between the two values for every image augmented.  If a single float is
             used, a value between `0.0` and the passed float is sampled.
             In order to ensure the value is always the same, please pass a tuple with
             two identical floats: `(0.5, 0.5)`.
+        value_range:  the range of values the incoming images will have.
+            Represented as a two number tuple written [low, high].
+            This is typically either `[0, 1]` or `[0, 255]` depending
+            on how your preprocessing pipeline is setup.  Defaults to
+            `[0, 255]`.
```
qlzh727, line 45: I didn't find the default value in the signature. Did I miss anything?

keras_cv/layers/preprocessing/random_hue.py
```diff
@@ -28,40 +28,45 @@ class RandomHue(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
...
+        self.value_range = value_range
 
     def get_random_transformation(self, image=None, label=None, bounding_box=None):
         del image, label, bounding_box
-        if self.factor[0] == self.factor[1]:
-            return self.factor[0]
-        return self._random_generator.random_uniform(
-            shape=(), minval=self.factor[0], maxval=self.factor[1], dtype=tf.float32
-        )
+        invert = preprocessing.random_inversion(self._random_generator)
```
qlzh727, line 57: I am bit confused here, so there are two randomness here? (one for +-1 and another from self.factor()?). Is this what we expect here?

qlzh727 (changes requested):

LukeWood:
> This is bit hard to see what has been changed by this PR. Could u update this to be self-contained.

Sorry - after factor is merged I will rebase again so that it can be self contained.

keras_cv/layers/preprocessing/random_hue.py
LukeWood: Good point - updated.

keras_cv/layers/preprocessing/random_hue.py
LukeWood: Ah - no, no value

keras_cv/layers/preprocessing/random_hue.py
LukeWood: I do think this is what we want, because hue adjustments can be in range [-1, 1], but we want end users to only pass [0, 1] so that they can think about magnitudes instead of thinking about transformed values.  Before, it was a worse UX to pass (0.5 0.5) and get a no-op.  

While this is slightly less performant, I believe it affords us a much cleaner end user API.

keras_cv/layers/preprocessing/random_hue.py
LukeWood: updated

keras_cv/layers/preprocessing/random_hue.py
```diff
@@ -28,36 +28,44 @@ class RandomHue(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
...
-            `factor`=`0.0`, `0.5` or `1.0` makes this layer perform a no-op operation.
-            `factor=0.25` and `factor=0.75` makes the image to have fully opposite
-            hue value. Values should be between `0.0` and `1.0`.
-            If a tuple is used, a `factor` is sampled
-            between the two values for every image augmented.  If a single float is
-            used, a value between `0.0` and the passed float is sampled.
-            In order to ensure the value is always the same, please pass a tuple with
-            two identical floats: `(0.5, 0.5)`.
+        factor: A tuple of two floats, a single float or a
+            `keras_cv.FactorSampler`.  `height_factor` controls the size of the
```
qlzh727, line 32: I think this should be `factor`.

qlzh727: And the whole docstring seems to be copied from other layer.

keras_cv/layers/preprocessing/random_hue.py
qlzh727: Note that when user specify (1.0, 1.0), the value we send to tf.image.adjust_hue() is either -1 or 1. In that case, the output is actually same as the input. To achieve the max hue adjustment, the delta should be either -0.5 or 0.5 I think.

keras_cv/layers/preprocessing/random_shear.py
```diff
@@ -90,10 +90,8 @@ def _get_shear_amount(self, constraint):
         if constraint is None:
             return None
 
-        negate = self._random_generator.random_uniform((), 0, 1, dtype=tf.float32) > 0.5
-        negate = tf.cond(negate, lambda: -1.0, lambda: 1.0)
-
-        return negate * constraint()
+        invert = preprocessing.random_inversion(self._random_generator)
```
qlzh727, line 93: Does this change belong to this PR?

qlzh727 (changes requested):

keras_cv/layers/preprocessing/random_hue.py
LukeWood: Sorry about that - forgot to update the text.  updated.

qlzh727 (changes requested):
Chatted offline, the value to tf.image.adjust_hue should be 0.5 (or -0.5) to achieve the max augmentation.

LukeWood:
> Chatted offline, the value to tf.image.adjust_hue should be 0.5 (or -0.5) to achieve the max augmentation.

Done

keras_cv/layers/preprocessing/random_hue.py
```diff
@@ -29,35 +29,41 @@ class RandomHue(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
...
         self.factor = preprocessing.parse_factor(
             factor,
         )
+        self.value_range = value_range
 
     def get_random_transformation(self, image=None, label=None, bounding_box=None):
         del image, label, bounding_box
-        return self.factor()
+        invert = preprocessing.random_inversion(self._random_generator)
+        return invert * self.factor() * 0.5
```
qlzh727, line 55: let's add a comment for that, since the 0.5 is bit magic here, which is used to scale the input value to the tf.image API call.

keras_cv/layers/preprocessing/random_hue_test.py
```diff
@@ -62,7 +55,7 @@ def test_adjust_full_opposite_hue(self):
             channel_min, tf.math.reduce_min(image, axis=-1), atol=1e-5, rtol=1e-5
         )
 
-        layer = preprocessing.RandomHue(factor=(0.75, 0.75))
+        layer = preprocessing.RandomHue(factor=(0.75, 0.75), value_range=(0, 255))
```
qlzh727, line 58: I think the 0.75 is no longer the max augment. Could u check the test?

keras_cv/layers/preprocessing/random_hue.py
LukeWood: Done - agreed, this is too confusing to not comment.  Thanks

keras_cv/layers/preprocessing/random_hue_test.py
LukeWood: Updated to remove.

qlzh727 (approved):

PR #275, opened 2022-04-06
Update solarization for v1

### LINKED ISSUE #270: Update `Solarize` to support factors for use in `add` and `threshold`

LukeWood:
fixes #270

### DIFF
--- a/keras_cv/layers/preprocessing/solarization.py
+++ b/keras_cv/layers/preprocessing/solarization.py
@@ -31,16 +31,21 @@ class Solarization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     - [RandAugment](https://arxiv.org/pdf/1909.13719.pdf)
 
     Args:
-        addition: (Optional) int or float.  If specified, this value is added to each
-            pixel before solarization and thresholding.  The addition value should be
-            scaled accoring to the value range (0, 255).  Defaults to 0.0.
-        threshold: (Optional) int or float. If specified, only pixel values above this
-            threshold will be solarized.
         value_range: a tuple or a list of two elements. The first value represents
             the lower bound for values in passed images, the second represents the
             upper bound. Images passed to the layer should have values within
             `value_range`. Defaults to `(0, 255)`.
-
+        addition_factor: (Optional)  A tuple of two floats, a single float or a
+            `keras_cv.FactorSampler`. For each augmented image a value is sampled
+            from the provided range. If a float is passed, the range is interpreted as
+            `(0, addition)`. If specified, this value is added to each pixel before
+            solarization and thresholding.  The addition value should be scaled
+            according to the value range (0, 255). Defaults to 0.0.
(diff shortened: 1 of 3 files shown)


keras_cv/layers/preprocessing/solarization.py
```diff
@@ -31,16 +31,21 @@ class Solarization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
...
-            threshold will be solarized.
         value_range: a tuple or a list of two elements. The first value represents
             the lower bound for values in passed images, the second represents the
             upper bound. Images passed to the layer should have values within
             `value_range`. Defaults to `(0, 255)`.
-
+        addition_factor: (Optional)  A tuple of two floats, a single float or a
+            `keras_cv.FactorSampler`. For each augmented image a value is sampled
+            from the provided range. If a float is passed, the range is interpreted as
+            `(0, addition)`. If specified, this value is added to each pixel before
```
qlzh727, line 41: (0, addition_factor)

keras_cv/layers/preprocessing/solarization.py
```diff
@@ -31,16 +31,21 @@ class Solarization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
...
+            solarization and thresholding.  The addition value should be scaled
+            according to the value range (0, 255). Defaults to 0.0.
+        threshold_factor: (Optional)  A tuple of two floats, a single float or a
+            `keras_cv.FactorSampler`. For each augmented image a value is sampled
+            from the provided range. If a float is passed, the range is interpreted as
+            `(0, threshold)`. If specified, only pixel values above this threshold will
```
qlzh727, line 47: same, (0, threshold_factor)

keras_cv/layers/preprocessing/solarization.py
```diff
@@ -59,31 +64,36 @@ class Solarization(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
...
     """
 
-    def __init__(self, value_range, addition=0.0, threshold=0.0, **kwargs):
-        super().__init__(**kwargs)
-        self.addition = addition
-        self.threshold = threshold
+    def __init__(self, value_range, addition_factor=0.0, threshold_factor=0.0, seed=None, **kwargs):
+        super().__init__(seed=seed, **kwargs)
+        self.addition_factor = preprocessing.parse_factor(
+            addition_factor, max_value=256, seed=seed, param_name="addition_factor"
```
qlzh727, line 77: shouldn't the max_value be 255?

qlzh727 (changes requested):

keras_cv/layers/preprocessing/solarization.py
LukeWood: I was thinking because we use the `<` operator so we should actually support all the way to 256.

Instead, let's opt to `<=` so we can keep the range to 255 as you suggested.  That is better, done.

qlzh727 (approved):

Issue #256, opened 2022-04-02
Update all KPLs to support `seed`


LukeWood:
We also need to update docstrings to support stateless_uniform

LukeWood:
FYI @qlzh727 we'll need to do this

qlzh727:
Ack. We already have that in the BaseImageAugmentLayer, and we should make sure all the subclass layer have that in the __init__().

PR #276, opened 2022-04-06
Add "seed" param to all the random augmentation layers

qlzh727:
Make sure we pass them to parent layer, as well as parse_factor(). Added tests for serialization as well.


qlzh727:
Update for https://github.com/keras-team/keras-cv/issues/256

LukeWood (approved):

LukeWood:
Looks good to me

PR #282, opened 2022-04-08
Bounding box support in MixUp augmentation

### LINKED ISSUE #261: Bounding box support in MixUp augmentation

quantumalaviya:
closes #261 
also edits the mix_up_demo.py to support `BaseImageAugmentationLayer`

The mixup is simply executed as a concatenation of bounding boxes across the mixed images. 

One possible issue is the repetition of bounding boxes on images that get mixed with themselves. Another thing to note (which probably is true by convention), the layer will expect the number of bounding boxes per image to be the same. This may require padding.

### DIFF
--- a/keras_cv/layers/preprocessing/mix_up.py
+++ b/keras_cv/layers/preprocessing/mix_up.py
@@ -17,7 +17,6 @@ import tensorflow as tf
 @tf.keras.utils.register_keras_serializable(package="keras_cv")
 class MixUp(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     """MixUp implements the MixUp data augmentation technique.
-
     Args:
         alpha: Float between 0 and 1.  Inverse scale parameter for the gamma
             distribution.  This controls the shape of the distribution from which the
@@ -25,7 +24,6 @@ class MixUp(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
             when training an imagenet1k classification model.
     References:
         [MixUp paper](https://arxiv.org/abs/1710.09412).
-
     Sample usage:
     ```python
     (images, labels), _ = tf.keras.datasets.cifar10.load_data()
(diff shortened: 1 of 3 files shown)


LukeWood:
> closes #261 also edits the mix_up_demo.py to support `BaseImageAugmentationLayer`
> 
> The mixup is simply executed as a concatenation of bounding boxes across the mixed images.
> 
> One possible issue is the repetition of bounding boxes on images that get mixed with themselves. Another thing to note (which probably is true by convention), the layer will expect the number of bounding boxes per image to be the same. This may require padding.

Thanks for the PR @quantumalaviya.  Can you provide a reference to another codebase/paper indicating that this logic is the correct way to handle bboxes in MixUp?  We can add it to the `references` section of the docstring.

quantumalaviya:
Sure, have a look now @LukeWood 
Mosaic with Mixup will mimic how it's used in SOTA detectors.

keras_cv/layers/preprocessing/mix_up.py
```diff
@@ -49,15 +50,26 @@ def _sample_from_beta(alpha, beta, shape):
...
                 f"Got: inputs = {inputs}"
             )
-        images, labels = self._update_labels(*self._mixup(images, labels))
+        images, lambda_sample, permutation_order = self._mixup(images)
+        if labels is not None:
+            images, labels = self._update_labels(
+                images, labels, lambda_sample, permutation_order
+            )
+            inputs["labels"] = labels
+        if bounding_boxes is not None:
```
LukeWood, line 67: You'll also need to consider the case where bounding_boxes AND labels are provided.  This is a supported use case.

LukeWood (changes requested):
Minor comment, other than that LGTM

keras_cv/layers/preprocessing/mix_up.py
quantumalaviya: It is covered. In case both are not `None`, both the `if` bodies will be traversed.

In fact, I have tested using this case as well. Have a look at the changes in the `mix_up_test. py` file. I hope that is what you mean.

LukeWood: I think we can also remove the `images` argument now from _update_labels and _update_bounding boxes!

LukeWood: But yes looks like it works!  I was just confused by the fact that it looks like _update_labels and _update_bounding_boxes updates images (though it doesn't actually!)

quantumalaviya: True, there was no use of `images` in  `_update_labels` and `_update_bounding_boxes`. Removed it now!

LukeWood: Thanks!

LukeWood:
Thanks @quantumalaviya !!!

PR #294, opened 2022-04-12
Remove layers that don't belong in RandAugment from RandAugment

LukeWood:
Some of these are mean't to be applied BEFORE rand augment - CutOut, and some of these are just too strong (Invert).

This PR is a result of a discussion with the original RandAugment author.

Updated image:
randaug

### DIFF
--- a/keras_cv/layers/preprocessing/rand_augment.py
+++ b/keras_cv/layers/preprocessing/rand_augment.py
@@ -58,10 +58,11 @@ class RandAugment(RandomAugmentationPipeline):
             Defaults to `0.15`.
         rate:  the rate at which to apply each augmentation.  This parameter is applied
             on a per-distortion layer, per image.  Should be in the range `[0, 1]`.
-            Defaults to `10/11`, as the original `RandAugment` paper includes an
-            Identity transform.  By setting the rate to 10/11 in our implementation, the
-            behavior is identical to sampling an Identity augmentation 10/11th of the
-            time.
+            To reproduce the original RandAugment paper results, set this to `10/11`.
+            The original `RandAugment` paper includes an Identity transform.  By setting
+            the rate to 10/11 in our implementation, the behavior is identical to
+            sampling an Identity augmentation 10/11th of the time.
+            Defaults to `1.0`.
 
     Usage:
     ```python
--- a/keras_cv/layers/preprocessing/rand_augment.py
+++ b/keras_cv/layers/preprocessing/rand_augment.py
@@ -58,9 +58,9 @@ class RandAugment(RandomAugmentationPipeline):
             Defaults to `0.15`.
         rate:  the rate at which to apply each augmentation.  This parameter is applied
             on a per-distortion layer, per image.  Should be in the range `[0, 1]`.
-            Defaults to `1/11`, as the original `RandAugment` paper includes an
-            Identity transform.  By setting the rate to 1/11 in our implementation, the
-            behavior is identical to sampling an Identity augmentation 1/11th of the
+            Defaults to `10/11`, as the original `RandAugment` paper includes an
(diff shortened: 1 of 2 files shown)


keras_cv/layers/preprocessing/rand_augment.py
```diff
@@ -76,7 +79,7 @@ def __init__(
         augmentations_per_image=3,
         magnitude=0.5,
         magnitude_stddev=0.15,
-        rate=1.0,
+        rate=10 / 11,
```
kartik4949, line 82: 1/11 or 10/11?

LukeWood: 10/11 because this is the rate we apply augmentations at.

LukeWood: Ha - no worries now I see you were looking at the docstring.  I fixed it - thanks Kartik!

kartik4949: got confused from doctoring, cool

keras_cv/layers/preprocessing/rand_augment.py
```diff
@@ -58,7 +58,10 @@ class RandAugment(RandomAugmentationPipeline):
             Defaults to `0.15`.
         rate:  the rate at which to apply each augmentation.  This parameter is applied
             on a per-distortion layer, per image.  Should be in the range `[0, 1]`.
-            Defaults to `1`.
+            Defaults to `10/11`, as the original `RandAugment` paper includes an
```
fchollet, line 61: Not 100% sure we should do it. This adds overhead, and the actual benefits are unclear. We could simply leave it to 1.0 and mention that 10 / 11 reproduces the implementation proposed by the paper.

fchollet (commented):
Thanks for the PR! Changes LGTM except the `rate` default value.

keras_cv/layers/preprocessing/rand_augment.py
LukeWood: Done - yeah, this is kind of weird and you are right that it is unclear if it actually even helps.

bhack: The ablation study was done only for  Wide-ResNet-28-2 models weretrained on CIFAR-10 using RandAugment (N = 3, M = 4)

immagine

Do we need to remove negative avg augmentations from the list? Do we know if this was confirmed for the other datasets/models?

bhack: As always.. too late :)

LukeWood: > The ablation study was done only for `Wide-ResNet-28-2` models trained on `CIFAR-10` using `RandAugment (N = 3, M = 4)`
> 
> immagine
> 
> Do we need to remove negative avg augmentations from the list? Do we know if this was confirmed for the other datasets/models?

Hey @bhack - it is not confirmed for other datasets.  I discussed this with the original paper author.  He highly recommended the set of augmentations we now have.

bhack: Ok thanks

LukeWood: > Ok thanks

I was very confused by the presence of negative values in the table in the paper too.  It is very confusing with respect to the point they're trying to show (that this is a good general purpose augmentation technique)

bhack: > I was very confused by the presence of negative values in the table in the paper too. It is very confusing with respect to the point they're trying to show (that this is a good general purpose augmentation technique)

Yes I suppose that also a more trivial augmentation policy like TA could perform better in some cases:

immagine

PR #314, opened 2022-04-13
Add FMix Preprocessing Layer

### LINKED ISSUE #254: Add FMix preprocessing layer
Paper: [FMix: Enhancing Mixed Sample Data Augmentation](https://arxiv.org/abs/2002.12047)
Code: https://github.com/ecs-vlc/FMix

(both tensorflow and pytorch implementation are provided).

fmix_example

fmix_3d

quantumalaviya:
closes #254 

Adding support for FMix.

Updated image:
fmix

### DIFF
--- a/keras_cv/layers/preprocessing/fmix.py
+++ b/keras_cv/layers/preprocessing/fmix.py
@@ -74,9 +74,9 @@ class FMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
 
     def _apply_fftfreq(self, h, w):
         # Applying the fourier transform across 2 dimensions (height and width).
-        fx = FMix.fftfreq(w)[: w // 2 + 1 + w % 2]
+        fx = FMix._fftfreq(w)[: w // 2 + 1 + w % 2]
 
-        fy = FMix.fftfreq(h)
+        fy = FMix._fftfreq(h)
         fy = tf.expand_dims(fy, -1)
 
         return tf.math.sqrt(fx * fx + fy * fy)
--- a/keras_cv/layers/preprocessing/fmix.py
+++ b/keras_cv/layers/preprocessing/fmix.py
@@ -201,8 +201,11 @@ class FMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     def _update_labels(self, labels, lambda_sample, permutation_order):
         labels_for_fmix = tf.gather(labels, permutation_order)
 
-        # for one-hot broadcast
-        lambda_sample = tf.expand_dims(lambda_sample, -1)
+        # for broadcasting
+        batch_size = tf.expand_dims(tf.shape(labels)[0], -1)
+        labels_rank = tf.rank(labels)
+        broadcast_shape = tf.concat([batch_size, tf.ones(labels_rank - 1, tf.int32)], 0)
+        lambda_sample = tf.reshape(lambda_sample, broadcast_shape)
 
         labels = lambda_sample * labels + (1.0 - lambda_sample) * labels_for_fmix
         return labels
--- a/keras_cv/layers/preprocessing/fmix.py
+++ b/keras_cv/layers/preprocessing/fmix.py
@@ -87,7 +87,7 @@ class FMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
         param_size = tf.concat(
             [tf.constant([ch]), tf.shape(freqs), tf.constant([2])], 0
         )
-        param = self._random_generator.random_uniform(param_size)
+        param = self._random_generator.random_normal(param_size)
 
         scale = tf.expand_dims(scale, -1)[None, :]
 
--- a/keras_cv/layers/preprocessing/fmix.py
+++ b/keras_cv/layers/preprocessing/fmix.py
@@ -48,38 +48,35 @@ class FMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
         self.seed = seed
 
     @staticmethod
-    def _sample_from_beta(alpha, beta):
-        sample_alpha = tf.random.gamma((), 1.0, beta=alpha)
-        sample_beta = tf.random.gamma((), 1.0, beta=beta)
+    def _sample_from_beta(alpha, beta, shape):
+        sample_alpha = tf.random.gamma(shape, 1.0, beta=alpha)
+        sample_beta = tf.random.gamma(shape, 1.0, beta=beta)
         return sample_alpha / (sample_alpha + sample_beta)
 
     def fftfreq(self, n, d=1):
         # An implementation of numpy.fft.fftfreq using Tensorflow.
 
-        results = tf.zeros(n, dtype=tf.int32)
         N = tf.convert_to_tensor((n - 1) / 2 + 1)
         results = tf.concat(
             [tf.range(N, dtype=tf.int32), tf.range(-(n // 2), 0, dtype=tf.int32)], 0
         )
         return results / (n * d)
 
-    def apply_fftfreq(self, h, w, z):
+    def apply_fftfreq(self, h, w):
         # Applying the fourier transform across 3 dimensions.
 
         fx = self.fftfreq(w)[: w // 2 + 1 + w % 2]
 
         fy = self.fftfreq(h)
         fy = tf.expand_dims(fy, -1)
-        fy = tf.expand_dims(fy, -1)
 
-        fz = self.fftfreq(z)[:, None]
-        return tf.math.sqrt(fx * fx + fy * fy + fz * fz)
+        return tf.math.sqrt(fx * fx + fy * fy)
 
-    def get_spectrum(self, freqs, decay_power, ch, h, w, z):
+    def get_spectrum(self, freqs, decay_power, ch, h, w):
         # Function to apply a low pass filter by decaying its high frequency components.
 
         scale = tf.ones(1) / tf.cast(
-            tf.math.maximum(freqs, tf.convert_to_tensor([1 / tf.reduce_max([w, h, z])]))
+            tf.math.maximum(freqs, tf.convert_to_tensor([1 / tf.reduce_max([w, h])]))
             ** decay_power,
             tf.float32,
         )
@@ -96,12 +93,12 @@ class FMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     def sample_mask_from_transform(self, decay, shape, ch=1):
         # Sampling low frequency map from fourier transform.
 
-        freqs = self.apply_fftfreq(shape[0], shape[1], shape[2])
-        spectrum = self.get_spectrum(freqs, decay, ch, shape[0], shape[1], shape[2])
+        freqs = self.apply_fftfreq(shape[0], shape[1])
+        spectrum = self.get_spectrum(freqs, decay, ch, shape[0], shape[1])
         spectrum = tf.complex(spectrum[:, 0], spectrum[:, 1])
 
-        mask = tf.math.real(tf.signal.irfft3d(spectrum, shape))
-        mask = mask[:1, : shape[0], : shape[1], : shape[2]]
+        mask = tf.math.real(tf.signal.irfft2d(spectrum, shape))
+        mask = mask[:1, : shape[0], : shape[1]]
 
         mask = mask - tf.reduce_min(mask)
         mask = mask / tf.reduce_max(mask)
@@ -168,19 +165,33 @@ class FMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
         shape = tf.shape(images)
         permutation_order = tf.random.shuffle(tf.range(0, shape[0]), seed=self.seed)
 
-        lambda_sample = FMix._sample_from_beta(self.alpha, self.alpha)
+        lambda_sample = FMix._sample_from_beta(self.alpha, self.alpha, (shape[0],))
 
-        masks = self.sample_mask_from_transform(self.decay_power, shape[:-1])
-        masks = self.binarise_mask(masks, lambda_sample, shape[:-1], self.max_soft)
+        # generate masks utilizing mapped calls
+        masks = tf.map_fn(
+            lambda x: self.sample_mask_from_transform(self.decay_power, shape[1:-1]),
+            tf.range(shape[0], dtype=tf.float32),
+        )
+
+        masks = tf.map_fn(
+            lambda i: self.binarise_mask(
+                masks[i], lambda_sample[i], shape[1:-1], self.max_soft
+            ),
+            tf.range(shape[0], dtype=tf.int32),
+            fn_output_signature=tf.float32,
+        )
         masks = tf.expand_dims(masks, -1)
-        fmix_images = tf.gather(images, permutation_order)
 
+        fmix_images = tf.gather(images, permutation_order)
         images = masks * images + (1.0 - masks) * fmix_images
 
         return images, lambda_sample, permutation_order
 
     def _update_labels(self, labels, lambda_sample, permutation_order):
         labels_for_fmix = tf.gather(labels, permutation_order)
+
+        # for one-hot broadcast
+        lambda_sample = tf.expand_dims(lambda_sample, -1)
 
         labels = lambda_sample * labels + (1.0 - lambda_sample) * labels_for_fmix
         return labels
--- a/keras_cv/layers/preprocessing/fmix.py
+++ b/keras_cv/layers/preprocessing/fmix.py
@@ -75,7 +75,6 @@ class FMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
     def _apply_fftfreq(self, h, w):
         # Applying the fourier transform across 2 dimensions (height and width).
         fx = FMix._fftfreq(w)[: w // 2 + 1 + w % 2]
-
         fy = FMix._fftfreq(h)
         fy = tf.expand_dims(fy, -1)
 
--- a/keras_cv/layers/preprocessing/fmix.py
+++ b/keras_cv/layers/preprocessing/fmix.py
@@ -1,0 +1,196 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class FMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """FMix implements the FMix data augmentation technique.
+
+    Args:
+        alpha: Float value for beta distribution.  Inverse scale parameter for the gamma
+            distribution.  This controls the shape of the distribution from which the
+            smoothing values are sampled.  Defaults 0.5, which is a recommended value
+            in the paper.
+        decay_power: A float value representing the decay power for frequency decay prop
+            1/f**d.  Defaults to 3, as recommended in the paper.
+        max_soft: Float value representing softening value between 0 and 0.5 which
+            smooths hard edges in the mask.  Defaults to 0.0.
+        seed: Integer. Used to create a random seed.
+    References:
+        [FMix paper](https://arxiv.org/abs/2002.12047).
+
+    Sample usage:
+    ```python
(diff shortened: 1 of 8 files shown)


quantumalaviya:
These changes are inspired by the MixUp layer and the official implementation. The functions are ported to Tensorflow for support. 
Also, currently, a batch will use very similar masks. This was particularly important to ensure the same intermediate tensor shapes and types (for vectorization)

quantumalaviya:
@LukeWood This took a bit longer than expected. Please review this when you get a chance!

LukeWood:
Wow!  This was faster than I expected.  Will review @quantumalaviya Thanks!

LukeWood:
A high level question... I wonder if the name `FourierCutMix` is more indicative of what is going on here.

LukeWood:
Or `FourierMix`

innat:
@LukeWood regarding the name, I think it's better to stick with the official name. We can think surely set a more indicative name but that might show disrespect to the official one. Also, the name `FMix` is known to many but `FourierMix`, people may get confused.

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,207 @@
...
+        self.seed = seed
+
+    @staticmethod
+    def _sample_from_beta(alpha, beta, shape):
+        sample_alpha = tf.random.gamma(shape, 1.0, beta=alpha)
+        sample_beta = tf.random.gamma(shape, 1.0, beta=beta)
+        return sample_alpha / (sample_alpha + sample_beta)
+
+    def fftfreq(self, n, d=1):
+        # An implementation of numpy.fft.fftfreq using Tensorflow.
```
innat, line 57: I think you should remove this line or give a reference. Officially they use numpy.

quantumalaviya:
I did consider naming it something more descriptive like FourierMix but decided to stick to `FMix` because there were no mentions of this in the paper. 
I do agree with @innat here but I can change it if that's important.

keras_cv/layers/preprocessing/fmix.py
quantumalaviya: Actually, I changed it to what the function does. This is more descriptive as well!

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,207 @@
...
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class FMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """FMix implements the FMix data augmentation technique.
+
+    Args:
+        alpha: Float value for beta distribution.  Inverse scale parameter for the gamma
+            distribution.  This controls the shape of the distribution from which the
+            smoothing values are sampled.  Defaults 0.5, which is a recommended value
+            in the paper.
+        decay_power: A float value representing the decay power for frequency decay prop
```
LukeWood, line 26: Lets remove abbreviations (unless prop isn't an abbreviation!)

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,207 @@
...
+            1/f**d.  Defaults to 3, as recommended in the paper.
+        max_soft: Float value representing softening value between 0 and 0.5 which
```
LukeWood, line 28: could we rename this to `softening` or `edge_softening`

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,207 @@
...
+        self.max_soft = max_soft
+        self.seed = seed
+
+    @staticmethod
+    def _sample_from_beta(alpha, beta, shape):
+        sample_alpha = tf.random.gamma(shape, 1.0, beta=alpha)
+        sample_beta = tf.random.gamma(shape, 1.0, beta=beta)
+        return sample_alpha / (sample_alpha + sample_beta)
+
+    def fftfreq(self, n, d=1):
```
LukeWood, line 56: This can be static: @staticmethod

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,207 @@
...
+        # This function returns the sample frequencies of a discrete fourier transform.
+
+        N = tf.convert_to_tensor((n - 1) / 2 + 1)
```
LukeWood, line 59: Rename `N` to something else to avoid upper casing.  Also will make code less confusing.  Basically it looks like `N` is sampling the number of frequencies required to capture all information in an FFT (because the result of the FFT is mirrored).

Do we need the convert to tensor call?

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,207 @@
...
+        results = tf.concat(
+            [tf.range(N, dtype=tf.int32), tf.range(-(n // 2), 0, dtype=tf.int32)], 0
```
LukeWood, line 61: I am a little confused what exactly is going on in this part.  If you don't know and this is taken from the original implementation, that is OK but if you DO know what is going on please leave some comments.

Why don't you add a docstring for this that follows what is in numpy's?

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+        return results / (n * d)
+
+    def apply_fftfreq(self, h, w):
+        # Applying the fourier transform across 3 dimensions.
```
LukeWood, line 66: What's the third dimension?  Channels?

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,207 @@
...
+        # Function to apply a low pass filter by decaying its high frequency components.
+
+        scale = tf.ones(1) / tf.cast(
+            tf.math.maximum(freqs, tf.convert_to_tensor([1 / tf.reduce_max([w, h])]))
+            ** decay_power,
+            tf.float32,
+        )
+
+        param_size = tf.concat(
+            [tf.constant([ch]), tf.shape(freqs), tf.constant([2])], 0
```
LukeWood, line 85: lets expand some of the abbreviations where possible.  `channel` is easier to read than `ch`

LukeWood (changes requested):

LukeWood:
Wow!  Didn't realize how complex the signal processing portion of this code was!  Very cool, I reviewed as much as I can - there could be some value in adding extra comments to the signal processing functions as signal processing knowledge in the deep learning domain is pretty rare.  Thanks!

keras_cv/layers/preprocessing/fmix.py
quantumalaviya: I added some comments. Please have a look and lmk if that is descriptive enough.

keras_cv/layers/preprocessing/fmix.py
quantumalaviya: Added this as part of the range itself.

quantumalaviya:
I have added some comments and changed a few more variable names. Each function already has a small line describing what it does. I assumed that's enough context for anyone looking? A detailed description is available in the paper, which I added as part of the reference!

innat:
cc. @MattPainter01
Mentioning matt, who is one of the authors of FMix and officially implemented the `tf` version of the `f-mix`. It would be great to have some feedback from him.

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,221 @@
...
+                f"Got: inputs = {inputs}"
+            )
+        images, lambda_sample, permutation_order = self._fmix(images)
+        if labels is not None:
+            labels = self._update_labels(labels, lambda_sample, permutation_order)
+            inputs["labels"] = labels
+        inputs["images"] = images
+        return inputs
+
+    def _augment(self, inputs):
```
bhack, line 167: Is this the only layer in KLPs where we have this behaviour?

quantumalaviya: `MixUp` is implemented in a similar way.

bhack: Oh yes it is seems that it was introduced previously also in `MIxUp` by @qlzh727 with https://github.com/keras-team/keras-cv/commit/2c42c48bc6a449799fa008f1a8a808ee1952c75c

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,221 @@
...
+
+        fx = FMix.fftfreq(w)[: w // 2 + 1 + w % 2]
+
+        fy = FMix.fftfreq(h)
+        fy = tf.expand_dims(fy, -1)
+
+        return tf.math.sqrt(fx * fx + fy * fy)
+
+    def get_spectrum(self, freqs, decay_power, channel, h, w):
+        # Function to apply a low pass filter by decaying its high frequency components.
```
LukeWood, line 86: Let's try to remove blanklines to create logical blocks

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,221 @@
...
+        mask = tf.math.real(tf.signal.irfft2d(spectrum, shape))
+        mask = mask[:1, : shape[0], : shape[1]]
+
+        mask = mask - tf.reduce_min(mask)
+        mask = mask / tf.reduce_max(mask)
+        return mask
+
+    def binarise_mask(self, mask, lam, in_shape, edge_softening=0.0):
+        # Create the final mask from the sampled values.
+
```
LukeWood, line 119: Let's remove these newlines after comments.  No need for them, other than that the newlines are great!

LukeWood (approved):

LukeWood:
I approved these changes, but I'm still not convinced of the name `FMix`.  I don't think the naming conventions of research are a strong enough signal to drive our naming as they often change over the years.  There are some cases where I believe the technique names are accurate enough to keep (i.e. RandAugment), but `FMix` is very generic and I don't think particularly clear.  FourierMix is clearer to me.

Thoughts on naming here @fchollet @qlzh727?

This layer performs CutMix using shapes sampled from fourier (aka frequency) domain.  To me, FourierMix sounds good, but I worry `Fourier.*` may become too generic as time goes on.  Perhaps `FMix` is actually good for this reason.

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,221 @@
...
+        self.seed = seed
+
+    @staticmethod
+    def _sample_from_beta(alpha, beta, shape):
+        sample_alpha = tf.random.gamma(shape, 1.0, beta=alpha)
+        sample_beta = tf.random.gamma(shape, 1.0, beta=beta)
+        return sample_alpha / (sample_alpha + sample_beta)
+
+    @staticmethod
+    def fftfreq(signal_size, sample_spacing=1):
```
LukeWood, line 59: lets make this private.

_fftfreq

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,221 @@
...
+            [
+                tf.range((signal_size - 1) / 2 + 1, dtype=tf.int32),
+                tf.range(-(signal_size // 2), 0, dtype=tf.int32),
+            ],
+            0,
+        )
+
+        return results / (signal_size * sample_spacing)
+
+    def apply_fftfreq(self, h, w):
```
LukeWood, line 75: make private

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,221 @@
...
+        # Applying the fourier transform across 2 dimensions (height and width).
+
+        fx = FMix.fftfreq(w)[: w // 2 + 1 + w % 2]
+
+        fy = FMix.fftfreq(h)
+        fy = tf.expand_dims(fy, -1)
+
+        return tf.math.sqrt(fx * fx + fy * fy)
+
+    def get_spectrum(self, freqs, decay_power, channel, h, w):
```
LukeWood, line 85: same here - make private

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,221 @@
...
+        param_size = tf.concat(
+            [tf.constant([channel]), tf.shape(freqs), tf.constant([2])], 0
+        )
+        param = self._random_generator.random_normal(param_size)
+
+        scale = tf.expand_dims(scale, -1)[None, :]
+
+        return scale * param
+
+    def sample_mask_from_transform(self, decay, shape, ch=1):
```
LukeWood, line 103: etc for the remaining helper methods.  We don't want these showing up on the public API Docs.

LukeWood:
> @MattPainter01

Would be great to get your code review if possible!

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+
+        return results / (signal_size * sample_spacing)
+
+    def apply_fftfreq(self, h, w):
+        # Applying the fourier transform across 2 dimensions (height and width).
+
+        fx = FMix.fftfreq(w)[: w // 2 + 1 + w % 2]
+
+        fy = FMix.fftfreq(h)
```
innat: @quantumalaviya linebreak in 79 isn't necessary, is it?

keras_cv/layers/preprocessing/fmix.py
```diff
@@ -0,0 +1,221 @@
...
+    def _augment(self, inputs):
+        raise ValueError(
+            "FMix received a single image to `call`.  The layer relies on "
+            "combining multiple examples, and as such will not behave as "
+            "expected.  Please call the layer with 2 or more samples."
+        )
+
+    def _fmix(self, images):
+        shape = tf.shape(images)
+        permutation_order = tf.random.shuffle(tf.range(0, shape[0]), seed=self.seed)
```
innat, line 172: @quantumalaviya In order to reproduce the o/p with a given seed value, I think it's needed to do as follows:

```python
tf.random.set_seed(seed=self.seed)
tf.random.shuffle(tf.range(0, shape[0]), seed=self.seed)
```
Can u please check?

MattPainter01:
This is really cool work! 

The implementation looks good to me, but unfortunately I don't have much experience with tf, so I can't really comment on the specifics... Sorry! 

Just a couple of high level comments:
- From what I remember, we found that the mask softening resulted in no real benefits for image tasks, and sometimes hindered learning. You might want to consider removing it, or adding a disclaimer to the comments. We did find that it helped somewhat on text tasks, but presumably that's not within the keras-cv domain. 
- Whilst you didn't implement it here, just thought I'd mention that the reformulation we tried also didn't help. So I wouldn't put any effort into implementing/allowing that either. The reformulation I refer to is [here](https://github.com/ecs-vlc/FMix/blob/e5991dca018882734c8ea63599f10dfbe67fa0ae/fmix.py#L81) in our repo, and I think was discussed in the v1 arxiv paper.
- I notice you're sticking to 2D? I see the early focus of keras-cv is images, but I wonder if it's planned to include video eventually?  I don't think we explored 3D FMix for video too much, but maybe it's something to consider for the future.


As for the name, the F in FMix was meant to stand for Fourier, as you all guessed. It was an oversight that we didn't include that somewhere in the paper. That being said, do feel free to rename your implementation to whatever best fits your repo/structure!

keras_cv/layers/preprocessing/fmix.py
quantumalaviya: The `BaseImageAugmentationLayer` takes care of this, I believe. It inherits the `BaseRandomLayer` to which the seed is passed.

quantumalaviya:
Thanks for the review @MattPainter01! 

@LukeWood Which of these changes do you suggest we adopt? 
As for the 3D support, I got rid of it after the commit to support varying masks. Although it may be desirable to support videos, are we focusing on that right now? If so, I don't mind making some changes to reintroduce it!

keras_cv/layers/preprocessing/fmix.py
quantumalaviya: Originally added it to separate the calculation of `fx` and `fy`. Do you suggest I remove it?

innat: I think so, not necessary to line break here.

keras_cv/layers/preprocessing/fmix.py
sebastian-sz: @quantumalaviya I think that to use that `seed` you would need to replace `tf.random` ops with `self._random_generator`.

Looking at the code of [BaseRandomLayer](https://github.com/keras-team/keras/blob/275a0707c066bd90501e3fb3a0c26b1b676c06d3/keras/engine/base_layer.py#L3342)
```python
self._random_generator = backend.RandomGenerator(
        seed, force_generator=force_generator)
```

quantumalaviya: Yes, that actually makes sense @sebastian-sz. I didn't realize I haven't used the `_random_generator` everywhere.

Correct me if I am wrong here but `backend.RandomGenerator` doesn't provide a shuffle function (although it claims it should be used for shuffling). Am I missing something here?

sebastian-sz: @quantumalaviya Yes you are right - this particular op does not seem to have Generator equivalent.

quantumalaviya: I believe @innat's approach should work then:

```python
tf.random.set_seed(seed=self.seed)
tf.random.shuffle(tf.range(0, shape[0]), seed=self.seed)
```

LukeWood: I don't think we want to be setting a seed inside of call like this actually.

LukeWood: For now, feel free to just pass `seed` to the `tf.random.shuffle` as you have done.

LukeWood:
> This is really cool work!
> 
> The implementation looks good to me, but unfortunately I don't have much experience with tf, so I can't really comment on the specifics... Sorry!
> 
> Just a couple of high level comments:
> 
> * From what I remember, we found that the mask softening resulted in no real benefits for image tasks, and sometimes hindered learning. You might want to consider removing it, or adding a disclaimer to the comments. We did find that it helped somewhat on text tasks, but presumably that's not within the keras-cv domain.
> * Whilst you didn't implement it here, just thought I'd mention that the reformulation we tried also didn't help. So I wouldn't put any effort into implementing/allowing that either. The reformulation I refer to is [here](https://github.com/ecs-vlc/FMix/blob/e5991dca018882734c8ea63599f10dfbe67fa0ae/fmix.py#L81) in our repo, and I think was discussed in the v1 arxiv paper.
> * I notice you're sticking to 2D? I see the early focus of keras-cv is images, but I wonder if it's planned to include video eventually?  I don't think we explored 3D FMix for video too much, but maybe it's something to consider for the future.
> 
> As for the name, the F in FMix was meant to stand for Fourier, as you all guessed. It was an oversight that we didn't include that somewhere in the paper. That being said, do feel free to rename your implementation to whatever best fits your repo/structure!

Thanks for your detailed review @MattPainter01.  I really appreciate it. 

I think I am more comfortable with the name `FourierMix` or `FourierCutMix`.  I'm trying to think if `FourierMix` could ever have a naming collision... but I think it is safe to use here.

`FMix` just doesn't match the other layers or the Keras ecosystem.  `FourierMix` also gives some context to users on the behavior.  Thanks

@quantumalaviya FYI

LukeWood:
> Thanks for the review @MattPainter01!
> 
> @LukeWood Which of these changes do you suggest we adopt? As for the 3D support, I got rid of it after the commit to support varying masks. Although it may be desirable to support videos, are we focusing on that right now? If so, I don't mind making some changes to reintroduce it!

3D will likely have to come at the BaseImageAugmentationLayer down the line.  No point in adding it to one layer for now; lets do it as a big effort down the line where we can support it from ALL layers

LukeWood:
> From what I remember, we found that the mask softening resulted in no real benefits for image tasks, and sometimes hindered learning. You might want to consider removing it, or adding a disclaimer to the comments. We did find that it helped somewhat on text tasks, but presumably that's not within the keras-cv domain.

Actually, @quantumalaviya I missed this point on my comment pass.  Should we remove softening?  I'd be in favor if the author @MattPainter01 found no benefit.

keras_cv/layers/preprocessing/fourier_mix.py
```diff
@@ -182,9 +165,7 @@ def _fourier_mix(self, images):
 
         # binarise masks utilizing mapped calls
         masks = tf.map_fn(
```
innat, line 167: @quantumalaviya is it possible to use `tf.vectorized_map` instead?

quantumalaviya: I was just wondering this. I am running the experiment rn!

LukeWood: Much better.  Performance comes second to correctness.

quantumalaviya:
@LukeWood mask softening has also been removed btw!

LukeWood:
Thank you for the wonderful contribution @quantumalaviya!  Looks good, I'll merge this following minor copyedits!

PR #322, opened 2022-04-15
Add citation blurb

LukeWood:
Hey @qlzh727 @fchollet, adding this to mirror https://github.com/keras-team/keras-io/pull/852/files (style matches those of Keras and KerasTuner).

Wanted to make sure to run this by everyone before pushing, feel free to chat if anything should be changed/adjusted. I will be writing a short 4~ page paper on KerasCV to submit to the "tooling track" for NeurIps or CVPR this Fall.  If that goes as planned we can swap this out when the time comes.


bhack:
We could add a Github action and CFF format to validate as in TF https://github.com/tensorflow/tensorflow/pull/54291

PR #363, opened 2022-04-25
Expose BaseImageAugmentationLayer as a user facing API.

### LINKED ISSUE #352: Expose `BaseImageAugmentationLayer` to users
I'd like to offer BaseImageAugmentationLayer as a KerasCV API.  This will allow our users to subclass this layer and use their own BaseImageAugmentationLayers within `RandomAugmentationPipeline`s.

Additionally: this is fantastic for us as a library because if people start using this layer, we may integrate their implementations into the library!

I will write a keras.io tutorial on how to do this.  This will double as a resource for repo contributors.

LukeWood:
closes #352


qlzh727 (approved):
Thanks.

Issue #296, opened 2022-04-12
DICE loss


innat:
It is one of the most used loss functions in semantic segmentation tasks. Unfortunately still not available in `keras.loss.*`.  It has been asked many times, [#3611](https://github.com/keras-team/keras/issues/3611), [#13085](https://github.com/keras-team/keras/issues/13085), [#9395](https://github.com/keras-team/keras/issues/9395), [#10890](https://github.com/keras-team/keras/issues/10890).

(Not sure, if it's fit here or `keras.loss.*`)

bhack:
https://github.com/tensorflow/addons/pull/2558#issuecomment-941220370

innat:
Okayyy, someone pushed. I think this function should be available here. It's widely used. 
cc. @gamenerd457

gamenerd457:
Interested

bhack:
@LukeWood If you want this please assign it to @gamenerd457 to port his PR here.

innat:
@LukeWood @qlzh727 
Could u please give some feedback on this and also https://github.com/keras-team/keras-cv/issues/297?

qlzh727:
Yea, I think we can host this here (adding this to keras repo might need more extended API review and requirement). We should put this to keras_cv/losses folder.

bhack:
@gamenerd457 Green light for a PR here.

LukeWood:
yeah seems like a great fit!

gamenerd457:
So should I make a pr here

LukeWood:
> So should I make a pr here

If the loss easily confirms to the Keras loss API (y_true, y_pred) then yes!

bhack:
@LukeWood The intention was to port/refactoring @gamenerd457's Addons PR https://github.com/tensorflow/addons/pull/2558/files

innat:
@gamenerd457 (cc. @bhack @LukeWood @qlzh727 ) 

> If the loss easily confirms to the Keras loss API (y_true, y_pred) then yes!

Some refactoring may be needed, (i.e. subclassing the `tf.keras.losses.Loss` class). I think you can start working on it, either you can move PR from tf-addons or create a new one here. The implementation confirms the keras loss API.

[...]

PR #371, opened 2022-04-26
Add Dice Loss

innat:
Close https://github.com/keras-team/keras-cv/issues/296

image

Ps. Tested on actual training, works fine so far for 2D and 3D cases; now needs some polish.

### DIFF
--- a/keras_cv/losses/dice.py
+++ b/keras_cv/losses/dice.py
@@ -1,49 +1,99 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import tensorflow as tf
 from tensorflow import keras
 
 from keras_cv.utils.fill_utils import gather_channels
-from keras_cv.utils.fill_utils import get_reduce_axes
 
 
 class Dice(keras.losses.Loss):
-    """Dice loss class for semantic segmentation task.
+    """Compute the dice loss between the labels and predictions.
 
-    Input shape:
-        ...
-    Output shape:
-        ...
+      Use this dice loss function when there are 2D and 3D semantic
+      segmentaiton task. We expect labels to be provided in a `one_hot`
+      representation.
 
-    Args:
-        ...
+      In the snippet below, there are `num_classes` times channels per
+      sample. The shape of `y_true` and `y_pred` are
+      `[batch_size, height, width, num_classes]` or
+      `[batch_size, height, widht, depth, num_classes]`.
 
-    Call arguments:
-        y_true:
-        y_pred:
+      Standalone usage:
 
-    Usages:
-    ```python
+      >>> y_true = tf.random.uniform([5, 10, 10, 3], 0, maxval=2, dtype=tf.int32)
+      >>> y_true = tf.cast(y_true, dtype=tf.float32)
+      >>> y_pred = tf.random.uniform([5, 10, 10, 3], 0, maxval=2, dtype=tf.int32)
+      >>> y_pred = tf.cast(y_pred, dtype=tf.float32)
+      >>> dice = Dice()
+      >>> dice(y_true, y_pred).numpy()
+      0.49238735
 
+      >>> # Calling with 'sample_weight'.
+      >>> dice(y_true, y_pred, sample_weight=tf.constant([[0.5, 0.5, 0.5,
+                                                           0.5, 0.5]])).numpy()
+      0.24619368
+
+      Usage with the `compile()` API:
+      ```python
+      model.compile(optimizer='adam', loss=keras_cv.losses.Dice())
     ```
-
     """
 
     def __init__(
         self,
         beta=1,
-        class_ids=None,
-        per_image=False,
-        smooth=keras.backend.epsilon(),
-        label_smoothing=0.0,
         from_logits=False,
+        class_ids=None,
+        label_smoothing=0.0,
+        per_sample=False,
+        epsilon=keras.backend.epsilon(),
+        name="dice",
         **kwargs
     ):
+        """Initializes `Dice` instance.
+
+        Args:
+            beta: A float or integer coefficient for balancing the precision
+                and recall. It determines the weight of recall and precision
+                in the combined score. If `beta < 1`, precisoin will be
+                doninated; if `beta > 1`, recall will be dominated. Default to `1`.
+            from_logits: Whether `y_pred` is expected to be a logits tensor. By
+                default, we assume that `y_pred` encodes a probability distribution.
+                Default to `False`.
+            class_ids: An interger or a lost of intergers within `range(num_classes)`
+                to evaluate the loss. If it's `None`, all classes will beu used to
+                calculate the loss. Default to `None`.
+            label_smoothing: Float in [0, 1]. When > 0, label values are smoothed,
+                meaning the confidence on label values are relaxed. For example, if
+                `0.1`, use `0.1 / num_classes` for non-target labels and
+                `0.9 + 0.1 / num_classes` for target labels. Default to `0.0`
+            per_sample: If `True`, the loss will be calculated for each sample in
+                batch and then averaged. Otherwise the loss will be calculated for
+                the whole batch. Default to `False`.
+            epsilon: Small float added to dice score to avoid dividing by zero.
+                Default to `keras.backend.epsilon()` or `1e-07`.
+            name: Optional name for the instance.
+            Defaults to 'dice'.
+        """
         super().__init__(**kwargs)
         self.beta = beta
-        self.class_ids = class_ids
-        self.per_image = per_image
         self.from_logits = from_logits
+        self.class_ids = class_ids
         self.label_smoothing = label_smoothing
-        self.smooth = smooth
+        self.per_sample = per_sample
+        self.epsilon = epsilon
 
     def call(self, y_true, y_pred):
         y_pred = tf.convert_to_tensor(y_pred)
@@ -62,13 +112,19 @@ class Dice(keras.losses.Loss):
             label_smoothing, _smooth_labels, lambda: y_true
         )
 
-        y_true, y_pred = tf.__internal__.smart_cond.smart_cond(
-            self.class_indexes == None,
-            lambda: (y_true, y_pred),
-            lambda: gather_channels(y_true, y_pred, indexes=self.class_ids),
-        )
+        if self.class_ids is not None:
+            y_true, y_pred = gather_channels(y_true, y_pred, indexes=self.class_ids)
 
-        axes = get_reduce_axes(self.per_image)
+        axes = (
+            tf.constant([1, 2])
+            if keras.backend.image_data_format() == "channels_last"
+            else tf.constant([2, 3])
+        )
+        axes = tf.__internal__.smart_cond.smart_cond(
+            self.per_sample,
+            lambda: axes,
+            lambda: tf.concat([tf.constant([0]), axes], axis=0),
+        )
 
         true_positive = keras.backend.sum(y_true * y_pred, axis=axes)
         false_positive = keras.backend.sum(y_pred, axis=axes) - true_positive
@@ -76,17 +132,17 @@ class Dice(keras.losses.Loss):
 
         # Type I and type II errors - f-score forumula
         power_beta = 1 + self.beta**2
-        numerator = power_beta * true_positive + self.smooth
+        numerator = power_beta * true_positive + self.epsilon
         denominator = (
             (power_beta * true_positive)
             + (self.beta**2 * false_negative)
             + false_positive
-            + self.smooth
+            + self.epsilon
         )
 
         dice_score = numerator / denominator
-        dice_score = tf.cond(
-            self.per_image,
+        dice_score = tf.__internal__.smart_cond.smart_cond(
+            self.per_sample,
             lambda: keras.backend.mean(dice_score, axis=0),
             lambda: keras.backend.mean(dice_score),
         )
--- a/keras_cv/losses/dice.py
+++ b/keras_cv/losses/dice.py
@@ -73,8 +73,8 @@ class Dice(keras.losses.Loss):
             from_logits: Whether `y_pred` is expected to be a logits tensor. By
                 default, we assume that `y_pred` encodes a probability distribution.
                 Default to `False`.
-            class_ids: An interger or a lost of intergers within `range(num_classes)`
-                to evaluate the loss. If it's `None`, all classes will beu used to
+            class_ids: An interger or a list of intergers within `range(num_classes)`
+                to evaluate the loss. If it's `None`, all classes will be used to
                 calculate the loss. Default to `None`.
             axis: An optional sequence of `int` specifying the axis to perform reduce
                 ops for raw dice score. For 2D model, it should be [1,2] or [2,3] for
--- a/keras_cv/losses/dice.py
+++ b/keras_cv/losses/dice.py
@@ -151,3 +151,17 @@ class Dice(keras.losses.Loss):
             dice_score = keras.backend.mean(dice_score)
 
         return 1 - dice_score
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "beta": self.beta,
+                "from_logits": self.from_logits,
+                "class_ids": self.class_ids,
+                "label_smoothing": self.label_smoothing,
+                "per_sample": self.per_sample,
+                "epsilon": self.epsilon,
+            }
+        )
+        return config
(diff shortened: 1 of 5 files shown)


qlzh727:
Thanks. Will take a look. (Do u still want to keep it as a draft or make it a formal PR?)

innat:
@qlzh727 This PR can be reviewed now. Though I will add some more test cases in the `dice_test.py`, feel free to provide some initial comments. Thanks.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,167 @@
...
+        num_classes = tf.cast(tf.shape(y_true)[-1], y_pred.dtype)
+        return y_true * (1.0 - label_smoothing) + (label_smoothing / num_classes)
+
+    def call(self, y_true, y_pred):
+        y_pred = tf.convert_to_tensor(y_pred)
+        y_true = tf.cast(y_true, y_pred.dtype)
+        label_smoothing = tf.convert_to_tensor(self.label_smoothing, dtype=y_pred.dtype)
+
+        if self.from_logits:
+            y_pred = tf.nn.softmax(y_pred)
```
innat, line 117: @qlzh727 In [MONAI](https://github.com/Project-MONAI/MONAI), they set some arguments regarding the last layer activation function ([HERE](https://github.com/Project-MONAI/MONAI/blob/dev/monai/losses/dice.py#L48-L49)) - which helps to properly transform [logits to probabilities.](https://github.com/Project-MONAI/MONAI/blob/dev/monai/losses/dice.py#L129-L140). But keras doesn't follow this design. Instead, it provides separate loss functions (binary cross-entropy, categorical cross-entropy). 

In the above (lines 116-117), we used only softmax for `logit=True`; not for sigmoid activation. So, to maintain the keras design, what could be the fixes here?

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,167 @@
...
+            y_true = self._smooth_labels(y_true, y_pred, label_smoothing)
+
+        if self.class_ids is not None:
+            y_true, y_pred = gather_channels(y_true, y_pred, indices=self.class_ids)
+
+        axes = (
+            tf.constant([1, 2])
+            if keras.backend.image_data_format() == "channels_last"
+            else tf.constant([2, 3])
+        )
```
innat, line 129: To support `3D`  modeling, I may need to do `tf.constant([1, 2, 3]) / tf.constant([2, 3, 4])`. But not sure how to set parameters here by maintaining the keras API design. There's no 3D version of [keras-loss](https://keras.io/api/losses/) to follow guidelines. So, need some advice here.  

An example below, (not a keras style though, asaik).

```python
if modeling == '2D':
   axes = tf.constant([1, 2])
elif modeling == '3D':
   axes = tf.constant([1, 2, 3])
else:
   raise ValueError('...')
```

innat: Thanks, @bhack. 
I was thinking about how to introduce `generalized-dice loss` here. Seeing [this](https://github.com/tensorflow/models/blob/master/official/projects/volumetric_models/losses/segmentation_losses.py#L38-L39) implementation, I think we can introduce both params if it's ok to keras-team. 

```python
def __init__(self,
      beta,
      from_logists,
      ....
      loss_type,  # 'generalized', 'adaptive', None.
      axis= (1, 2, 3)  # (1, 2, 3) for 3D, (1, 2) for 2D.
):
```

innat: Added the `axis` param.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,215 @@
...
+                `generalized` or `adaptive`; otherwise compute original dice score.
+                Default to `None`.
+            label_smoothing: Float in [0, 1]. When > 0, label values are smoothed,
+                meaning the confidence on label values are relaxed. For example, if
+                `0.1`, use `0.1 / num_classes` for non-target labels and
+                `0.9 + 0.1 / num_classes` for target labels. Default to `0.0`.
+            per_sample: If `True`, the loss will be calculated for each sample in
+                batch and then averaged. Otherwise the loss will be calculated for
+                the whole batch. Default to `False`.
+            epsilon: Small float added to dice score to avoid dividing by zero.
```
qlzh727, line 94: I think this one is probably not needed. We can just use keras.backend.epsilon() directly.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,167 @@
...
+
+        self.class_ids = class_ids
+
+    def _smooth_labels(self, y_true, y_pred, label_smoothing):
+        num_classes = tf.cast(tf.shape(y_true)[-1], y_pred.dtype)
+        return y_true * (1.0 - label_smoothing) + (label_smoothing / num_classes)
+
+    def call(self, y_true, y_pred):
+        y_pred = tf.convert_to_tensor(y_pred)
+        y_true = tf.cast(y_true, y_pred.dtype)
```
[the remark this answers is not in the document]
qlzh727: Good question, I think its probably better to have 2 separate class for categorical and sparse categorical. Usually loss/metric can be created from a raw string in the model.compile. If we have a flag to config whether the losses intake one_hot or not, then user are forced to create loss instance.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,215 @@
...
+            y_pred = tf.nn.softmax(y_pred)
+
+        if tf.cast(label_smoothing, dtype=tf.bool):
+            y_true = self._smooth_labels(y_true, y_pred, label_smoothing)
+
+        if self.class_ids is not None:
+            y_true, y_pred = gather_channels(y_true, y_pred, indices=self.class_ids)
+
+        # loss calculation: Fβ-score (in terms of Type I and type II erro
+        true_positive = keras.backend.sum(y_true * y_pred, axis=self.axis)
```
qlzh727, line 149: I think we are hiding most of the keras.backend functions since they are just simple wrapper of tf ops. Please replace this with tf.math ops instead.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,215 @@
...
+
+        if self.loss_type == "generalized":
+            dice_score = self._generalized_dice_score(y_true, numerator, denominator)
+        elif self.loss_type == "adaptive":
+            dice_score = self._adaptive_dice_score(numerator, denominator)
+        else:
+            dice_score = numerator / denominator
+
+        if self.per_sample:
+            dice_score = keras.backend.mean(dice_score, axis=0)
```
qlzh727, line 170: Keras loss should return scalar tensor, so per_sample result is bit anti-pattern here.

keras_cv/utils/fill_utils.py
```diff
@@ -80,3 +82,20 @@ def fill_rectangle(images, centers_x, centers_y, widths, heights, fill_values):
 
     images = tf.where(is_rectangle, fill_values, images)
     return images
+
+
+def gather_channels(*matrices, indices=None):
```
qlzh727, line 87: This one doesn't seem have to anything to do with fill. We might just move it to the loss module for now, until there is any reuse case.

qlzh727 (changes requested):
Thanks for the PR, and sorry for the late reply.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,215 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_cv.utils.fill_utils import gather_channels
+
+
+class Dice(keras.losses.Loss):
```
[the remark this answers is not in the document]
innat: Done.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,215 @@
...
+        Args:
+            beta: A float or integer coefficient for balancing the precision
+                and recall. It determines the weight of recall and precision
+                in the combined score. If `beta < 1`, precisoin will doninate;
+                if `beta > 1`, recall will dominate. Default to `1`.
+            from_logits: Whether `y_pred` is expected to be a logits tensor. By
+                default, we assume that `y_pred` encodes a probability distribution.
+                Default to `False`.
+            class_ids: An interger or a lost of intergers within `range(num_classes)`
+                to evaluate the loss. If it's `None`, all classes will beu used to
```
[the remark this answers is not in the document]
innat: Done.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,215 @@
...
+
+        Args:
+            beta: A float or integer coefficient for balancing the precision
+                and recall. It determines the weight of recall and precision
+                in the combined score. If `beta < 1`, precisoin will doninate;
+                if `beta > 1`, recall will dominate. Default to `1`.
+            from_logits: Whether `y_pred` is expected to be a logits tensor. By
+                default, we assume that `y_pred` encodes a probability distribution.
+                Default to `False`.
+            class_ids: An interger or a lost of intergers within `range(num_classes)`
```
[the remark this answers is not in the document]
innat: Done.

keras_cv/losses/dice.py
innat: Done.

keras_cv/utils/fill_utils.py
innat: Relocated.
Done.

keras_cv/losses/dice.py
innat: This one is the smoothing parameter and sometimes users may want to set their own. WDYT? But it can be removed of course. As a default value, I use `backend.epsilon` though.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,215 @@
...
+        name="dice",
+        **kwargs,
+    ):
+        """Initializes `Dice` instance.
+
+        Args:
+            beta: A float or integer coefficient for balancing the precision
+                and recall. It determines the weight of recall and precision
+                in the combined score. If `beta < 1`, precisoin will doninate;
+                if `beta > 1`, recall will dominate. Default to `1`.
```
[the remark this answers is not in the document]
innat: Info Added. 
It should be greater than zero to any positive number.

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,215 @@
...
+                `channel_first` format respectively.
+            loss_type: An optional `str` specifying the type of the dice score to
+                compute. Compute generalized or adaptive dice score if loss type is
+                `generalized` or `adaptive`; otherwise compute original dice score.
+                Default to `None`.
+            label_smoothing: Float in [0, 1]. When > 0, label values are smoothed,
+                meaning the confidence on label values are relaxed. For example, if
+                `0.1`, use `0.1 / num_classes` for non-target labels and
+                `0.9 + 0.1 / num_classes` for target labels. Default to `0.0`.
+            per_sample: If `True`, the loss will be calculated for each sample in
```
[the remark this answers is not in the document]
innat: This param is inspired by [THIS](https://github.com/qubvel/segmentation_models/blob/94f624b7029deb463c859efbd92fa26f512b52b8/segmentation_models/losses.py#L70) implementation.

> per_image: If **True** loss is calculated for each image in batch and then averaged,
else loss is calculated for the whole batch.

The original implementation is in python, we just translate it into tf. In practice, we found that depending on the task, enabling it to `True` makes an improvement in training. So, I kept it there. WDYT?

keras_cv/losses/dice.py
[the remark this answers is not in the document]
innat: Ok. And about [this](https://github.com/keras-team/keras-cv/pull/371#discussion_r861804809), should we use another new class, `BinaryDice`?

keras_cv/losses/dice.py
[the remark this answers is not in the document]
innat: I did a quick look at the reference implementation but what decision should I make next for consistency? Can you elaborate on this?

The `class_id` coded [HERE](https://github.com/keras-team/keras/blob/6b3916813d761c75de61f89bde19538add1a8e2d/keras/metrics/metrics.py#L781), take only 1 class id and [slice from indices](https://github.com/keras-team/keras/blob/6b3916813d761c75de61f89bde19538add1a8e2d/keras/utils/metrics_utils.py#L613-L615), whereas, in above dice implementation, it gathers 1 and more within class indices within the valid range. Also, I'm not sure if the existing [Precision](https://github.com/keras-team/keras/blob/6b3916813d761c75de61f89bde19538add1a8e2d/keras/metrics/metrics.py#L750) (and similar others) are compatible for semantic segmentation task! I tried to run it in actual training and ended up with `InvalidArgumentError`.

keras_cv/losses/dice.py
qlzh727: I think keras will probably keep the existing keras format. We don't apply any activation in the loss function. User should do this in their model instead. If user would like to apply sigmoid activation, they should change the model to include sigmoid activation, and specify a loss with from_logits=False.

keras_cv/losses/dice.py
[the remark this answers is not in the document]
qlzh727: Replied on the other thread. We should have different class for binary/categorcial/sparse target input.

lucasdavid:
I can see `axis=[1,2]` in this and some other classes in this repository.
Don't know if this has been discussed somewhere else before, but I would like to suggest that we treated all axes as positional, except for the `batch` and `channels` axes. This way, every single method/procedure would handle multi-dimentional cases (2D, 3D etc) transparently.

This can be achieved by reshaping the 2d,3d,4d,...,nd signal into a `(batch, positions, channels)` one, and reducing over `positions`:
```py
def some_metric(y_true, y_pred, axis=None):
  if axis is not None:
    # business as usual...
    return

  y_true = squeeze(y_true, dims=3, axis=axis)      # 2d, 3d, 4d, ..., nd -> 3d
  y_pred = squeeze(y_pred, dims=3, axis=axis)    # 2d, 3d, 4d, ..., nd -> 3d

  # calculate metric result
  # ...
  return tf.reduce_mean(result, axis=1)

def squeeze(y, dims: int = 3):
  shape = tf.shape(y)

  if dims not in (2, 3):
    raise ValueError(f'Illegal value for parameter dims=`{dims}`. Can only squeeze '
                     'positional signal, resulting in a tensor with rank 2 or 3.')

  new_shape = [shape[0], -1]
  if dims == 3:  # keep channels.
    new_shape += [shape[-1]]
  return tf.reshape(y, new_shape)
```

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,517 @@
...
+            from_logits: Whether `y_pred` is expected to be a logits tensor. By
+                default, we assume that `y_pred` encodes a probability distribution.
+                Default to `False`.
+            class_ids: An interger or a list of intergers within `range(num_classes)`
+                to evaluate the loss. If it's `None`, all classes will be used to
+                calculate the loss. Default to `None`.
+            axis: An optional sequence of `int` specifying the axis to perform reduce
+                ops for raw dice score. For 2D model, it should be [1,2] or [2,3]
+                for the `channels_last` or `channels_first` format respectively. And for
+                3D mdoel, it should be [1,2,3] or [2,3,4] for the `channels_last` or
```
quantumalaviya, line 146: should be model on line 146

LukeWood:
Thanks for your patience @innat, wanted to merge a Loss first before this.  Ill revisit this now that we have done don

bhack:
This is the oldest open PR. What is its status?

LukeWood:
> This is the oldest open PR. What is its status?

To be honest it simply hasn’t been prioritized due to it not being a part of one of our flagship workflows (OD w/ RetinaNet, classification training in a reusable template with SOTA augmentation, and segmentation map prediction with model TBD)

one thing to double check is that the loss reduction matches the construct set forth in the rest of the losses (see smoothL1, focal as examples).

we can revisit this, but at the moment we are focusing effort on delivering those flagship workflows.

LukeWood:
> > This is the oldest open PR. What is its status?
> 
> To be honest it simply hasn’t been prioritized due to it not being a part of one of our flagship workflows (OD w/ RetinaNet, classification training in a reusable template with SOTA augmentation, and segmentation map prediction with model TBD)
> 
> one thing to double check is that the loss reduction matches the construct set forth in the rest of the losses (see smoothL1, focal as examples).
> 
> we can revisit this, but at the moment we are focusing effort on delivering those flagship workflows.

That explanation said @innat, feel free to rebase and re request a review.  It’s almost ready so happy to prioritize this.

innat:
@LukeWood I may not able to work on it due to tight schedule. If it's prioritized, someone may need to take it from here. ( Same goes to [Jaccard PR](https://github.com/keras-team/keras-cv/pull/449). )

DavidLandup0:
Can I take over this one and #449?

DavidLandup0:
Awesome, thanks! I'll be working on these in the upcoming days then.
What's left is the documentation, test cases and implementing the suggestions from the comments, right?

innat:
> Awesome, thanks! I'll be working on these in the upcoming days then. What's left is the documentation, test cases and implementing the suggestions from the comments, right?

Sort of.

DavidLandup0:
@innat Could you share the training script you mentioned if you still have it?

innat:
I am not sure which one you're referring to. Could you please more precise?  ( I'm in leave, so if I've anything that is related, I will share in upcoming days.)

DavidLandup0:
Oh, I meant this:
> Ps. Tested on actual training, works fine so far for 2D and 3D cases; now needs some polish.

Sorry to bug you while you're on leave - enjoy your time off! :)

tanzhenyu (commented):
Thanks for the PR!
A general comment -- is it possible to start with a simple one, say binary dice, test it with our existing models? This would help a LOT in speed-up the review process.

DavidLandup0:
Of course! I'll get an example up and running as soon as I finish the test cases for the other PR

DavidLandup0:
@tanzhenyu added an  example run in the new PR #968 based on your training script for DeepLabV3

PR #389, opened 2022-04-28
Whitelisted format and linting files and folders for bash scripts

### LINKED ISSUE #386: `shell/lint.sh` fails with exit code 1 without giving proper error
Following the issue discussed in [#104](https://github.com/keras-team/keras-nlp/issues/104) for Keras-NLP (which is now resolved), there is a need for whitelisting directories and files which are needed to be linted and formatted, avoiding virtual environment directory.

adhadse:
Resolves #386.
Now I have whitelisted files and folders that are specifically needed for linting and formatting. 
Hey @LukeWood, This PR is ready for review!!!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@ __pycache__/
 .vscode/
 .devcontainer/
 .coverage
+
+#Virtual environment and IDE files
+venv/
+.idea/
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,3 +35,4 @@ ignore =
 known_first_party = keras_cv,tests
 default_section = THIRDPARTY
 line_length = 88
+profile = black
--- a/shell/format.sh
+++ b/shell/format.sh
@@ -1,3 +1,8 @@
 #!/bin/bash
-isort --sl --profile=black .
-black .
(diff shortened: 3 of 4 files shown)


LukeWood:
Thanks for the PR. 

Any benefit to doing this instead of just running `black .`?

adhadse:
The problem has to do with virtual environment folder that often user creates in the project repo. This leads to formatting and linting performed on code which is not of our concern. 

This eliminates that problem, making our scripts robust.

LukeWood:
Can we exclude the problematic venv dir instead?

adhadse:
The problem is there is nothing really canonical there for a name, the directory could be anything. I tried to do the same, and I realized it was wrong. That's why this [PR](https://github.com/keras-team/keras-nlp/pull/105#pullrequestreview-938483411) was closed for good

LukeWood:
Thanks for the PR @adhadse.  I'd rather actually lint the entire repo.  We will continue adding more directoires over time, and as such I'd like to ensure we enforce lint and format checks for any new directories.

I would recommend placing your venv somewhere outside of the repo.  I think this is probably the best, low-tech solution.

PR #394, opened 2022-04-29
Move LICENSE into root path

zhiqwang:
As described in https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#determining-the-location-of-your-license , the LICENSE should be in the root of the repository, and GitHub will do some better showing for it.


LukeWood:
This currently matches the rest of our repos (it is done this way in order to work with some Google pipelines).  I'm not sure we want to duplicate it to both locations.

zhiqwang:
Hi @LukeWood , 

> This currently matches the rest of our repos (it is done this way in order to work with some Google pipelines).

Got it, feel free to close this PR if there are further concerns.

LukeWood:
> Honestly I will suggest to move also other files like `.github/CONTRIBUTING.md` etc..

yeah, upon further thought KerasCV is also OSS first so we should move all of these to their github positions.  I will likely have to configure some internal pipelines to make this work.  We can merge this though

PR #396, opened 2022-04-29
Update and correct the grammatical errors in README.md

frostbyte012:
Update the README.md file and correct grammatical errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,12 @@
 
 KerasCV is a repository of modular building blocks (layers, metrics, losses, data-augmentation) that
 applied computer vision engineers can leverage to quickly assemble production-grade, state-of-the-art
-training and inference pipelines for common use cases such as image classification, object detection,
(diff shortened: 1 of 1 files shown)


LukeWood:
These are actually grammatically correct.  "Backwards" is a very weird case in English but is actually used this way commonly in software.  Both "We must preserve backwards compatibility" and "The playstation 2 is backwards compatible with the playstation 1" are correct phrases.   Commas are also not used in subclauses if only two items are given: "object detectiton and image classification".  However once we add 3 it is required: "Object detection, segmentation mask prediction, and image classification"

hope this helps!

bhack:
Something better to grasp the nuance:
https://grammarist.com/usage/backward-vs-backwards/

frostbyte012:
Thanks a lot @bhack sir and @LukeWood sir, I got to know the difference and that helped me a lot.

PR #397, opened 2022-04-29
Update dropblock test

### LINKED ISSUE #344: Re-examine dropblock parameter names
i.e. https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/regularization/dropblock_2d.py#L142

I think `rate` is probably actually more descriptive.

Similarly, https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/regularization/dropblock_2d.py#L143 dropblock_size can probably just be block_size.

LukeWood:
fixes #344


LukeWood:
@sebastian-sz

PR #399, opened 2022-04-29
Fix tests for Dropblock2d.

sebastian-sz:
This is a follow on #397 . The parameter names have been changed, but the test's remained the same, which resulted in failing tests.


LukeWood:
"It is a simple change, what could go wrong" - me 30 minutes ago.  Lesson learned, thanks @sebastian-sz

PR #407, opened 2022-05-02
Add AugMix

### LINKED ISSUE #40: AugMix image augmentation
Paper: https://arxiv.org/abs/1912.02781 Cited by 308
Code: 
- src 1: https://github.com/google-research/augmix
- src 2: https://github.com/AakashKumarNain/AugMix_TF2

quantumalaviya:
closes #40 
Moreover, added util functions from keras to support translation and rotation layers.

download


quantumalaviya:
Currently, I have only implemented the main layers used in the implementation ( color, contrast, brightness, sharpness is not implemented)

I have also re-coded `Posterization` and `Solarization` as part of this layer because tensor inputs weren't compatible.

Lastly, the layer isn't vectorized. I believe it's better this way for now and we can further vectorized the layer in a follow-up layer (if possible)

quantumalaviya:
Also, this implementation is highly inspired by @szacho's wonderful implementation. 

Thanks!

LukeWood:
> Also, this implementation is highly inspired by @szacho's wonderful implementation.
> 
> Thanks!

Be sure to include this in references!  Thanks!

LukeWood:
> Currently, I have only implemented the main layers used in the implementation ( color, contrast, brightness, sharpness is not implemented)
> 
> I have also re-coded `Posterization` and `Solarization` as part of this layer because tensor inputs weren't compatible.
> 
> Lastly, the layer isn't vectorized. I believe it's better this way for now and we can further vectorized the layer in a follow-up layer (if possible)

Thanks for the PR @quantumalaviya.  I will review it as soon as I can.

examples/layers/preprocessing/aug_mix_demo.py
```diff
@@ -0,0 +1,66 @@
...
+
+    num_classes = ds_info.features["label"].num_classes
+
+    train_ds = (
+        train_ds.map(lambda x, y: resize(x, y, num_classes=num_classes))
+        .shuffle(10 * BATCH_SIZE)
+        .batch(BATCH_SIZE)
+    )
+
+    augmix = preprocessing.AugMix([0, 255])
```
innat, line 49: set explicitly `value_range=(0, 255)`.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,295 @@
...
+        self.width = width
+        self.depth = depth
+        self.alpha = alpha
+        self.seed = seed
+        self.auto_vectorize = False
+
+    @staticmethod
+    def _sample_from_dirichlet(alpha):
+        gamma_sample = tf.random.gamma(shape=(), alpha=alpha)
+        return gamma_sample / tf.reduce_sum(gamma_sample, -1, keepdims=True)
```
innat, line 76: set explicitly `axis=-1`.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,295 @@
...
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class AugMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Performs the AugMix data augmentation technique.
+
+    Args:
+        value_range: the range of values the incoming images will have.
+            Represented as a two number tuple written [low, high].
+            This is typically either `[0, 1]` or `[0, 255]` depending
+            on how your preprocessing pipeline is setup.
+        severity: an integer representing the level of strength of
```
LukeWood, line 30: Let's try to follow the `magnitude` pattern here: receiving either a FactorSampler, float from [0, 1], or a range.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,295 @@
...
+
+import tensorflow as tf
+
+from keras_cv import layers
+from keras_cv.utils import preprocessing
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class AugMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Performs the AugMix data augmentation technique.
```
LukeWood, line 23: Can you provide a high level overview of the AugMix technique?  It would be great to go over "what a chain is", what the technique does, etc.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,295 @@
...
+            This is typically either `[0, 1]` or `[0, 255]` depending
+            on how your preprocessing pipeline is setup.
+        severity: an integer representing the level of strength of
+            augmentations (between 1 and 10). Defaults to 3.
+        width: an integer representing the number of different chains to
+            be mixed. Defaults to 3.
+        depth: an integer representing the number of transformations in
+            the chains. A negative value enables stochastic depth uniformly
+            in [1,3]. Defaults to -1.
+        alpha: a float value used as the probability coefficients for the
```
LukeWood, line 37: are both distributions meant to be parameterized by the same alpha parameter?  Fine if so, just want to confirm.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,295 @@
...
+        return gamma_sample / tf.reduce_sum(gamma_sample, axis=-1, keepdims=True)
+
+    @staticmethod
+    def _sample_from_beta(alpha, beta):
+        sample_alpha = tf.random.gamma((), 1.0, beta=alpha)
+        sample_beta = tf.random.gamma((), 1.0, beta=beta)
+        return sample_alpha / (sample_alpha + sample_beta)
+
+    def _sample_level(self, level, maxval, dtype, shape=()):
+        level = self._random_generator.random_uniform(
```
LukeWood, line 85: Let's use the factor sampler API to do this.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,295 @@
...
+            lambda depth_level, image_aug: tf.less(depth_level, depth),
+            self._loop_on_depth,
+            [depth_level, image_aug],
+        )
+        result += tf.gather(chain_mixing_weights, curr_chain) * image_aug
+        curr_chain += 1
+        return image, chain_mixing_weights, curr_chain, result
+
+    def _auto_contrast(self, image):
+        return layers.AutoContrast(self.value_range)(image)
```
LukeWood, line 119: We'll want to construct a single auto_contrast layer in the constructor instead.

If you construct a new one in the call method this can be problematic and lead to duplicate constructions.

prefer:
self.auto_contrast = layers.AutoContrast(value_range=self.value_range)

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def _equalize(self, image):
+        shape = image.shape
```
LukeWood, line 122: same comment here.

Do we have to call image.set_shape(shape)?  If so, this is a bug in equalization and we should fix it.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,295 @@
...
+        image = layers.Equalization(self.value_range)(image)
+        image.set_shape(shape)
+        return image
+
+    def _posterize(self, image):
```
LukeWood, line 127: Can we use the Posterize layer here?

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,295 @@
...
+    def _solarize(self, image):
+        threshold = tf.cast(
+            self._sample_level(self.severity, 255, tf.int32), tf.float32
+        )
+
+        image = preprocessing.transform_value_range(
+            image, original_range=self.value_range, target_range=(0, 255)
+        )
+        result = tf.clip_by_value(image, 0, 255)
+        result = tf.where(result < threshold, result, 255 - result)
```
LukeWood, line 163: same question here, can we use the solarize layer here?

LukeWood (changes requested):
Some preliminary comments @quantumalaviya!

Great work, this is coming along great.  Definitely a substantial effort, and happy to review again once we have the comments above addressed.  Thanks!

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: I added a new implementation here because posterize requires static values as inputs.

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Same reason as posterize.

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Yeah, we might need to. The equalization layer in itself causes a lot more problems here. I am still debating having an entirely new implementation (like solarize and posterize). Not only this, equalization also triggers retracing and really hurts the performance if it occurs a lot of times.

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Yes, that's the way it has been done in the paper.

keras_cv/layers/preprocessing/aug_mix.py
LukeWood: I’ve noticed this too.  Would you mind opening a GitHub issue with your findings on Equalization?  We should fix our implementation!

quantumalaviya:
Are we keeping this on hold until #431 is resolved? @LukeWood

LukeWood:
It's not the worst thing in the world to repeat yourself occasionally in code, but if possible we should use our layers' _augment_image methods with a static provided `transformation`.

If this isn't possible let me know

LukeWood:
It's not the worst thing in the world to repeat yourself occasionally in code, but if possible we should use our layers' _augment_image methods with a static provided `transformation`.

If this isn't possible let me know

quantumalaviya:
I  can't exactly think of a way to reduce the lines of code here. I considered using it for the `sample_level` calls but I don't think there will be a considerable reduction. To my understanding, this method was more useful in cases where labels and augmentations had to share a common transformation.

If you have something specific in mind, let me know.

LukeWood:
> 

Sounds good, let me play around with it a bit more!  Thanks!

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Sorry, missed this message. This should be possible, I'll look into it and send an update!

LukeWood: Thanks for the update.  Let's update Posterization to integrate these changes, and see if that works.

keras_cv/layers/preprocessing/aug_mix.py
LukeWood: I think we can fix this by using `.shape` on equalization instead of `tf.shape`.  Because we are running the code at trace time, presumably with static channel #s, this actually should be fine.  Let's open a PR to do this, then rebase this on that change when it is done.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,313 @@
...
+        self.depth = depth
+        self.alpha = alpha
+        self.seed = seed
+        self.auto_vectorize = False
+        self.severity = severity
+
+        # initialize layers
+        self.auto_contrast = layers.AutoContrast
+        self.equalize = layers.Equalization
+        self.random_shear = layers.RandomShear
```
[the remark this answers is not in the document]
quantumalaviya: Yeah, this needs to be invoked (I think this was recommended by you in one of the first reviews)

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,313 @@
...
+
+    @staticmethod
+    def _sample_from_beta(alpha, beta):
+        sample_alpha = tf.random.gamma((), 1.0, beta=alpha)
+        sample_beta = tf.random.gamma((), 1.0, beta=beta)
+        return sample_alpha / (sample_alpha + sample_beta)
+
+    def _sample_level(self, level, maxval, dtype):
+        level = (
+            preprocessing.parse_factor(
```
LukeWood, line 97: we don't want to parse a factor every call to `_sample_level()`

Is level always a single number in these calls?  Is it dynamic in that there are many calls?

If so, we may need to use a different abstraction here (maybe just directly compute the samples instead - similar to _sample_from_beta and _sample_from_dirichlet).

We definitely dont want to parse a factor every invocation.

keras_cv/layers/preprocessing/aug_mix.py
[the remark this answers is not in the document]
LukeWood: I see, perhaps I miscommunicated there.  Sorry!

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,313 @@
...
+            lambda depth_level, image_aug: tf.less(depth_level, depth),
+            self._loop_on_depth,
+            [depth_level, image_aug],
+        )
+        result += tf.gather(chain_mixing_weights, curr_chain) * image_aug
+        curr_chain += 1
+        return image, chain_mixing_weights, curr_chain, result
+
+    def _auto_contrast(self, image):
+        return self.auto_contrast(self.value_range)(image)
```
LukeWood, line 133: yeah, we can't construct a new one every call (and same to all others).  Let's avoid calling any constructors in the `call()` method

LukeWood (changes requested):

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: So just to be clear, what has been implemented right now is correct, right?

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: This is resolved now! #475

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Yes, I believe we should do this. Levels are calculated dynamically.

I think initially this wasn't implemented using parse_factor and this was due to a review comment. I will simply revert back!

LukeWood: I imagine what what I mean't to ask with the comment is whether or not we use the factor API in the constructor.  If it is not possible, that is totally fine.

quantumalaviya: Yeah, I don't think it's possible. We need different level values for each call.

LukeWood (changes requested):

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,311 @@
...
+        result = tf.clip_by_value(image, 0, 255)
+        result = tf.where(result < threshold, result, 255 - result)
+        return preprocessing.transform_value_range(
+            result, original_range=(0, 255), target_range=self.value_range
+        )
+
+    def _shear_x(self, image):
+        x = self._sample_level(self.severity, 0.3, tf.float32)
+        x *= preprocessing.random_inversion(self._random_generator)
+        transform_x = self.random_shear._format_transform(
```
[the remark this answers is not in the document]
quantumalaviya: I have made the changes. Please see if that is what you're looking for.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+
+import tensorflow as tf
+
+from keras_cv import layers
+from keras_cv.utils import preprocessing
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class AugMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Performs the AugMix data augmentation technique.  AugMix mixes several
```
LukeWood, line 23: First line of docstring should be a standalone sentence followed by two newlines:

```
    """Performs the AugMix data augmentation technique.

    AugMix....
```

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+    chains of augmentations together using weights sampled from a Dirichlet
+    distribution.  A single chain consist of a series of individual augmentations.
```
LukeWood, line 25: Good docstring, I wonder if we can split it into two sections - a paragraph of high level overview and one with more details

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+    The resultant image is further mixed with the original image to form the
+    final augmented image.
+
+    Args:
+        value_range: the range of values the incoming images will have.
+            Represented as a two number tuple written [low, high].
```
LukeWood, line 31: because we call it a tuple lets use a tuple (low, high)

(0, 1), (0, 255), etc

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+            This is typically either `[0, 1]` or `[0, 255]` depending
+            on how your preprocessing pipeline is setup.
+        severity: A tuple of two floats, a single float or a `keras_cv.FactorSampler`.
+            A value is sampled from the provided range.  If a float is passed, the
+            range is interpreted as `(0, severity)`. This value represents the
+            level of strength of augmentations.  Defaults to 0.3.
+        width: an integer representing the number of different chains to
```
LukeWood, line 38: as a docstring reader it is not clear why multiple chains are used

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+        alpha: a float value used as the probability coefficients for the
+            Beta and Dirichlet distributions. Defaults to 1.0.
+        seed: Integer. Used to create a random seed.
+
+    References:
+        [AugMix paper](https://arxiv.org/pdf/1912.02781)
+        [Official Code](https://github.com/google-research/augmix)
+        [Unoffial TF Code](https://github.com/szacho/augmix-tf)
+
+    Sample usage:
```
LukeWood, line 52: change to `Usage:`

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+    distribution.  A single chain consist of a series of individual augmentations.
+    The resultant image is further mixed with the original image to form the
+    final augmented image.
+
+    Args:
+        value_range: the range of values the incoming images will have.
+            Represented as a two number tuple written [low, high].
+            This is typically either `[0, 1]` or `[0, 255]` depending
+            on how your preprocessing pipeline is setup.
+        severity: A tuple of two floats, a single float or a `keras_cv.FactorSampler`.
```
LukeWood, line 34: what is the range of values that are acceptable for severity?

(0, 1)?

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+        **kwargs,
+    ):
+        super().__init__(seed=seed, **kwargs)
+        self.value_range = value_range
+        self.width = width
+        self.depth = depth
+        self.alpha = alpha
+        self.seed = seed
+        self.auto_vectorize = False
+        self.severity = severity
```
LukeWood, line 77: You'll need to parse a factor here I believe (according to the docstring)

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+        return self.equalize(image)
+
+    def _posterize(self, image):
+        image = preprocessing.transform_value_range(
+            images=image,
+            original_range=self.value_range,
+            target_range=[0, 255],
+        )
+
+        bits = self._sample_level(self.severity, 3, tf.int32)
```
LukeWood, line 145: I think that we could produce a clearer abstraction than "_sample_level(self.severity, 3, dtype)" for these.

I think what I like is the following:


```
self.severity_factor = preprocessing.parse_factor(severity, ...other args)
```

then in methods:

```
bits = tf.cast(self.severity_factor() * MAX_BITS, tf.int32)
```

This unifies the meaning of severity across all augmentations to be in the scale (0, 1), it also lets us use configurable factor samplers.  it is also more readable in my opinion.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+        image = tf.cast(image, self.compute_dtype)
+        return preprocessing.transform_value_range(
+            images=image,
+            original_range=[0, 255],
+            target_range=self.value_range,
+        )
+
+    def _rotate(self, image):
+        angle = tf.expand_dims(
+            self._sample_level(self.severity, 30, tf.float32), axis=0
```
LukeWood, line 158: same comment as above, also should we expose maximum rotation value as a parameter?

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+        shape = tf.cast(tf.shape(image), tf.float32)
+
+        return preprocessing.transform(
+            tf.expand_dims(image, 0),
+            preprocessing.get_rotation_matrix(angle, shape[0], shape[1]),
+        )[0]
+
+    def _solarize(self, image):
+        threshold = tf.cast(
+            self._sample_level(self.severity, 255, tf.int32), tf.float32
```
LukeWood, line 169: same as above, just leaving a comment so you have a checklist

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+            image, original_range=self.value_range, target_range=(0, 255)
+        )
+        result = tf.clip_by_value(image, 0, 255)
+        result = tf.where(result < threshold, result, 255 - result)
+        return preprocessing.transform_value_range(
+            result, original_range=(0, 255), target_range=self.value_range
+        )
+
+    def _shear_x(self, image):
+        x = self._sample_level(self.severity, 0.3, tf.float32)
```
LukeWood, line 182: same as above, just leaving a comment so you have a checklist

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,310 @@
...
+            self._loop_on_width,
+            [image, chain_mixing_weights, curr_chain, result],
+        )
+        result = weight_sample * image + (1 - weight_sample) * result
+        return result
+
+    def augment_label(self, label, transformation=None):
+        return label
+
+    def get_config(self):
```
LukeWood, line 300: make sure to add a test to serialization_test

LukeWood (changes requested):

LukeWood:
Thanks for the hard work on this one @quantumalaviya !  it will be a wonderful feature soon!

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Similar to the last comment?

```
    """Performs the AugMix data augmentation technique.

    AugMix....
```

LukeWood: Yes but I mean even further, so 3 total paragraphs.

one sentence, one high level overview, one detail heavy description

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: I was hoping this would explain that multiple chains are required:
```
    AugMix mixes several chains of augmentations together using weights sampled
    from a Dirichlet distribution
```

As for "why" multiple chains are used, I believe we can leave it out of the scope of this implementation and is better left for the paper to explain.

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Yes, I'll edit it to:
```
This value represents the level of strength of augmentations and is in the range [0, 1]. 
```

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: In the official implementation, these values are hardcoded. Do you think we should expose it as a parameter? What about the maximum values for other layers used in AugMix?

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Understood. Can you give me an idea as to what the detail-heavy description should cover? 

Outside the high-level overview, I think the AugMix image (from the paper) is very descriptive. Users that do require that sort of description may be better off just reading the paper (and seeing the image/gif)

keras_cv/layers/preprocessing/aug_mix.py
LukeWood: Gotcha, good to know.  Let’s keep them hard coded unless we get a feature request.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,301 @@
...
+        value_range: the range of values the incoming images will have.
+            Represented as a two number tuple written (low, high).
+            This is typically either `(0, 1)` or `(0, 255)` depending
+            on how your preprocessing pipeline is setup.
+        severity: A tuple of two floats, a single float or a `keras_cv.FactorSampler`.
+            A value is sampled from the provided range.  If a float is passed, the
+            range is interpreted as `(0, severity)`. This value represents the
+            level of strength of augmentations and is in the range [0, 1].
+            Defaults to 0.3.
+        width: an integer representing the number of different chains to
```
LukeWood, line 40: perhaps instead of `width` we can come up with a more descriptive name.

`chains`, `num_chains`, something of the sort?  Thoughts?

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,301 @@
...
+            be mixed. Defaults to 3.
+        depth: an integer representing the number of transformations in
```
LukeWood, line 42: `chain_depth` perhaps is more descriptive.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,301 @@
...
+            the chains. A negative value enables stochastic depth uniformly
+            in [1,3]. Defaults to -1.
```
LukeWood, line 44: instead of [-1] can we default to [1, 3] and accept either a single int or a range of ints for

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class AugMix(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Performs the AugMix data augmentation technique.
+
+    AugMix mixes several chains of augmentations together using weights sampled
+    from a Dirichlet distribution.  A single chain consist of a series of
+    individual augmentations.  The resultant image is further mixed with the
+    original image to form the final augmented image.
```
LukeWood, line 28: Another paragraph documenting what augmentations augmix uses and a brief overview of some of the benefits of the technique would be appreciated.  Keep in mind these docstrings are what render in our public docs - they're more than just an in-editor docstring.

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,301 @@
...
+
+    References:
+        [AugMix paper](https://arxiv.org/pdf/1912.02781)
+        [Official Code](https://github.com/google-research/augmix)
+        [Unoffial TF Code](https://github.com/szacho/augmix-tf)
+
+    Sample Usage:
+    ```python
+    (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    augmix = keras_cv.layers.preprocessing.mix_up.AugMix([0, 255])
```
LukeWood, line 57: be sure to run examples inside of your terminal.  This won't run.

Also, prefer `keras_cv.layers.AugMix()` over `keras_cv.layers.preprocessing.`

keras_cv/layers/preprocessing/aug_mix.py
LukeWood: I think I like num_chains.

keras_cv/layers/preprocessing/aug_mix.py
LukeWood: i.e.

"A single chain is produced by........ {explanation}"

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,301 @@
...
+        super().__init__(seed=seed, **kwargs)
+        self.value_range = value_range
+        self.width = width
+        self.depth = depth
+        self.alpha = alpha
+        self.seed = seed
+        self.auto_vectorize = False
+        self.severity = severity
+        self.severity_factor = preprocessing.parse_factor(
+            self.severity, min_value=0.01, param_name="severity", seed=self.seed
```
LukeWood, line 81: I should've made min_value and max_value a required parameter.  This is bad for readability.

Can you add max_value in and in the future I can make this required for all layers?

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,301 @@
...
+        )
+
+        # initialize layers
+        self.auto_contrast = layers.AutoContrast(value_range=self.value_range)
+        self.equalize = layers.Equalization(value_range=self.value_range)
+
+    @staticmethod
+    def _sample_from_dirichlet(alpha):
+        gamma_sample = tf.random.gamma(shape=(), alpha=alpha)
```
LukeWood, line 90: Should we create a gamma sampler in core and use it here?  This seems exactly in scope of the sampler

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -0,0 +1,301 @@
...
+            shape=(), minval=0, maxval=8, dtype=tf.int32
+        )
+        image_aug = self._apply_op(image_aug, op_index)
+        depth_level += 1
+        return depth_level, image_aug
+
+    def _loop_on_width(self, image, chain_mixing_weights, curr_chain, result):
+        image_aug = tf.identity(image)
+        depth = tf.cond(
+            tf.greater(self.depth, 0),
```
LukeWood, line 110: I think this is really confusing due to the check for -1.

It would be a lot simpler if you could write:

```
# in constructor parse depth and assign it to a pair of ints always
def sample_depth(self):
    return self._random_generator.random_uniform(shape=(), minval=self.depth[0], maxval=self.depth[1], dtype=tf.int32)
depth = self.sample_depth()
```

LukeWood (changes requested):

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Do we want this changed from the original implementation and the paper terminology? 

I do agree with you on this however, this would make it more descriptive.

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: Same question as the last comment about `width` to `num_chains`

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: This is fixed. 

That being said, since AugMix is an operation-heavy task, it takes a lot of time. Any user who does try this will be slightly confused as to why it's taking that long. I am wondering:

1) Is this a potential bottleneck in the implementation of this layer that's slowing it down? 
2) Can I add the code from the demo file here? It seems to work fine

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: We can do that, yes. Gamma distribution is in every interpolative augmentation technique I've seen thus far. This can be used to implement the beta distribution in MixUp, FourierMix, and AugMix as well. I think I can work on this after the merge of this PR and use it to sample for other augmentations as part of a follow-up PR.

quantumalaviya:
Also fixed sample usages in FMix and MixUp in docstrings

examples/layers/preprocessing/aug_mix_demo.py
```diff
@@ -0,0 +1,66 @@
...
+import tensorflow as tf
+import tensorflow_datasets as tfds
+
+from keras_cv.layers import preprocessing
+
+IMG_SIZE = (224, 224)
+BATCH_SIZE = 64
+
+
+def resize(image, label, num_classes=10):
```
LukeWood, line 31: feel free to add the utils I created for demos here.  

Here's an example: https://github.com/keras-team/keras-cv/blob/master/examples/layers/preprocessing/channel_shuffle_demo.py

LukeWood (changes requested):
minor changes, think I believe it will look pretty good.

keras_cv/layers/preprocessing/aug_mix.py
quantumalaviya: changed to image[:100]

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -89,8 +89,8 @@ def __init__(
             seed=self.seed,
         )
 
-        if isinstance(self.depth, int):
-            self.depth = [self.depth, self.depth]
```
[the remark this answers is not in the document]
quantumalaviya: done!

LukeWood:
Thank you @quantumalaviya for the contribution!  This is great, and the API looks good.

In the future, we may want to refactor the helper methods to be reused in their corresponding KPLs, but this is an implementation detail not an API change - so this is great

Issue #330, opened 2022-04-19
Hausdorff Distance / Loss function


innat:
From wiki

> In mathematics, the Hausdorff distance, or Hausdorff metric, also called Pompeiu–Hausdorff distance, **measures how far two subsets of a metric space are from each other**. It turns the set of non-empty compact subsets of a metric space into a metric space in its own right.

It can be used for point cloud similarity metrics, keypoint detection, segmentation task, etc. For example, in an ongoing kaggle comp [**UW-Madison GI Tract Image Segmentation**](https://www.kaggle.com/competitions/uw-madison-gi-tract-image-segmentation/overview/evaluation), a 3D version of Hausdorff distance is used as one of the metrics to evaluate. 

> Hausdorff distance is a method for calculating the distance between segmentation objects A and B, by calculating the furthest point on object A from the nearest point on object B.

I think it's a useful loss function to have. Here are some references implementation:

- in [tf.keras](https://stackoverflow.com/questions/61897779/how-to-use-hausdorff-metric-in-keras)
- in [pytorch](https://github.com/javiribera/locating-objects-without-bboxes/blob/e51f75ec925f67bb4054ab4a6e9b20d1d8e689cc/object-locator/losses.py#L120)

blackhat-coder:
Hi @innat, I'm interested in implementing this

innat:
@qlzh727 @LukeWood WDYT?

bhack:
It is not in the Keras loss API format but we could be inspired also by other TF ecosystem official implementations:

https://github.com/tensorflow/graphics/blob/master/tensorflow_graphics/nn/loss/hausdorff_distance.py

blackhat-coder:
Hi @innat
Currently there's no `Losses` Folder in `keras-cv` to house loss functions
Do I have to create that ? 

[113](https://github.com/keras-team/keras-cv/issues/113) @LukeWood stated to implement IoU Loss to set an example for loss functions

innat:
Yes, I think so. You need to create a loss folder.

blackhat-coder:
Hi, sorry I haven't made a PR on this issue, I've worked on it but currently waiting for [Dice loss PR](https://github.com/keras-team/keras-cv/pull/371) to get merged to avoid conflicts from me creating `losses`.
I'm also writing test cases (noob ✌).

innat:
Don't worry about that. You can safely do your own. No conflict will happen on this.

LukeWood:
[...]

PR #414, opened 2022-05-05
Hausdorff loss

blackhat-coder:
issue [330](https://github.com/keras-team/keras-cv/issues/330)

Hausdorff distance is the maximum distance of a set to the nearest point in the other set.

Capture

Other variations of hausdorff distance include [weighted hausdorff distance](https://arxiv.org/pdf/1806.07564.pdf)

### DIFF
--- a/keras_cv/losses/hausdorff_test.py
+++ b/keras_cv/losses/hausdorff_test.py
@@ -1,0 +1,47 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+import numpy as np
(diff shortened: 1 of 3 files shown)


tanzhenyu:
@blackhat-coder can you help us describe how this will be used? And more importantly, which paper uses this loss that improves model quality

bhack:
@tanzhenyu  https://paperswithcode.com/paper/weighted-hausdorff-distance-a-loss-function

Also check https://paperswithcode.com/paper/weighted-hausdorff-distance-a-loss-function

keras_cv/losses/hausdorff_test.py
```diff
@@ -0,0 +1,47 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+import numpy as np
+from keras_cv.losses.hausdorff import Hausdorff
+
+
+class HausdorffTest(tf.test.TestCase):
```
tanzhenyu, line 20: can you add a test with a fully convolutional network that compiles with this loss?

tanzhenyu (changes requested):

LukeWood:
Thanks for the PR!  For now we're going to limit contributions to things on our critical path; it is not clear to me that this loss independently produces value for users.  We will definitely revisit this loss should a model architecture that makes use of this loss be introduced to the package.

Perhaps if this is a standalone loss that provides great value it could belong in core Keras; but we'd need to discuss what unique issue this loss solves.

Feel free to discuss on the issue thread if you would really like to see this loss introduced!

PR #428, opened 2022-05-11
Workaround tf.function issue

LukeWood:


keras_cv/layers/preprocessing/random_augmentation_pipeline.py
```diff
@@ -83,15 +83,28 @@ def _augment(self, inputs):
...
             selected_op = self._random_generator.random_uniform(
                 (), minval=0, maxval=len(self.layers), dtype=tf.int32
             )
 
-            branch_fns = []
-            for (i, layer) in enumerate(self.layers):
-                branch_fns.append((i, lambda: layer(inputs)))
+            # Warning!!!
+            # DO NOT REPLACE WITH A FOR LOOP
+            # Autograph has an edge case where capturing python for loop
```
bhack, line 100: Is this tracked in a ticket?

bhack: @LukeWood I want add a test in TF let me know what issue we had here?

LukeWood: This is actually a python quirk.

references to local variables in lambdas get shadowed.  Removing the curry makes this behavior apparent.  Can copy paste some snippets when I get home if you need more info; cheers

bhack: > Can copy paste some snippets when I get home if you need more info

Yes It could be useful.

LukeWood: Basically...

>>> fns = []
>>> fns = [lambda: print(i) for i in range(3)]
>>> for f in fns:
...     f()
2
2
2
fixed it with:

>>> def curry_fn(i):
...     return lambda: print(i)
...
>>> fns = [curry_fn(i) for i in range(3)]
>>> for f in fns:
...     f()
...
0
1
2


And if you wrap the entire thing in a tf.function it will print

0
1
2

because tf.function DOES produce a scope for each loop iteration.

bhack: How I can reproduce your gist? As this gives me the same results:

```python
import tensorflow as tf
@tf.function
def test():
  print(tf.executing_eagerly())
  ns = []
  fns = [lambda: print(i) for i in range(3)]
  for f in fns:
    f()


def curry_fn(i):
  return lambda: print(i)

@tf.function
def test2():
  print(tf.executing_eagerly())
  fns = [curry_fn(i) for i in range(3)]
  for f in fns:
    f()

tf.config.run_functions_eagerly(False)
test()
test2()
tf.config.run_functions_eagerly(True)
test()
test2()
```

bhack:
Also you could simply your first python impl with:

```python
  fns = [lambda i=i: print(i) for i in range(3)]
```

keras_cv/layers/preprocessing/random_augmentation_pipeline.py
LukeWood: you will only encounter the issue when you do NOT wrap test() in a tf.function

bhack: @LukeWood I don't think so as not wrapping the test in a `tf.function` gives exactly the same results as in this example.
I suppose that manipulating `tf.config.run_functions_eagerly` was already enough as a demonstration.

PR #430, opened 2022-05-12
Fix the wrong comment  about Autograph bug

bhack:
It was a "not safe" python code but it was mistakenly debugged as an autograph bug

https://github.com/keras-team/keras-cv/pull/428#discussion_r871297012

### DIFF
--- a/keras_cv/layers/preprocessing/random_augmentation_pipeline.py
+++ b/keras_cv/layers/preprocessing/random_augmentation_pipeline.py
@@ -95,12 +95,10 @@ class RandomAugmentationPipeline(
                 (), minval=0, maxval=len(self.layers), dtype=tf.int32
             )
 
-            # Warning!!!
-            # DO NOT REPLACE WITH A FOR LOOP
-            # Autograph has an edge case where capturing python for loop
-            # variables is inconsistent between eager and graph execution
-            # by using a list comprehension and currying, we mitigate
-            # our code against both of these cases.
+            # Warning: 
+            # In case of refactoring don't close over the loop variable
+            # https://docs.microsoft.com/en-us/archive/blogs/ericlippert/closing-over-the-loop-variable-considered-harmful
+    
             branch_fns = [
                 (i, self._curry_call_layer(inputs, layer))
                 for (i, layer) in enumerate(self.layers)


bhack:
See also:
https://discuss.python.org/t/make-lambdas-proper-closures/10553

keras_cv/layers/preprocessing/random_augmentation_pipeline.py
```diff
@@ -95,12 +95,11 @@ def _augment():
...
-            # DO NOT REPLACE WITH A FOR LOOP
-            # Autograph has an edge case where capturing python for loop
-            # variables is inconsistent between eager and graph execution
-            # by using a list comprehension and currying, we mitigate
-            # our code against both of these cases.
+            # Warning:
+            # Do not replace the currying function with a lambda.
+            # Originally we used a lambda, but due to Python's
+            # lack of loop level scope this causes unexpected
+            # behavior running outside of graph mode.
```
bhack, line 102: @LukeWood  @mdanatag I cannot reproduce this claim about `unexpected behavior running outside of graph mode`.

I have the same output in graph and eager mode with your gists https://github.com/keras-team/keras-cv/pull/428#discussion_r871297012

mdanatg: Yes, I think this is purely a Python gotcha, and as far as I can see, it happens consistently irrespective of graph mode.

My suggestion for comment would be something in the lines of: "Note: not using a lambda for currying to avoid loop variable closures, see <insert link here>".

bhack: @mdanatg As I need to open another PR probably how can we be a little more specific with the claim since this works correctly:

```python
import tensorflow as tf
@tf.function
def test():
  print(tf.executing_eagerly())
  ns = []
  fns = [lambda i=i: print(i) for i in range(3)]
  for f in fns:
    f()


def curry_fn(i):
  return lambda: print(i)

@tf.function
def test2():
  print(tf.executing_eagerly())
  fns = [curry_fn(i) for i in range(3)]
  for f in fns:
    f()

tf.config.run_functions_eagerly(False)
test()
test2()
tf.config.run_functions_eagerly(True)
test()
test2()
```

LukeWood: > @LukeWood @mdanatg I cannot reproduce this claim about `unexpected behavior running outside of graph mode`.
> 
> I have the same output in graph and eager mode with your gists [#428 (comment)](https://github.com/keras-team/keras-cv/pull/428#discussion_r871297012)

It is not about eager mode, it is about python execution vs tensorflow.  Remove the @tf.function annotation on test().

Thanks

bhack: I think it is pure python and not python vs TensorFlow. I've already replied in the other thread: https://github.com/keras-team/keras-cv/pull/428#discussion_r871713450 but you can see the execution here without `@tf.function`

```python
import tensorflow as tf

def test():
  print(tf.executing_eagerly())
  ns = []
  fns = [lambda: print(i) for i in range(3)]
  for f in fns:
    f()


def curry_fn(i):
  return lambda: print(i)

@tf.function
def test2():
  print(tf.executing_eagerly())
  fns = [curry_fn(i) for i in range(3)]
  for f in fns:
    f()

tf.config.run_functions_eagerly(False)
test()
test2()
tf.config.run_functions_eagerly(True)
test()
test2()
```

```python
True
2
2
2
False
0
1
2
True
2
2
2
True
0
1
2
```

LukeWood: Yes, it is pure python vs tensorflow autograph.

bhack: Is not the `def test()` in https://github.com/keras-team/keras-cv/pull/430#discussion_r871708447
like the he original gist that you have shared?

PR #432, opened 2022-05-12
Fix the comment about the lambda variable scope in a loop

bhack:
This fix the comment about the right root cause of the original bug as it was not related to Tensoflow, autograph, eager or graph mode but it was a pure python bug/quirks of the original implementation:

https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result

See more at https://github.com/keras-team/keras-cv/pull/430#pullrequestreview-971282078

/cc @mdanatg @martin-gorner

As this is the 2nd PR please comment inline with a review and don't edit the string directly with a commit. 
I am happy to reproduce this with any Gist you can share and eventually contribute/extend an autograph tests in TF if confirmed.

Thanks

### DIFF
--- a/keras_cv/layers/preprocessing/random_augmentation_pipeline.py
+++ b/keras_cv/layers/preprocessing/random_augmentation_pipeline.py
@@ -95,11 +95,10 @@ class RandomAugmentationPipeline(
                 (), minval=0, maxval=len(self.layers), dtype=tf.int32
             )
 
-            # Warning:
-            # Do not replace the currying function with a lambda.
-            # Originally we used a lambda, but due to Python's
-            # lack of loop level scope this causes unexpected
-            # behavior running outside of graph mode.
+            # Note:
+            # If you replace the currying function with a lambda please
+            # take care of the variable scope.
+            # https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result
             branch_fns = [
                 (i, self._curry_call_layer(inputs, layer))
                 for (i, layer) in enumerate(self.layers)


LukeWood:
The point I make in the comment is that it *does* work as expected in graph mode, even without the currying.  The reason it leaked into our build is that in batch mode it got compiled by @tf.function, masking the broken behavior.  That is the relation to graph mode.

bhack:
> The point I make in the comment is that it _does_ work as expected in graph mode, even without the currying. The reason it leaked into our build is that in batch mode it got compiled by @tf.function, masking the broken behavior. That is the relation to graph mode.

Do you have a gist or a Colab  to reproduce your claim? As with your original gist example I had the same wrong result in garph mode and pure python.

LukeWood:
```
import tensorflow as tf

def test():
  fns = []
  for i in range(3):
    fns.append(lambda: print(i))
  for f in fns:
    f()

test()


@tf.function
def test():
  fns = []
  for i in range(3):
    fns.append(lambda: print(i))
  for f in fns:
    f()

test()
```

LukeWood:
Looks like there is a discrepency between for loop and list comprehension in tf.function as well.

The reality is there is a discrepency in graph mode and pure python.  We should rely on currying.  Feel free to add a link to the information about the python closures, but don't remove the other.  thanks

bhack:
Ok now I can reproduce it with this gist (not with your original one). 
So I can investigate if we could have a TF issue/test or not as it will be more useful for the TF ecosystem then an untracked inline comment.

Can you confirm that we have also the correct result both in pure Python and TF when the python code is correct and at the same time still using lambda?

```python
import tensorflow as tf

def test():
  fns = []
  for i in range(3):
    fns.append(lambda i=i: print(i))
  for f in fns:
    f()

test()


@tf.function
def test():
  fns = []
  for i in range(3):
    fns.append(lambda i=i: print(i))
  for f in fns:
    f()

test() 
```

LukeWood:
> Ok now I can reproduce it with this gist (not with your original one). So I can investigate if we could have a TF issue/test or not as it will be more useful for the TF ecosystem then an untracked inline comment.
> 
> Can you confirm that we have also the correct result both in pure Python and TF when the python code is correct and at the same time still using lambda?
> 
> ```python
> import tensorflow as tf
> 
> def test():
>   fns = []
>   for i in range(3):
>     fns.append(lambda i=i: print(i))
>   for f in fns:
>     f()
> 
> test()
> 
> 
> @tf.function
> def test():
>   fns = []
>   for i in range(3):
>     fns.append(lambda i=i: print(i))
>   for f in fns:
>     f()
> 
> test() 
> ```

run the code and you will see the first one fails.  So no, it is not consistent between python and autograph.

LukeWood:
> Ok now I can reproduce it with this gist (not with your original one). So I can investigate if we could have a TF issue/test or not as it will be more useful for the TF ecosystem then an untracked inline comment.
> 
> Can you confirm that we have also the correct result both in pure Python and TF when the python code is correct and at the same time still using lambda?
> 
> ```python
> import tensorflow as tf
> 
> def test():
>   fns = []
>   for i in range(3):
>     fns.append(lambda i=i: print(i))
>   for f in fns:
>     f()
> 
> test()
> 
> 
> @tf.function
> def test():
>   fns = []
>   for i in range(3):
>     fns.append(lambda i=i: print(i))
>   for f in fns:
>     f()
> 
> test() 
> ```

I see, adding lambda i=i works.  Regardless, the comment is valid and a warning is valid.  You can add the link, but there is a relation to autograph as to how this bug got into the main build.  We can merge a new link, but I want to keep the old comment.

LukeWood:
Did some more digging after cloning locally.

The link also goes far beyond our character limit, I'm really not convinced this PR is a value add.  I think the original communicates the issue.  

I am happy to propagate the issue to tf.autograph and provide reproduction snippets, but I think the comment serves the purpose.

Thank you for the Pr and concern, but I think the description is currently accurate.

bhack:
> I see, adding lambda i=i works. 

It is not me, it is the official [Python FAQ on how to write correct python code](https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result).

Nevermind, I will open a ticket myself in TF but IMHO we don't need to use comment to not write wrong Python code.

LukeWood:
> > I see, adding lambda i=i works.
> 
> It is not me, it is the official [Python FAQ on how to write correct python code](https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result).
> 
> Nevermind, I will open a ticket myself in TF but IMHO we don't need to use inline comments to not write wrong Python code.

I wrote the comment because the incorrect code DOES work in graph mode with a for loop.  This is a valid warning.

Thank you for your concern on this issue

bhack:
> I wrote the comment because the incorrect code DOES work in graph mode with a for loop.

Yes what I meant this is the bug:
Incorrect python code worked correctly in graph mode

But as we don't run only in graph mode if we write it CORRECTLY in pure python it will also work perfectly in graph mode.

So the generic claim `Do not replace the currying function with a lambda.` is not valid as with lambda it works in all the cases if wrote correctly in pure Python. 

But as you like, you are the maintainer, I am only expressing my rationale. 

I will come back later if we could fix this in TF.

bhack:
For who is interested we are checking this at upstream level with:

https://github.com/tensorflow/tensorflow/issues/56089

PR #434, opened 2022-05-13
RandomChoice layer

### LINKED ISSUE #408: Support `RandomChoice` - randomly pick from list!
**Short Description**

Let's say in your training, you want to use all of the follows:

- `cutmix-mixup-fmix-mosaic` 
- `gaussian_blur-zoom_blur-motion_blur-median-blur`
- `optical_distortion-elastic-transform-grid_distortion`
- `equalization-clahe`
- `grid_mask-cutput`
- etc.

then it's required to randomly pick one among others (in general) in a single pass during training, for example: `[cutmix-zoom_blur-optical_distortion-clahe-cutput]`. So, it would be nice to have a functionality to feasibly do this task.

**Papers**
Most of the top used libraries/frameworks offer similar strategies.

**Existing Implementations**
- https://pytorch.org/vision/main/generated/torchvision.transforms.RandomChoice.html

**Other Information**
In the title, the name `RandomChoice` is picked for demonstration purposes. The thing is, we may need to introduce or modify any of the existing functionality if it's reasonable to take.

chjort:
Implementation of a `RandomChoice` layer. Closes #408.

* `RandomChoice` will at each call _uniformly_ sample a _single_ layer from a list of layers, and call it on the inputs.
* `RandomAugmentationPipeline` has been refactored to use `RandomChoice` internally.

Right now `RandomChoice` is placed under the `keras_cv.layers.preprocessing` module, however it is not really a preprocessing layer, so it might be worth thinking about if it should be located elsewhere.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,4 +1,4 @@
-name: Tests
+nane: Tests
 
 on:
   push:
@@ -32,7 +32,7 @@ jobs:
           ${{ runner.os }}-pip-
     - name: Install dependencies
       run: |
-        pip install tf-nightly
+        pip install tensorflow==2.9.0rc2
         pip install -e ".[tests]" --progress-bar off --upgrade
     - name: Test with pytest
       run: |
@@ -60,7 +60,7 @@ jobs:
           ${{ runner.os }}-pip-
     - name: Install dependencies
       run: |
-        pip install tf-nightly
+        pip install tensorflow==2.9.0rc2
         pip install -e ".[tests]" --progress-bar off --upgrade
     - name: Lint
       run: bash shell/lint.sh
@@ -76,7 +76,7 @@ jobs:
         python-version: 3.7
     - name: Install dependencies
       run: |
-        pip install tf-nightly
+        pip install tensorflow==2.9.0rc2
         python -m pip install --upgrade pip setuptools wheel twine
     - name: Build and publish
       env:
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,4 +1,4 @@
-nane: Tests
+name: Tests
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


keras_cv/layers/preprocessing/random_choice.py
```diff
@@ -0,0 +1,102 @@
...
+        self.auto_vectorize = auto_vectorize
+        self.seed = seed
+
+    def _curry_call_layer(self, inputs, layer):
+        def call_layer():
+            return layer(inputs)
+
+        return call_layer
+
+    def _augment(self, inputs, *args, **kwargs):
```
LukeWood, line 69: I think we actually want to do this on a sample-wise basis, not a batch-wise basis.

As of now, it looks like this does everything on a batch-wise basis.  Thoughts?

chjort: It is on a sample-wise basis. The `_augment` method of `BaseImageAugmentationLayer` is passed single samples as inputs. It is mapped to a batch in the `_batch_augment`.

Sample-wise basis is also tested for here: https://github.com/keras-team/keras-cv/pull/434/files#diff-ba41275c34609917e7f01d01bfe4f36d015082795580f91ab30bc3b9aa4cece2R71

LukeWood: Oops!  Yep, you're right.  Sorry!

LukeWood:
Looks like nightly is failing; I'll merge this when 2.9 stabilizes.  Thanks!

bhack:
> Looks like nightly is failing; I'll merge this when 2.9 stabilizes.  Thanks!

I think that, as we are configured now, we are diverging every day "2.10" nightly releases. 

Do want to track TF 2.9rc2/2.9 in Keras-cv master or "2.10" nightly/master?

.github/workflows/actions.yml
```diff
@@ -1,4 +1,4 @@
-name: Tests
+nane: Tests
```
bhack, line 1: typo: `name`

bhack:
:vertical_traffic_light: Green light

PR #435, opened 2022-05-15
Introduces MaybeApply layer.

### LINKED ISSUE #422: `MaybeApply` high order abstraction
We should consider an implementation for a `MaybeApply` layer:

```
maybe_solarize = MaybeApply(Solarization(), rate=0.5)
# maybe_solarize solarizes half of the images passed to it
```

https://github.com/chjort/chambers/blob/b796f507b7353fb141b5c8b026b811a023d303e6/chambers/augmentations/image_augmentations.py#L514

Notes:
- we can rely on this in RandomAugmentationPipeline
- this must be element-wise, not batch wise!

sebastian-sz:
Closes #422 .

Maybe Apply layer is a wrapper around native Keras layers or `BaseImageAugmentationLayer` that applies, specified operation to random samples in a batch.

#### Example usage
```python
images = tf.random.stateless_uniform(shape=(5, 2, 2, 1), seed=[0, 1])

zero_out = tf.keras.layers.Lambda(lambda x: 0 * x)
maybe_apply = MaybeApply(layer=zero_out, rate=0.5, seed=1234)

outputs = maybe_apply(images)
print(outputs)  # Random samples in a batch have been zero'ed out.
```

#### Performance / overhead
The layer introduces an overhead for the layer it wraps. It looks like the layer wrapped in `MaybeApply` takes ~2x the time to execute. I'm not sure if any improvements here can be made - I tried to implement the same behaviour using `tf.gather_nd` + `tf.scatter_nd_update` but got similar or even worse results for larger batches.

Below are latency (ms) measurements for Solarization and Posterization. Both were benchmarked with XLA to avoid `tf.function` retracing.

Solarization + MaybeApply Overhead
Posterization + MaybeApply overhead

#### Known Issues
The layer throws error in XLA with `auto_vectorize=True`, regardless of what layer it wraps. I'm not really sure why
```python
InvalidArgumentError: Reading input as constant from a dynamic tensor is not yet supported. Xla shape: s32[<=32]
```
but it works with `auto_vectorize=False`.

### DIFF
--- a/keras_cv/layers/preprocessing/maybe_apply.py
+++ b/keras_cv/layers/preprocessing/maybe_apply.py
@@ -1,0 +1,120 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class MaybeApply(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Apply provided layer to random elements in a batch.
+
+    Args:
+        layer: a keras Layer or BaseImageAugmentationLayer. This layer will be applied
+            to randomly chosen samples in a batch.
+        rate: controls the frequency of applying the layer. 1.0 means all elements in
+            a batch will be modified. 0.0 means no elements will be modified.
+            Defaults to 0.5.
+        auto_vectorize: bool, whether to use tf.vectorized_map or tf.map_fn for
+            batched input. Setting this to True might give better performance but
+            currently doesn't work with XLA. Defaults to False.
+        seed: integer, controls random behaviour.
+
+    Example usage:
+        # Let's declare an example layer that will set all image pixels to zero.
+        zero_out = tf.keras.layers.Lambda(lambda x: 0 * x)
+
+        # Create a small batch of random, single-channel, 2x2 images:
+        images = tf.random.stateless_uniform(shape=(5, 2, 2, 1), seed=[0, 1])
+        print(images[..., 0])
+        # <tf.Tensor: shape=(5, 2, 2), dtype=float32, numpy=
+        # array([[[0.08216608, 0.40928006],
+        #         [0.39318466, 0.3162533 ]],
+        #
+        #        [[0.34717774, 0.73199546],
+        #         [0.56369007, 0.9769211 ]],
+        #
+        #        [[0.55243933, 0.13101244],
+        #         [0.2941643 , 0.5130266 ]],
+        #
+        #        [[0.38977218, 0.80855536],
+        #         [0.6040567 , 0.10502195]],
+        #
+        #        [[0.51828027, 0.12730157],
+        #         [0.288486  , 0.252975  ]]], dtype=float32)>
+
+        # Apply the layer with 50% probability:
+        maybe_apply = MaybeApply(layer=zero_out, rate=0.5, seed=1234)
+        outputs = maybe_apply(images)
+        print(outputs[..., 0])
+        # <tf.Tensor: shape=(5, 2, 2), dtype=float32, numpy=
+        # array([[[0.        , 0.        ],
+        #         [0.        , 0.        ]],
+        #
+        #        [[0.34717774, 0.73199546],
+        #         [0.56369007, 0.9769211 ]],
+        #
+        #        [[0.55243933, 0.13101244],
+        #         [0.2941643 , 0.5130266 ]],
+        #
+        #        [[0.38977218, 0.80855536],
+        #         [0.6040567 , 0.10502195]],
+        #
+        #        [[0.        , 0.        ],
+        #         [0.        , 0.        ]]], dtype=float32)>
+
+        # We can observe that the layer has been randomly applied to 2 out of 5 batches.
+    """
+
+    def __init__(self, layer, rate=0.5, auto_vectorize=False, seed=None, **kwargs):
+        super().__init__(seed=seed, **kwargs)
+
+        if not (0 <= rate <= 1.0):
+            raise ValueError(f"rate must be in range [0, 1]. Received rate: {rate}")
+
+        self._layer = layer
+        self._rate = rate
+        self.auto_vectorize = auto_vectorize
+        self.seed = seed
+
+    def augment_image(self, image, transformation=None):
+        should_apply = transformation > 1.0 - self._rate
+        return tf.cond(should_apply, lambda: self._layer(image), lambda: image)
+
+    def get_random_transformation(self, image=None, label=None, bounding_box=None):
+        return self._random_generator.random_uniform(shape=())
+
+    def augment_label(self, label, transformation=None):
+        should_apply = transformation > 1.0 - self._rate
+        return tf.cond(
+            should_apply, lambda: self._layer.augment_label(label), lambda: label
+        )
+
+    def augment_bounding_box(self, bounding_box, transformation=None):
+        should_apply = transformation > 1.0 - self._rate
+        return tf.cond(
+            should_apply,
+            lambda: self._layer.augment_bounding_box(bounding_box),
+            lambda: bounding_box,
+        )
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "rate": self._rate,
+                "layer": self._layer,
+                "seed": self.seed,
+                "auto_vectorize": self.auto_vectorize,
+            }
+        )
+        return config
--- a/keras_cv/layers/preprocessing/maybe_apply.py
+++ b/keras_cv/layers/preprocessing/maybe_apply.py
@@ -31,7 +31,7 @@ class MaybeApply(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
 
     Example usage:
         # Let's declare an example layer that will set all image pixels to zero.
-        zero_out = tf.keras.layers.Lambda(lambda x: 0 * x)
+        zero_out = tf.keras.layers.Lambda(lambda x: {"images": 0 * x["images"]})
 
         # Create a small batch of random, single-channel, 2x2 images:
         images = tf.random.stateless_uniform(shape=(5, 2, 2, 1), seed=[0, 1])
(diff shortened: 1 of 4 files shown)


bhack:
> It looks like the layer wrapped in MaybeApply takes ~2x the time to execute

You could try to trace/profile the code to collect more insight:
https://www.tensorflow.org/api_docs/python/tf/profiler/experimental/Trace

sebastian-sz:
@bhack 
Running with XLA for both `MaybeApply` and raw layer I'm getting:
`Your program is NOT input-bound because only 0.0% of the total step time sampled is waiting for input.`   
And it says that almost entire time is spent on `Host Compute Time`. Still this is e.g. `12.2 ms` for native layer and `25.3 ms` for `MaybeApply` wrapped.

However, running without XLA: I'm getting a different message for `MaybeApply`:
`Your program is POTENTIALLY input-bound because 36.5% of the total step time sampled is spent on 'All Others' time (which could be due to I/O or Python execution or both).`

I am not reading any data from the disk, so I'd assume it could be Python's overhead? Not sure.

bhack:
Do you have a gist with tf.profile to reproduce this?

sebastian-sz:
@bhack this is what I'm using:
[benchmark_maybe_apply.zip](https://github.com/keras-team/keras-cv/files/8694939/benchmark_maybe_apply.zip)

bhack:
> Known Issues
The layer throws error in XLA with auto_vectorize=True, regardless of what layer it wraps. I'm not really sure why
InvalidArgumentError: Reading input as constant from a dynamic tensor is not yet supported. Xla shape: s32[<=32]

Yes it Is:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tf2xla/xla_op_kernel.cc#L138-L146

Do you have, can you post, Tensorboard screenshots for the ops section for plain vs maybe wrapper?

The ops tables are like these:
https://github.com/keras-team/keras-cv/issues/165#issuecomment-1061225089

More in general with XLA is also interesting to check how well the graph ops that we have used in the implementaton are optimized-fused or not with the HLO dump (not all the possible ops permutation sequences have the same XLA coverage and optimizzation quality). 
See:
https://github.com/keras-team/keras-cv/issues/141#issuecomment-1043716578

bhack:
>@bhack this is what I'm using:
[benchmark_maybe_apply.zip](https://github.com/keras-team/keras-cv/files/8694939/benchmark_maybe_apply.zip)

You cannot do like you have done when you don't use a Keras model. If you don't want to use a model check:
https://www.tensorflow.org/tensorboard/graphs#graphs_of_tffunctions

But as this bug is still open https://github.com/tensorflow/tensorboard/issues/1961 it is better that you use a model.

sebastian-sz:
@bhack Instead of running the layer, I should wrap it in a `tf.keras.Model`, compile and run `.predict` ?

I'm still struggling to create a HLO graph. Will post when I gather more information.

bhack:
> @bhack Instead of running the layer, I should wrap it in a tf.keras.Model, compile and run .predict ?

Yes as the graph is on model build:
https://github.com/keras-team/keras/blob/master/keras/engine/training.py#L401-L405

>I'm still struggling to create a HLO graph. Will post when I gather more information.

https://www.tensorflow.org/xla#inspect_compiled_programs

LukeWood:
> Closes #422 .
> 
> Maybe Apply layer is a wrapper around native Keras layers or `BaseImageAugmentationLayer` that applies, specified operation to random samples in a batch.
> 
> #### Example usage
> ```python
> images = tf.random.stateless_uniform(shape=(5, 2, 2, 1), seed=[0, 1])
> 
> zero_out = tf.keras.layers.Lambda(lambda x: 0 * x)
> maybe_apply = MaybeApply(layer=zero_out, rate=0.5, seed=1234)
> 
> outputs = maybe_apply(images)
> print(outputs)  # Random samples in a batch have been zero'ed out.
> ```
> 
> #### Performance / overhead
> The layer introduces an overhead for the layer it wraps. It looks like the layer wrapped in `MaybeApply` takes ~2x the time to execute. I'm not sure if any improvements here can be made - I tried to implement the same behaviour using `tf.gather_nd` + `tf.scatter_nd_update` but got similar or even worse results for larger batches.
> 
> Below are latency (ms) measurements for Solarization and Posterization. Both were benchmarked with XLA to avoid `tf.function` retracing.
> 
> Solarization + MaybeApply Overhead Posterization + MaybeApply overhead
> 
> #### Known Issues
> The layer throws error in XLA with `auto_vectorize=True`, regardless of what layer it wraps. I'm not really sure why
> 
> ```python
> InvalidArgumentError: Reading input as constant from a dynamic tensor is not yet supported. Xla shape: s32[<=32]
> ```
> 
> but it works with `auto_vectorize=False`.

I feel like we need a page in our docs for performance related artifacts like these!  These are great to have.

keras_cv/layers/preprocessing/maybe_apply.py
```diff
@@ -0,0 +1,120 @@
...
+            raise ValueError(f"rate must be in range [0, 1]. Received rate: {rate}")
+
+        self._layer = layer
+        self._rate = rate
+        self.auto_vectorize = auto_vectorize
+        self.seed = seed
+
+    def augment_image(self, image, transformation=None):
+        should_apply = transformation > 1.0 - self._rate
+        return tf.cond(should_apply, lambda: self._layer(image), lambda: image)
```
LukeWood, line 91: This actually won't work with labels and bounding boxes, as we'll need to make sure they receive the same transformation.

We should do this on the `_augment` level.  Check out PR https://github.com/keras-team/keras-cv/pull/434

sebastian-sz: Done

sebastian-sz:
@bhack Following your suggestion on using the `tf.keras.Model` class I did the benchmarks again using `.predict_on_batch` and `model.compile(jit_compile=...)`.

When using `tf.keras.Model` the wrapper layer runs in a similar amount of time as the regular layer.
XLA mean latency (ms)
Eager mean latency (ms)

[benchmark_model.zip](https://github.com/keras-team/keras-cv/files/8708483/benchmark_model.zip)

bhack:
@sebastian-sz Can you post the same graph without XLA?

As we still don't expose an API to control the XLA compilation x layer in Keras/Keras-cv (see https://github.com/keras-team/keras/issues/16403) using just `model.compile(jit_compile=...).` is very risky cause any single ops not supported by XLA will let the model compilation to fail  (see https://github.com/keras-team/keras-cv/pull/146#issuecomment-1062288950).

Extra: Without extending Keras-cv layers test for XLA compilation we will never know the exact list of layers with one or more usupported XLA ops. 
Also the list/inventory in TF Docs hasn't been updated for years: https://github.com/tensorflow/tensorflow/issues/14798#issuecomment-1047796247

bhack:
P.s. Just to clarify I meant in graph mode as `model.compile` default but without XLA instread of `model.compile` eager that I suppose  produced the eager graph in your previous post.

sebastian-sz:
@bhack 
>Edit:
Checking your ZIP and the XLA boolean I suppose that your 2nd graph is [graph mode without XLA](https://github.com/keras-team/keras/blob/v2.8.0/keras/engine/training.py#L553-L557) instead of eager mode also at it seems to me too fast to be eager. Right?

Thanks, yes, my bad. Added option for `run_eagerly`. Eager is slower than graph / XLA (as expected) and the difference between layers is also small:
Corrected eager mode mean latency  ms

> @sebastian-sz Can you post the same graph without XLA?

I'm not sure I follow - Is the above graph what you requested?

bhack:
> I'm not sure I follow - Is the above graph what you requested?

No It was the previous one labeled as eager but It was in graph mode without jit compile

keras_cv/layers/preprocessing/maybe_apply_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+class MaybeApplyTest(tf.test.TestCase, parameterized.TestCase):
+    rng = tf.random.Generator.from_non_deterministic_state()
+
+    @parameterized.parameters([-0.5, 1.7])
+    def test_raises_error_on_invalid_rate_parameter(self, invalid_rate):
+        with self.assertRaises(ValueError):
+            MaybeApply(rate=invalid_rate, layer=ZeroOut())
+
+    def test_works_with_batched_input(self):
```
LukeWood, line 44: can you pass a seed so that this test is not potentially flaky?  Given, it is 1/2^32 flakiness, but still may as well seed it.

LukeWood (changes requested):
2 minor comments then good to go.

keras_cv/layers/preprocessing/maybe_apply_test.py
sebastian-sz: Added seed to `rng` on line 37.

LukeWood: Thanks!  does this seed the layer too?

sebastian-sz: You are right. Added `seed` param to layer as well.

keras_cv/layers/preprocessing/maybe_apply.py
```diff
@@ -0,0 +1,105 @@
...
+            to randomly chosen samples in a batch.
+        rate: controls the frequency of applying the layer. 1.0 means all elements in
+            a batch will be modified. 0.0 means no elements will be modified.
+            Defaults to 0.5.
+        auto_vectorize: bool, whether to use tf.vectorized_map or tf.map_fn for
+            batched input. Setting this to True might give better performance but
+            currently doesn't work with XLA. Defaults to False.
+        seed: integer, controls random behaviour.
+
+    Example usage:
```
LukeWood, line 32: unindent Example usage: section and wrap code in:

```python
```

It eventually renders as markdown on keras.io :)

keras_cv/layers/preprocessing/maybe_apply.py
```diff
@@ -0,0 +1,105 @@
...
+# limitations under the License.
+import tensorflow as tf
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class MaybeApply(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Apply provided layer to random elements in a batch.
+
+    Args:
+        layer: a keras Layer or BaseImageAugmentationLayer. This layer will be applied
```
LukeWood, line 22: "Layer should not modify the size of provided inputs"

LukeWood (approved):

keras_cv/layers/preprocessing/maybe_apply.py
sebastian-sz: Added this to `Args: layer:` section in the docstring.

keras_cv/layers/preprocessing/maybe_apply.py
sebastian-sz: Yes you are right, I forgot about it. Unindented the Example section and added backticks to code.

LukeWood:
Looks good to me @sebastian-sz Thanks for the contribution!

PR #444, opened 2022-05-19
Add get_config() to COCOMetrics and add test coverage

LukeWood:


keras_cv/metrics/serialization_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def custom_compare(obj1, obj2):
+    if isinstance(obj1, core.FactorSampler):
+        return obj1.get_config() == obj2.get_config()
+    else:
+        return obj1 == obj2
+
+
+def config_equals(config1, config2):
```
qlzh727, line 28: Just curious why this is needed? (we can update after this PR to unblock user)

qlzh727 (approved):

PR #453, opened 2022-05-20
Implement clear error messages to CutMix and MixUp

LukeWood:


keras_cv/layers/preprocessing/cut_mix.py
```diff
@@ -122,6 +123,23 @@ def _update_labels(self, images, labels, lambda_sample, permutation_order):
         labels = lambda_sample * labels + (1.0 - lambda_sample) * cutout_labels
         return images, labels
 
+    def _validate_inputs(self, inputs):
```
bhack, line 126: Is this only relative to cut_mix or it could be used to other layers?

LukeWood: Only these.  For the other layers it would be included in base.  These two are special cases

Issue #131, opened 2022-02-11
Implement RandomResizedCrop layer


sayakpaul:
Randomly resized cropping is pretty much a standard augmentation transformation that is used to train vision models. Recently, a team at Facebook also showed why it's often necessary to include this transformation for better generalization [1].

An implementation of this layer in PyTorch is available as [`RandomResizedCrop`](https://pytorch.org/vision/main/generated/torchvision.transforms.RandomResizedCrop.html).

**References**:

[1] https://arxiv.org/abs/2106.05121

LukeWood:
Yeah, I agree RandomResizedCrop should be included.

innat:
@sayakpaul 
how about this one ?https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_crop/

@LukeWood If the above layer is the desired layer for the above request, then you may want to take a look at [this](https://github.com/keras-team/keras/blob/v2.8.0/keras/layers/preprocessing/image_preprocessing.py#L225-L226).

LukeWood:
that layer doesn't fill the need because the Resize portion is also important.  RandomResizedCrop basically stretch _and_ crop images randomly.

innat:
something like this? 

```python
tf.keras.layers.RandomCrop
tf.keras.layers.Resizing
```

LukeWood:
So: passing `Resizing` followed by `RandomCrop` does work, but you temporarily have a tensor with dimension `None, None` for `width, height` after the resizing.  This may be wasteful/undesirable to work with, so I think this may be a useful KPL to add it as a single operation .   It may also be more efficient.

innat:
Sorry, didn't catch your concern. Could you elaborate? Like, are you saying that after passing the input to resize a layer, the shape of the output tensor becomes `(100, 100, 3)` to `(None, None, 3)`?

sayakpaul:
We do also need to remember that this layer needs to handle more than 1 sample i.e., an arbitrary number of images having a uniform shape.

LukeWood:
> Sorry, didn't catch your concern. Could you elaborate? Like, are you saying that after passing the input to resize a layer, the shape of the output tensor becomes `(100, 100, 3)` to `(None, None, 3)`?

Precisely, because each individual image should be augmented independently.  So they will have random sizes.

LukeWood:
They’d be a RaggedTensor really, which is a bad UX

sayakpaul:
[...]

PR #457, opened 2022-05-28
Adds Random Resized Crop

beresandras:
This PR implements the Random Resized Crop augmentation minimally.

Example augmentations of the elephant image from the [keras.io guide](https://keras.io/guides/keras_cv/custom_image_augmentations/#overriding-augmentimage):
aug_image
aug_image2

Corresponding issue: #131 

Details to discuss:
* interface:
  * currently an `area_factor` and `aspect_ratio_factor` is used, both between 0 and 1, following the [API design guidelines](https://github.com/keras-team/keras-cv/blob/master/.github/API_DESIGN.md). They are translated to new areas and aspect ratios via new_value = 1 +/- factor(). The maximal area_factor of exactly 1.0 can lead to a crop with 0% or 200% area.
  * we could also use min/max area and min/max aspect ratio, similarly to the [torchvision implementation](https://pytorch.org/vision/stable/generated/torchvision.transforms.RandomResizedCrop.html#torchvision.transforms.RandomResizedCrop). This might be easier to understand, and would enable the user to set area = 8% - 100%, aspect_ratio = 3/4 - 4/3, corresponding to the commonly used [Inception crops (Section 6)](https://arxiv.org/abs/1409.4842).
  * Do we even want crops to go outside the image?
* crop method: currently the crop + resize is implemented with the `ImageProjectiveTransformV3` operation. It could be done with tf.image.crop_and_resize() as well.
* batched implementation: I proposed a [batched implementation](https://github.com/beresandras/image-augmentation-layers-keras/blob/master/augmentations.py#L30:L77) originally in the issue, as it is not difficult to do, however, it does not fit well with the current `BaseImageAugmentationLayer API`.
* log-uniform aspect ratio distribution: When having aspect ratios e.g. between 1:2 and 2:1 (0.5 and 2.0) one might want to keep the distribution centered at 1.0. To achieve that, we could use a log-uniform sampling between the two value, similarly to the torchvision implementation. This would somehow require a default log-uniform factor sampler, or a hardcoded sampling. I am not sure how this should be implemented.
* aliasing: we should not allow image areas above 4 (2x-2x zoom out along both sides), as they would lead to aliasing, and the ImageProjectiveTransformV3 does not have an antialias argument, in contrast to [tf.image.resize()](https://www.tensorflow.org/api_docs/python/tf/image/resize). 

TODO:
* tests

### DIFF
--- a/keras_cv/layers/preprocessing/random_resized_crop.py
+++ b/keras_cv/layers/preprocessing/random_resized_crop.py
@@ -19,25 +19,10 @@ from keras_cv.utils import preprocessing
 
 
 @tf.keras.utils.register_keras_serializable(package="keras_cv")
-class RandomShear(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
-    """Randomly shears an image.
+class RandomResizedCrop(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Randomly crops an image and resizes it to its original resolution.
     Args:
-        x_factor: A tuple of two floats, a single float or a
-            `keras_cv.FactorSampler`. For each augmented image a value is sampled
-            from the provided range. If a float is passed, the range is interpreted as
-            `(0, x_factor)`.  Values represent a percentage of the image to shear over.
-             For example, 0.3 shears pixels up to 30% of the way across the image.
-             All provided values should be positive.  If `None` is passed, no shear
-             occurs on the X axis.
-             Defaults to `None`.
-        y_factor: A tuple of two floats, a single float or a
-            `keras_cv.FactorSampler`. For each augmented image a value is sampled
-            from the provided range. If a float is passed, the range is interpreted as
-            `(0, y_factor)`. Values represent a percentage of the image to shear over.
-            For example, 0.3 shears pixels up to 30% of the way across the image.
-            All provided values should be positive.  If `None` is passed, no shear
-            occurs on the Y axis.
-            Defaults to `None`.
+        TODO
         interpolation: interpolation method used in the `ImageProjectiveTransformV3` op.
              Supported values are `"nearest"` and `"bilinear"`.
              Defaults to `"bilinear"`.
@@ -52,8 +37,8 @@ class RandomShear(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
 
     def __init__(
         self,
-        x_factor=None,
-        y_factor=None,
+        area_factor=0.0,
+        aspect_ratio_factor=0.0,
         interpolation="bilinear",
         fill_mode="reflect",
         fill_value=0.0,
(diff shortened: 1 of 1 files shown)


beresandras:
Sample code to test it:
```python
import matplotlib.pyplot as plt
import tensorflow as tf

from keras_cv.layers.preprocessing.random_resized_crop import RandomResizedCrop

SIZE = (300, 300)
elephants = tf.keras.utils.get_file(
    "african_elephant.jpg", "https://i.imgur.com/Bvro0YD.png"
)
elephants = tf.keras.utils.load_img(elephants, target_size=SIZE)
elephants = tf.keras.utils.img_to_array(elephants) / 255

augmenter = RandomResizedCrop(area_factor=0.75, aspect_ratio_factor=0.25)
elephants = augmenter(elephants).numpy()

plt.imsave("aug_image.png", elephants)
```

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,147 @@
...
+import tensorflow as tf
+
+from keras_cv.utils import preprocessing
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class RandomResizedCrop(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
+    """Randomly crops an image and resizes it to its original resolution.
+    Args:
+        TODO
```
LukeWood, line 25: lets document these factors, great job!!!

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,147 @@
...
+        aspect_ratio_factor=0.0,
+        interpolation="bilinear",
+        fill_mode="reflect",
+        fill_value=0.0,
+        seed=None,
+        **kwargs,
+    ):
+        super().__init__(seed=seed, **kwargs)
+        self.area_factor = preprocessing.parse_factor(
+            area_factor, param_name="area_factor", seed=seed
```
LukeWood, line 50: are there high and low bounds on this value?  Be sure to document them, and pass them to `min` and `max` in parse_factor

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,147 @@
...
+        self.seed = seed
+
+    def get_random_transformation(self, image=None, label=None, bounding_box=None):
+        # random area and aspect ratio
+        random_area = (
+            1.0
+            + preprocessing.random_inversion(self._random_generator)
+            * self.area_factor()
+        )
+        random_aspect_ratio = (
```
LukeWood, line 73: great use of the utils!

LukeWood (changes requested):
This looks wonderful!  Great work

LukeWood:
@beresandras let me know if you need help adding test cases!  This is looking fantastic!  Thanks for your swift contribution!

LukeWood:
FYI @beresandras I updated the API a bit.

LukeWood:
Some notes:

- we need to include a crop size
- we want to resize images first, then crop second.

LukeWood:
Hey @beresandras are you intending to finish this PR up?  If not, we can try to find someone to run it across the finish line!

No worries either way.

beresandras:
Yes, I will finish it in the next few days.

beresandras:
@LukeWood unfortunately it seems that I won't be able to finish the PR for another week. Please try to find someone to complete it if you can.

AdityaKane2001:
@sayakpaul @LukeWood 

I would like to work on this one. I'll create a PR asap.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,162 @@
...
+# limitations under the License.
+import warnings
+
+import tensorflow as tf
+
+from keras_cv.utils import preprocessing
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class RandomResizedCrop(tf.keras.__internal__.layers.BaseImageAugmentationLayer):
```
LukeWood, line 22: lets use keras_cv.layers.BaseImageAugmentationLayer going forward

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,162 @@
...
+            `factor=0.0` makes this layer perform a no-op operation, while a value of
+            1.0 doubles the size of the image.  If a single float is used, a value
+            between `0.0` and the passed float is sampled.  In order to ensure the value
+            is always the same, please pass a tuple with two identical floats:
+            `(0.5, 0.5)`.
+        crop_size: a tuple of two integers used as the target size to crop images to.
+        interpolation: interpolation method used in the `ImageProjectiveTransformV3` op.
+             Supported values are `"nearest"` and `"bilinear"`.
+             Defaults to `"bilinear"`.
+        aspect_ratio_factor: (Optional) A tuple of two floats, a single float or
```
LukeWood, line 36: I think aspect ratios respect the aspect ratio convention, 1.0 does not change the image.  aspect_ratio=4/3 shifts aspect ratio by 4/3, etc.

LukeWood: As such, we should default this ratio to (1, 1) so it performs a no-op.  I don't think distorting aspect ratio should be a default behavior.

Perhaps we make this argument required?

LukeWood (changes requested):
Minor changes

LukeWood:
Thanks for the contribution!  Closing in favor of https://github.com/keras-team/keras-cv/pull/499

PR #462, opened 2022-06-05
Add pre-commit hooks

bhack:
https://github.com/keras-team/keras-cv/pull/14#issuecomment-1146745495

- Add pre-commit hook
- Opitional pre-commit in devcontainer envs
- Install locales in devcontainer envs for lint script communication with IDE (UTF-8)
- Move some linting args to the config to achieve an increased IDEs compatibility
- Add an optional arg to the linting script to just check staged files when we are in a pre-commit "mode"
- Add new Vscode (May 2022) extensions for `black` and `isort`
 
/cc @LukeWood @haifeng-jin


LukeWood:
Are .githooks automatically symlinked by the devcontainer?

.devcontainer/devcontainer.json
```diff
@@ -13,17 +13,17 @@
...
 	],
-
 	"features": {
 		"git": {
-			"version": "os-provided",
-		},
+			"version": "os-provided"
+		}
 	},
+	"postCreateCommand": "git config core.hooksPath .githooks"
```
bhack, line 27: > Are .githooks automatically symlinked by the devcontainer?

Yes It is this line. 
For who is using the manual setup we need to put the same command in the markdown section we have at:

https://github.com/keras-team/keras-cv/blob/master/.github/CONTRIBUTING.md#setup-environment

LukeWood:
Personally I prefer pre commit hooks to be opt in - largely because I find them counter productive (I often commit code that would fail the hooks to save a working copy).  

Maybe we should leave them as opt in.

.github/CONTRIBUTING.md
```diff
@@ -63,6 +63,7 @@ You can achieve this by running the following commands:
 gh repo fork keras-team/keras-cv --clone --remote
 cd keras-cv
 pip install ".[tests]"
+git config core.hooksPath .githooks
```
LukeWood, line 66: Can we make this opt-in and a separate step.

LukeWood:
Thanks for the PR.

Can you update the PR description to include a list of all the changes you are intending to make here?

bhack:
> Personally I prefer pre commit hooks to be opt in - largely because I find them counter productive (I often commit code that would fail the hooks to save a working copy).
> 
> Maybe we should leave them as opt in.

It is that if we move the check to the `pre-push` hook instead of `pre-commit` the logic to detect files to be pushed it will be really more complex.
 precommit wheel have [this logic already implemented](https://github.com/samskiter/pre-commit/blob/master/pre_commit/resources/pre-push-tmpl`) but i suppose it will be complex if we want to, be more relex, to understand on what branch you are pushing (e.g. if was it a branch not related to a PR? etc..).

The problem with containers / devcontainer / codespaces vs a local setup is that often these environments are used as ephemeral and then you have to remember, every time you create/launch a container / codespace, to reinstall the hooks so many are used to have a ready environment with this type of setup.

.github/CONTRIBUTING.md
bhack: I did but this undermines our purpose of reducing all the time we already spend with formatting requests.

At least can we, in a subsequent PR, avoid to run all the jobs if the linting one is failing? 
It is really a waste of resources, including environmental ones

LukeWood:
> It is that if we move the check to the pre-push hook instead of pre-commit the logic to detect files to be pushed it will be really more complex.

I think we are miscommunicating.  I'm only suggesting we remove automatic setup from the dev-container.  I am not suggesting we move to pre-push.

.github/CONTRIBUTING.md
bhack: > At least can we, in a subsequent PR, avoid to run all the jobs if the linting one is failing?
It is really a waste of resources, including environmental ones

It could be quite simple if we will use the `needs` syntax in Github Actions:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

bhack:
> I think we are miscommunicating. I'm only suggesting we remove automatic setup from the dev-container. I am not suggesting we move to pre-push.

We are miscommunicating cause they are two independent topics:
- the pre-push part of the comment was an alternative candidate to your "optional" request to not totally weak the scope of this PR that is to minimize the re-format comment requests on PR and on PR branches. What I meant is that: also now you already know that you need to lint the code before pushing as it is in the docs but many people don't execute this step. It is not something theoretical it is in the repo stats (and also in the TF repo history if we want to have longer accumulated stats).

- The second point was instead was related to the user expectation and the ephemeral nature of a developer environment created by devcontainers vs your local and "more permanent" setup.

bhack:
I put it optional because I don't have too much time to invest in this small infra PR but in any case the rationale I have expressed remains valid.

.github/CONTRIBUTING.md
bhack: I've added a new PR https://github.com/keras-team/keras-cv/pull/464

shell/lint.sh
```diff
@@ -1,25 +1,35 @@
 #!/bin/bash
-isort --sl -c --profile=black .
+files="."
```
LukeWood, line 2: can you explain this change a bit more thoroughly - maybe through comments?  

Add a usage comment perhaps, showing both ways to use the script.

bhack: Do you mean files variabile handling?

LukeWood: Now that there are two ways to execute the script we should have a comment at the top indicating how to use the script.

bhack: Done

shell/lint.sh
```diff
@@ -1,14 +1,12 @@
 #!/bin/bash
-# We could optionally pass a list of file as arg ($1)
+# Usage: # lint.sh can be used without arguments to lint the entire project: 
+#
```
LukeWood, line 3: looks like a formatting issue here

shell/lint.sh
```diff
@@ -1,14 +1,12 @@
...
-if [ ! -z "$1" ]
-  then
-    files=$(eval $1)
-    if [ -z "$files" ]
-      then
-        exit 0
-    fi
+if [ $# -ne 0  ]
+  then 
+    files=$(eval $@)
```
LukeWood: is $(eval $@) needed here?  shouldn't the shell expand these?

LukeWood:
Thanks @bhack

PR #464, opened 2022-06-06
Make tests depend on linting

bhack:
With this we could minimize the tests runs when the contributor doesn't format  the code locally with the script, pre-commit hook or the VsCode integration we offer in the repository and is going to totally rely on the CI job.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -7,6 +7,7 @@ on:
     types: [created]
 jobs:
   test:
+    needs: [format]
     name: Test the code
     runs-on: ${{ matrix.os }}
     strategy:


LukeWood:
thanks for the PR

Any reason to not run them in parallel?

LukeWood:
to me, triggering CI on non-formatted code is still valuable.  

This can reduce developer round trips, as it can make you aware of two issues you didn't know about at once.

bhack:
Yes but when you are lazy abusing the CI for fix you code formatting issues you are always run tests also if you have not formated the code.

As we don't merge without formatting check It Is ok it Is better to fix it earlier and then execute tests.

bhack:
> CI’s purpose is to save developers' time.

I don't think it is the main scope of the CI but suppose that it is the case right now linting is jut 5s over 51s of preparing the environment for linting.

PR #475, opened 2022-06-08
Fixing Equalization Shape Calculation During Trace Time

quantumalaviya:


keras_cv/layers/preprocessing/equalization.py
```diff
@@ -101,7 +101,7 @@ def augment_image(self, image, transformation=None):
         image = tf.cast(image, tf.int32)
         image = tf.vectorized_map(
             lambda channel: self.equalize_channel(image, channel),
-            tf.range(tf.shape(image)[-1]),
+            tf.range(image.shape[-1]),
```
LukeWood, line 104: Can we add a test case to equalization.py that asserts that the output shape of the layer is the same as the input shape both in and outside of a model?  Right now, equalization outputs a shape:

(batch, height, width, None)

for some reason, but your PR should fix this.

quantumalaviya:
@LukeWood Turns out that the problem was with `vectorized_map` instead of the `tf.shape` call. I have changed it to `map_fn` and it seems to work now!

PR #483, opened 2022-06-10
fix typo in Devcontainer

bhack:


bhack:
We could prepare a specific github Action later with:
https://github.com/devcontainers/ci/blob/main/docs/github-action.md
/cc @haifeng-jin

Issue #328, opened 2022-04-18
Mscoco: Sort bboxes before picking top detections


bhack:
bboxes are not sorted before picking top max_detections: https://github.com/keras-team/keras-cv/issues/222#issuecomment-1089073856

PR #492, opened 2022-06-14
Fix major numerical issue in COCOMeanAveragePrecision

LukeWood:
Noticed this bug during numerical testing, now I have sufficient test coverage of this and have fixed the issue.

In short:
```
ground_truths = utils.filter_boxes(...)
```

was overwriting every single other class if a single one was missing.  Now this is fixed

### DIFF
--- a/keras_cv/metrics/coco/recall.py
+++ b/keras_cv/metrics/coco/recall.py
@@ -98,6 +98,11 @@ class COCORecall(keras.metrics.Metric):
         num_thresholds = len(iou_thresholds)
         num_categories = len(class_ids)
 
+        if any([c < 0 for c in class_ids]):
+            raise ValueError(
+                "class_ids must be positive.  Got " f"class_ids={class_ids}"
+            )
+
         self.true_positives = self.add_weight(
             name="true_positives",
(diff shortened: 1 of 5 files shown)


keras_cv/metrics/coco/recall.py
```diff
@@ -151,6 +156,8 @@ def update_state(self, y_true, y_pred, sample_weight=None):
             dtype=self.compute_dtype,
         )
 
+        y_pred = utils.sort_bounding_boxes(y_pred, axis=bounding_box.XYXY.CONFIDENCE)
```
bhack, line 159: Is this going to fix also https://github.com/keras-team/keras-cv/issues/328 ?

LukeWood: yep

bhack: Do we have gained only -0.01  with al these fixes?

LukeWood: yep

LukeWood: in extreme cases though it saves us potentially a result that should be 1.0 showing as 0.0

PR #493, opened 2022-06-14
Migrate RandomRotation to Keras_CV

divyashreepathihalli:
Issue link : https://github.com/keras-team/keras-cv/issues/470


keras_cv/layers/preprocessing/preprocessing_utils.py
```diff
@@ -24,3 +25,141 @@ def ensure_tensor(inputs, dtype=None):
...
+                tf.sin(angles)[:, None],
+                tf.cos(angles)[:, None],
+                y_offset[:, None],
+                tf.zeros((num_angles, 2), tf.float32),
+            ],
+            axis=1,
+        )
+
+
+def transform(
```
LukeWood, line 84: some of these utils already exist here: https://github.com/keras-team/keras-cv/blob/master/keras_cv/utils/preprocessing.py#L213

LukeWood, line 84: some of these utils already exist here: https://github.com/keras-team/keras-cv/blob/master/keras_cv/utils/preprocessing.py#L213

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -0,0 +1,225 @@
...
+from keras_cv.layers.preprocessing import preprocessing_utils as utils
+from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
+    BaseImageAugmentationLayer,
+)
+
+H_AXIS = -3
+W_AXIS = -2
+
+IMAGES = "images"
+LABELS = "labels"
```
LukeWood, line 28: are these used anywhere?

LukeWood, line 28: are these used anywhere?

LukeWood (changes requested):
Thanks for the PR!  Some early stage comments.

keras_cv/layers/preprocessing/preprocessing_utils.py
divyashreepathihalli: Thanks for pointing it out @LukeWood. I have updated it.

keras_cv/layers/preprocessing/random_rotation.py
divyashreepathihalli: They aren't, sorry - removed them.

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -0,0 +1,220 @@
...
+import numpy as np
+import tensorflow as tf
+
+from keras_cv import bounding_box
+from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
+    BaseImageAugmentationLayer,
+)
+from keras_cv.utils import preprocessing
+
+H_AXIS = -3
```
qlzh727, line 24: Should we use 1 and 2 here (NHWC)? what's the reason for using reverse index here?

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -0,0 +1,220 @@
...
+            )
+        image = tf.expand_dims(image, 0)
+        image_shape = tf.shape(image)
+        h = image_shape[H_AXIS]
+        w = image_shape[W_AXIS]
+        # origin coordinates, all the points on the image are rotated around
+        # this point
+        origin_x, origin_y = int(h / 2), int(w / 2)
+        angle = transformation["angle"]
+        angle = -angle
```
qlzh727, line 155: Any reason we are doing a negation here?

keras_cv/layers/preprocessing/random_rotation_test.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertEqual(expected_output.shape, actual_output.shape)
+
+    def test_random_rotation_inference(self):
+        input_images = np.random.random((2, 5, 8, 3)).astype(np.float32)
+        expected_output = input_images
+        layer = RandomRotation(0.5)
+        actual_output = layer(input_images, training=False)
+        self.assertAllClose(expected_output, actual_output)
+
+    def test_distribution_strategy(self):
```
qlzh727, line 35: We usually don't directily test layers with tf.distribute. Should we have a model test for this?

keras_cv/layers/preprocessing/random_rotation_test.py
```diff
@@ -0,0 +1,122 @@
...
+        with strat.scope():
+            layer = RandomRotation(0.5)
+            output = strat.run(lambda: layer(input_images, training=True))
+        values = output.values
+        self.assertAllEqual(2, len(values))
+
+    def test_config_with_custom_name(self):
+        layer = RandomRotation(0.5, name="image_preproc")
+        config = layer.get_config()
+        layer_1 = RandomRotation.from_config(config)
```
qlzh727, line 49: any reason for layer_1 here? I didn't see layer_2.

keras_cv/layers/preprocessing/random_rotation_test.py
```diff
@@ -0,0 +1,122 @@
...
+        bboxes = tf.convert_to_tensor([[200, 200, 400, 400], [100, 100, 300, 300]])
+        # 180 rotation.
+        layer = RandomRotation(factor=(0.5, 0.5), bounding_box_format="xyxy")
+        output_bbox = layer.augment_bounding_boxes(
+            input_image,
+            bboxes,
+            transformation=layer.get_random_transformation(),
+        )
+        expected_output = np.asarray(
+            [[111, 112, 312, 312], [212, 211, 412, 412]]
```
qlzh727, line 99: I was somehow expecting the x/y are the same after rotation, is it because of int casting?

keras_cv/layers/preprocessing/random_rotation_test.py
```diff
@@ -0,0 +1,122 @@
...
+        ).astype(np.int32)
+        expected_output = np.reshape(expected_output, (2, 4))
+        self.assertAllClose(expected_output, output_bbox)
+
+    def test_augment_bbox_dict_input(self):
+        input_image = np.random.random((512, 512, 3)).astype(np.float32)
+        bboxes = tf.convert_to_tensor([[200, 200, 400, 400], [100, 100, 300, 300]])
+        input = {"images": input_image, "bounding_boxes": bboxes}
+        # 180 rotation.
```
qlzh727, line 108: I don't think this is 180 degree rotation.

qlzh727 (changes requested):

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -0,0 +1,220 @@
...
+          wrapping around to the opposite edge.
+        - *nearest*: `(a a a a | a b c d | d d d d)` The input is extended by
+          the nearest pixel.
+      interpolation: Interpolation mode. Supported values: `"nearest"`,
+        `"bilinear"`.
+      seed: Integer. Used to create a random seed.
+      fill_value: a float represents the value to be filled outside the
+        boundaries when `fill_mode="constant"`.
+    Bounding_box_format: The format of bounding boxes of input dataset.
+        Supported formats are -`"xyxy"`,`"rel_xyxy"`, `"xyWH"`, `"center_xyWH"`.
```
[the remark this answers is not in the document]
LukeWood: I agree.  The way I implemented this in the bounding box utilities is by checking the keys for the converter functions.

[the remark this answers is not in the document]
divyashreepathihalli: I can point to this function - https://github.com/keras-team/keras-cv/blob/master/keras_cv/bounding_box/converters.py#:~:text=def%20convert_format(boxes%2C%20source%2C%20target%2C%20images%3DNone%2C%20dtype%3D%22float32%22)%3A
or update it once user guides are updated.

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -0,0 +1,220 @@
...
+        - *wrap*: `(a b c d | a b c d | a b c d)` The input is extended by
+          wrapping around to the opposite edge.
+        - *nearest*: `(a a a a | a b c d | d d d d)` The input is extended by
+          the nearest pixel.
+      interpolation: Interpolation mode. Supported values: `"nearest"`,
+        `"bilinear"`.
+      seed: Integer. Used to create a random seed.
+      fill_value: a float represents the value to be filled outside the
+        boundaries when `fill_mode="constant"`.
+    Bounding_box_format: The format of bounding boxes of input dataset.
```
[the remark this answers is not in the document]
divyashreepathihalli: done

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -0,0 +1,220 @@
...
+            bounding_boxes = bounding_box.convert_format(
+                bounding_boxes, source=self.bounding_box_format, target="xyxy"
+            )
+        image = tf.expand_dims(image, 0)
+        image_shape = tf.shape(image)
+        h = image_shape[H_AXIS]
+        w = image_shape[W_AXIS]
+        # origin coordinates, all the points on the image are rotated around
+        # this point
+        origin_x, origin_y = int(h / 2), int(w / 2)
```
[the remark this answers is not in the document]
divyashreepathihalli: int(<scalar tensor>) returns an int.

keras_cv/layers/preprocessing/random_rotation.py
divyashreepathihalli: negating the angle makes sure the rotation is anti clockwise.

keras_cv/layers/preprocessing/random_rotation_test.py
divyashreepathihalli: there is layer and layer_1

keras_cv/layers/preprocessing/random_rotation_test.py
divyashreepathihalli: good catch! it was 30 degrees, I have updated it to 180.

keras_cv/layers/preprocessing/random_rotation_test.py
```diff
@@ -0,0 +1,122 @@
...
+        layer = RandomRotation(factor=(0.0833, 0.0833), bounding_box_format="xyxy")
+        output_bbox = layer(input)
+        expected_output = np.asarray([[179, 135, 452, 408], [42, 98, 316, 372]]).astype(
```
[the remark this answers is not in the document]
divyashreepathihalli: Done!

keras_cv/layers/preprocessing/random_rotation_test.py
divyashreepathihalli: there is layer and layer_1

keras_cv/layers/preprocessing/random_rotation_test.py
divyashreepathihalli: I am not sure I understand, can you explain a little bit more?

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -73,10 +73,9 @@ class RandomRotation(BaseImageAugmentationLayer):
       seed: Integer. Used to create a random seed.
       fill_value: a float represents the value to be filled outside the
         boundaries when `fill_mode="constant"`.
-      bounding_box_format: The format of bounding boxes of input dataset.
-        Refer <a href=
-        "https://github.com/keras-team/keras-cv/blob/master/keras_cv/bounding_box/converters.py"
-        >here</a> for more details on supported bounding box formats
+      bounding_box_format: The format of bounding boxes of input dataset. Refer
```
LukeWood, line 76: This is good for now, in the future we can link to API docs generated from bounding_box/__init__.py

divyashreepathihalli: ACK

keras_cv/layers/preprocessing/random_rotation_test.py
LukeWood: I think layer_reconstructed would be a good name

divyashreepathihalli: Done

LukeWood (approved):
looks good to me, lets wait for @qlzh727

keras_cv/layers/preprocessing/random_rotation_test.py
qlzh727: I mean the result after rotation should be [111, 111, 312, 312]? Not sure why there is a off by one here for to corner.

keras_cv/layers/preprocessing/random_rotation.py
qlzh727: i see. We probably should comment this somehow.

keras_cv/layers/preprocessing/random_rotation.py
qlzh727: i see. Since I didn't find the negation part in image, we probably should document it here.

qlzh727 (approved):

keras_cv/layers/preprocessing/random_rotation_test.py
```diff
@@ -0,0 +1,122 @@
...
+        ).astype(np.float32)
+        expected_output = np.reshape(expected_output, (5, 5, 1))
+        self.assertAllClose(expected_output, output_image)
+
+    def test_augment_image(self):
+        input_image = np.reshape(np.arange(0, 25), (5, 5, 1)).astype(np.float32)
+        # 180 rotation.
+        layer = RandomRotation(factor=(0.5, 0.5))
+        output_image = layer.augment_image(
+            input_image, transformation=layer.get_random_transformation()
```
LukeWood, line 63: Instead of get random transformation can we call the layer?

LukeWood (changes requested):
Can we also add an example to `examples/layers/preprocessing` that loads an image and bounding box, augments it, and shows the result?  

The RandomShear PR has an example of this you can base this on.

keras_cv/layers/preprocessing/random_rotation_test.py
divyashreepathihalli: @luke this test is specifically to verify augment_image(). So I am testing that separately.

LukeWood: Ideally, tests should be structured to use user facing APIs instead of relying on implementation details.  Call will do the same thing here, and if it does not that is symptomatic of a greater error at play

divyashreepathihalli: Cool, In that case the tests are already covered. I will remove this.

keras_cv/layers/preprocessing/random_rotation_test.py
divyashreepathihalli: Oh yes, this could be because of  int casting

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -235,20 +235,20 @@ def call(self, inputs, training=True):
     def _augment(self, inputs):
         image = inputs.get(IMAGES, None)
         label = inputs.get(LABELS, None)
-        bounding_box = inputs.get(BOUNDING_BOXES, None)
+        bounding_boxes = inputs.get(BOUNDING_BOXES, None)
         transformation = self.get_random_transformation(
```
LukeWood, line 239: hey Divya, can you update _augment to introduce the following:
```
    def _augment(self, inputs):
        image = inputs.get(IMAGES, None)
        label = inputs.get(LABELS, None)
        bounding_box = inputs.get(BOUNDING_BOXES, None)
        transformation = self.get_random_transformation(
            image=image, label=label, bounding_boxes=bounding_box
        )
        image = self.augment_image(image, transformation=transformation, bounding_boxes=bounding_box, label=label)
        result = {IMAGES: image}
        if label is not None:
            label = self.augment_target(label, transformation=transformation, bounding_boxes=bounding_box, label=label, image=image)
            result[LABELS] = label
        if bounding_box is not None:
            bounding_box = self.augment_bounding_boxes(
                bounding_box, transformation=transformation, bounding_boxes=bounding_box, label=label, image=image
            )
            result[BOUNDING_BOXES] = bounding_box
        return result
```

This includes the other label types into the other augment_* methods.

LukeWood (changes requested):
You'll also need to update the testing layer in MaybeApply to integrate these changes as well.

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -0,0 +1,224 @@
...
+        # find readjusted coordinates of bounding box to represent it in corners
+        # format
+        min_cordinates = tf.math.reduce_min(out, axis=1)
+        max_cordinates = tf.math.reduce_max(out, axis=1)
+        bboxes_out = tf.concat([min_cordinates, max_cordinates], axis=1)
+        # cordinates cannot be float values, it is casted to int32
+        bboxes_out = bounding_box.convert_format(
+            bboxes_out, source="xyxy", target=self.bounding_box_format
+        )
+        bboxes_out = tf.cast(bboxes_out, bbox_dtype)
```
LukeWood, line 206: I agree these casts make the behavior less surprising, but I wonder if they are worth it.  Perhaps casting to `self.compute_dtype` is more consistent.

LukeWood (changes requested):
Few more minor changes, then I think it LGTM.

I also think a visualization of the bounding boxes similar to the RandomShear bounding box PR would be very useful in convincing me of the correctness of the operation.

keras_cv/layers/preprocessing/random_rotation.py
divyashreepathihalli: casting to `self.compute_dtype` converts it bbox output to float, even if bbox input is int. And for bbox coordinates -float is not expected.

LukeWood (approved):

keras_cv/layers/preprocessing/random_rotation.py
LukeWood: Actually I'm not sure that I agree with this: float is a reasonable datatype for bounding boxes.  rel_* bounding boxes are floats, many COCO annotations have coordinates set as floats, rescaling operations output bounding boxes as floats. 

My concern is that if we don't cast them in all layers the behavior is now inconsistent

LukeWood: Actually I'm not sure that I agree with this: float is a reasonable datatype for bounding boxes.  rel_* bounding boxes are floats, many COCO annotations have coordinates set as floats, rescaling operations output bounding boxes as floats. 

My concern is that if we don't cast them in all layers the behavior is now inconsistent

LukeWood: Actually I'm not sure that I agree with this: float is a reasonable datatype for bounding boxes.  rel_* bounding boxes are floats, many COCO annotations have coordinates set as floats, rescaling operations output bounding boxes as floats. 

My concern is that if we don't cast them in all layers the behavior is now inconsistent

LukeWood (changes requested):
lets sort out the issue with casting first

keras_cv/layers/preprocessing/random_rotation.py
divyashreepathihalli: If the input format is float, we will be returning float, but if the input format is int, then returning a float value would not make sense to me. If the bboxes are int values for an image of size 512x512, what does a float value of coordinates indicate? what location on the image?

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -201,9 +202,11 @@ def augment_bounding_boxes(self, bounding_boxes, transformation, **kwargs):
         bboxes_out = tf.concat([min_cordinates, max_cordinates], axis=1)
         # cordinates cannot be float values, it is casted to int32
         bboxes_out = bounding_box.convert_format(
-            bboxes_out, source="xyxy", target=self.bounding_box_format
+            bboxes_out,
+            source="xyxy",
+            target=self.bounding_box_format,
```
LukeWood, line 207: in keras we prefer full names to abbreviations: `bounding_boxes_out`

Let's update all of these to say `bounding_boxes` instead of bboxes

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -138,7 +140,6 @@ def augment_bounding_boxes(self, bounding_boxes, transformation, **kwargs):
         image = None
         if "image" in kwargs:
             image = kwargs.get("image")
-        bbox_dtype = bounding_boxes.dtype
         if self.bounding_box_format is None:
             raise ValueError(
                 "bounding_box_format has not been specified during layer "
```
LukeWood, line 145: lets provide more actionable feedback to users.  Errors should be self correction; ideally without a trip to the docs.

```
"`RandomRotation()` was called with bounding boxes, but no `bounding_box_format` was specified in the constructor.  Please specify a bounding box format in the constructor.  i.e. `RandomRotation(bounding_box_format='xyxy')`
```

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -138,7 +140,6 @@ def augment_bounding_boxes(self, bounding_boxes, transformation, **kwargs):
         image = None
```
LukeWood, line 140: can be shortened to:

image = kwargs.get("image", None)

LukeWood (changes requested):
Thanks Divya, some minor style comments now that the major design points are sorted

keras_cv/layers/preprocessing/random_rotation.py
divyashreepathihalli: Done

keras_cv/layers/preprocessing/random_rotation.py
divyashreepathihalli: Done

keras_cv/layers/preprocessing/random_rotation.py
divyashreepathihalli: Done

LukeWood:
Thanks @divyashreepathihalli!  Feel free to contribute yxyx if you’re blocked, and feel free to add a bounding box demo script example!

divyashreepathihalli:
> 
Thanks Luke! I'll do that.

PR #499, opened 2022-06-16
Random Resized Crop

### LINKED ISSUE #131: Implement RandomResizedCrop layer

AdityaKane2001:
@sayakpaul @LukeWood 

Creating this draft PR to initiate discussions regarding RandomResizedCrop layer. Code is copied over from #457. Please let me know if you would prefer to start from scratch. 

I will study and make relevant changes to the code shortly according to the comments in #457. 

/auto Closes #131


AdityaKane2001:
@sayakpaul @LukeWood 

Please feel free to review this PR now.

I went through the implementation and refactored the same. I believe this implementation can be easily done using [`tf.image.crop_resize`](https://www.tensorflow.org/api_docs/python/tf/image/crop_and_resize) as mentioned in the [comment](https://github.com/keras-team/keras-cv/pull/457#issue-1251744408) in #457. I have jotted down a [gist](https://colab.research.google.com/gist/AdityaKane2001/279a7235577450af727af396b0c35c7d/rrc_using_cropandresize.ipynb) of the same. 

`ImageProjectiveTransformV3` seems a bit of over the top to me, since the same result can be achieved with `tf.image.crop_resize` but with better readability. So I have replaced `ImageProjectiveTransformV3` with `tf.image.crop_resize`. Here is a [gist](https://colab.research.google.com/gist/AdityaKane2001/fb9f287e8458f97e0075bffdb9af641f/scratchpad.ipynb) using the updated implementation.

Both [`tf.image.crop_resize`](https://www.tensorflow.org/api_docs/python/tf/image/crop_and_resize) and [`tf.raw_ops.ImageProjectiveTransformV3`](https://www.tensorflow.org/api_docs/python/tf/raw_ops/ImageProjectiveTransformV3) are vectorized themselves. So batching and unbatching images and boxes (or transform) arguments is redundant. Is there some way to avoid doing that?

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -15,33 +15,27 @@
...
-            `factor` controls the extent to which the image is resized.
-            `factor=0.0` makes this layer perform a no-op operation, while a value of
-            1.0 doubles the size of the image.  If a single float is used, a value
-            between `0.0` and the passed float is sampled.  In order to ensure the value
-            is always the same, please pass a tuple with two identical floats:
-            `(0.5, 0.5)`.
-        crop_size: a tuple of two integers used as the target size to crop images to.
+        target_size: A uple of two integers used as the target size to crop 
+            images to.
+        aspect_ratio_factor: (Optional) A tuple of two floats. Represents the
```
LukeWood: I wonder if there is a such thing as a "sensible default" here.  Does torchvision use 3/4 and 4/3, or was that a number we picked?  Unsure on this as of now.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -15,33 +15,27 @@
...
-            between `0.0` and the passed float is sampled.  In order to ensure the value
-            is always the same, please pass a tuple with two identical floats:
-            `(0.5, 0.5)`.
-        crop_size: a tuple of two integers used as the target size to crop images to.
+        target_size: A uple of two integers used as the target size to crop 
+            images to.
+        aspect_ratio_factor: (Optional) A tuple of two floats. Represents the 
+            lower and upper bound for the aspect ratio of the cropped image 
+            before resizing it to `target_size`. Defaults to (3./4., 4./3.).
+        area_factor: (Optional) A tuple of two floats, a single float or
```
LukeWood: also unsure if this should be provided by default, or required

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -50,103 +44,86 @@ class RandomResizedCrop(BaseImageAugmentationLayer):
...
+    def __init__(self,
+                 target_size,
+                 aspect_ratio_factor=(3. / 4., 4. / 3.),
+                 area_factor=(0.08, 1.0),
+                 interpolation="bilinear",
+                 fill_mode="reflect",
+                 fill_value=0.0,
+                 seed=None,
+                 **kwargs):
+        super(RandomResizedCrop, self).__init__(seed=seed, **kwargs)
```
LukeWood, line 57: in python3 you can use super().__init__(...) instead of super(RandomResizedCrop, self)

This holds for all cases.  Before everyone did this to be backwards compatible with python2.

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: @LukeWood 

Yes torchvision uses the same. I believe these defaults are also used in the original paper, but I'm not sure.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -50,103 +44,86 @@ class RandomResizedCrop(BaseImageAugmentationLayer):
...
-            self.aspect_ratio_factor())
-
-        # corresponding height and width (1 = original height/width)
-        new_height = tf.sqrt(random_area / random_aspect_ratio)
-        new_width = tf.sqrt(random_area * random_aspect_ratio)
-
-        # random offsets for the crop, inside or outside the image
-        height_offset = self._random_generator.random_uniform(
+        area_factor = self.area_factor()
+        aspect_ratio = tf.random.uniform((), minval=self.aspect_ratio_factor[0],
```
LukeWood: this should be sampled from the aspect ratio factor, right?  Perhaps we should make this a factor sampler?

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -50,103 +44,86 @@ class RandomResizedCrop(BaseImageAugmentationLayer):
...
-        new_width = tf.sqrt(random_area * random_aspect_ratio)
-
-        # random offsets for the crop, inside or outside the image
-        height_offset = self._random_generator.random_uniform(
+        area_factor = self.area_factor()
+        aspect_ratio = tf.random.uniform((), minval=self.aspect_ratio_factor[0], 
+            maxval=self.aspect_ratio_factor[1], dtype=tf.float32)
+
+        new_height = tf.clip_by_value(tf.sqrt(area_factor / aspect_ratio), 0.0,
+                                    1.0)  # to avoid unwanted/unintuitive effects
```
LukeWood: we should also be able to support zoomed in resizes- as in they can be larger than the original image

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -50,103 +44,86 @@ class RandomResizedCrop(BaseImageAugmentationLayer):
...
-            aspect_ratio_factor, param_name="aspect_ratio_factor", seed=seed)
-        if resize_factor == 0.0 and aspect_ratio_factor == 0.0:
-            warnings.warn(
-                "RandomResizedCrop received both `area_factor=0.0` and "
-                "`aspect_ratio_factor=0.0`. As a result, the layer will perform no "
-                "augmentation.")
+    def __init__(self,
+                 target_size,
+                 aspect_ratio_factor=(3. / 4., 4. / 3.),
+                 area_factor=(0.08, 1.0),
```
LukeWood, line 51: shouldn't we also include zoomed images?  So, (0.5, 2)

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -50,103 +44,86 @@ class RandomResizedCrop(BaseImageAugmentationLayer):
...
-            0.0,
-        ])
-        image = preprocessing.transform(
-            images=image,
-            transforms=transform,
-            interpolation=self.interpolation,
-            fill_mode=self.fill_mode,
-            fill_value=self.fill_value,
+        # See bit.ly/tf_crop_resize for more details
+        augmented_image = tf.image.crop_and_resize(
```
LukeWood, line 118: nice use of tf.image - but you won't be using fill mode if you do this.  Note that our `preprocessing.transform` function does actually support mirror fill modes, while tf.image.crop_and_resize() will not.

keras_cv/layers/preprocessing/random_resized_crop_test.py
```diff
@@ -18,12 +18,26 @@
 
 
 class RandomResizedCropTest(tf.test.TestCase):
-    def test_preserves_output_shape(self):
+    def train_augments_image(self):
```
LukeWood, line 20: lets add a test case where the area factor is 2.0

LukeWood (changes requested):

sayakpaul:
@AdityaKane2001 @LukeWood 

If we're using code from https://github.com/keras-team/keras-cv/pull/457 let's add the author of https://github.com/keras-team/keras-cv/pull/457 as a [co-author of the commit](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)?

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: Factor sampler does not support values >1. Hence I kept is as a tuple.

AdityaKane2001:
@sayakpaul @LukeWood 

I don't understand what do we mean by having `area_factor` > 1.0. How would the image look like in that case? Torchvision simply outputs the entire image if `area_factor` > 1. Gist [here](https://colab.research.google.com/gist/AdityaKane2001/20c6d5a370d0f8e8d421e0ad3dd32219/scratchpad.ipynb).

My general approach to this PR is to make it consistent with the torchvision implementation. The earlier implementation in #457 was a bit different from it, both in terms of behavior and API signature.

> If we're using code from https://github.com/keras-team/keras-cv/pull/457 let's add the author of https://github.com/keras-team/keras-cv/pull/457 as a [co-author of the commit](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)?

I had included [code credits](https://github.com/keras-team/keras-cv/pull/499/commits/8237608ee14a74d9b60d18e65eee5aef8ec0bb63#diff-6834fc0ba79a728ed2b63f05a521251d5edc3c057ad1320022cfb83f2449bd2aR25) for the earlier PR in the initial commits, but I removed it later since I refactored a major chunk of the code. But if we will be going with `preprocessing.transform`, then I'll include the credits again.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+import tensorflow as tf
+
+from keras_cv.layers import BaseImageAugmentationLayer
+from keras_cv.utils import preprocessing
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class RandomResizedCrop(BaseImageAugmentationLayer):
+    """
+    Randomly crops a part of an image and resizes it to provided size.
```
sayakpaul, line 25: Should we briefly jot down the approach that we're taking here. There are many implementation of RandomResizedCrop. I believe if we jot down our summary here it'd be easier for the community to understand what we're doing.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Args:
+        target_size: A uple of two integers used as the target size to crop
+            images to.
+        aspect_ratio_factor: (Optional) A tuple of two floats. Represents the
+            lower and upper bound for the aspect ratio of the cropped image
+            before resizing it to `target_size`. Defaults to (3./4., 4./3.).
+        area_factor: (Optional) A tuple of two floats, a single float or
+            `keras_cv.FactorSampler`. Represents the lower and upper bound for
```
sayakpaul, line 34: `keras_cv.FactorSampler`. Represents the lower and upper bound for

```suggestion
            `keras_cv.FactorSampler`. Represents the lower and upper bounds for
```

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+            warnings.warn(
+                "RandomResizedCrop received both `area_factor=0.0` and "
+                "`aspect_ratio_factor=0.0`. As a result, the layer will perform no "
+                "augmentation.")
+
+    def get_random_transformation(self,
+                                  image=None,
+                                  label=None,
+                                  bounding_box=None):
+        area_factor = self.area_factor()
```
sayakpaul, line 80: In case `area_factor` was a float, then would `preprocessing.parse_factor()` ensure it gets converted to a function? If so, what kind?

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+
+        y1 = height_offset
+        y2 = height_offset + new_height
+        x1 = width_offset
+        x2 = width_offset + new_width
+
+        return [[y1, x1, y2, x2]]
+
+    def augment_image(self, image, transformation):
+        image = tf.expand_dims(image, axis=0)
```
sayakpaul, line 114: Does `BaseImageAugmentationLayer` automatically make this a vectorized function? Otherwise, should we add a check to see if `image` already has a leading batch dimension? Happy to review other resources to make my understanding of the architecture more refined.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+        boxes = transformation
```
sayakpaul, line 115: The output of `get_random_transformation()`? How does it internally work though? I would have expected to see the actual image sizes being used to calculate the modified aspect ratios. Do you mind taking me through your approach in a bit more detail? 

@LukeWood do we have documentation on how the `BaseImageAugmentationLayer` is supposed to be extended? A couple of things seem a bit abstract to me. For example, for layers, I would have expected to see `call()` method be implemented but I understand why that's not the case here. But not sure how `transformation` from `augment_image()` is propagated and used.

keras_cv/layers/preprocessing/random_resized_crop_test.py
sayakpaul: +1

keras_cv/layers/serialization_test.py
```diff
@@ -36,54 +36,96 @@ def config_equals(config1, config2):
...
+        }),
+        ("ChannelShuffle", preprocessing.ChannelShuffle, {
+            "seed": 1
+        }),
+        ("CutMix", preprocessing.CutMix, {
+            "seed": 1
+        }),
+        ("Equalization", preprocessing.Equalization, {
+            "value_range": (0, 255)
+        }),
```
sayakpaul, line 50: Needs fixing?

keras_cv/layers/serialization_test.py
```diff
@@ -36,54 +36,96 @@ def config_equals(config1, config2):
...
         ("Grayscale", preprocessing.Grayscale, {}),
-        ("GridMask", preprocessing.GridMask, {"seed": 1}),
-        ("MixUp", preprocessing.MixUp, {"seed": 1}),
+        ("GridMask", preprocessing.GridMask, {
+            "seed": 1
+        }),
+        ("MixUp", preprocessing.MixUp, {
+            "seed": 1
+        }),
```
sayakpaul, line 57: Same.

keras_cv/layers/serialization_test.py
```diff
@@ -36,54 +36,96 @@ def config_equals(config1, config2):
...
+            },
         ),
         (
             "Posterization",
             preprocessing.Posterization,
-            {"bits": 3, "value_range": (0, 255)},
+            {
+                "bits": 3,
+                "value_range": (0, 255)
+            },
```
sayakpaul, line 72: Same and elsewhere.

sayakpaul (changes requested):

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: Sure. The basic approach in the current implementation is same as one would expect: randomly crop the images and resize to given target size. I'll add a line regarding that.

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: Yes. `preprocessing.parse_factor()` returns a `FactorSampler`: 
https://github.com/keras-team/keras-cv/blob/9da8e27bbfea0e860efc287006913e320b1aa4cb/keras_cv/utils/preprocessing.py#L117-L120

This object can be called later to get a random value in the specified range. However it does not support >1 max values at the moment.

keras_cv/layers/serialization_test.py
AdityaKane2001: If it's okay I'll fix the formatting errors when all other things are ironed out to avoid redundant efforts.

sayakpaul: Yes, that's a better approach.

keras_cv/layers/preprocessing/random_resized_crop.py
sayakpaul: Which paper are you referring to @AdityaKane2001? Inception?

keras_cv/layers/preprocessing/random_resized_crop.py
sayakpaul: I think it would be beneficial to also expand a little bit on our approach of randomly cropping an image in the first place. Some implementations sample lower and upper bounds on the coordinates which are then used for cropping. But it varies. Hence, I am suggesting it.

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: My bad. The original paper did not have any defaults. I recollected this snippet from SimCLR:
image

keras_cv/layers/preprocessing/random_resized_crop.py
sayakpaul: Thank you! That helps.

sayakpaul:
Regarding https://github.com/keras-team/keras-cv/pull/499#issuecomment-1159536076, how far Luke's comment in https://github.com/keras-team/keras-cv/pull/499#discussion_r901056559 impacts this? 

I'd prefer using `tf.image.crop_and_resize()` since the code does read cleaner and crisper with it. Conceptually also, it becomes simpler.

AdityaKane2001:
> Regarding https://github.com/keras-team/keras-cv/pull/499#issuecomment-1159536076, how far Luke's comment in https://github.com/keras-team/keras-cv/pull/499#discussion_r901056559 impacts this?

@LukeWood's comments point towards the direction of `ImageProjectiveTransformV3`. It is definitely more robust but comes at the cost of reduced readability. Also I'm not sure how will images with area_factor=2 look like, and whether that is required. 

> I'd prefer using tf.image.crop_and_resize() since the code does read cleaner and crisper with it. Conceptually also, it becomes simpler.

I agree.

keras_cv/layers/preprocessing/random_resized_crop.py
LukeWood: yeah, I don't think we want any defaults.

LukeWood:
> @sayakpaul @LukeWood
> 
> Please feel free to review this PR now.
> 
> I went through the implementation and refactored the same. I believe this implementation can be easily done using [`tf.image.crop_resize`](https://www.tensorflow.org/api_docs/python/tf/image/crop_and_resize) as mentioned in the [comment](https://github.com/keras-team/keras-cv/pull/457#issue-1251744408) in #457. I have jotted down a [gist](https://colab.research.google.com/gist/AdityaKane2001/279a7235577450af727af396b0c35c7d/rrc_using_cropandresize.ipynb) of the same.
> 
> `ImageProjectiveTransformV3` seems a bit of over the top to me, since the same result can be achieved with `tf.image.crop_resize` but with better readability. So I have replaced `ImageProjectiveTransformV3` with `tf.image.crop_resize`. Here is a [gist](https://colab.research.google.com/gist/AdityaKane2001/fb9f287e8458f97e0075bffdb9af641f/scratchpad.ipynb) using the updated implementation.
> 
> Both [`tf.image.crop_resize`](https://www.tensorflow.org/api_docs/python/tf/image/crop_and_resize) and [`tf.raw_ops.ImageProjectiveTransformV3`](https://www.tensorflow.org/api_docs/python/tf/raw_ops/ImageProjectiveTransformV3) are vectorized themselves. So batching and unbatching images and boxes (or transform) arguments is redundant. Is there some way to avoid doing that?

No there is not.  But, under the hood vectorized_map will vectorize this for us.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+                 area_factor=(0.08, 1.0),
+                 interpolation="bilinear",
+                 fill_mode="reflect",
+                 fill_value=0.0,
+                 seed=None,
+                 **kwargs):
+        super(RandomResizedCrop, self).__init__(seed=seed, **kwargs)
+
+        self.target_size = target_size
+        self.aspect_ratio_factor = aspect_ratio_factor
```
LukeWood, line 60: you can specify min and max values for factors:

https://github.com/keras-team/keras-cv/blob/master/keras_cv/utils/preprocessing.py#L97

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+        area_factor = self.area_factor()
+        aspect_ratio = tf.random.uniform((),
+                                         minval=self.aspect_ratio_factor[0],
+                                         maxval=self.aspect_ratio_factor[1],
+                                         dtype=tf.float32)
+
+        new_height = tf.clip_by_value(
+            tf.sqrt(area_factor / aspect_ratio), 0.0,
+            1.0)  # to avoid unwanted/unintuitive effects
+        new_width = tf.clip_by_value(tf.sqrt(area_factor * aspect_ratio), 0.0,
```
LukeWood, line 89: just so I understand, what would be unintuitive about making the new with bigger than 1.0?   Is there a reason we crop first then resize instead of resizing first then cropping?

You could resize the image by 0.5, then randomly crop a 10x10 section.

OR resize an image by 2.0, then crop to 10x10.

Does this explain why I think we can support area_factor > 1.0?

keras_cv/layers/preprocessing/random_resized_crop.py
LukeWood: @sayakpaul https://keras.io/guides/keras_cv/custom_image_augmentations/

LukeWood (changes requested):

keras_cv/layers/preprocessing/random_resized_crop.py
sayakpaul: Completely forgot about this! Thanks.

keras_cv/layers/serialization_test.py
LukeWood: Do you know why these errors are occurring yet?

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: I'm not sure how that would make a difference, since area_factor and aspect_ratio are sampled randomly. Since the aforementioned "10x10" is also calculated on the fly using random values, it would not matter in a distribution of images  whether we resize first or not. Please correct me if I'm wrong.

LukeWood: intuitively I would think area_factor would resize the image, then aspect ratio fact would distort the image, then the crop operation would happen last.

I guess this is not how it is done in the rest of the ML world?

keras_cv/layers/serialization_test.py
AdityaKane2001: Actually, no. I wanted to raise a separate issue regarding this. Not sure if this is due to Windows. In worst case, I'll sync my branch with master and run formatting on colab. I hope that will sort things out.

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: No I believe the convention I told is generally followed, at least that's what I'm familiar with. @sayakpaul might have a better idea of this.

keras_cv/layers/preprocessing/random_resized_crop.py
LukeWood: parse_factor supports max_value and min_value.

use these to support values > 1.

LukeWood:
Ok @AdityaKane2001 you are right that crop first is smarter.  I just did some additional thought work and they are indeed equivalent, but your way is more efficient.

carry on with the crop first API

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+            lower and upper bounds for the aspect ratio of the cropped image
+            before resizing it to `target_size`. Defaults to (3./4., 4./3.).
+        area_factor: (Optional) A tuple of two floats, a single float or
+            `keras_cv.FactorSampler`. Represents the lower and upper bounds for
+            the area relative to the original image of the cropped image before
+            resizing it to `target_size`. Defaults to (0.08, 1.0).
+        interpolation: interpolation method used in the `ImageProjectiveTransformV3` op.
+             Supported values are `"nearest"` and `"bilinear"`.
+             Defaults to `"bilinear"`.
+        fill_mode: fill_mode in the `ImageProjectiveTransformV3` op.
```
tanzhenyu, line 40: Will this render correctly, or do we need tf.raw_ops.XXX?

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class RandomResizedCrop(BaseImageAugmentationLayer):
+    """
+    Randomly crops a part of an image and resizes it to provided size.
+
+    Args:
+        target_size: A tuple of two integers used as the target size to crop
+            images to.
+        aspect_ratio_factor: (Optional) A tuple of two floats. Represents the
+            lower and upper bounds for the aspect ratio of the cropped image
```
tanzhenyu, line 31: Maybe clarify what the aspect ratio mean here?

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,140 @@
...
+            before resizing it to `target_size`. Defaults to (3./4., 4./3.).
+        area_factor: (Optional) A tuple of two floats, a single float or
+            `keras_cv.FactorSampler`. Represents the lower and upper bounds for
+            the area relative to the original image of the cropped image before
```
tanzhenyu, line 35: Same here, it'd be more readable if we let users know what an example look like for (0.08, 1.0)

keras_cv/layers/preprocessing/random_resized_crop.py
LukeWood: it won't render in keras.io properly, you do need tf.raw_ops. to get automatic linking.

keras_cv/layers/preprocessing/random_resized_crop.py
LukeWood: good point, IIUC it is width/height

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: @LukeWood do we need to have these args, since we will be using `tf.image.crop_resize` now?

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: Added some explanation to the docstring, but not sure if that's good enough. Please take a look.

AdityaKane2001: Added some explanation to the docstring, but not sure if that's good enough. Please take a look.

AdityaKane2001:
@sayakpaul @LukeWood @tanzhenyu 

I have made the changes as discussed.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    Randomly crops a part of an image and resizes it to provided size. This
+    implementation takes an intuitive approach, where we crop the images to a
+    random height and width, and then resize them. To do this, we first sample a
+    random value for area using `area_factor` and a value for aspect ratio using
+    `aspect_ratio_factor`. Further we get the new height and width by
+    dividing and multiplying the old height and width by the random area
+    respectively. We then sample offsets for height and width and clip them such
+    that the cropped area does not exceed image boundaries. Finally we do the
+    actual cropping operation and resize the image to `target_size`.
```
sayakpaul, line 33:
```suggestion
    Randomly crops a part of an image and resizes it to provided size. In
    this implementation, we crop the images to a random height and width, and
    then resize them to the given resolution. To do this, we first sample a
    random value for area using `area_factor` and a value for aspect ratio using 
    `aspect_ratio_factor`. Further, we get the new height and width by dividing
    and multiplying the old height and width by the random area respectively. We
    then sample offsets for height and width and clip them such that the cropped
    area does not exceed image boundaries. Finally, we do the actual cropping
    operation and resize the image to `target_size`.
```

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+
+        y1 = height_offset
+        y2 = height_offset + new_height
+        x1 = width_offset
+        x2 = width_offset + new_width
+
+        return [[y1, x1, y2, x2]]
+
+    def augment_image(self, image, transformation):
```
sayakpaul, line 128: @LukeWood @tanzhenyu @AdityaKane2001 I am wondering if we should support an inference behaviour of this layer?

[`RandomCrop`](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_crop/) has a separate inference-time behaviour:

> At inference time, and during training if an input image is smaller than the target size, the input will be resized and cropped so as to return the largest possible window in the image that matches the target aspect ratio. If you need to apply random cropping at inference time, set training to True when calling the layer.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,155 @@
...
+        image = tf.expand_dims(image, axis=0)
+        boxes = transformation
+
+        # See bit.ly/tf_crop_resize for more details
```
sayakpaul, line 132: Full links supported for now?

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,155 @@
...
+        augmented_image = tf.image.crop_and_resize(
+            image,  # image shape: [B, H, W, C]
+            boxes,  # boxes: (1, 4) in this case; represents area
+            # to be cropped from the original image
+            [0],  # box_indices: maps boxes to images along batch axis
+            # [0] since there is only one image
+            self.target_size,  # output size
```
sayakpaul, line 139: Great job describing these!

keras_cv/layers/preprocessing/random_resized_crop_test.py
```diff
@@ -0,0 +1,42 @@
...
+import tensorflow as tf
+
+from keras_cv.layers import preprocessing
+
+
+class RandomResizedCropTest(tf.test.TestCase):
+    def train_augments_image(self):
+        # Checks if original and augmented images are different
+
+        input_image_shape = (4, 300, 300, 3)
```
sayakpaul, line 23: Should we define the numbers in separate variables for better readability? 

```py
height, width = 300
num_channels = 3
batch_size = 4

...
```

keras_cv/layers/preprocessing/random_resized_crop_test.py
```diff
@@ -0,0 +1,42 @@
...
+        image = tf.random.uniform(shape=input_image_shape)
+
+        layer = preprocessing.RandomResizedCrop((224, 224))
```
sayakpaul, line 26: Same for this?

```py
resize_to = (224, 224)

...
```

keras_cv/layers/preprocessing/random_resized_crop_test.py
```diff
@@ -0,0 +1,42 @@
...
+        output = layer(image, training=True)
+
+        output_image_resized = tf.image.resize(output, input_image_shape)
+
+        self.assertNotEqual(image, output_image_resized)
+
+    def test_preserves_image(self):
```
sayakpaul, line 33: Should we also add a test for grayscale images (channels = 1)?

sayakpaul (changes requested):
Looking good @AdityaKane2001! Just some minor comments and questions.

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: Since we're subclassing `BaseImageAugmentationLayer`, I believe at test time the image is left unchanged.

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: Not sure about comments, so kept it as it is.

keras_cv/layers/preprocessing/random_resized_crop.py
sayakpaul: Ah nice. Should we add a note in the docstrings then so that the readers are aware of this behaviour?

keras_cv/layers/preprocessing/random_resized_crop.py
sayakpaul: @LukeWood

keras_cv/layers/preprocessing/random_resized_crop_test.py
AdityaKane2001: Added a `train_` test for the same.

keras_cv/layers/preprocessing/random_resized_crop.py
LukeWood: good point, comments may not support full links yet.  I'll see if we can disable the lint check for this.

keras_cv/layers/preprocessing/random_resized_crop_test.py
```diff
@@ -0,0 +1,44 @@
...
+        layer = preprocessing.RandomResizedCrop(
+            target_size=(224, 224), area_factor=(0.8, 1.0)
+        )
+        output = layer(image, training=True)
+
+        output_image_resized = tf.image.resize(output, input_image_shape)
+
+        self.assertNotEqual(image, output_image_resized)
+
+    def test_preserves_image(self):
```
LukeWood, line 35: add a grayscale example as @sayakpaul suggested, then I think I feel pretty good about this PR!  

Perhaps a seeded random behavior test would be nice too - something like sample random noise as an image batch: (2, 512, 512, 3), crop randomly (200, 200), then assert that they are not equal.  With proper seeding this should work fine.

LukeWood (changes requested):

AdityaKane2001:
@LukeWood 

I feel we should keep the relevant defaults, as the optimum values for these arguments are not obvious to someone who is not familiar with this. @sayakpaul WDYT?

sayakpaul:
> @LukeWood
> 
> I feel we should keep the relevant defaults, as the optimum values for these arguments are not obvious to someone who is not familiar with this. @sayakpaul WDYT?

Agreed.

LukeWood:
> @LukeWood
> 
> I feel we should keep the relevant defaults, as the optimum values for these arguments are not obvious to someone who is not familiar with this. @sayakpaul WDYT?
> 
> Also @LukeWood thanks for the correction in `call`, I missed that earlier.

I see, are these values widely optimal across almost all use cases?

sayakpaul:
> I see, are these values widely optimal across almost all use cases?

We could say that, yes. At least most of the pre-training schemes (supervised, self-supervised, semi-supervised, vision-text, etc.) I have studied these are the default values.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,184 @@
...
+            The area represents the percentage of the original image that the crop taken
+            will be. Represents the lower and upper bounds for the area relative to the
+            original image of the cropped image before resizing it to `target_size`.
+        aspect_ratio_factor: (Optional) A tuple of two floats, a single float or
+            `keras_cv.FactorSampler`. Aspect ratio means the ratio of width to
+            height of the cropped image. In the context of this layer, the aspect ratio
+            sampled represents a value to distort the aspect ratio by.
+            Represents the lower and upper bound
+            for the aspect ratio of the cropped image before resizing it to
+        aspect_ratio_factor: (Optional) A tuple of two floats. Represents the
```
tanzhenyu, line 51: Isn't this a duplicate?

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,184 @@
...
+            lower and upper bounds for the aspect ratio of the cropped image
+            before resizing it to `target_size`. Defaults to (3./4., 4./3.).
+        area_factor: (Optional) A tuple of two floats, a single float or
+            `keras_cv.FactorSampler`. The ratio of area of the cropped part to
+            that of original image is sampled using this factor. Represents the
+            lower and upper bounds for the area relative to the original image
+            of the cropped image before resizing it to `target_size`.
+            `target_size`.  Defaults to (1.0, 1.0), or a no-op.
+        interpolation: (Optional) A string specifying the sampling method for
+            resizing.
```
tanzhenyu, line 61: "Default to `bilinear`"

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,184 @@
...
+        seed: (Optional) Integer. Used to create a random seed.
```
tanzhenyu, line 62: Let's always provide a default value in the docstring so users can easily find it?

AdityaKane2001: Defaults to None, so adding that.

keras_cv/layers/preprocessing/random_resized_crop_test.py
AdityaKane2001: I guess `test_train_augments_image` already does that.

AdityaKane2001:
@sayakpaul @LukeWood @tanzhenyu 

I think I have made all the requested changes. However, if I have left out any of them by mistake, please let me know. If everything's alright, I'll run the formatter so that this can be merged.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,184 @@
...
+
+    Args:
+        target_size: A tuple of two integers used as the target size to ultimately crop
+            images to.
+        area_factor: (Optional) A tuple of two floats, a single float or
+            `keras_cv.FactorSampler`. The ratio of area of the cropped part to
+            that of original image is sampled using this factor. Represents the
+            lower and upper bounds for the area relative to the original image
+            of the cropped image before resizing it to `target_size`.
+            `target_size`.  Defaults to (0.08, 1.0).
```
LukeWood, line 45: The more I think the more I think we don't want a default for this.

Also, lets rename this to `crop_area_factor`.  We can recommendd 0.08 and 1.0, but a crop area of 0.08 is actually really aggressive for classification and might not be a sensible default there as it is for self supervised learning.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,184 @@
...
+        aspect_ratio_factor: (Optional) A tuple of two floats, a single float or
```
LukeWood, line 46: I also don't think we can take a default here because this will be surprising to users.  That being said, a user SHOULD use this, so I suggest we provide no default and require this argument.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,184 @@
...
+            area_factor,
+            max_value=1.0,
+            param_name="area_factor",
+            seed=seed,
+        )
+
+        self.interpolation = interpolation
+        self.seed = seed
+
+        if area_factor == 0.0 and aspect_ratio_factor == 0.0:
```
LukeWood, line 99: I think that an area factor of 0.0 is actually nonsense, right?  A no-op should be 1.0, and a no-op for aspect ratio is also 1.0.

LukeWood (changes requested):
Minor changes

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: @LukeWood In that case do we want to have a less aggressive number there? I had used 0.25 for training the smaller RegNets (200 MFLOP variant) and it seemed to work fine.

AdityaKane2001: At the time I had experimented with other values as well and found 0.25 to be good for that case.

LukeWood: I don't think we want a default for this because I don't think there will be a generic best value for this.

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: @LukeWood do you mean to say for `aspect_ratio_factor`?

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: Got it, makes sense. 

@sayakpaul WDYT?

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: @LukeWood In the case of a single float or integer, what should be the max (or min) value for `aspect_ratio_factor`? Currently we are doing this:
```py
if isinstance(aspect_ratio_factor, tuple):
  min_aspect_ratio = min(aspect_ratio_factor)
  max_aspect_ratio = max(aspect_ratio_factor)
elif isinstance(aspect_ratio_factor, core.FactorSampler):
  pass
else:
  raise ValueError(
      "Expected `aspect_ratio` to be tuple or FactorSampler. Received "
      f"aspect_ratio_factor={aspect_ratio_factor}."
  )
```

If we want to add a case for single ints and floats, we'll need to decide on a predefined max or min.

LukeWood: lets just drop integer and float support for aspect ratio factor and make them pass a tuple.

AdityaKane2001: Awesome, thanks for clarifying!

AdityaKane2001:
@sayakpaul @LukeWood 

I have pushed the discussed changes. Could you please take a look at the checks in `__init__`? I'm not sure if the ones I have implemented are good enough.

keras_cv/layers/preprocessing/random_resized_crop.py
sayakpaul: @LukeWood has an interesting PoV here. I honestly haven't ever thought of default values at this depth. Luke seems to have a point. 

But let's please do one thing -- write a comprehensive guide around this layer when it's out detailing the thought process and suggest some recommended values for different tasks. The reason why I'm stressing about this is that RRC is one of the most influential layers in the domain so we'd want to ensure our users have a good guide on using it. 

WDYT? @AdityaKane2001 @LukeWood

sayakpaul:
@LukeWood when you get a moment, could you please assign the "GSOC 2022" label?

keras_cv/layers/preprocessing/random_resized_crop.py
AdityaKane2001: Sounds good. 

I think this guide can be based upon seminal literature in vision. That will give a good idea about good defaults IMO.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,244 @@
...
+        else:
+            raise ValueError(
+                "Expected `aspect_ratio` to be tuple or FactorSampler. Received "
+                f"aspect_ratio_factor={aspect_ratio_factor}."
+            )
+
+        self.target_size = target_size
+        self.aspect_ratio_factor = preprocessing.parse_factor(
+            aspect_ratio_factor,
+            max_value=max_aspect_ratio,
```
LukeWood, line 87: max_value should be a constant that is used for validation.  Check out how it is used inside of parse_factor.

The goal of max_value is to prevent users from passing values outside of the value range.

LukeWood: In this case, please use min_value=0, max_value=None.

Reference: https://github.com/keras-team/keras-cv/blob/master/keras_cv/utils/preprocessing.py#L109

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,244 @@
...
+    def _check_class_arguments(
+        self, target_size, crop_area_factor, aspect_ratio_factor
+    ):
+        if not isinstance(target_size, tuple):
+            raise ValueError(
+                "`target_size` must be tuple of two integers."
+                f"Received target_size={target_size}"
+            )
+
+        if len(target_size) != 2:
```
LukeWood, line 186: error messages match, add this as an and clause to the check above.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,244 @@
...
+            raise ValueError(
+                "`target_size` must be tuple of two integers."
+                f"Received target_size={target_size}"
+            )
+
+        if not isinstance(target_size[0], int):
```
LukeWood, line 192: same for these two.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,244 @@
...
+                f"Received target_size={target_size}"
+            )
+
+        if not isinstance(target_size[1], int):
+            raise ValueError(
+                "`target_size` must be tuple of two integers."
+                f"Received target_size={target_size}"
+            )
+
+        if isinstance(aspect_ratio_factor, core.NormalFactorSampler):
```
LukeWood, line 204: Using a NormalFactorSampler here is fine - you can remove these checks.  

Please add checks for `aspect_ratio_factor` and `crop_area_factor` being single integers.  Single values are not valid configurations.

check that it is either:
- a factor sampler
- a tuple of two values

LukeWood: Oh!  I missed that, let me go review your get_config

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,244 @@
...
+            )
+
+        if isinstance(crop_area_factor, core.NormalFactorSampler):
+            raise ValueError(
+                "`crop_area_factor` is an instance of NormalFactorSampler."
+                "Expected `crop_area_factor` to be a tuple, ConstantFactorSampler or UniformFactorSampler."
+            )
+
+    def get_config(self):
+        if isinstance(self.crop_area_factor, core.ConstantFactorSampler):
```
LukeWood, line 217: you can drop all of these checks.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,244 @@
...
+            aspect_upper = self.aspect_ratio_factor.value
+        else:
+            aspect_lower = self.aspect_ratio_factor.lower
+            aspect_upper = self.aspect_ratio_factor.upper
+
+        config = super().get_config()
+        config.update(
+            {
+                "target_size": self.target_size,
+                "crop_area_factor": (area_lower, area_upper),
```
LukeWood, line 235: you can literally pass self.crop_area_factor and self.aspect_area_factor in place of these.

Factors are already serializable.

LukeWood (changes requested):

examples/layers/preprocessing/random_resized_crop_demo.py
```diff
@@ -0,0 +1,33 @@
...
+Finally, they are shown using matplotlib.
+"""
+
+from examples.layers.preprocessing import demo_utils
+from keras_cv.layers.preprocessing import RandomResizedCrop
+
+
+def main():
+    many_elephants = demo_utils.load_elephant_tensor(output_size=(300, 300))
+    layer = RandomResizedCrop((224, 224), (0.08, 1.0), (3.0 / 4.0, 4.0 / 3.0))
```
LukeWood, line 27: always use named arguments for non-singular arguments that take the same format:

```
target_size=(224, 224), area_crop_factor=(0.08, 1.0), aspect_ratio_factor=(3/4, 4/3)
```

fixed this for you, but just wanted to let you know this is a better python style in general - not just in Keras.


@AdityaKane2001.

AdityaKane2001:
@LukeWood I'll add the checks in a separate PR then.

PR #480, opened 2022-06-10
Drop Path

AdityaKane2001:
@sayakpaul @LukeWood 

Implemented DropPath layer. I have used [timm](https://github.com/rwightman/pytorch-image-models/blob/7c67d6aca992f039eece0af5f7c29a43d48c00e4/timm/models/layers/drop.py#L135-L152) and [tensorflow/tpu](https://github.com/tensorflow/tpu/blob/b24729de804fdb751b06467d3dce0637fa652060/models/official/amoeba_net/network_utils.py#L164-L174) as source of truth and included timm in references as well. 

The tests were inspired by DropBlock2D's tests since both of them have (to some extent) similar functionallity.

I believe the docstring can be improved but I'm not really sure what to write. Open to suggestions.

Edit: Incorporated changes to `StochasticDepth` as discussed. Kept the argument names unchanged for the moment.

/auto Closes #130

PR #505, opened 2022-06-20
Squeeze and Excite block

### LINKED ISSUE #167: Squeeze and Excitement Block
https://arxiv.org/abs/1709.01507

AdityaKane2001:
Added SE Block. As discussed, this one can be done after #480 and #499 are merged.

/auto Closes #167


keras_cv/layers/regularization/squeeze_excite.py
```diff
@@ -0,0 +1,81 @@
...
+
+        if filters <= 0 or not isinstance(filters, int):
+            raise ValueError(f"`filters` should be a positive integer. Got "
+                             f" {filters}")
+
+        self.ratio = ratio
+        self.se_filters = int(self.filters * self.ratio)
+
+        self.ga_pool = layers.GlobalAveragePooling2D()
+        self.squeeze_reshape = layers.Reshape((1, 1, self.filters))
```
innat, line 59: Let's do as follows to get rid of reshape layer.

```python
self.ga_pool = layers.GlobalAveragePooling2D(keepdims=True)
```

AdityaKane2001:
@sayakpaul  @LukeWood

Marking this as ready to review as #480 is now merged.

sayakpaul:
@LukeWood this needs to have the GSoC2022 label.

keras_cv/layers/regularization/squeeze_excite_test.py
```diff
@@ -0,0 +1,39 @@
...
+        with self.assertRaisesRegex(
+                ValueError, "`ratio` should be a float"
+                "between 0 and 1. Got (.*?)"):
+            layer = SqueezeAndExciteBlock2D(8, ratio=1)
+
+    def raises_invalid_ratio_error(self):
+        with self.assertRaisesRegex(
+                ValueError, "`filters` should be a positive"
+                " integer. Got (.*?)"):
+            layer = SqueezeAndExciteBlock2D(-8.7)
```
sayakpaul, line 39: Should we also add a test to check if the number of input and output filters are the same?

sayakpaul (changes requested):

keras_cv/layers/regularization/squeeze_excite_test.py
AdityaKane2001: I think `maintains_shape` takes care of that.

keras_cv/layers/regularization/squeeze_excite.py
```diff
@@ -0,0 +1,81 @@
...
+		filters: Number of input and output filters. The number of input and
+			output filters is same.
+		ratio: Ratio for bottleneck filters. Number of bottleneck filters = 
+			filters * ratio. Defaults to 0.25.
+	Usage:
+
+	```python
+	# (...)
+	input = tf.ones((1, 5, 5, 16), dtype=tf.float32)
+	x = tf.keras.layers.Conv2D(16, (3,3))(input)
```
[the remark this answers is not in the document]
AdityaKane2001: Added space between (3,3) manually.

AdityaKane2001:
@sayakpaul @LukeWood 

I have made the changes, please review again.

sayakpaul (approved):
LGTM given the linting tests pass!

LukeWood (changes requested):
Minor changes, thanks!

keras_cv/layers/regularization/squeeze_excite.py
```diff
@@ -0,0 +1,73 @@
...
+        self.ratio = ratio
+        self.bottleneck_filters = int(self.filters * self.ratio)
+
+        self.global_average_pool = layers.GlobalAveragePooling2D(keepdims=True)
+        self.squeeze_conv = layers.Conv2D(
+            self.bottleneck_filters,
+            (1, 1),
+            activation="relu",
+        )
+        self.excite_conv = layers.Conv2D(self.filters, (1, 1), activation="sigmoid")
```
LukeWood, line 61: Do we need to support `squeeze_activation`, `excite_activation` parameters?

AdityaKane2001: image
This implementation is based on the original paper (https://arxiv.org/pdf/1709.01507.pdf), so I think ReLU and Sigmoid are more or less mandatory. But would like to hear your and @sayakpaul's thoughts on the same.

LukeWood: MobileNetv3 uses SE blocks with HardSigmoid - we should support this level of parameterization whether or not the original paper used that or not.  

Keep in mind ML research moves fast, and while research is great to reference it should not be treated as a hard source of truth.  Another examples of this is that ML research has a tendency to give things strange names that stick for a year or two, then are left behind in favor of a more idiomatic name.  This has happened numerous times, and as such we pick names based on idiomaticness rather than their name in the research - excluding some niche cases like RandAugment.

AdityaKane2001: Makes sense. Thanks for the detailed explanation. I'll make the changes and re-request a review.

sayakpaul: `squeeze_activation` is a better naming than `excite_activation` IMO.

AdityaKane2001: @sayakpaul I agree but I guess we need to have both of those (one for squeeze conv and other for excite conv).

sayakpaul: > However, in that case, in the get_config method squeeze_activation or excite_activation will be keras Layers, which won't be serializable. 

Why won't they be serializable? 

I think we could support strings that can be directly passed to `activation` or we could allow users to pass functions rather than layers.

sayakpaul: Definitely. For one moment, I got confused with `tf.keras.initializers.*`. We can pass a class under it (such as `VarianceScaling`) to arguments like `kernel_initializer`. But this is not a layer like you mentioned.

LukeWood: > > I think we could support strings that can be directly passed to activation or we could allow users to pass functions rather than layers.
> 
> Sure, since the functional versions of activations are a strict superset (I believe) of the layer versions of the same, it would be favorable.
> 
> Just as a side note, it is interesting to see if one can get layers to be serializable whose arguments are layer objects. I tried it just now but it isn't as simple as it seems 😅

Can you paste a snippet of the issue here?  I think layers to the constructor should be serializable - it should call their get_config() under the hood if I remember correctly.

AdityaKane2001: Sure. 

Code which raises error:
```py
    def get_config(self):
        config = {
            "filters": self.filters,
            "ratio": self.ratio,
            "squeeze_activation": self.squeeze_activation, # instance of, say, tf.keras.activations.ReLU()
            "excite_activation": self.excite_activation,
        }
        base_config = super().get_config()
        return dict(list(base_config.items()) + list(config.items()))
```
error log: shared on https://codeshare.io/3AMwy6


My workaround:
```py
def get_config(self):
        squeeze_activation_serialized = (
            serialize(self.squeeze_activation)
            if hasattr(self.squeeze_activation, "__call__")
            else self.squeeze_activation
        )
        excite_activation_serialized = (
            serialize(self.excite_activation)
            if hasattr(self.squeeze_activation, "__call__")
            else self.excite_activation
        )

        config = {
            "filters": self.filters,
            "ratio": self.ratio,
            "squeeze_activation": squeeze_activation_serialized,
            "excite_activation": excite_activation_serialized,
        }
        base_config = super().get_config()
        return dict(list(base_config.items()) + list(config.items()))

```

keras_cv/layers/regularization/squeeze_excite_test.py
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv.layers import SqueezeAndExciteBlock2D
+
+
+class SqueezeAndExciteBlock2DTest(tf.test.TestCase):
```
LukeWood, line 20: please add a test case covering custom activations: https://github.com/keras-team/keras/issues/4076#issue-183230193

LukeWood (changes requested):
one comment then LGTM

keras_cv/layers/regularization/squeeze_excite_test.py
AdityaKane2001: @LukeWood 
Added test. 
Please see #544 before this one. I have changed the docstring assuming that one is merged before this one.

keras_cv/layers/regularization/squeeze_excite.py
```diff
@@ -25,12 +24,14 @@ class SqueezeAndExciteBlock2D(layers.Layer):
...
+            output filters is same.
         ratio: Ratio for bottleneck filters. Number of bottleneck filters =
-                filters * ratio. Defaults to 0.25.
-        squeeze_activation: (Optional) String or function denoting activation to
+            filters * ratio. Defaults to 0.25.
+        squeeze_activation: (Optional) String, function or tf.keras.activations.*
+            instance denoting activation to
             be applied after squeeze convolution. Defaults to `relu`.
-        excite_activation: (Optional) String or function denoting activation to
+        excite_activation: (Optional)String, function or tf.keras.activations.*
```
LukeWood, line 33: replace function with "callable" (or tf.keras.Layer), replace tf.keras.activations.* with "tf.keras.activations.Activation" -> we have a base class for these.

Also you forgot a space after optional :)

AdityaKane2001: @LukeWood even with the change in serialization_test, it errors out for activation functions.

AdityaKane2001: This one is a bit weird. Here's the test case:
```py
        (
            "SqueezeAndExciteBlock2D",
            regularization.SqueezeAndExciteBlock2D,
            {
                "filters": 16,
                "ratio": 0.25,
                "squeeze_activation": tf.keras.layers.ReLU(),
                "excite_activation": tf.keras.activations.relu,
            },
        ),

```
Error trace: https://codeshare.io/3AMwqv

I have inserted the print statements as follows:
```py
def custom_compare(obj1, obj2):
    print("#" * 20, "custom_compare")

    if isinstance(obj1, (core.FactorSampler, tf.keras.layers.Layer)):
        print(obj1.get_config())
        print(obj2.get_config())
        return config_equals(obj1.get_config(), obj2.get_config())
    else:
        return obj1 == obj2


def config_equals(config1, config2):
    print("#"*20, "config_equals")
    print(config1)
    print(config2)
    for key in list(config1.keys()) + list(config2.keys()):
        v1, v2 = config1[key], config2[key]
        if not custom_compare(v1, v2):
            return False
    return True

```

AdityaKane2001: Note that the function is getting "auto-serialized":

```
original: {'name': 'squeeze_and_excite_block2d', 'trainable': True, 'dtype': 'float32', 'filters': 16, 'ratio': 0.25, 'squeeze_activation': <keras.layers.activation.relu.ReLU object at 0x0000025EC9F39208>, 'excite_activation': <function relu at 0x0000025EF179B318>}
restored: {'name': 'squeeze_and_excite_block2d', 'trainable': True, 'dtype': 'float32', 'filters': 16, 'ratio': 0.25, 'squeeze_activation': <keras.layers.activation.relu.ReLU object at 0x0000025EF1D86088>, 'excite_activation': 'relu'}
```

LukeWood: The paste is empty.

LukeWood: This code is still working: it is serializing and deserializing a relu activation.  the code should not be changed to work around the test case.  The test case should be changed to work around the code 👍

AdityaKane2001: Could you try this one: https://www.codepile.net/pile/nP5Nnyl4

AdityaKane2001:
@LukeWood I have a workaround for the failing tests:
```py
def custom_compare(obj1, obj2):
    if isinstance(obj1, (core.FactorSampler, tf.keras.layers.Layer)):
        return config_equals(obj1.get_config(), obj2.get_config())
    elif inspect.isfunction(obj1):
        return tf.keras.activations.serialize(obj1) == obj2
    elif inspect.isfunction(obj2):
        return obj1 == tf.keras.activations.serialize(obj2)
    else:
        return obj1 == obj2

```

LukeWood:
> @LukeWood I have a workaround for the failing tests:
> 
> ```python
> def custom_compare(obj1, obj2):
>     if isinstance(obj1, (core.FactorSampler, tf.keras.layers.Layer)):
>         return config_equals(obj1.get_config(), obj2.get_config())
>     elif inspect.isfunction(obj1):
>         return tf.keras.activations.serialize(obj1) == obj2
>     elif inspect.isfunction(obj2):
>         return obj1 == tf.keras.activations.serialize(obj2)
>     else:
>         return obj1 == obj2
> ```
> 
> The tests were failing when the passed object is a function. So, we check if it is a function and serialize it if it is. I'll replace `activations.serialize` by `keras.utils.serialize_keras_object`.

Awesome!  Thanks for investigating this.  Let’s update so the tests pass!

AdityaKane2001:
@LukeWood Done.

AdityaKane2001:
@sayakpaul @LukeWood 

Even though this is a valid workaround at the moment, it would be interesting to see if there is any concrete solution to serializing and deserializing of _functions_. WDYT? Or is there some existing solution to this?

AdityaKane2001:
@sayakpaul  @LukeWood 

Could you please take a look at this one when you get a moment? TIA

AdityaKane2001:
@sayakpaul @LukeWood

Could you please take a look at this one? Please let me know if there's anything you'd like me to change.

LukeWood (changes requested):

LukeWood:
Thank you for the PR Aditya, merged

PR #510, opened 2022-06-22
added yxyx  and rel_yxyx bounding box format support in bounding box utils

divyashreepathihalli:


keras_cv/bounding_box/converters_test.py
```diff
@@ -20,9 +20,13 @@
 from keras_cv import bounding_box
 
 xyxy_box = tf.constant([[10, 10, 110, 110], [20, 20, 120, 120]], dtype=tf.float32)
+yxyx_box = tf.constant([[10, 10, 110, 110], [20, 20, 120, 120]], dtype=tf.float32)
```
LukeWood, line 23: now that we have a distinction between the first X and y why don't we update one of the 10s to a 15, and one of the 20s to a 25, and one of the 120s to a 122.

That way we can have coverage of these swaps.

LukeWood (changes requested):
One comment on test coverage, other than that LGTM

Thank you for the PR Divya!

LukeWood (approved):
Thanks Divya!  Ready to merge when you are

divyashreepathihalli:
> 

Thanks Luke, but I don't have the permission to merge.

PR #513, opened 2022-06-22
Adding RetinaNet FPN

quantumalaviya:
This is a follow-up PR to #504. This implementation uses the FPN layer from #504 to implement the modified FPN used in RetinaNet.


quantumalaviya:
@LukeWood This is the RetinaNet FPN PR we discussed offline. There is some repetition of code from #504. Have a look when you get a chance!

quantumalaviya:
Alternatively, this layer can have boolean arguments representing whether levels 2, 6, and 7 are to be considered (instead of passing pyramid levels in a list).

PR #516, opened 2022-06-23
Remove "tf.keras.__internal__.layers" from all files

AdityaKane2001:
@LukeWood 

Removed `__internal__` layer subclassing from all files as well as the contributing guide.

### DIFF
--- a/keras_cv/__init__.py
+++ b/keras_cv/__init__.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from keras_cv import layers
 from keras_cv import metrics
 from keras_cv import models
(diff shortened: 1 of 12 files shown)


keras_cv/__init__.py
```diff
@@ -21,6 +20,9 @@
 from keras_cv.core import FactorSampler
 from keras_cv.core import NormalFactorSampler
 from keras_cv.core import UniformFactorSampler
+from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
```
LukeWood, line 23: we probably don't need this in keras_cv/__init__.py

LukeWood (changes requested):

Issue #496, opened 2022-06-16
Add MLP Mixer classes to the `models` module


sayakpaul:
MLP Mixers [1] are a cool architecture that does not involve convs, attention, or transformers -- just `Dense` layers 🤯! I have worked with MLP Mixer before (see [here](https://github.com/sayakpaul/MLP-Mixer-CIFAR10) and [here](https://github.com/sayakpaul/MLPMixer-jax2tf)). Therefore, would love to contribute them to `keras_cv.models`. 

## References

[1] MLP-Mixer: An all-MLP Architecture for Vision: https://arxiv.org/abs/2105.01601

LukeWood:
MLPMixer is a very unique architecture!  I think it would be great to offer this from keras_cv.models

sayakpaul:
> i wonder if it is worth offering the patch extractor as an API, or if it is too specific to MLPMixer.


In my experience of working with similar architectures (i.e., the ones that don't reply on convs to process images and build representations) usually benefit from the patch extractor module. For the first iteration, I will include the patch extractor in the model itself. But can do the changes necessary to offer it as a separate API.

LukeWood:
sounds good, as long as we write it as an independent layer it shouldn’t be hard to migrate it to the layers API should it turn out to be useful to do so.

bhack:
We could also expore if it is useful to access to surveys to have a good overview of the specific domain on reusable components/API design:

https://arxiv.org/abs/2111.04060

https://arxiv.org/abs/2112.04453

bhack:
We could also expore if it is useful to access to surveys to have a good overview of the specific domain on reusable components/API design:

https://arxiv.org/abs/2111.04060

https://arxiv.org/abs/2112.04453

![IMG_20220616_163457.jpg](https://user-images.githubusercontent.com/1710528/174094079-829ce418-f8b7-44d4-8530-73bdc35490ad.jpg)

PR #529, opened 2022-06-26
Add MLP Mixer models

sayakpaul:
# What does this PR do?

Closes https://github.com/keras-team/keras-cv/issues/496

**TODOs**:

- [x] Add model script
- [x] Add tests
- [x] Add / modify rest of the files

**Some things to note regarding models like MLP Mixer**:

* Unlike CNNs, their feature maps are of shape: `(batch_size, sequence_length, hidden_dimension)`. 
* MLP Mixers require the input resolution to be divisible by the patch size. 
* The model architecture includes a reshape operation after patchification + projection. This is why we cannot specify `None` (`(None, None, 3)` for example) input shapes whilst initializing the model. 
* The model script is such that even if we decided to port [the original (and official) pre-trained parameters](https://github.com/google-research/vision_transformer) to TF / Keras, they would be easily compatible. 
 
@LukeWood


keras_cv/models/mlp_mixer.py
```diff
@@ -0,0 +1,279 @@
...
+    patch_size,
+    num_blocks,
+    hidden_dim,
+    tokens_mlp_dim,
+    channels_mlp_dim,
+    include_rescaling,
+    include_top,
+    classes=None,
+    weights=None,
+    input_shape=(224, 224, 3),
```
LukeWood, line 90: Hey Sayak, Is there a reason for (224, 224, 3) other than it being the imagenet size?

keras_cv/models/mlp_mixer.py
```diff
@@ -0,0 +1,279 @@
...
+    pooling=None,
+    classifier_activation="softmax",
+    name=None,
+    **kwargs,
+):
+    """Instantiates the MLP Mixer architecture.
+
+    Reference:
+    - [MLP-Mixer: An all-MLP Architecture for Vision (NeurIPS 2021)](
```
LukeWood, line 99: links can go on one line now.

keras_cv/models/mlp_mixer.py
```diff
@@ -0,0 +1,279 @@
...
+        )
+
+    elif pooling == "avg":
+        x = layers.GlobalAveragePooling1D(name="avg_pool")(x)
+    elif pooling == "max":
+        x = layers.GlobalMaxPooling1D(name="max_pool")(x)
+
+    model = keras.Model(inputs, x, name=name, **kwargs)
+
+    if weights is not None:
```
LukeWood, line 196: You mention in this Pr we could port the original weights - do they use the same rescaling scheme we do here?  If so this would be a wonderful correctness check!

LukeWood (changes requested):
Minor comments, mainly around input size.

All around great implementation @sayakpaul!

keras_cv/models/mlp_mixer.py
sayakpaul: Yes, it's a sensible default that's why. Could have been (384, 384) too. But we need non `None` spatial resolutions here as there's a `Reshape()` operation dependent on it.

keras_cv/models/mlp_mixer.py
sayakpaul: Fixed.

sayakpaul:
@LukeWood I don't understand why the test would fail. I just checked locally with `pytest keras_cv/models/models_test.py` and it was all green.

I didn't change a line after https://github.com/keras-team/keras-cv/pull/529/commits/6cd3bfda4073e72abea31a9d2808509a5a4d663d. I will just do an empty commit to see if it has any effect.

keras_cv/models/mlp_mixer.py
LukeWood: Gotcha, I wonder if it makes sense to have this be a required argument instead of providing a default.

LukeWood: I suppose it depends on how we anticipate the architecture to be used!

sayakpaul: I provided a sensible default. Having it as a required argument won't hurt anything I guess.

sayakpaul: > I suppose it depends on how we anticipate the architecture to be used!

I won't expect this architecture to do good at tasks other than image classification. All the other tasks such as segmentation, detection benefit from hierarchical features but MLP-Mixer is an isotropic architecture meaning it maintains a single resolution of the feature maps throughout its existence.

LukeWood: lets make sure to discuss how to compute some alternative values for valid configurations in the docstrings.

LukeWood (approved):
I think this looks good, I wonder, can we have basic training script in examples just as a sanity check that it can run with fit?  It doesn't have to be anything special for the time being and we can upgrade it when we begin creating our weight reproduction scripts.

sayakpaul:
> I think this looks good, I wonder, can we have basic training script in examples just as a sanity check that it can run with fit?

@LukeWood where do you want me to keep it? Also, does something like the following work (haven't tested, might contain errors)? 


```py
from keras_cv.models import MLPMixerB16
from tensorflow import keras
import tensorflow as tf 

# constants
batch_size = 10
height = width = 224
num_channels = 3

# model
mixer = MLPMixerB16(include_rescaling=True, include_top=True, classes=10)
mixer.compile(optimizer="adam", loss="sparse_categorical_crossentropy", metrics=["accuracy"])

# data
random_images = tf.random.normal((batch_size, height, width, num_channels))
random_labels = tf.random.uniform((batch_size, ), 0, 5, dtype="int32")

# train
mixer.fit(
   random_images, random_labels, 
   validation_data=(random_images, random_labels), 
   epochs=3
)
```

sayakpaul:
@LukeWood ready for another review.

LukeWood:
> > > I think this looks good, I wonder, can we have basic training script in examples just as a sanity check that it can run with fit?
> > 
> > 
> > @LukeWood where do you want me to keep it? Also, does something like the following work (haven't tested, might contain errors)?
> > ```python
> > from keras_cv.models import MLPMixerB16
> > from tensorflow import keras
> > import tensorflow as tf 
> > 
> > # constants
> > batch_size = 10
> > height = width = 224
> > num_channels = 3
> > 
> > # model
> > mixer = MLPMixerB16(include_rescaling=True, include_top=True, classes=10)
> > mixer.compile(optimizer="adam", loss="sparse_categorical_crossentropy", metrics=["accuracy"])
> > 
> > # data
> > random_images = tf.random.normal((batch_size, height, width, num_channels))
> > random_labels = tf.random.uniform((batch_size, ), 0, 5, dtype="int32")
> > 
> > # train
> > mixer.fit(
> >    random_images, random_labels, 
> >    validation_data=(random_images, random_labels), 
> >    epochs=3
> > )
> > ```
> 
> Yeah that looks good to me.

perhaps just examples/models/basic/mlp_mixer.py or something?  we might move/remove it later once we have training scripts that produce nice weights, but for now this is just a sanity check.  thank you!

sayakpaul:
@LukeWood maybe in a separate PR? Would be an easier review I believe.

PR #537, opened 2022-06-27
Add an augment_label method to the RandomGaussianBlur augmentation layer

### LINKED ISSUE #538: Add `with_labels_test.py` to the augmentation directory
This should be a parameterized absl test similar to `layers/serialization_test.py`.

It can live in the directory: https://github.com/keras-team/keras-cv/tree/master/keras_cv/layers/preprocessing

You'll basically want to test that all the layers that should run with labels can run with labels.

ianjjohnson:
# What does this PR do?

Makes RandomGaussianBlur and RandomJpeg override the augment_label method. Without overriding this method, I found that these layers were throwing a NotImplementedError when applied to a tensor that included labels.

Fixes #538 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood

### DIFF
--- a/keras_cv/layers/preprocessing/random_gaussian_blur.py
+++ b/keras_cv/layers/preprocessing/random_gaussian_blur.py
@@ -83,6 +83,9 @@ class RandomGaussianBlur(BaseImageAugmentationLayer):
 
         return tf.squeeze(blurred, axis=0)
 
+    def augment_label(self, label, transformation=None, **kwargs):
+        return label
+
     @staticmethod
     def get_kernel(factor, filter_size):
         x = tf.cast(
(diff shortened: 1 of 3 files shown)


ianjjohnson:
I have since realized that there are other layers that don't implement `augment_label`, and the docstrings in BaseImageAugmentationLayer suggest that an implementation is required. 

So perhaps this is WAI, but is the expected behavior of these layers that they will fail to augment data if it has labels?
Either way, _some_ layers seem to overwrite `augment_label` with a no-op, and some don't. Should we make this consistent?

LukeWood:
Thanks @ianjjohnson for the catch!

PR #539, opened 2022-06-27
Fix DenseNet Docstring

### LINKED ISSUE #528: DenseNet docstring should be fixed
There are some minor inconsistencies in DenseNet compared to the other layers and keras applications:

Firstly, in the argument description for `include_top`, it says,
```
include_top: whether to include the fully-connected
layer at the top of the network.  If provided, num_classes must be provided.
```
However, for keras_cv there is no `num_classes` attribute. We use `classes`. Either the `classes` attribute name or this description needs to be changed. Personally, I feel that `num_classes` is more descriptive.

Secondly, the DenseNet docstring is copied to all versions of the DenseNet architecture. This may be confusing in some cases. For example, the `name` is supposed to default to "DenseNet" according to the docstring. However, in practice, it changes for every DenseNet version. We can opt to have a base docstring that is edited in each case (see the [EfficientNet implementation from Keras](https://github.com/keras-team/keras/blob/07e13740fd181fc3ddec7d9a594d8a08666645f6/keras/applications/efficientnet.py#L145)).

Lastly, the DenseNet docstring can be edited to follow the same formatting as followed across the other keras_cv layers

quantumalaviya:
# What does this PR do?

Fixes #528 

## Before submitting
- [x] Was this discussed/approved via a Github issue? #528

## Who can review?
@LukeWood

### DIFF
--- a/keras_cv/models/densenet.py
+++ b/keras_cv/models/densenet.py
@@ -15,10 +15,8 @@
 """DenseNet models for KerasCV.
 
 Reference:
-  - [Densely Connected Convolutional Networks]
-    (https://arxiv.org/abs/1608.06993)
-  - [Based on the Original keras.applications DenseNet]
-    (https://github.com/keras-team/keras/blob/master/keras/applications/densenet.py)
+  - [Densely Connected Convolutional Networks](https://arxiv.org/abs/1608.06993)
+  - [Based on the Original keras.applications DenseNet](https://github.com/keras-team/keras/blob/master/keras/applications/densenet.py)
 """
 
 import tensorflow as tf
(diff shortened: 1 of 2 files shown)


keras_cv/models/densenet.py
```diff
@@ -15,8 +15,10 @@
 """DenseNet models for KerasCV.
 
 Reference:
-  - [Densely Connected Convolutional Networks](https://arxiv.org/abs/1608.06993)
-  - [Based on the Original keras.applications DenseNet](https://github.com/keras-team/keras/blob/master/keras/applications/densenet.py)
+  - [Densely Connected Convolutional Networks]
```
LukeWood, line 18: I think that the newlines will break these links in keras.io

keras_cv/models/densenet.py
```diff
@@ -26,6 +28,42 @@
 
 BN_AXIS = 3
 
+BASE_DOCSTRING = """Instantiates the {name} architecture.
+
+    Reference:
+        - [Densely Connected Convolutional Networks (CVPR 2017)]
```
LukeWood, line 34: Condense links to 1 line

LukeWood (approved):
Approved with minor comments.

LukeWood (approved):

Issue #530, opened 2022-06-26
Port RegNets from leras.applications


AdityaKane2001:
@sayakpaul @LukeWood I'm interested in porting [RegNets](https://github.com/keras-team/keras/blob/master/keras/applications/regnet.py) from keras to keras_cv.

LukeWood:
Awesome!  Go ahead!

Issue #37, opened 2022-01-11
ResNeXT block/layer


LukeWood:

LukeWood:
We won't expose these as an API.

PR #558, opened 2022-07-02
Porting over ResNets from keras.applications

### LINKED ISSUE #78: Port ResNet50 to the new model architecture style

AdityaKane2001:
@sayakpaul @LukeWood 

Ported over ResNets from keras applications. Most of the code is same as it was with minor changes as discussed in #476. 

Both `keras.applications.resnet.py` and `keras.applications.resnet_v2.py`  have been combined into one file since they belong to the same family of models. `block3` and `stack3` from `resnet.py` was removed as it was redundant. 

/auto Closes #78


keras_cv/models/resnet.py
```diff
@@ -0,0 +1,537 @@
...
+    name="ResNet",
+    include_top=True,
+    weights="None",
+    input_shape=(None, None, 3),
+    pooling=None,
+    num_classes=1000,
+    classifier_activation="softmax",
+    **kwargs,
+):
+    """Instantiates the ResNet, ResNetV2, and ResNeXt architecture.
```
sayakpaul, line 73: Do we have an example of ResNeXt?

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,537 @@
...
+        )
+        x = layers.Activation("relu", name="conv1_relu")(x)
+
+    x = layers.MaxPooling2D(3, strides=2, padding="same", name="pool1_pool")(x)
+
+    x = stack_fn(x)
+
+    if preact:
+        x = layers.BatchNormalization(axis=BN_AXIS, epsilon=1.001e-5, name="post_bn")(x)
+        x = layers.Activation("relu", name="post_relu")(x)
```
sayakpaul, line 146: Would it make sense to add a comment here denoting the differences between ResNet v1 and v2?

sayakpaul: Nevermind, I see that in the docstring.

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,537 @@
...
+
+    if weights is not None:
+        model.load_weights(weights)
+
+    return model
+
+
+# v1 block
+def V1Block(filters, kernel_size=3, stride=1, conv_shortcut=True, name=None):
+    """A residual block.
```
sayakpaul, line 172: Would it make sense to add a shot description of how the v1 block differs from the v2 block?

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,537 @@
...
+        x = V1Block(filters, stride=stride1, name=name + "_block1")(x)
+        for i in range(2, blocks + 1):
+            x = V1Block(filters, conv_shortcut=False, name=name + "_block" + str(i))(x)
+        return x
+
+    return apply
+
+
+def V2Block(filters, kernel_size=3, stride=1, conv_shortcut=False, name=None):
+    """A residual block.
```
sayakpaul, line 247:
```suggestion
    """A residual block (v2).
```

sayakpaul (changes requested):
LGTM with some comments.

keras_cv/models/resnet.py
bhack: If we have It will cover https://github.com/keras-team/keras-cv/issues/37

AdityaKane2001: I removed the Resnext block from the file, I'll change the docstring as well. I thought it would be best that we support ResNetv1&2  for the time being, as ResneXt is not supported by `keras.applications` as of now either.

AdityaKane2001:
@sayakpaul 

Could you please share your thoughts on what I should write in the docstring regarding the difference between V1 and V2 blocks? Asking this because it is a bit difficult to convincingly explain the architecture and its motivation in a few words IMO.

sayakpaul:
> Could you please share your thoughts on what I should write in the docstring regarding the difference between V1 and V2 blocks? Asking this because it is a bit difficult to convincingly explain the architecture and its motivation in a few words IMO.

I think the difference is the way the residual block is constructed in the two blocks. One doesn't use identity mapping and one does (keeping residual branch clean). 

Maybe something along those lines?

AdityaKane2001:
> Hi Aditya -- this PR references https://github.com/keras-team/keras-cv/issues/530, which is about regnet, not resnet. Can you please either update https://github.com/keras-team/keras-cv/issues/530 if this was a typo on the issue, or create a new issue for resnet and link it here?
 Thanks!

Pardon me, updated the link.

keras_cv/models/resnet.py
LukeWood: great, yeah update the docstring please!

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,537 @@
...
+    pooling=None,
+    num_classes=1000,
+    classifier_activation="softmax",
+    **kwargs,
+):
+    """Instantiates the ResNet, ResNetV2, and ResNeXt architecture.
+    Args:
+      stack_fn: a function that returns output tensor for the
+        stacked residual blocks.
+      preact: whether to use pre-activation or not
```
LukeWood, line 77: lets not abbreviate this.  use_preactivation would be easier to read

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,537 @@
...
+        (True for ResNetV2, False for ResNet).
+      include_rescaling: whether or not to Rescale the inputs.If set to True,
```
LukeWood, line 79: missing space before if

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,537 @@
...
+        into, only to be specified if `include_top` is True, and
+        if no `weights` argument is specified.
+      classifier_activation: A `str` or callable. The activation function to use
+        on the "top" layer. Ignored unless `include_top=True`. Set
+        `classifier_activation=None` to return the logits of the "top" layer.
+        When loading pretrained weights, `classifier_activation` can only
+        be `None` or `"softmax"`.
+      **kwargs: For backwards compatibility only.
+
+        Returns:
```
LukeWood, line 108: is the indentation right for Returns?

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,537 @@
...
+    img_input = layers.Input(shape=input_shape)
+
+    if include_rescaling:
+        img_input = layers.Rescaling(1 / 255.0)(img_input)
+
+    x = layers.Conv2D(
+        64, 7, strides=2, use_bias=True, padding="same", name="conv1_conv"
+    )(img_input)
+
+    if not preact:
```
LukeWood, line 134: why is this here for the not case?

keras_cv/models/resnet.py
LukeWood: a comment would be helpful too.  I agree - the resnet code is harder to follow due to the reuse

keras_cv/models/resnet.py
LukeWood: +1

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,537 @@
...
+):
+    """Instantiates the ResNet50 architecture."""
+
+    def stack_fn(x):
+        x = V1Stack(64, 3, stride1=1, name="conv2")(x)
+        x = V1Stack(128, 4, name="conv3")(x)
+        x = V1Stack(256, 6, name="conv4")(x)
+        return V1Stack(512, 3, name="conv5")(x)
+
+    return ResNet(
```
LukeWood, line 346: meta comment: Is it even worth resharing the ResNet body across v1 and v2?   

To me it makes the code HARDER to read by adding an extra level of indirection

LukeWood (changes requested):
Minor changes and a meta comment

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,547 @@
...
+"""
+
+
+def ResNet(
+    stack_fn,
+    use_preactivation,
+    include_rescaling,
+    name="ResNet",
+    include_top=True,
+    weights="None",
```
ianjjohnson, line 73: This looks like it was accidentally stringified

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,547 @@
...
+            be applied.
+      num_classes: optional number of classes to classify images
+        into, only to be specified if `include_top` is True, and
+        if no `weights` argument is specified.
+      classifier_activation: A `str` or callable. The activation function to use
+        on the "top" layer. Ignored unless `include_top=True`. Set
+        `classifier_activation=None` to return the logits of the "top" layer.
+        When loading pretrained weights, `classifier_activation` can only
+        be `None` or `"softmax"`.
+      **kwargs: For backwards compatibility only.
```
ianjjohnson, line 114: In the VGG and DenseNet impls in keras_cv, we use kwargs as a pass-through for params to Keras.model(). I think we should remain consistent with that here.

Either way, I don't think we need to include them for backwards-compatability as this is being moved to a new namespace.

keras_cv/models/resnet.py
```diff
@@ -0,0 +1,547 @@
...
+    # Create model.
+    model = tf.keras.Model(inputs, x, name=name)
+
+    if weights is not None:
+        model.load_weights(weights)
+
+    return model
+
+
+# v1 block
```
ianjjohnson, line 178: nit: I think we can drop this (same on line 313)

ianjjohnson (changes requested):

LukeWood:
@fchollet any strong opinion on refactoring this to not use a shared ResNet base function?

Personally, I think having a file for ResNet v1 and a file for ResNet v2 is more idiomatic/readable as it decreases levels of indirection.

AdityaKane2001:
@sayakpaul  @LukeWood 

Should I follow the structure as in [keras/applications/regnet.py](https://github.com/keras-team/keras/blob/master/keras/applications/regnet.py)? Here we define the model configuration as a `dict` at the beginning of the file and then just access it later according to the individual model. Rest of the logic is covered in the main function.

The most important  reason, as far as I understand, for keeping `stack_fn` like that is to keep the model specific things in the individual model variant function and  the model agnostic functionality in the main ResNet function. Thus I think the above solution does this while keeping the code reasonably readable.

keras_cv/models/resnet_v1.py
```diff
@@ -0,0 +1,357 @@
...
+        "stackwise_filters": [64, 128, 256, 512],
+        "stackwise_blocks": [3, 4, 23, 3],
+        "stackwise_strides": [1, 2, 2, 2],
+    },
+    "ResNet152": {
+        "stackwise_filters": [64, 128, 256, 512],
+        "stackwise_blocks": [3, 8, 36, 3],
+        "stackwise_strides": [1, 2, 2, 2],
+    },
+}
```
sayakpaul, line 40: I think keeping the configs in the form of a dictionary helps provide a clear separation between the different parameters.

keras_cv/models/resnet_v1.py
```diff
@@ -0,0 +1,357 @@
...
+    Returns:
+      Output tensor for the residual block.
+    """
+    if name is None:
+        name = f"v1_block_{backend.get_uid('v1_block')}"
+
+    def apply(x):
+        if conv_shortcut:
+            shortcut = layers.Conv2D(4 * filters,
+                                     1,
```
sayakpaul, line 102: Do we want to spell out the arguments too to keep it consistent with the rest?

keras_cv/models/resnet_v1.py
```diff
@@ -0,0 +1,357 @@
...
+                                                 epsilon=1.001e-5,
+                                                 name=name + "_0_bn")(shortcut)
+        else:
+            shortcut = x
+
+        x = layers.Conv2D(filters, 1, strides=stride, name=name + "_1_conv")(x)
+        x = layers.BatchNormalization(axis=BN_AXIS,
+                                      epsilon=1.001e-5,
+                                      name=name + "_1_bn")(x)
+        x = layers.Activation("relu", name=name + "_1_relu")(x)
```
sayakpaul, line 115: Possible to delegate this to a separate function called something like `conv_bn_relu()`?  I am suggesting it because it's been used many times. Or would it make the code more complicated?

keras_cv/models/resnet_v1.py
```diff
@@ -0,0 +1,357 @@
...
+    include_rescaling,
+    include_top,
+    num_classes=1000,
+    weights=None,
+    input_shape=(None, None, 3),
+    pooling=None,
+    classifier_activation="softmax",
+    name="resnet50",
+    **kwargs,
+):
```
sayakpaul, line 285: Since this doesn't contain pre-trained weights let's have the `num_classes` and related argument as optional or required. An example is here: https://github.com/keras-team/keras-cv/pull/529/files.

sayakpaul (changes requested):
Looking good! Overall the code reads very clear! 

My comments are the same for V2 as V1.

LukeWood:
@ianjjohnson would you be able to perform a pass on this as I am on sick leave?

thanks!

keras_cv/models/resnet_v1.py
```diff
@@ -0,0 +1,357 @@
...
+        raise ValueError(
+            "The `weights` argument should be either `None` or the path to the "
+            "weights file to be loaded. Weights file not found at location: {weights}"
+        )
+
+    if include_top and not num_classes:
+        raise ValueError(
+            "If `include_top` is True, you should specify `num_classes`. "
+            f"Received: num_classes={num_classes}")
+
```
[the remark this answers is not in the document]
ianjjohnson: I think this is a reasonable warning, and if we include it here we should also update other KerasCV models to do the same (in this PR or a separate PR is fine with me)

keras_cv/models/resnet_v2.py
```diff
@@ -0,0 +1,365 @@
...
+    "ResNet152V2": {
+        "stackwise_filters": [64, 128, 256, 512],
+        "stackwise_blocks": [3, 8, 36, 3],
+        "stackwise_strides": [1, 2, 2, 2],
+    },
+}
+
+BN_AXIS = 3
+
+BASE_DOCSTRING = """Instantiates the {name} architecture.
```
ianjjohnson, line 44: (Same as in V1) -- please format and use this docstring as the __doc__ attribute for the exported model methods.

keras_cv/models/resnet_v2.py
```diff
@@ -0,0 +1,365 @@
...
+    name="ResNetV2",
+    include_top=True,
+    weights=None,
+    input_shape=(None, None, 3),
+    pooling=None,
+    num_classes=1000,
+    classifier_activation="softmax",
+    **kwargs,
+):
+    """Instantiates the ResNet and ResNetV2 architecture.
```
ianjjohnson, line 184: nit: V2 only

keras_cv/models/resnet_v2.py
```diff
@@ -0,0 +1,365 @@
...
+    stackwise_filters,
+    stackwise_blocks,
+    stackwise_strides,
+    include_rescaling,
+    name="ResNetV2",
+    include_top=True,
+    weights=None,
+    input_shape=(None, None, 3),
+    pooling=None,
+    num_classes=1000,
```
ianjjohnson, line 180: (Same as V1) -- let's make this default to `None`

keras_cv/models/resnet_v2.py
```diff
@@ -0,0 +1,365 @@
...
+    return apply
+
+
+def ResNetV2(
+    stackwise_filters,
+    stackwise_blocks,
+    stackwise_strides,
+    include_rescaling,
+    name="ResNetV2",
+    include_top=True,
```
ianjjohnson, line 176: For this internal method, and the V1 equivalent, let's default `include_top` to `None` for consistency with the external API.

While this doesn't change behavior, it makes the behavior a bit more explicit

keras_cv/models/resnet_v2.py
```diff
@@ -0,0 +1,365 @@
...
+    include_top,
+    num_classes=1000,
+    weights=None,
+    input_shape=(None, None, 3),
+    pooling=None,
+    classifier_activation="softmax",
+    name="resnet50v2",
+    **kwargs,
+):
+    """Instantiates the ResNet50V2 architecture."""
```
ianjjohnson, line 294: (Same as V1) -- these docstrings can be removed since they will be overwritten by a `setattr` call

ianjjohnson (changes requested):
Thank you for this PR! We are excited to have ResNet support in KerasCV :-)

My comments are mostly minor changes -- overall this looks good.

AdityaKane2001:
@sayakpaul @ianjjohnson Thanks for the reviews! I'll make the changes and get back to you tomorrow.

keras_cv/models/resnet_v2.py
ianjjohnson: My apologies -- I misspoke here. I meant let's make this a required parameter with no default.

keras_cv/models/resnet_v2.py
```diff
@@ -0,0 +1,366 @@
...
+                of the last convolutional block, and thus the output of the model will
+                be a 2D tensor.
+            - `max` means that global max pooling will be applied.
+        name: (Optional) name to pass to the model.  Defaults to "{name}".
+    Returns:
+      A `keras.Model` instance.
+"""
+
+
+def V2Block(filters, kernel_size=3, stride=1, conv_shortcut=False, name=None):
```
ianjjohnson, line 82: For consistency with your changes in V1, let's drop the V2 suffix from these as well in this version

keras_cv/models/resnet_v2.py
```diff
@@ -0,0 +1,366 @@
...
+    return apply
+
+
+def ResNetV2(
+    stackwise_filters,
+    stackwise_blocks,
+    stackwise_strides,
+    include_rescaling,
+    name="ResNetV2",
+    include_top=None,
```
ianjjohnson, line 170: (Same here as in my other comment -- let's make this required instead of default `None`)
again -- my mistake I misspoke on my first comment.

ianjjohnson (approved):
This LGTM. I'd like to get a quick look from @LukeWood before we merge.

Thank you for this great contribution!

keras_cv/models/resnet_v2.py
LukeWood: +1

LukeWood (approved):
Agree with Ian's minor comments.  Other than that LGTM.  Please address, then we can merge

keras_cv/models/resnet_v1.py
```diff
@@ -0,0 +1,360 @@
...
+        x = layers.Dense(
+            num_classes, activation=classifier_activation, name="predictions"
+        )(x)
+    else:
+        if pooling == "avg":
+            x = layers.GlobalAveragePooling2D(name="avg_pool")(x)
+        elif pooling == "max":
+            x = layers.GlobalMaxPooling2D(name="max_pool")(x)
+
+    inputs = img_input
```
LukeWood, line 262: you need to assign `inputs = img_inputs` above before you override it in line 231

keras_cv/models/resnet_v1.py
```diff
@@ -0,0 +1,360 @@
...
+            f"Received: num_classes={num_classes}"
+        )
+
+    if include_top and pooling:
+        raise ValueError(
+            f"`pooling` must be `None` when `include_top=True`."
+            f"Received pooling={pooling} and include_top={include_top}. "
+        )
+
+    img_input = layers.Input(shape=input_shape)
```
LukeWood, line 228: add `inputs = img_input` here

LukeWood, line 228: my recommendation would be to instead of reuse img_input, instead do:

img_input = layers.Input()
x = img_input

if include_rescaling:
    x = layers.Rescaling(1/255.0)(x)

x = layers.Conv2D(x)

This is less error prone than the approach you use here.

LukeWood (changes requested):
spotted some bugs, please update and take a look.

ianjjohnson:
> spotted some bugs, please update and take a look.

I've created #603 to prevent this kind of bug in the future. Thank you for catching this!

AdityaKane2001:
@LukeWood @ianjjohnson @sayakpaul 

Fixed the bug. Please take a look.

sayakpaul:
LGTM

AdityaKane2001:
@LukeWood @ianjjohnson 

Rebased, tested and formatted the code.

LukeWood (approved):
LGTM

PR #559, opened 2022-07-03
Fixed #425 with Lambda layer

### LINKED ISSUE #425: RandomAugmentationPipeline(layers=[]) does not work
This might be a valid use case.

AdityaKane2001:
@LukeWood 

Fixed #425 using a Lambda layer in case of `layers=[]` or `layers=None`. 

/auto Closes #425

### DIFF
--- a/keras_cv/layers/preprocessing/random_augmentation_pipeline.py
+++ b/keras_cv/layers/preprocessing/random_augmentation_pipeline.py
@@ -51,8 +51,8 @@ class RandomAugmentationPipeline(BaseImageAugmentationLayer):
     Args:
         layers: a list of `keras.Layers`.  These are randomly inputs during
             augmentation to augment the inputs passed in `call()`.  The layers passed
-            should subclass `BaseImageAugmentationLayer`. Passing `layers=[]` or
-            `layers=None` would result in a no-op.
+            should subclass `BaseImageAugmentationLayer`. Passing `layers=[]`
+            would result in a no-op.
         augmentations_per_image: the number of layers to apply to each inputs in the
             `call()` method.
         rate: the rate at which to apply each augmentation.  This is applied on a per
@@ -78,7 +78,13 @@ class RandomAugmentationPipeline(BaseImageAugmentationLayer):
         self.augmentations_per_image = augmentations_per_image
         self.rate = rate
         self._layers_empty = False
-        if list(layers) == [] or layers is None:
+
+        if layers is None:
+            raise ValueError(
+                f"Received layers={layers}. Expected `layers` to be a list."
+            )
+
+        if list(layers) == []:
             layers = [tf.keras.layers.Lambda(lambda x: x)]
             self._layers_empty = True
         self.layers = layers
--- a/keras_cv/layers/preprocessing/random_augmentation_pipeline.py
+++ b/keras_cv/layers/preprocessing/random_augmentation_pipeline.py
@@ -79,11 +79,6 @@ class RandomAugmentationPipeline(BaseImageAugmentationLayer):
         self.rate = rate
         self._layers_empty = False
 
-        if layers is None:
-            raise ValueError(
-                f"Received layers={layers}. Expected `layers` to be a list."
-            )
-
         if list(layers) == []:
             layers = [tf.keras.layers.Lambda(lambda x: x)]
             self._layers_empty = True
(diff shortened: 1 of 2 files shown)


keras_cv/layers/preprocessing/random_augmentation_pipeline.py
```diff
@@ -51,7 +51,8 @@ class RandomAugmentationPipeline(BaseImageAugmentationLayer):
     Args:
         layers: a list of `keras.Layers`.  These are randomly inputs during
             augmentation to augment the inputs passed in `call()`.  The layers passed
-            should subclass `BaseImageAugmentationLayer`.
+            should subclass `BaseImageAugmentationLayer`. Passing `layers=[]` or
+            `layers=None` would result in a no-op.
```
LukeWood, line 55: lets not support None

LukeWood (changes requested):

keras_cv/layers/preprocessing/random_augmentation_pipeline.py
```diff
@@ -78,7 +78,13 @@ def __init__(
         self.augmentations_per_image = augmentations_per_image
         self.rate = rate
         self._layers_empty = False
-        if list(layers) == [] or layers is None:
+
+        if layers is None:
```
LukeWood, line 82: I think we can just drop this check.  There’s no need for this, if users pass layers=none it will error out anyways in call.

AdityaKane2001:
@LukeWood Made requested changes.

keras_cv/layers/preprocessing/random_augmentation_pipeline.py
```diff
@@ -99,11 +105,15 @@ def _augment(self, inputs):
 
     def get_config(self):
         config = super().get_config()
+        if self._layers_empty:
```
LukeWood, line 108: This looks like a hack to make unit tests work.

Does passing self.layers change any behavior other than making the UT fail?  If so, just update the unit test to take a single layer.

AdityaKane2001: @LukeWood 

No, it is so that we don't serialize the Lambda layer. The layer will always generate a new Lambda layer object when it gets layers=[]. 

It also does pass the UT, but that's not the main motive.

LukeWood:
Another question:

Can an if statement in call() fix this?  

Something like `if layers == []` or whatever then just return inputs?

AdityaKane2001:
@LukeWood 

> Something like if layers == [] or whatever then just return inputs?

Yes, I that can be done as well. Missed that 🤦‍♂️ Sorry for the fuss with Lambda layer. 

> but is not serializing the lambda layer worth the code readability hit - I guess I am asking what the use case of not serializing the lambda layer is. Is it just to save a little storage in a savedmodel?

Still, to answer the question: serialization of Lambda layer is not an intended effect. Unless the user knows the source code, it is not possible for them to figure out why a Lambda layer was serialized with the model. Also, it is mentioned on https://keras.io/api/layers/core_layers/lambda/ to avoid serializing Lambda layer, hence I thought it would be best to avoid serializing Lambda layer entirely. 

Having said that, IMO what you suggested in https://github.com/keras-team/keras-cv/pull/559#discussion_r916095560 is clearly the best solution to go for. I'll make the relevant changes.

keras_cv/layers/preprocessing/random_augmentation_pipeline.py
```diff
@@ -76,6 +77,7 @@ def __init__(
         super().__init__(**kwargs, seed=seed, force_generator=True)
         self.augmentations_per_image = augmentations_per_image
         self.rate = rate
+
```
LukeWood, line 80: remove spurious newline

keras_cv/layers/preprocessing/random_augmentation_pipeline.py
```diff
@@ -76,6 +77,7 @@ def __init__(
...
         self.layers = layers
```
LukeWood, line 81: lets do:

self.layers = list(layers)

keras_cv/layers/preprocessing/random_augmentation_pipeline.py
```diff
@@ -85,6 +87,9 @@ def __init__(
         )
 
     def _augment(self, inputs):
+        if list(self.layers) == []:
```
LukeWood, line 90: remove list(self.layers) here

LukeWood (changes requested):
lets move the list() call to the constructor

AdityaKane2001:
@LukeWood done.

LukeWood (approved):

PR #564, opened 2022-07-07
Adding Non-Max Suppression Layer

### LINKED ISSUE #461: NonMaxSupression Layer
This should return a KerasCV bounding box format instead of a custom object.

https://github.com/LukeWood/reef-net/blob/master/reef_net/model.py#L247

quantumalaviya:
# What does this PR do?
This PR will add an NMS layer that supports the KerasCV bounding box API. The layer is inspired by [Luke Wood's RetinaNet implementation](https://github.com/LukeWood/reef-net/blob/5f1a3fb8b963fb024a3c49da7ffd43bdccb636cd/reef_net/layers/decode_predictions.py#L7). The biggest change is the support for dynamically computed batch size and removal of the anchor box calculations from this layer (this should be done outside the NMS layer).

Fixes #461 

## Before submitting
      Pull Request section?
      to it if that's the case. (#461)

## Who can review?
@LukeWood

### DIFF
--- a/keras_cv/bounding_box/non_max_suppression.py
+++ b/keras_cv/bounding_box/non_max_suppression.py
@@ -34,7 +34,7 @@ class NonMaxSuppression(tf.keras.layers.Layer):
         bounding_box_format: a case-sensitive string which is one of `"xyxy"`,
             `"rel_xyxy"`, `"xyWH"`, `"center_xyWH"`, `"yxyx"`, `"rel_yxyx"`. The
             position and shape of the bounding box will be followed by the class and
-            confidencevalues (in that order). This is required for proper ranking of
+            confidence values (in that order). This is required for proper ranking of
             the bounding boxes. Therefore, each bounding box is defined by 6 values.
             For detailed information on the supported format, see the
             [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
@@ -147,8 +147,8 @@ class NonMaxSuppression(tf.keras.layers.Layer):
             )
 
             # iterate through the boxes and append it to TensorArray
-            for i in range(nmsed_boxes.valid_detections[i]):
-                boxes = boxes.write(boxes.size(), boxes_recombined[i])
+            for j in range(nmsed_boxes.valid_detections[i]):
+                boxes = boxes.write(boxes.size(), boxes_recombined[j])
 
         # stacking to create a tensor
         boxes = boxes.stack()
--- a/keras_cv/bounding_box/non_max_suppression.py
+++ b/keras_cv/bounding_box/non_max_suppression.py
@@ -1,0 +1,195 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+from keras_cv.bounding_box.converters import convert_format
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class NonMaxSuppression(tf.keras.layers.Layer):
+    """
+    Implements the non-max suppression layer.
+
+    Non-maximal suppression is used to suppress potentially repeated boxes by:
+    1) picking the highest ranked boxes
+    2) pruning away all boxes that have a high IoU with the chosen boxes.
+
+    References:
+        - [Yolo paper](https://arxiv.org/pdf/1506.02640)
+
+    Args:
+        num_classes: an integer representing the number of classes that a bounding
+            box can belong to.
+        bounding_box_format: a case-sensitive string which is one of the supported
+            formats:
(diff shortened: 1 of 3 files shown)


LukeWood (changes requested):
Thank you for the great PR.  I've left you some minor comments.

keras_cv/bounding_box/non_max_suppression.py
```diff
@@ -0,0 +1,195 @@
...
+    ):
+        super().__init__(**kwargs)
+        self.num_classes = num_classes
+        self.bounding_box_format = bounding_box_format
+        self.confidence_threshold = confidence_threshold
+        self.nms_iou_threshold = nms_iou_threshold
+        self.max_detections = max_detections
+        self.max_detections_per_class = max_detections_per_class
+
+    def call(self, images, predictions):
```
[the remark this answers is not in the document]
quantumalaviya: Yeah, doesn't have to be relative for the tf nms op (both yxyx and rel_yxyx are fine).

We can make images optional but they're required for when the user passes relative formats.

keras_cv/bounding_box/non_max_suppression.py
```diff
@@ -0,0 +1,195 @@
...
+    1) picking the highest ranked boxes
+    2) pruning away all boxes that have a high IoU with the chosen boxes.
+
+    References:
+        - [Yolo paper](https://arxiv.org/pdf/1506.02640)
+
+    Args:
+        num_classes: an integer representing the number of classes that a bounding
+            box can belong to.
+        bounding_box_format: a case-sensitive string which is one of the supported
```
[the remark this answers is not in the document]
quantumalaviya: I agree. Can you link the docs here? I can't find the documentation for such a list.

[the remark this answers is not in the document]
LukeWood: https://keras.io/api/keras_cv/bounding_box/formats/

keras_cv/bounding_box/non_max_suppression.py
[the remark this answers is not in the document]
LukeWood: Oh good point, we can actually update the converters to support no-images rel->rel conversions.

keras_cv/bounding_box/non_max_suppression.py
```diff
@@ -0,0 +1,181 @@
...
+                    tf.expand_dims(
+                        nmsed_boxes.nmsed_scores[i][:num_detections], axis=-1
+                    ),
+                ],
+                axis=-1,
+            )
+
+            # iterate through the boxes and append it to TensorArray
+            for i in range(nmsed_boxes.valid_detections[i]):
+                boxes = boxes.write(boxes.size(), boxes_recombined[i])
```
LukeWood, line 151: you're shadowing i here.

for i in range(...[i]).

You need to use another variable, like j.

Then I think you can replace boxes.size() with j.

keras_cv/bounding_box/non_max_suppression.py
```diff
@@ -0,0 +1,181 @@
...
+        # converting all boxes to the original format
+        boxes = convert_format(
+            tf.expand_dims(boxes, axis=0),
+            source="yxyx",
+            target=self.bounding_box_format,
+            images=images,
+        )[0]
+
+        # using cumulative sum to calculate row_limits for ragged tensor
+        row_limits = tf.cumsum(nmsed_boxes.valid_detections)
```
LukeWood, line 165: clever use of this

LukeWood (changes requested):
Biggest concern is the variable shadowing

keras_cv/bounding_box/non_max_suppression.py
```diff
@@ -0,0 +1,181 @@
...
+            scores,
+            self.max_detections_per_class,
+            self.max_detections,
+            self.iou_threshold,
+            self.confidence_threshold,
+            clip_boxes=False,
+        )
+
+        # output will be a ragged tensor because num_boxes will change across the batch
+        return self._encode_to_ragged(nmsed_boxes, images)
```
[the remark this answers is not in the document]
quantumalaviya: Yes, we can cover this as part of a follow-up PR.

keras_cv/bounding_box/non_max_suppression.py
quantumalaviya: That's a great catch, thanks!

That being said, I have left the `boxes.size()` as it is to mimic the `append` function whereas `j` will reset for every batch.

keras_cv/bounding_box/non_max_suppression.py
quantumalaviya: Thanks!

keras_cv/bounding_box/non_max_suppression.py
```diff
@@ -147,8 +147,8 @@ def _encode_to_ragged(self, nmsed_boxes, images):
             )
 
             # iterate through the boxes and append it to TensorArray
-            for i in range(nmsed_boxes.valid_detections[i]):
-                boxes = boxes.write(boxes.size(), boxes_recombined[i])
+            for j in range(nmsed_boxes.valid_detections[i]):
+                boxes = boxes.write(boxes.size(), boxes_recombined[j])
```
LukeWood, line 151: any reason to keep using boxes.size() here now?

quantumalaviya: This is because j will be reset for every batch, we need an append function

LukeWood: Ohhh I see, I'm actually surprised that this works in graph mode.  What a wonderful trick, thanks for showing me this.

LukeWood (approved):

LukeWood:
Thank you Sarvagya for the wonderful contribution.

@quantumalaviya

PR #567, opened 2022-07-08
Added visualization feature to demo utils

divyashreepathihalli:
Demo of loading VOC data and visualizing using bounding box.
vis_demo


examples/layers/preprocessing/bounding_box/demo_utils.py
```diff
@@ -0,0 +1,122 @@
...
+from keras_cv import bounding_box
+
+
+def resize(image, label, img_size=(224, 224), num_classes=10):
+    image = tf.image.resize(image, img_size)
+    label = tf.one_hot(label, num_classes)
+    return {"images": image, "labels": label}
+
+
+def load_oxford_dataset(
```
LukeWood, line 29: you can remove the oxford functions from the bounding box utils

examples/layers/preprocessing/bounding_box/demo_utils.py
```diff
@@ -0,0 +1,122 @@
...
+            "objects": {
+                "bbox": tf.cast(-1, tf.float32),
+                "is_difficult": None,
+                "is_truncated": None,
+                "label": None,
+                "pose": None,
+            },
+        },
+    )
+    return next(iter(dataset.take(1)))
```
LukeWood, line 81: any reason we're taking next(iter()) here?  We should be showing off how to use the KPLs on the entire dataset, not just a batch.

Let's return the dataset

LukeWood, line 81: I recommend mapping the dataset before returning in the following way:

```
def package_to_dict(inputs):
    outputs = {"images": ds["image"], "bounding_boxes": ds["objects"]["bbox"]}
    return outputs
```

LukeWood (changes requested):
Minor changes

also, you can't use `examples.layers.preprocessing.classification.demo_utils` if people want to run the scripts from any directory.   then they have to run from the top level which is a bit annoying to users, lets revert `import examples.layers.preprocessing.classification.demo_utils as demo_utils` => demo_utils

examples/layers/preprocessing/bounding_box/demo_utils.py
divyashreepathihalli: Done!

examples/layers/preprocessing/bounding_box/demo_utils.py
divyashreepathihalli: Done!

divyashreepathihalli: Done!

LukeWood:
Thank you Divya!

PR #573, opened 2022-07-11
Migrate VGG19 to keras_cv.models

ianjjohnson:
# What does this PR do?

Creates a keras_cv.model version of VGG19, ported from keras.applications.
This follows the paradigm set forth by #476 in that it:

- Makes include_preprocessing required
- Does not provide default weights (yet)
- Drops support for channels_first input shape
- Supports kwargs to Keras.model

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood @tanzhenyu

### DIFF
--- a/keras_cv/models/models_test.py
+++ b/keras_cv/models/models_test.py
@@ -18,11 +18,13 @@ from absl.testing import parameterized
 from tensorflow.keras import backend
 
 from keras_cv.models import densenet
+from keras_cv.models import vgg19
 
 MODEL_LIST = [
     (densenet.DenseNet121, 1024),
     (densenet.DenseNet169, 1664),
     (densenet.DenseNet201, 1920),
+    (vgg19.VGG19, 512),
 ]
 
 
--- a/keras_cv/models/vgg19.py
+++ b/keras_cv/models/vgg19.py
@@ -1,0 +1,178 @@
+# Copyright 2022 The KerasCV Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""VGG19 model for KerasCV.
+
+Reference:
+  - [Very Deep Convolutional Networks for Large-Scale Image Recognition](
+      https://arxiv.org/abs/1409.1556) (ICLR 2015)
+"""
+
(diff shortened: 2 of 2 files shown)


keras_cv/models/vgg19.py
```diff
@@ -0,0 +1,176 @@
...
+    input_shape=(None, None, 3),
+    pooling=None,
+    classifier_activation="softmax",
+    name="VGG19",
+    **kwargs,
+):
+    """Instantiates the VGG19 architecture.
+
+    Reference:
+    - [Very Deep Convolutional Networks for Large-Scale Image Recognition](
```
LukeWood, line 42: docstrings can have links on one line now.  I updated the lint rules to support this, so feel free to put this in a single line.  It breaks keras.io to have it on multiple lines in some circumstances.

keras_cv/models/vgg19.py
```diff
@@ -0,0 +1,176 @@
...
+  - [Very Deep Convolutional Networks for Large-Scale Image Recognition](
+      https://arxiv.org/abs/1409.1556) (ICLR 2015)
+"""
+
+import tensorflow as tf
+from tensorflow import keras
+from tensorflow.keras import layers
+
+
+def VGG19(
```
LukeWood, line 28: meta-comment:

VGG architectures aren't SUPER useful at this point, BUT a lot of style transfer models use VGG19.  As such, I think this is a good inclusion, but we probably don't need to do the same for VGG16.

keras_cv/models/vgg19.py
```diff
@@ -0,0 +1,176 @@
...
+        )(x)
+    else:
+        if pooling == "avg":
+            x = layers.GlobalAveragePooling2D()(x)
+        elif pooling == "max":
+            x = layers.GlobalMaxPooling2D()(x)
+
+    model = keras.Model(inputs, x, name=name, **kwargs)
+
+    if weights is not None:
```
LukeWood, line 174: lets add the error checking done on the file existence checks too.

https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/densenet.py#L202

LukeWood (changes requested):
Thanks for the PR @ianjjohnson !  Looks pretty good, only some minor changes

keras_cv/models/vgg19.py
ianjjohnson: this is already done on line 76

keras_cv/models/vgg19.py
ianjjohnson: sg -- thanks

keras_cv/models/vgg19.py
ianjjohnson: Done

keras_cv/models/models_test.py
```diff
@@ -18,11 +18,13 @@
 from tensorflow.keras import backend
 
 from keras_cv.models import densenet
+from keras_cv.models import vgg19
```
LukeWood, line 21: make sure to update __init__.py to import your VGG model!

LukeWood (approved):

LukeWood:
Looks good, thanks!

PR #584, opened 2022-07-13
Initial basic training template (densenet / cifar10)

ianjjohnson:
# What does this PR do?

This creates an initial template for training KerasCV model architectures.
It currently supports
- Failure recovery
- Storing weights to GCS
- Basic tensorboard output to GCS
- A simple custom learning rate schedule
- Very basic metadata output

It's not ready for real use at this point, but let's talk through the structure and high-level approach, and then we can add features from here.

Some open questions:
- How generic do we want to make training scripts (I'd like to avoid too much code duplication but also recognize that it's important to avoid unnecessary indirection, especially since these should mostly stand alone)
- Should we have a single copy of the model setup / training / weight storage code, and make each training script implement just data augmentation and a training loop? Or should we embrace some code duplication so that training scripts can be more easily customized

@LukeWood @tanzhenyu

### DIFF
--- a/examples/training/classification/densenet.py
+++ b/examples/training/classification/densenet.py
@@ -31,12 +31,18 @@ from utils import load_cifar10_dataset
 import keras_cv
 from keras_cv.models import DenseNet121
 
+import wandb
+from wandb.keras import WandbCallback
+
 _GCS_BUCKET = flags.DEFINE_string("gcs_bucket", None, "Name of GCS Bucket")
 _EXPERIMENT_ID = flags.DEFINE_string(
     "experiment_id", None, "An experiment name (preferably a git commit hash)"
 )
 _AUTHOR = flags.DEFINE_string(
     "author", None, "The GitHub username of the author of this training script"
+)
+_WANDB_PROJECT = flags.DEFINE_string(
+    "wandb_project", None, "Optional, a WandB project for exporting training data"
 )
 
 NUM_CLASSES = 10
@@ -87,8 +93,13 @@ def main(argv):
         mode="max",
     )
     tensorboard = TensorBoard(log_dir=gcs_tensorboard_path)
+    callbacks = [gcs_backup, local_checkpoint, tensorboard]
 
-    with tf.distribute.MirroredStrategy().scope():
+    if _WANDB_PROJECT.value:
+        wandb.init(_WANDB_PROJECT.value, entity="keras-team-testing")
+        callbacks.append(WandbCallback())
+
+    with tf.device("/cpu:0"):
         model = DenseNet121(
             include_rescaling=True,
             include_top=True,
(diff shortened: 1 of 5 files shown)


examples/training/classification/densenet.py
```diff
@@ -53,6 +68,7 @@ def main(argv):
     gcs_tensorboard_path = gcs_path_base + "logs/"
 
     train, test = load_cfar10_dataset(BATCH_SIZE)
+    train = train.map(augment)
```
LukeWood, line 72: add num_parallel_calls=tf.data.AUTOTUNE to map calls

examples/training/classification/densenet.py
```diff
@@ -53,6 +68,7 @@ def main(argv):
     gcs_tensorboard_path = gcs_path_base + "logs/"
 
     train, test = load_cfar10_dataset(BATCH_SIZE)
```
LukeWood, line 71: cfar => cifar

examples/training/classification/densenet.py
```diff
@@ -36,10 +37,24 @@
...
-EPOCHS = 10
+EPOCHS = 250
 BATCH_SIZE = 32
 WEIGHTS_PATH = "weights.hdf5"
 
+AUGMENT_LAYERS = [
+    keras_cv.layers.RandAugment(value_range=(0, 255), magnitude=0.7),
+    keras_cv.layers.CutMix(),
+    keras_cv.layers.MixUp(),
+    keras_cv.layers.RandomFlip(),
```
LukeWood, line 48: believe it or not, order matters.

RandomFlip() => first
RandAugment() => second
RandomCutout() => third
CutMix() => fourth
MixUp => fifth

examples/training/classification/densenet.py
```diff
@@ -0,0 +1,113 @@
...
+    # In case storing weights in GCS fails, the weights are also stored locally.
+    os.system(
+        "gsutil cp {local_path} {remote_path}".format(
+            local_path=WEIGHTS_PATH, remote_path=gcs_weights_path
+        )
+    )
+
+    metadata = {
+        "experiment_id": _EXPERIMENT_ID.value,
+        "gcs_weights_path": gcs_weights_path,
```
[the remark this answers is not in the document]
ianjjohnson: I've added author, and a dict of evaluation metrics (which is currently just accuracy)

examples/training/classification/densenet.py
ianjjohnson: Good to know -- is there a good reference I can look at about why order is important here?

ianjjohnson: (Such a reference would probably be good to point to in the training script)

LukeWood: not that I know of actually... 🤔

ianjjohnson:
I am closing this while we think more about how these scripts should be structured.

PR #597, opened 2022-07-18
Include with_targets to RandomResizedCrop

LukeWood:


keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -191,6 +191,9 @@ def _check_class_arguments(
                 f"aspect_ratio_factor={aspect_ratio_factor}"
             )
 
+    def augment_target(self, augment_target, **kwargs):
```
ianjjohnson, line 194: Should we make this a default implementation in the superclass?

ianjjohnson (approved):
Thanks Luke!

PR #600, opened 2022-07-20
Improves `serialization_test.py` to cover all argument with default values

atuleu:
# What does this PR do?

While working on `RandomRotation` I remarked that the serialization of some layers is inaccurate, as they are not saving some default argument. By example, in the case of `RandomRotation`, it is `bounding_box_format`, but for `RandomJpegQuality` it is `seed`.

There is a check for the presence of `seed`, but only if it is passed in the `init_args` of serialization test. If a test configuration for a new layer with a `seed` parameter is added , but no `seed` parameter is added in the `serialization_test.py` `init_args`, then the test is skipped.

This PR solves this issue, by looking at the init parameters of the layers, and ensuring all of them are returned by the `get_config()` method of the added layer. Therefore this would be a regression test if a new parameter is added to an existent layer,  but no additional key is added to `get_config()` return value. In that case the parameter check would be covered, without the need to add the new parameter to `serialization_test.py`. In fact with this version, `init_args` in `serialization_test.py` should only contain non-default argument of each layers and nothing else. In contrast with the current version where it should contain _non default values_ for each of the arguments to cover all serializations scenarios.

List of missing parameters in serialization fixed by this PR:
 * `RandomRotation` : `bounding_box_format`
 * `RandomJpegQuality` : `seed`
 * `Solarization` : `seed`
 * `RandomAugmentationPipeline` : `auto_vectorized`

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood and @tanzhenyu 

### DIFF
--- a/keras_cv/layers/preprocessing/__init__.py
+++ b/keras_cv/layers/preprocessing/__init__.py
@@ -21,7 +21,6 @@ from tensorflow.keras.layers import RandomContrast
 from tensorflow.keras.layers import RandomCrop
 from tensorflow.keras.layers import RandomFlip
 from tensorflow.keras.layers import RandomHeight
-from tensorflow.keras.layers import RandomRotation
 from tensorflow.keras.layers import RandomTranslation
 from tensorflow.keras.layers import RandomWidth
 from tensorflow.keras.layers import RandomZoom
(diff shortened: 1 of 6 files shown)


LukeWood (approved):

LukeWood:
Thank you @atuleu for catching this!!!

PR #603, opened 2022-07-20
Add a test case to enforce rescaling behavior

### LINKED ISSUE #601: Test case to ensure include_rescaling works
We need to add a test to models_test.py to cover this case?  

This unit test should check that for all models, if include_rescaling=True a Rescaling layer is present in the models, and if it is false a Rescaling layer is NOT true.

This came up in https://github.com/keras-team/keras-cv/pull/558

ianjjohnson:
# What does this PR do?

Adds a new test case for models to verify that rescaling does occur when specified and does not occur when not specified.

Fixes #601

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood @tanzhenyu

### DIFF
--- a/keras_cv/models/models_test.py
+++ b/keras_cv/models/models_test.py
@@ -283,6 +283,42 @@ class ApplicationsTest(tf.test.TestCase, parameterized.TestCase):
 
         backend.clear_session()
 
+    @parameterized.parameters(*MODEL_LIST)
+    def test_rescaling(self, app, _):
+        for include_rescaling in [True, False]:
+            if "Mixer" not in app.__name__:
+                model = app(
+                    input_shape=(224, 224, 3),
+                    include_top=True,
+                    num_classes=1000,
+                    include_rescaling=include_rescaling,
+                    weights=None,
+                )
+            elif "MixerB16" in app.__name__ or "MixerL16" in app.__name__:
+                model = app(
+                    input_shape=(224, 224, 3),
+                    patch_size=(16, 16),
+                    include_top=True,
+                    num_classes=1000,
+                    include_rescaling=include_rescaling,
+                    weights=None,
+                )
+            elif "MixerB32" in app.__name__:
+                model = app(
+                    input_shape=(224, 224, 3),
+                    patch_size=(32, 32),
+                    include_top=True,
+                    num_classes=1000,
+                    include_rescaling=include_rescaling,
+                    weights=None,
+                )
+
+            if include_rescaling:
+                self.assertIsNotNone(model.get_layer(name="rescaling"))
+            else:
+                with self.assertRaises(ValueError):
+                    model.get_layer(name="rescaling")
+
 
 def _get_output_shape(model_fn):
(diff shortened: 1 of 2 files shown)


keras_cv/models/models_test.py
```diff
@@ -283,6 +283,42 @@ def test_application_variable_input_channels(self, app, last_dim):
...
+                    input_shape=(224, 224, 3),
+                    include_top=True,
+                    num_classes=1000,
+                    include_rescaling=include_rescaling,
+                    weights=None,
+                )
+            elif "MixerB16" in app.__name__ or "MixerL16" in app.__name__:
+                model = app(
+                    input_shape=(224, 224, 3),
+                    patch_size=(16, 16),
```
LukeWood, line 300: Lets update these tests to remove hard coded checks in the test body.  

Instead, lets introduce support for "kwargs" in MODEL_LIST as an extra parameter.

So for models with no extra params you'll have:

```
    (darknet.DarkNet53, 512, {}),
```

For models that do require more info:

```
    (mlp_mixer.MLPMixerL16, 1024, {"patch_size": (16, 16))
```

Unit tests will take:

```
    def test_rescaling(self, app, args):
```

and model construction will be done:

```
model = app(
                    input_shape=(224, 224, 3),
                    include_top=True,
                    num_classes=1000,
                    include_rescaling=include_rescaling,
                    weights=None,
                    **extra_args
                )
```

LukeWood (changes requested):
readability improvement that will be more important as time goes on, other than that LGTM!

keras_cv/models/models_test.py
```diff
@@ -283,6 +283,42 @@ def test_application_variable_input_channels(self, app, last_dim):
...
+                model = app(
+                    input_shape=(224, 224, 3),
+                    patch_size=(32, 32),
+                    include_top=True,
+                    num_classes=1000,
+                    include_rescaling=include_rescaling,
+                    weights=None,
+                )
+
+            if include_rescaling:
```
LukeWood, line 316: lets have one test with & without:

```
test_rescaling_true
```

```
test_rescaling_false
```

to ensure we hit both cases for all models.

LukeWood:
Thanks for the swift PR, some minor changes requested!

keras_cv/models/models_test.py
ianjjohnson: I realized these cases are covered by existing tests, so I moved the assertions into those cases.

keras_cv/models/models_test.py
ianjjohnson: I've adopted the args model, which has cleaned up this code quite a bit. It's still imperfect due to the output_shape checks, but I think this is a big improvement.

ianjjohnson:
Note that I also uncovered that VGG needs to have a specified input_shape due to the dense layer needing an input shape hint. This required a small change to the tests.

LukeWood (approved):
Thank you Ian!!!

PR #612, opened 2022-07-22
Fixed converters so that rel->rel don't require images

### LINKED ISSUE #604: The bounding box format converters should support rel->rel conversions without images
As for now, xyxy is the default intermediate step in any conversion. When it comes to rel->rel conversions, this isn't ideal because if no image is passed, a `RequiresImagesException` will be raised.

quantumalaviya:
# What does this PR do?
A small fix so that rel->rel conversions don't require images

Fixes #604 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood

### DIFF
--- a/keras_cv/bounding_box/converters.py
+++ b/keras_cv/bounding_box/converters.py
@@ -236,6 +236,10 @@ def convert_format(boxes, source, target, images=None, dtype="float32"):
     if source == target:
         return boxes
 
+    if "rel" in source and "rel" in target:
+        source = source[4:]
+        target = target[4:]
+
     boxes, images, squeeze = _format_inputs(boxes, images)
 
     to_xyxy_fn = TO_XYXY_CONVERTERS[source]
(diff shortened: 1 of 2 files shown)


keras_cv/bounding_box/converters.py
```diff
@@ -236,6 +236,10 @@ def convert_format(boxes, source, target, images=None, dtype="float32"):
     if source == target:
         return boxes
 
+    if "rel" in source and "rel" in target:
+        source = source[4:]
```
LukeWood, line 240: can we make this check "startsWith" instead of "in"?  

also, can we use python string.replace()?   It supports a `count` param:

source = source.replace("rel_", "", 1)

LukeWood (changes requested):
Please add a unit test for a case where both are `rel_` and images is *not* provided.

LukeWood (approved):

PR #613, opened 2022-07-22
Style changes to NMS layer

LukeWood:
- remove symbol import of convert_format(), it is important to keep `bounding_box` as it serves as the subject of the "sentence"
- improve readability by better publicizing the bounding_box.convert_format() call in the `call()` method
- refactor into an additional helper method

# Who can review

@quantumalaviya

### DIFF
--- a/keras_cv/layers/object_detection/non_max_suppression.py
+++ b/keras_cv/layers/object_detection/non_max_suppression.py
@@ -124,13 +124,13 @@ class NonMaxSuppression(tf.keras.layers.Layer):
         # output will be a ragged tensor because num_boxes will change across the batch
         boxes = self._encode_to_dense_output_tensor(nmsed_boxes)
         # converting all boxes to the original format
-        boxes = bounding_box.convert_format(
+        boxes = self._encode_to_ragged(boxes, nmsed_boxes.valid_detections)
+        return bounding_box.convert_format(
             boxes,
             source="yxyx",
             target=self.bounding_box_format,
             images=images,
         )
-        return self._encode_to_ragged(boxes, nmsed_boxes.valid_detections)
 
     def _encode_to_dense_output_tensor(self, nmsed_boxes):
         boxes = tf.TensorArray(tf.float32, size=0, dynamic_size=True)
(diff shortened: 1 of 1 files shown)


keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -121,12 +121,18 @@ def call(self, predictions, images=None):
             self.confidence_threshold,
             clip_boxes=False,
         )
-
         # output will be a ragged tensor because num_boxes will change across the batch
-        return self._encode_to_ragged(nmsed_boxes, images)
+        boxes = self._encode_to_dense_output_tensor(nmsed_boxes)
+        # converting all boxes to the original format
+        boxes = bounding_box.convert_format(
```
quantumalaviya, line 127: Since images are batched, the boxes need to be batched as well. This small fix may help:

```python
        boxes = bounding_box.convert_format(
            tf.expand_dims(boxes, axis = 0),
            source="yxyx",
            images=images,
        )[0]
```

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -121,12 +121,18 @@ def call(self, predictions, images=None):
             self.confidence_threshold,
             clip_boxes=False,
         )
-
         # output will be a ragged tensor because num_boxes will change across the batch
-        return self._encode_to_ragged(nmsed_boxes, images)
+        boxes = self._encode_to_dense_output_tensor(nmsed_boxes)
```
quantumalaviya, line 125: Maybe this can have a different name? This function combines the boxes with the classes/scores and appends all the bounding boxes to a TensorArray.

This is fine too though.

keras_cv/layers/object_detection/non_max_suppression.py
LukeWood: Moved to be done after ragged encoding.

keras_cv/layers/object_detection/non_max_suppression.py
LukeWood: thanks for the feedback, I picked => _decode_nms_boxes_to_tensor

PR #615, opened 2022-07-25
Add Focal Loss

### LINKED ISSUE #319: Focal Loss
Focal loss was used to train [RetinaNet](https://arxiv.org/pdf/1708.02002v2.pdf).

From [Papers with code](https://paperswithcode.com/method/focal-loss):
> A Focal Loss function addresses class imbalance during training in tasks like object detection.

Reference Keras Implementation:
https://keras.io/examples/vision/retinanet/#implementing-smooth-l1-loss-and-focal-loss-as-keras-custom-losses 

What do you think about this component? I think that maybe it could also benefit other models.

quantumalaviya:
# What does this PR do?
Draft PR for Focal Loss

Fixes #319 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood

### DIFF
--- a/keras_cv/losses/focal.py
+++ b/keras_cv/losses/focal.py
@@ -13,9 +13,8 @@
 # limitations under the License.
 
 import tensorflow as tf
-from tensorflow import keras
-
 import tensorflow.keras.backend as K
+from tensorflow import keras
 
 
 @tf.keras.utils.register_keras_serializable(package="keras_cv")
@@ -24,7 +23,7 @@ class FocalLoss(tf.losses.Loss):
 
     Focal loss is a modified cross-entropy designed to perform better with
     class imbalance. For this reason, it's commonly used with object detectors.
-    
+
     Args:
         alpha: a float value between 0 and 1 representing a weighting factor
             used to deal with class imbalance. Positive classes and negative
@@ -57,7 +56,13 @@ class FocalLoss(tf.losses.Loss):
     """
 
     def __init__(
-        self, alpha = 0.25, gamma = 2, from_logits = False, label_smoothing = 0, reduction="none", name="FocalLoss"
+        self,
+        alpha=0.25,
+        gamma=2,
+        from_logits=False,
+        label_smoothing=0,
+        reduction="none",
+        name="FocalLoss",
     ):
         super().__init__(reduction=reduction, name=name)
         self._alpha = float(alpha)
@@ -67,7 +72,9 @@ class FocalLoss(tf.losses.Loss):
 
     def _smooth_labels(self, y_true, y_pred):
         num_classes = tf.cast(tf.shape(y_true)[-1], y_pred.dtype)
-        return y_true * (1.0 - self.label_smoothing) + (self.label_smoothing / num_classes)
+        return y_true * (1.0 - self.label_smoothing) + (
+            self.label_smoothing / num_classes
+        )
 
     def call(self, y_true, y_pred):
         y_pred = tf.convert_to_tensor(y_pred)
--- a/keras_cv/losses/focal.py
+++ b/keras_cv/losses/focal.py
@@ -60,7 +60,6 @@ class FocalLoss(tf.keras.losses.Loss):
         gamma=2,
         from_logits=False,
         label_smoothing=0,
-        reduction="auto",
         **kwargs,
     ):
         super().__init__(reduction=reduction, **kwargs)
--- a/keras_cv/losses/focal.py
+++ b/keras_cv/losses/focal.py
@@ -17,7 +17,7 @@ import tensorflow.keras.backend as K
 
 
 @tf.keras.utils.register_keras_serializable(package="keras_cv")
-class FocalLoss(tf.losses.Loss):
+class FocalLoss(tf.keras.losses.Loss):
     """Implements Focal loss
 
     Focal loss is a modified cross-entropy designed to perform better with
@@ -39,18 +39,18 @@ class FocalLoss(tf.losses.Loss):
             `0.9 + 0.1 / num_classes` for target labels. Default to `0.0`.
 
     References:
-        [Focal Loss paper](https://arxiv.org/abs/1708.02002)
+        - [Focal Loss paper](https://arxiv.org/abs/1708.02002)
 
     Standalone usage:
     ```python
-        y_true = tf.random.uniform([10], 0, maxval=4)
-        y_pred = tf.random.uniform([10], 0, maxval=4)
-        loss = FocalLoss()
-        loss(y_true, y_pred).numpy()
+    y_true = tf.random.uniform([10], 0, maxval=4)
+    y_pred = tf.random.uniform([10], 0, maxval=4)
+    loss = FocalLoss()
+    loss(y_true, y_pred).numpy()
     ```
     Usage with the `compile()` API:
     ```python
-        model.compile(optimizer='adam', loss=keras_cv.losses.FocalLoss())
+    model.compile(optimizer='adam', loss=keras_cv.losses.FocalLoss())
     ```
     """
 
@@ -60,7 +60,7 @@ class FocalLoss(tf.losses.Loss):
         gamma=2,
         from_logits=False,
         label_smoothing=0,
-        reduction="none",
+        reduction="auto",
         name="FocalLoss",
     ):
         super().__init__(reduction=reduction, name=name)
@@ -80,9 +80,7 @@ class FocalLoss(tf.losses.Loss):
         y_true = tf.cast(y_true, y_pred.dtype)
 
         if self.from_logits:
-            probs = tf.nn.sigmoid(y_pred)
-        else:
-            probs = y_pred
+            y_pred = tf.nn.sigmoid(y_pred)
 
         if self.label_smoothing:
             y_true = self._smooth_labels(y_true, y_pred)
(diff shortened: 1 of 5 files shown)


LukeWood (changes requested):

LukeWood:
@quantumalaviya lets add unit tests to this

keras_cv/losses/focal.py
```diff
@@ -0,0 +1,109 @@
...
+    Standalone usage:
+    ```python
+        y_true = tf.random.uniform([10], 0, maxval=4)
+        y_pred = tf.random.uniform([10], 0, maxval=4)
+        loss = FocalLoss()
+        loss(y_true, y_pred).numpy()
+    ```
+    Usage with the `compile()` API:
+    ```python
+        model.compile(optimizer='adam', loss=keras_cv.losses.FocalLoss())
```
LukeWood, line 53: unindent the python code by one level

keras_cv/losses/focal.py
```diff
@@ -0,0 +1,109 @@
...
+        num_classes = tf.cast(tf.shape(y_true)[-1], y_pred.dtype)
+        return y_true * (1.0 - self.label_smoothing) + (
+            self.label_smoothing / num_classes
+        )
+
+    def call(self, y_true, y_pred):
+        y_pred = tf.convert_to_tensor(y_pred)
+        y_true = tf.cast(y_true, y_pred.dtype)
+
+        if self.from_logits:
```
LukeWood, line 82: can you replace this with:

```
if self.from_logits:
    y_pred = tf.nn.sigmoid(y_pred)
```

and remove the probs variable entirely?

LukeWood: ah, nevermind I see that line 91 re-uses this.  But perhaps it would be more performant to use the probabilities directly since we already compute them with sigmoid here?

LukeWood (changes requested):
minor comments, looks correct logically but haven't deeply verified it.

keras_cv/losses/focal.py
```diff
@@ -0,0 +1,108 @@
...
+            y_pred = tf.nn.sigmoid(y_pred)
+
+        if self.label_smoothing:
+            y_true = self._smooth_labels(y_true, y_pred)
+
+        cross_entropy = K.binary_crossentropy(
+            y_true, y_pred, from_logits=self.from_logits
```
LukeWood, line 89: it looks like you're invoking from_logits twice.  You already apply sigmoid in line 83, so this should just be false.

keras_cv/losses/focal.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+        alpha = tf.where(tf.equal(y_true, 1.0), self._alpha, (1.0 - self._alpha))
+        pt = y_true * y_pred + (1.0 - y_true) * (1.0 - y_pred)
+        loss = alpha * tf.pow(1.0 - pt, self._gamma) * cross_entropy
+
+        return tf.reduce_mean(loss, axis=-1)
```
LukeWood, line 96: you can actually drop the reduction, Keras does this for you.   So a loss should return a Tensor of size [batch_size].

This is actually required to support sample weight.

keras_cv/losses/focal_test.py
```diff
@@ -0,0 +1,67 @@
...
+    def test_output_shape(self):
+        y_true = tf.cast(
+            tf.random.uniform(shape=[2, 5], minval=0, maxval=2, dtype=tf.int32),
+            tf.float32,
+        )
+        y_pred = tf.random.uniform(shape=[2, 5], minval=0, maxval=1, dtype=tf.float32)
+
+        focal_loss = FocalLoss()
+
+        self.assertAllEqual(focal_loss(y_true, y_pred).shape, [])
```
LukeWood, line 30: you'll need to update the target shape to be [2]

keras_cv/losses/focal.py
LukeWood: nevermind, I see you are already doing this.

keras_cv/losses/focal.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+
+    def __init__(
+        self,
+        alpha=0.25,
+        gamma=2,
+        from_logits=False,
+        label_smoothing=0,
+        reduction="auto",
+        name="FocalLoss",
```
LukeWood, line 64: add **kwargs and propagate to __init__

LukeWood, line 64: drop the name argument and just rely on **kwargs

LukeWood (changes requested):
minor changes, but LGTM overall.

keras_cv/losses/focal.py
quantumalaviya: That's true. Thanks for catching this!

keras_cv/losses/focal_test.py
quantumalaviya: When reduction is auto, it aggregates over the dimensions. It could be a 4D array and the output will still have a shape of [].

quantumalaviya: If you prefer keeping reduction="none" by default, I can do that. It may go against the other Keras losses though.

keras_cv/losses/focal.py
```diff
@@ -0,0 +1,102 @@
...
+    ```
+    """
+
+    def __init__(
+        self,
+        alpha=0.25,
+        gamma=2,
+        from_logits=False,
+        label_smoothing=0,
+        reduction="auto",
```
LukeWood, line 63: reduction is already auto by default in keras.losses.Loss, can drop this and rely on **kwargs here too

keras_cv/losses/focal_test.py
LukeWood: my bad, sorry this comment is outdated

LukeWood (changes requested):
last comment, then LGTM

keras_cv/losses/focal.py
quantumalaviya: I thought about this but decided against it going after Keras losses. 

I think letting **kwargs deal with it is more ideal, however.

keras_cv/losses/focal.py
```diff
@@ -0,0 +1,101 @@
...
+
+        if self.label_smoothing:
+            y_true = self._smooth_labels(y_true)
+
+        cross_entropy = K.binary_crossentropy(
+            y_true, y_pred, from_logits=self.from_logits
+        )
+
+        alpha = tf.where(tf.equal(y_true, 1.0), self._alpha, (1.0 - self._alpha))
+        pt = y_true * y_pred + (1.0 - y_true) * (1.0 - y_pred)
```
LukeWood, line 86: y_pred here is still the logits if `from_logits=True`, whereas if `from_logits=False` we will be using non-logits here.

We should apply sigmoid ourselves and pass from_logits=False to K.binary_crossentropy)

keras_cv/losses/focal.py
```diff
@@ -0,0 +1,101 @@
...
+        super().__init__(**kwargs)
+        self._alpha = float(alpha)
+        self._gamma = float(gamma)
+        self.from_logits = from_logits
+        self.label_smoothing = label_smoothing
+
+    def _smooth_labels(self, y_true):
+        return y_true * (1.0 - self.label_smoothing) + 0.5 * self.label_smoothing
+
+    def call(self, y_true, y_pred):
```
LukeWood, line 74: I think we need some numerical unit tests.  Can we construct some values and manually compute the expected FocalLoss values?

keras_cv/losses/focal.py
```diff
@@ -0,0 +1,101 @@
...
+        alpha=0.25,
+        gamma=2,
+        from_logits=False,
+        label_smoothing=0,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self._alpha = float(alpha)
+        self._gamma = float(gamma)
+        self.from_logits = from_logits
```
LukeWood, line 68: please add a unit test comparing the result for `from_logits=True` and `from_logits=False`, where in the from_logits=True case we apply sigmoid to the same data.

So basically...

```
y_true = np.random.uniform((8, 10, ...))
y_logits = np.random.uniform((8, 10, ...), min_value=-1000, max_value=1000)

y_pred = softmax(y_logits)

assertEqual(focal_loss_on_logits(y_true, y_logits), focal_loss(y_true, y_pred))
```

LukeWood (changes requested):
Found some bugs that indicates to me that we need more unit tests, please add the suggested unit tests.  Thanks

keras_cv/losses/focal.py
quantumalaviya: Ah, true. My bad again.

I'll just convert after the calculation of `binary_crossentropy`.

keras_cv/losses/focal.py
quantumalaviya: This is already done for `from_logits=True`. I'll add one for `from_logits=False`.

LukeWood (approved):
Thank you Sarvagya, this LGTM

LukeWood:
@quantumalaviya can we discuss the exact differences to model garden here a bit?  I'd like to attempt to introduce this loss to be used in their library, so I'd like to make sure we are away of any numerical differences.  Model garden source: https://pastebin.com/UXcRiPCT

PR #618, opened 2022-07-25
Add an initial Densenet/Imagenet classification training script

ianjjohnson:
# What does this PR do?

This adds an initial DenseNet/ImageNet training script which can be extended and modified for other tasks, models, and datasets.

At this point in time, dataloading is done manually, although eventually we should do it with dataloaders.

@LukeWood


examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+## Setup and flags
+
+"""
+
+flags.DEFINE_string("imagenet_path", None, "Directory from which to load Imagenet.")
+flags.DEFINE_string(
+    "backup_path", None, "Directory which will be used for training backups"
+)
+flags.DEFINE_string(
+    "weights_path", None, "Directory which will be used to store weight checkpoints."
```
LukeWood, line 55: do we need a Tensorboard path?

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+)
+
+FLAGS = flags.FLAGS
+FLAGS(sys.argv)
```
LukeWood, line 59: didn't know you could do this.  no action needed - but very nice to making CLI flags work with a scripted approach instead of having to define a main method.

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+
+NUM_CLASSES = 1000
+BATCH_SIZE = 256
+IMAGE_SIZE = (300, 300)
```
LukeWood, line 63: standard imagenet size is 224x224.

I don't actually know if there is a real historical reason for this....

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+EPOCHS = 250
+
+"""
+## Data loading
+This guide uses the
+[Imagenet dataset](https://www.tensorflow.org/datasets/catalog/imagenet2012). Note that this requires manual download, and does not work out-of-the-box with TFDS.
+To get started, we first load the dataset from a command-line specified directory where ImageNet is stored as TFRecords.
```
LukeWood, line 70: did you have a chance to ask @fchollet  how he prepared the TFRecords?

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+    keys_to_features = {
+        image_key: tf.io.FixedLenFeature((), tf.string, ""),
+        label_key: tf.io.FixedLenFeature([], tf.int64, -1),
+    }
+    parsed = tf.io.parse_single_example(example, keys_to_features)
+
+    # Decode and resize image
+    image_bytes = tf.reshape(parsed[image_key], shape=[])
+    image = tf.io.decode_jpeg(image_bytes, channels=3)
+    image = tf.image.resize(image, IMAGE_SIZE)
```
LukeWood, line 87: keras smart_resize tends to outperform tf.image.resize in terms of final model scores if I remember correctly.

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+
+    # Decode label
+    label = tf.cast(tf.reshape(parsed[label_key], shape=()), dtype=tf.int32) - 1
```
LukeWood, line 90: -1 => I assume this is because we are going from [1, 80] => [0, 79]?

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+
+"""
+## Data loading
+This guide uses the
+[Imagenet dataset](https://www.tensorflow.org/datasets/catalog/imagenet2012). Note that this requires manual download, and does not work out-of-the-box with TFDS.
+To get started, we first load the dataset from a command-line specified directory where ImageNet is stored as TFRecords.
+"""
+
+
+def parse_imagenet_example(example, IMAGE_SIZE):
```
LukeWood, line 74: does this need to be a parameter if the variable name matches the global name?

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+    ]
+    validation_filenames = [
+        f"{FLAGS.imagenet_path}/validation-{i:05d}-of-00128" for i in range(0, 128)
+    ]
+
+    train_dataset = tf.data.TFRecordDataset(filenames=train_filenames)
+    validation_dataset = tf.data.TFRecordDataset(filenames=validation_filenames)
+
+    train_dataset = train_dataset.map(
+        lambda x: parse_imagenet_example(x, IMAGE_SIZE),
```
LukeWood, line 107: if you remove the IMAGE_SIZE param we can drop the lambda

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+    )
+    validation_dataset = validation_dataset.map(
+        lambda x: parse_imagenet_example(x, IMAGE_SIZE),
+        num_parallel_calls=tf.data.AUTOTUNE,
+    )
+
+    return train_dataset.batch(BATCH_SIZE), validation_dataset.batch(BATCH_SIZE)
+
+
+train, test = load_imagenet_dataset()
```
LukeWood, line 118: eventually we will want to move this into a keras_cv.loader, but that can be a follow up PR.

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+"""
+Next, we augment our dataset. We define a set of augmentation layers and then apply them to our input dataset using the `apply_augmentation` method from our KerasCV training utils.
+"""
+
+AUGMENT_LAYERS = [
+    keras_cv.layers.RandomFlip(),
```
LukeWood, line 126: + layers.RandomlyResizedCrop()
+ layers.CutMix()

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+
+@tf.function
+def augment(img, label):
+    inputs = {"images": img, "labels": label}
+    for layer in AUGMENT_LAYERS:
+        inputs = layer(inputs)
+    return inputs["images"], inputs["labels"]
+
+
+train = train.map(augment, num_parallel_calls=tf.data.AUTOTUNE).prefetch(
```
LukeWood, line 140: split into:

```
train = train.map(augment, num_parallel_calls=tf.data.AUTOTUNE)
train = train.prefetch(tf.data.AUTOTUNE)
```

LukeWood, line 140: also prefetch val ds

examples/training/classification/densenet/imagenet/basic_training.py
LukeWood, line 118: lets rename these to train_ds, test_ds

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+    tf.data.AUTOTUNE
+)
+
+
+"""
+Now we can begin training our model. We begin by loading a DenseNet model from KerasCV.
+"""
+
+
+def get_model():
```
LukeWood, line 150: lets consider inlining this

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+        )
+    )
+
+
+"""
+Next, we pick a loss function. Here we use a built-in Keras loss function, so we simply specify it as a string.
+"""
+
+
+def get_loss_fn():
```
LukeWood, line 179: introduce label_smoothing - which will require you to call `keras.losses.CategoricalCrossentropy(label_smoothing=0.1)`.

label_smoothing improves performance when used in conjunction with cutmix

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+"""
+As a last piece of configuration, we configure callbacks for the method. We use EarlyStopping, BackupAndRestore, and a model checkpointing callback.
+"""
+
+
+def get_callbacks():
+    return [
+        EarlyStopping(patience=10),
```
LukeWood, line 199: patience=30

LukeWood (changes requested):
few comments, but LGTM overall.

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,224 @@
...
+        BackupAndRestore(FLAGS.backup_path),
```
LukeWood, line 200: lets use the Tensorboard callback

examples/training/classification/densenet/imagenet/basic_training.py
sebastian-sz: @LukeWood If not `tf.image` maybe a `tf.keras.layers.Resizing(..., crop_to_aspect_ratio=True)` instead?

from [the docs](https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing/image/smart_resize):

>Warning: [tf.keras.preprocessing.image.smart_resize](https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing/image/smart_resize) is not recommended for new code. Prefer [tf.keras.layers.Resizing](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Resizing), which provides the same functionality as a preprocessing layer and adds [tf.RaggedTensor](https://www.tensorflow.org/api_docs/python/tf/RaggedTensor) support. See the [preprocessing layer guide](https://www.tensorflow.org/guide/keras/preprocessing_layers) for an overview of preprocessing layers.

LukeWood: ah yes, thanks Sebastian.  @ianjjohnson lets use a tf.keras.layers.Resizing(..., crop_to_aspect_ratio=True) here.

examples/training/classification/densenet/imagenet/basic_training.py
ianjjohnson: Yeah this is pretty nifty. I really wanted to avoid having the main method approach for these flags.

examples/training/classification/densenet/imagenet/basic_training.py
ianjjohnson: Updated.

examples/training/classification/densenet/imagenet/basic_training.py
ianjjohnson: I presume so, yes. But I'll need to sync with fchollet about how the encoding was set up to verify. I've adapted this method from his ImageNet example.

examples/training/classification/densenet/imagenet/basic_training.py
ianjjohnson: Switched to 224x224

examples/training/classification/densenet/imagenet/basic_training.py
ianjjohnson: Good call -- thank you both.

examples/training/classification/densenet/imagenet/basic_training.py
ianjjohnson: I find this a bit more coherent with the rest of the script, especially for future expansions which may include pretraining. Keeping as much config as possible out of the main training loop will make it easier to follow, and isolate model-specific parameters to this section of the script.

What are your thoughts?

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,234 @@
...
+
+import tensorflow as tf
+from absl import flags
+from keras.callbacks import BackupAndRestore
+from keras.callbacks import EarlyStopping
+from keras.callbacks import ModelCheckpoint
+from keras.callbacks import TensorBoard
+from keras.layers import Resizing
+from keras.losses import CategoricalCrossentropy
+from keras.optimizers import Adam
```
innat, line 24: @ianjjohnson I think it would be nice to import keras from tensorflow. I've encountered many practitioners who just mix old and new keras in the same environment and get import errors.

ianjjohnson: Thanks for calling this out -- done.

examples/training/classification/densenet/imagenet/basic_training.py
LukeWood: Hmm, good point.  I think I'd drop all of these methods and prefer:

``
model = DenseNet121(
        include_rescaling=True,
        include_top=True,
        num_classes=NUM_CLASSES,
        input_shape=IMAGE_SIZE + (3,),
)
optimizer = optimizers.Adam(
        learning_rate=optimizers.schedules.PolynomialDecay(
            initial_learning_rate=0.005,
            decay_steps=train_ds.cardinality().numpy() * EPOCHS / 2,
            end_learning_rate=0.0001,
        )
    )
model.compile(
    optimizer=optimizer, 
    loss=losses.CategoricalCrossEntropy(label_smoothing=0.1), 
    metrics=['accuracy'],
    callbacks=get_callbacks()
)
```

in terms of readability.   My reasoning is that all of the model configuration and training is localized, but the callbacks (which are probably the least important) are non-localized.

That being said...  I dont have a particularly strong opinion and the line I am drawing is 100% arbitrary so I'm open to anything here.

File: examples/training/classification/densenet/imagenet/basic_training.py [hunk 17]
```diff
@@ -0,0 +1,230 @@
...
+        callbacks.ModelCheckpoint(FLAGS.weights_path),
+        callbacks.TensorBoard(log_dir=FLAGS.tensorboard_path),
+    ]
+
+
+"""
+We can now compile the model and fit it to the training dataset.
+"""
+
+with tf.distribute.MirroredStrategy().scope():
```
LukeWood, line 215: Nit: we should be able to JUST construct the model under scope and unindent the rest.

```
with strategy.scope():
    model = get_model()
  
model.compile()

model.fit(train_ds, ...)
```

@rchao to confirm this.

rchao: Correct, we should expect this to work. Please let me know if not.

examples/training/classification/densenet/imagenet/basic_training.py
AdityaKane2001: @LukeWood @ianjjohnson 

I believe the usage varies from case to case and in this particular case Luke's suggestion feels natural to me.
 
Personally, I prefer using the other approach (involving many `get_` functions) in research-like scenarios where the model, loss and optimizer might change frequently and so the code needs to be flexible (as a result of being modular).

ianjjohnson: I think that it makes our training scripts a bit more flexible / easily hackable if we use the `get_xyz` approach. It also enforces a bit of structure which will be good as we expand to including community-contributed training scripts.

Having a paradigm where this is all configured as one big block of model config code leaves us too much wiggle room to end up with ugly code as we get into more complex training methods including pretraining.

(Again, just my opinion. Happy to compromise on this)

AdityaKane2001: @ianjjohnson 

> as we expand to including community-contributed training scripts.

In this context, `get_xyz` seems less powerful IMO. When the repo starts accepting training scripts, a more rigid structure might be better, something akin to Flax's [`TrainState`](https://flax.readthedocs.io/en/latest/api_reference/flax.training.html#flax.training.train_state.TrainState). However the maintenance and readability of such scripts might be worrisome. 

As far as I understand this is just a starting point and we can change the paradigms later in the case of full scale training scripts. So looking at this like any other Keras example at the moment might be the right approach IMO. 

Having said that, I am a bit of rookie with respect to all this. 😅  So please correct me if I'm wrong.

ianjjohnson: I'd like to avoid having a really prescriptive structure like `TrainState` because we'd like these scripts to serve as an example of model training out-of-the-box with Keras+KerasCV. Adding extra structure beyond basic Python methods would obfuscate the library usage a bit.

We're still trying to work out what the best approach is, so there's no right or wrong

LukeWood: > Having a paradigm where this is all configured as one big block of model config code leaves us too much wiggle room to end up with ugly code as we get into more complex training methods including pretraining.

One point in favor of `get_xyz` is that it makes it easier for us to spot small mistakes in code reviews.  I'm happy with either scenario here, lets go with Ian's approach for now.

examples/training/classification/densenet/imagenet/basic_training.py
```diff
@@ -0,0 +1,230 @@
...
+from absl import flags
+from tensorflow.keras import callbacks
+from tensorflow.keras import layers
+from tensorflow.keras import losses
+from tensorflow.keras import optimizers
+
+import keras_cv
+from keras_cv.models import DenseNet121
+
+"""
```
LukeWood, line 26: this actually does have to go at the top of the file.

LukeWood (approved):

LukeWood:
Thanks @ianjjohnson !

PR #627, opened 2022-07-28
Add pre-trained weight loading (starting with DenseNet121)

ianjjohnson:
This adds the overall structure for pre-trained weight loading, starting with some initial weights for DenseNet121. (These weights aren't very good yet -- they get 63% top-1 validation accuracy on ImageNet -- but they serve as a basis to work on the API)

@LukeWood

### DIFF
--- a/keras_cv/models/weights.py
+++ b/keras_cv/models/weights.py
@@ -29,7 +29,8 @@ def parse_weights(weights, include_top, model_type):
     raise ValueError(
         "The `weights` argument should be either `None`, a the path to the "
         "weights file to be loaded, or the name of pre-trained weights from "
-        "`weights.py`.  Invalid `weights` argument: {weights}"
+        "https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/weights.py. "
+        "Invalid `weights` argument: {weights}"
     )
 
 
--- a/keras_cv/models/weights.py
+++ b/keras_cv/models/weights.py
@@ -1,0 +1,38 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


keras_cv/models/weights.py
```diff
@@ -0,0 +1,47 @@
...
+        "weights file to be loaded, or the name of pre-trained weights from "
+        "`weights.py`.  Invalid `weights` argument: {weights}"
+    )
+
+
+BASE_PATH = "https://storage.googleapis.com/keras-cv/models"
+WEIGHTS_CONFIG = {
+    "densenet121": {
+        # Current best weights are imagenet-v0
+        "imagenet": "5c51af007f7f3722b50d9390db4b2082962d5ba1ab5d184c3f531f3886200894",
```
LukeWood, line 40: we also need to index by pretraining approach

LukeWood (changes requested):
just need to support pretraining approach, then LGTM

keras_cv/models/weights.py
```diff
@@ -0,0 +1,47 @@
...
+        return data_utils.get_file(
+            origin=f"{BASE_PATH}/{model_type}/{weights}.h5",
+            cache_subdir="models",
+            file_hash=WEIGHTS_CONFIG[model_type][weights],
+        )
+
+    raise ValueError(
+        "The `weights` argument should be either `None`, a the path to the "
+        "weights file to be loaded, or the name of pre-trained weights from "
+        "`weights.py`.  Invalid `weights` argument: {weights}"
```
LukeWood, line 32: lets provide users with a link here.  For now, we can link to the literal github file, in the future we can link to a keras.io documentation page.

If we don't want to include this yet, you can leave TODO(ianjjohnson) in a comment to fix this with a keras.io link

LukeWood (approved):
approved with a single change, thanks Ian!

LukeWood (approved):

PR #634, opened 2022-07-30
Add IoU loss

quantumalaviya:
# What does this PR do?
Adding IoU loss

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood


quantumalaviya:
Although a draft PR exists for this (#449), I discussed the API, tests, and the name with @LukeWood, and we decided that there should be a new PR for this.

keras_cv/losses/iou_loss.py
```diff
@@ -0,0 +1,93 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import tensorflow as tf
+
+from keras_cv.bounding_box import compute_iou
```
LukeWood, line 18: prefer:

```
...
bounding_box.compute_iou()
```

keras_cv/losses/iou_loss.py
```diff
@@ -0,0 +1,93 @@
...
+    loss = IoULoss(bounding_box_format = "xyWH")
+    loss(y_true, y_pred).numpy()
+    ```
+    Usage with the `compile()` API:
+    ```python
+    model.compile(optimizer='adam', loss=keras_cv.losses.IoULoss())
+    ```
+    """
+
+    def __init__(self, bounding_box_format, mode="log", **kwargs):
```
tanzhenyu, line 58: Do you anticipate both inputs will be from different formats?

keras_cv/losses/iou_loss.py
```diff
@@ -0,0 +1,93 @@
...
+                "IoULoss expects mode to be one of 'linear', 'square' or 'log' "
+                f"Received mode={self.mode}, "
+            )
+
+    def call(self, y_true, y_pred):
+        y_pred = tf.convert_to_tensor(y_pred)
+        y_true = tf.cast(y_true, y_pred.dtype)
+
+        ious = bounding_box.compute_iou(y_true, y_pred, self.bounding_box_format)
+        iou = tf.reduce_mean(ious, axis=[-2, -1])
```
tanzhenyu, line 74: This seems to suggest y_pred and y_true have specific shapes, maybe document it here?

tanzhenyu (commented):
Thanks for the PR!

keras_cv/losses/iou_loss.py
LukeWood: Even a stage above this: we need to raise error messages if the boxes do not comply to the spec we expect

keras_cv/losses/iou_loss.py
LukeWood: Yeah, we need to support all bounding box formats.  This is critical to our API consistency, and our continued effort to minimize silent errors.

keras_cv/losses/iou_loss.py
```diff
@@ -0,0 +1,93 @@
...
+            - `"squared"`. The loss will be calculated as 1 - iou<sup>2</sup>
+            - `"log"`. The loss will be calculated as -ln(iou)
+            Defaults to "log".
+
+    References:
+        - [UnitBox paper](https://arxiv.org/pdf/1608.01471)
+
+    Sample Usage:
+    ```python
+    y_true = tf.random.uniform([5, 10, 6], minval=0, maxval=10, dtype=tf.dtypes.int32)
```
LukeWood, line 47: y_true should not have format [5, 10, 6] -> it should have (5, 10, 5).

no confidence on ground truths.

Also, please use tuples for shapes - thanks!

keras_cv/losses/iou_loss.py
```diff
@@ -0,0 +1,93 @@
...
+        self.mode = mode
+
+        if self.mode not in ["linear", "square", "log"]:
+            raise ValueError(
+                "IoULoss expects mode to be one of 'linear', 'square' or 'log' "
+                f"Received mode={self.mode}, "
+            )
+
+    def call(self, y_true, y_pred):
+        y_pred = tf.convert_to_tensor(y_pred)
```
LukeWood, line 70: is convert_to_tensor required here?

keras_cv/losses/iou_loss.py
```diff
@@ -0,0 +1,93 @@
...
+        y_true = tf.cast(y_true, y_pred.dtype)
```
LukeWood, line 71: prefer casting both to self.compute_dtype

LukeWood, line 71: lets do the following:

raise value error if y_true.shape does not match [None, None 4+]
same for y_pred

keras_cv/losses/iou_loss.py
LukeWood: lets replace this with:

```
mean_iou = tf.reduce_mean(ious, axis=[-2, -1])
```

keras_cv/losses/iou_loss_test.py
```diff
@@ -0,0 +1,78 @@
...
+        y_pred = [
+            [0, 0, 5, 6, 4, 0.9],
+            [0, 0, 7, 3, 1, 0.76],
+            [4, 5, 5, 6, 4, 0.04],
+            [2, 1, 3, 3, 7, 0.48],
+        ]
+
+        iou_loss = IoULoss(bounding_box_format="xywh")
+
+        self.assertAllClose(iou_loss(y_true, y_pred), 2.0311017)
```
LukeWood, line 78: how did we come up with this value, can we actually document this?

keras_cv/losses/iou_loss_test.py
```diff
@@ -0,0 +1,78 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv.losses.iou_loss import IoULoss
+
+
+class IoUTest(tf.test.TestCase):
+    def test_output_shape(self):
```
LukeWood, line 21: can you add test cases over the reduction strategies, in particular `reduction='none'` is good to test.

LukeWood (changes requested):
Thanks for the PR!  This will be useful in YOLO models.

@quantumalaviya minor comments, but overall LGTM

keras_cv/losses/iou_loss.py
quantumalaviya: Does that mean I should add two `bounding_box_format` variables to the constructor, i.e. one for true and one for pred?

LukeWood: No, It’s fine as is feel free to resolve

keras_cv/losses/iou_loss.py
quantumalaviya: Raising errors inside the call method? Which method would you recommend for this?

keras_cv/losses/iou_loss_test.py
quantumalaviya: This was a simple IoU calculation to check the value, should I add a comment for this?

keras_cv/losses/iou_loss.py
LukeWood: raise ValueError().

if you use`y_true.shape` instead of `tf.shape(y_true)` it will just happen at trace time, not at graph invocation time.  This allows us to perform validation at runtime

keras_cv/losses/iou_loss_test.py
LukeWood: yeah please add a comment just explaining where it came from.

keras_cv/losses/iou_loss.py
quantumalaviya: Losses don't seem to have a `self.compute_dtype`. See https://github.com/keras-team/keras/blob/v2.9.0/keras/losses.py#L34-L197. Let me know if you meant something else!

keras_cv/losses/iou_loss_test.py
quantumalaviya: `test_output_shape_reduction_none` covers this, right?

keras_cv/losses/iou_loss.py
quantumalaviya: Yeah, in case we need to support regular lists and numpy matrices? This seems to be the norm throughout most keras loss implementations

PR #639, opened 2022-08-01
Add input-tensor param to all KerasCV.models

### LINKED ISSUE #638: Re-introduce `input_tensor` in the keras_cv.models codebase

ianjjohnson:
Fixes #638

This is necessary for high performance of KerasCV models used as backbones in RetinaNet and other applications

@LukeWood

### DIFF
--- a/keras_cv/models/csp_darknet.py
+++ b/keras_cv/models/csp_darknet.py
@@ -24,6 +24,7 @@ import tensorflow as tf
 from tensorflow import keras
 from tensorflow.keras import layers
 
+from keras_cv.models import utils
 from keras_cv.models.__internal__.darknet_utils import CrossStagePartial
 from keras_cv.models.__internal__.darknet_utils import DarknetConvBlock
 from keras_cv.models.__internal__.darknet_utils import DarknetConvBlockDepthwise
@@ -112,13 +113,7 @@ def CSPDarkNet(
     base_channels = int(width_multiplier * 64)
     base_depth = max(round(depth_multiplier * 3), 1)
 
-    if input_tensor is None:
-        inputs = layers.Input(shape=input_shape)
-    else:
-        if not keras.backend.is_keras_tensor(input_tensor):
-            inputs = layers.Input(tensor=input_tensor, shape=input_shape)
-        else:
-            inputs = input_tensor
+    inputs = utils.parse_model_inputs(input_shape, input_tensor)
 
     x = inputs
     if include_rescaling:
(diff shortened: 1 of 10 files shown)


keras_cv/models/csp_darknet.py
```diff
@@ -109,7 +112,13 @@ def CSPDarkNet(
     base_channels = int(width_multiplier * 64)
     base_depth = max(round(depth_multiplier * 3), 1)
 
-    inputs = layers.Input(shape=input_shape)
+    if input_tensor is None:
```
LukeWood, line 115: looks like we can use a utility function for this.  

Let's do:
keras_cv/models/utils.py

```
def parse_model_input(input_shape, input_tensor):
    if input_tensor is None:
       ....
    # return based on these values
```

keras_cv/models/csp_darknet.py
```diff
@@ -40,6 +40,7 @@ def CSPDarkNet(
     num_classes=None,
```
LukeWood, line 40: lets add a unit test to models_test.py to cover this and make sure we can construct a model (like the backbone) using this API.

LukeWood (changes requested):
minor comments, otherwise LGTM.

Thanks for the quick PR!

keras_cv/models/csp_darknet.py
ianjjohnson: Done -- I've added a test and verified that without this new parameter, the test fails.

keras_cv/models/utils.py
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Utility functions for models"""
+
+from tensorflow import keras
+from tensorflow.keras import layers
+
+
+def parse_model_inputs(input_shape, input_tensor):
```
LukeWood, line 21: last thing we could do here is add test coverage over parse_model_inputs in `utils_test.py`

test:
- it constructs a keras.Input for input_shape
- it returns the tensor of the input_tensor is a keras tensor

Then LGTM, sorry for the late batch of comments.  Just thought of this after seeing the helper.

LukeWood (changes requested):
added a comment, sorry about that!

keras_cv/models/utils.py
ianjjohnson: Added a test case using assertIs to make sure the input tensor is returned when possible, and otherwise the shape matches up correctly

LukeWood (approved):

LukeWood:
Thanks @ianjjohnson - looks like a merge conflict is blocking merge

ianjjohnson:
> Thanks @ianjjohnson - looks like a merge conflict is blocking merge

I've fixed the merge conflict. Once CI passes I'll merge.

PR #646, opened 2022-08-02
Add a transfer learning example to demonstrate the weights API

ianjjohnson:
This simple example shows how to use the weights API in KerasCV.models and how to use pre-trained weights for (very simple) transfer learning.

This script requires the latest GitHub version of keras-cv as it requires the code from #627 

@LukeWood


ianjjohnson:
Closing because I screwed up the source branch.

PR #648, opened 2022-08-03
Switch to tensorboard.dev for training metadata handling

ianjjohnson:
This changes update_training_history.py to upload tensorboard logs to tensorboard.dev, and then use tensorboard.dev's API to load relevant metrics from training logs.

This also switches the script to support both flags and CLI input to give the user some flexibility.

@LukeWood

### DIFF
--- a/shell/weights/update_training_history.py
+++ b/shell/weights/update_training_history.py
@@ -1,84 +1,46 @@
 import json
 import os
 import sys
-
 import tensorboard as tb
+
 from absl import flags
 
 flags.DEFINE_string("tensorboard_logs_path", None, "Path to tensorboard logs to load")
 flags.DEFINE_string("training_script_path", None, "Path to the training script")
-flags.DEFINE_string(
-    "weights_version",
-    None,
-    "The version of the training script used to produce the latest weights. For example, v0",
-)
-flags.DEFINE_string(
-    "pr_number", None, "The PR number that merged the training script into KerasCV"
-)
-flags.DEFINE_string(
-    "pr_author", None, "The GitHub username of the author of the training script"
-)
+flags.DEFINE_string("weights_version", None, "The version of the training script used to produce the latest weights. For example, v0")
+flags.DEFINE_string("pr_number", None, "The PR number that merged the training script into KerasCV")
+flags.DEFINE_string("pr_author", None, "The GitHub username of the author of the training script")
 
 FLAGS = flags.FLAGS
 FLAGS(sys.argv)
 
-weights_version = FLAGS.weights_version or input(
-    "Input the weights version for your script\n"
-)
+weights_version = FLAGS.weights_version or input("Input the weights version for your script\n")
 
-training_script_path = FLAGS.training_script_path or input(
-    "Input the path to your training script\n"
-)
+training_script_path = FLAGS.training_script_path or input("Input the path to your training script\n")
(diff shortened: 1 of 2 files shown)


shell/weights/update_training_history.py
```diff
@@ -1,54 +1,92 @@
...
+    "weights_version",
     None,
-    "Path to the JSON file for run history for the producing training script",
+    "The version of the training script used to produce the latest weights. For example, v0",
+)
+flags.DEFINE_string(
+    "pr_number", None, "The PR number that merged the training script into KerasCV"
+)
+flags.DEFINE_string(
+    "pr_author", None, "The GitHub username of the author of the training script"
```
LukeWood, line 19: looks like this isn't inserted into the metadata anywhere

shell/weights/update_training_history.py
```diff
@@ -1,54 +1,92 @@
...
-# Find the single validation logs in the supplied logs dir
-files = os.listdir(FLAGS.logs_path)
-# Tensorboard logs are named based on timestamp, so the largest name alphabetically is the most recent.
-most_recent_file = max(files)
+weights_version = FLAGS.weights_version or input(
+    "Input the weights version for your script\n"
+)
 
-filename = FLAGS.logs_path + "/" + most_recent_file
+training_script_path = FLAGS.training_script_path or input(
```
LukeWood, line 29: should we also allow an input for author?

shell/weights/update_training_history.py
```diff
@@ -1,54 +1,92 @@
...
-print(f"Max validation accuracy: {max_validation_accuracy}")
+new_results = {
+    "validation_accuracy": max_validation_accuracy,
+    "epochs_trained": training_epochs,
+    "tensorboard_logs": f"https://tensorboard.dev/experiment/{tensorboard_experiment_id}/",
+}
 
-new_results = {FLAGS.weights_version: {"validation_accuracy": max_validation_accuracy}}
+pr_number = FLAGS.pr_number or input(
+    "Input the PR number that merged your script into KerasCV\n"
```
LukeWood, line 77: should this be a link instead of a number?

shell/weights/update_training_history.py
```diff
@@ -1,54 +1,92 @@
...
+# Check if the JSON file already exists
+results_file = open(training_script_json_path, "r")
 results_string = results_file.read()
 results = json.loads(results_string) if results_string != "" else {}
 results_file.close()
 
-results_file = open(FLAGS.training_script_json, "w")
-results[FLAGS.weights_version] = {"validation_accuracy": max_validation_accuracy}
+results_file = open(training_script_json_path, "w")
+results[weights_version] = new_results
```
LukeWood, line 90: can we write the PR result to the results file too?

LukeWood (changes requested):

LukeWood:
Thanks for the PR!  A few updates

shell/weights/update_training_history.py
ianjjohnson: Good call -- I've kept the flag as pr_number but the user input version is a full link.

LukeWood: seems reasonable

LukeWood (approved):

LukeWood:
Cool, great use of the integrations

PR #651, opened 2022-08-03
Silence E402 linter error

ianjjohnson:
E402 disallows imports from being anywhere other than the top of a file. This is necessary for Keras.io tutorial-style training scripts.

@LukeWood

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,7 @@ max-line-length = 88
 max-doc-length = 200
 per-file-ignores =
     ./keras_cv/__init__.py:E402, F401
-    ./keras_cv/examples/training/*:E402
+    ./examples/**/*:E402
     **/__init__.py:F401
 ignore =
     # Conflicts with black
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -16,6 +16,7 @@ max-line-length = 88
 max-doc-length = 200
 per-file-ignores =
     ./keras_cv/__init__.py:E402, F401
+    ./keras_cv/examples/training/*:E402
```
LukeWood, line 19: You'll need to do: ./examples/*

./keras_cv/examples doesn't exist

LukeWood: Lets do all examples, you might need to do examples/**/* to achieve this.

ianjjohnson: Hmm. I thought I had tested this but it turned out I hadn't correctly caused the error in the file I was testing with. I've updated this to ./examples/**/* and it now seems to be working correctly.

LukeWood:
Thanks @ianjjohnson!  We also need to remove the copyright notice from the files in examples to make them work with tutobooks, I can handle this.

PR #658, opened 2022-08-04
Improves `shell/format.sh` for local environment setups

atuleu:
# What does this PR do?

Changes introduced in 854e0b6b9e50b8cf12bd3666e7713ffe2aacd300 ( #563 ) made formatting very slow when using local vitual environment ( a local directory such as `venv` or `.direnv`). This is because the `setup.cfg` was overriding the default skip rules of `isort` ( which normally includes such local subdirectories) to solely excludes `keras_cv/__init__.py`. Therefore isort would sort all module from tensorflow ( that you would expect to be present in you local venv when developping for keras-cv)

This commit fixes this quality of life issue, by disabling with command comment `isort` for the very few first lines of `keras_cv/__init__.py`. It is also more robust, as the subsequent import statement in that file will still be automatically sorted by `isort`

## Before submitting
      Pull Request section?
      to it if that's the case. 
    _No but it is a very simple issue with a straightforward fix_
- [ ] Did you write any new necessary tests? _No, but it is not easily testable_

## Who can review?
@LukeWood @tanzhenyu @adhadse

### DIFF
--- a/keras_cv/__init__.py
+++ b/keras_cv/__init__.py
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# isort:off
 from keras_cv import version_check
 
 version_check.check_tf_version()
-
+# isort:on
 
 from keras_cv import layers
 from keras_cv import losses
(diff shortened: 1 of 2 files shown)


LukeWood (approved):

LukeWood:
Thank you @atuleu!  Your continued contributions are greatly appreciated!

PR #659, opened 2022-08-04
Restructure training scripts to avoid duplication

ianjjohnson:
This makes training scripts no-longer model-specific, but instead configurable per-model.
Users can still contribute improvements to scripts via:
- New scripts
- Improvements to existing scripts
- New parameters which improve on script performance

The only major outstanding question I have is: what should we do with historical training metadata when a script is improved? Perhaps the metadata for each run should include a script version based on a commit hash?

Let me know your thoughts.

(This also makes a few tweaks to basic_training to make it modular and switch to a DecreaseLROnPlateau strategy for learning rate).

@LukeWood


examples/training/classification/basic_imagenet_training.json
```diff
@@ -0,0 +1,25 @@
+{
+    "script_authors": ["ianjjohnson"]
+    "densenet121": {
```
LukeWood, line 3: so every model will land here basically?

examples/training/classification/basic_imagenet_training.json
```diff
@@ -0,0 +1,25 @@
+{
+    "script_authors": ["ianjjohnson"]
```
LukeWood, line 2: I do like the idea of "authors" to attribute credit to the entire group that wrote it.

examples/training/classification/basic_imagenet_training.py
```diff
@@ -161,7 +172,8 @@ def augment(img, label):
 strategy = tf.distribute.MirroredStrategy()
 
 with strategy.scope():
-    model = DenseNet121(
+    model = eval(f"models.{FLAGS.model_name}")
```
LukeWood, line 175: can we use __dict__ here too?

examples/training/classification/basic_imagenet_training.py
```diff
@@ -196,7 +203,7 @@ def augment(img, label):
 """
 
 
-metrics = [metrics.CategoricalAccuracy()]
+metrics = ["accuracy"]
```
LukeWood, line 206: I think Francois wanted to use a specific metric instance right?  lets keep it in line with the other ones

examples/training/classification/basic_imagenet_training.py
```diff
@@ -206,6 +213,9 @@ def augment(img, label):
 
 
 callbacks = [
+    callbacks.ReduceLROnPlateau(
```
LukeWood, line 216: patience=20 will generalize better.  Augmentation is very stochastic.

examples/training/classification/basic_imagenet_training.py
```diff
@@ -206,6 +213,9 @@ def augment(img, label):
...
+        monitor="val_loss", factor=0.3, patience=5, min_lr=0.0001
+    ),
     callbacks.EarlyStopping(patience=30),
```
LukeWood, line 219: lets increase patience now that we use callbacks.ReduceLROnPlateau()

LukeWood (changes requested):
Some comments, thanks Ian!

examples/training/classification/basic_imagenet_training.json
ianjjohnson: Right -- every model trained using this script.

examples/training/classification/basic_imagenet_training.py
ianjjohnson: Yes, but it turns out that this throws an error because if you do it with a real object you have to do it in the same strategy scope as you instantiate the model.

I'll just update this to re-open the scope when instantiating this.

examples/training/classification/basic_imagenet_training.py
ianjjohnson: I've updated this to 40 (basically meaning two consecutive learning rate drops without improvement)

examples/training/classification/basic_imagenet_training.py
ianjjohnson: Done.

examples/training/classification/basic_imagenet_training.py
```diff
@@ -47,22 +47,32 @@
...
-    "backup_path", None, "Directory which will be used for training backups"
+    "backup_path", None, "Directory which will be used for training backups."
 )
 flags.DEFINE_string(
     "weights_path", None, "Directory which will be used to store weight checkpoints."
 )
 flags.DEFINE_string(
     "tensorboard_path", None, "Directory which will be used to store tensorboard logs."
 )
+flags.DEFINE_integer("batch_size", 256, "Batch size for training and evaluation.")
```
LukeWood, line 63: I'm not a huge fan of "basic_imagenet_training.py"

can we use the structured `examples/training/classification/imagenet/basic_training.py`, include a README providing an overview of the task, say we use .tfrecord loading, etc.

Then in subdirectories we can have specific scripts: "ResNet50"

LukeWood: basic_imagenet_training.py is a good name for a keras.io tutorial - but we can do that in keras.io when we move there.

ianjjohnson: Done -- restructured again as discussed

LukeWood (approved):

LukeWood:
looks good to me!  Thanks Ian!

PR #661, opened 2022-08-04
Use aliasing for weights config to avoid duplicate weight files

ianjjohnson:
@LukeWood


LukeWood (approved):
looks like code needs a reformat, but LGTM.

Issue #675, opened 2022-08-08
Introduce utility functions to convert between -1 padded bounding box tensors and ragged bounding box tensors


LukeWood:
@divyashreepathihalli this will help us support Ragged inputs to the BaseImageAugmentationLayer

divyashreepathihalli:
https://github.com/keras-team/keras-cv/pull/677

PR #677, opened 2022-08-08
Add ragged bounding box support and utils to pad and filter ragged bounding boxes to keras cv

divyashreepathihalli:
https://github.com/keras-team/keras-cv/issues/675 
keras_cv should now be supporting batched/unbatched bounding box Tensor/RaggedTensor.
utils added to keras_cv.bounding_box
- pad with sentinels(bounding_boxes, default_value)- Pads the given bounding box tensor with `sentinel_value`.

- filter_sentinels(boudning_boxes, default_value) - Filters `sentinel_value`  padded bounding box `Tensor` to a `tf.RaggedTensor`.

- add_class_id(boudning_boxes, class_id) - Add class ID to the innermost Tensor or RaggedTensor representing bounding boxes.


keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,7 +293,14 @@ def _augment(self, inputs):
         return result
 
     def _batch_augment(self, inputs):
-        return self._map_fn(self._augment, inputs)
+        if BOUNDING_BOXES in inputs:
```
LukeWood, line 296: if users provide us with already padded bounding boxes, we should return padded bounding boxes.  As such, we need to use the `_format_inputs()` pattern here, and store whether or not the bounding box inputs were padded or not.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,7 +293,14 @@ def _augment(self, inputs):
...
+            inputs[BOUNDING_BOXES] = self._pad_bounding_boxes(inputs[BOUNDING_BOXES])
+        augmented_batch = self._map_fn(self._augment, inputs)
+        if BOUNDING_BOXES in inputs:
+            augmented_batch[BOUNDING_BOXES] = self._filter_padded_boxes(
+                augmented_batch[BOUNDING_BOXES]
+            )
+        return augmented_batch
```
LukeWood, line 303: I think what we should do is add a new return value to _format_inputs that stores whether or not bounding boxes were ragged when passed in.

Let's convert the return type of `_format_inputs()` to be two dictionaries: the inputs dict, and a metadata dict containing "if inputs are single image", "if bboxes were ragged', etc.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -333,3 +340,36 @@ def _ensure_inputs_are_compute_dtype(self, inputs):
                 self.compute_dtype,
             )
         return inputs
+
+    def _pad_bounding_boxes(self, bounding_boxes):
```
LukeWood, line 344: lets make this  a stable API in `keras_cv/bounding_box` under utilities.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -333,3 +340,36 @@ def _ensure_inputs_are_compute_dtype(self, inputs):
...
+        if not isinstance(bounding_boxes, tf.RaggedTensor):
+            bounding_boxes = tf.RaggedTensor.from_tensor(bounding_boxes)
+        pad_value = tf.cast(-1, dtype=self.compute_dtype)
+        padded_bounding_box = bounding_boxes.to_tensor(pad_value)
+        return padded_bounding_box
+
+    def _filter_padded_boxes(self, bounding_boxes):
+        def is_padded(box):
```
LukeWood, line 352: this is_padded check seems unstable; we should check the class ID and see if it is -1 instead of relying on a logical and like this

LukeWood (changes requested):
Thank you for the PR Divya!  This is coming along great!

LukeWood:
Perhaps a useful example for ragged encoding is https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/object_detection/non_max_suppression.py#L172

keras_cv/bounding_box should have an API to decode a masked tensor to a ragged tensor

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: sounds good!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: done!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: Done

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: look a different approach here like below
```
rt = tf.ragged.constant([[[1,0],[2,0],[3,0]],[[2,0],[0,3]],[[1,0]]])
# convert to dense tensor
dt = rt.to_tensor()
# drop padded
tf.RaggedTensor.from_tensor(dt, lengths=rt.nested_row_lengths())
```

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +291,49 @@ def _augment(self, inputs):
...
     def _batch_augment(self, inputs):
-        return self._map_fn(self._augment, inputs)
+        augmented_batch = self._map_fn(self._augment, inputs)
+        return augmented_batch
 
     def _format_inputs(self, inputs):
         if tf.is_tensor(inputs):
             # single image input tensor
-            return {IMAGES: inputs}, False, False
+            meta_data = {"is_dict": False, "use_targets": False}
```
LukeWood, line 300: this clause should be reformulated as a guard clause:

```
        if tf.is_tensor(inputs):
            # single image input tensor
            meta_data = {"is_dict": False, ... other_keys: false}
            inputs = {IMAGES: inputs}
            return inputs, metadata

       if ...

Comment (LukeWood) [inline keras_cv/layers/preprocessing/base_image_augmentation_layer.py:300]: nit: metadata is a singular word, so meta_data -> metadata

File: keras_cv/layers/preprocessing/base_image_augmentation_layer.py [hunk 6]
```diff
@@ -293,33 +291,49 @@ def _augment(self, inputs):
...
+            inputs = {IMAGES: inputs}
```
LukeWood, line 301: lets define metadata above the if statements.

```
if not isinstance(inputs, dict):
    raise ValueError(
                f"Expect the inputs to be image tensor or dict. Got {inputs}"
     )
metadata = {"is_dict": True"}

if TARGETS in inputs:
     ....
     metadata["use_targets"] = True

if BOUNDING_BOXES in inputs:
    ...

return inputs metadata

File: keras_cv/layers/preprocessing/base_image_augmentation_layer.py [hunk 7]
```diff
@@ -293,33 +291,49 @@ def _augment(self, inputs):
...
         else:
             raise ValueError(
                 f"Expect the inputs to be image tensor or dict. Got {inputs}"
             )
+        if BOUNDING_BOXES in inputs:
+            if isinstance(inputs[BOUNDING_BOXES], tf.RaggedTensor):
+                meta_data["is_bounding_boxes_ragged"] = True
+                meta_data["ragged_row_lengths"] = inputs[
+                    BOUNDING_BOXES
+                ].nested_row_lengths()
```
LukeWood, line 318: one caveat to this approach: what if the number of bounding boxes changes in the layer?  For example, RandomCrop() will remove some bounding boxes

This may be a point in favor of shipping a utility to go from padded Tensor -> ragged

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +291,49 @@ def _augment(self, inputs):
...
+                inputs[BOUNDING_BOXES] = inputs[BOUNDING_BOXES].to_tensor()
```
LukeWood, line 319: For consistency across the codebase: to_tensor(default_value=-1)

This makes class_id=-1 which is indicative that it is a padded sentinel

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +291,49 @@ def _augment(self, inputs):
...
+                meta_data["is_bounding_boxes_ragged"] = False
+        return inputs, meta_data
 
-    def _format_output(self, output, is_dict, use_targets):
-        if not is_dict:
+    def _format_output(self, output, meta_data):
+        if not meta_data["is_dict"]:
             return output[IMAGES]
-        elif use_targets:
+        elif meta_data["use_targets"]:
```
LukeWood, line 327: lets split these checks for readability:

```
if not metadata["is_dict"]:
...

if metadata["use_targets"]:
   ....

return output
```

When using a guard clause (an early if with a return statement) it is easier to mentally follow if the following statements are followed by a newline and a new set of ifs

LukeWood (changes requested):
Thank you @divyashreepathihalli !  This is a great effort, few minor comments.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: Do you think adding a zero area filter would be good? And augmentations never drop any boxes, they would just get clipped to 0?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: done

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: irrespectively it would still end up returning the same values from my understanding. It would not hit any other if conditions and it would return the same values.
is there any other reason?

divyashreepathihalli: done!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: done

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: done!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: done!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: done

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: done!

keras_cv/bounding_box/utils.py
```diff
@@ -8,7 +8,7 @@
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either exness or implied.
```
LukeWood, line 11: make sure to revert this spurious change

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,44 @@ def _format_outputs(boxes, squeeze):
     if squeeze:
         return tf.squeeze(boxes, axis=0)
     return boxes
+
+
+def pad_with_sentinels(bounding_boxes):
```
LukeWood, line 92: we will need to write a docstring for this, describing why we need it and how to use it.  This is actually a user facing API.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,44 @@ def _format_outputs(boxes, squeeze):
...
+    # padded ragged bounding box tensor with -1s to create a tensor
+    # This enables vectorized map execution
+    return bounding_boxes.to_tensor(-1)
+
+
+def filter_sentinels(bounding_boxes):
```
LukeWood, line 98: same here, docstring.  we should document what bounding_boxes is (a -1 padded tensor with axis 5 representing class), and document exactly what this does.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,44 @@ def _format_outputs(boxes, squeeze):
...
+    # filter the padded bounding boxes
+    def is_sentinel(box):
+        return tf.cond(box[4] == -1, lambda: False, lambda: True)
```
LukeWood, line 101: I think you can remove the tf.cond and just write box[4] == -1

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,44 @@ def _format_outputs(boxes, squeeze):
...
+
+    def drop_padded_boxes(bounding_boxes):
+        mask = tf.map_fn(is_sentinel, bounding_boxes, fn_output_signature=tf.bool)
```
LukeWood, line 104: can you replace this entire map_fn() call with `bounding_boxes[:, 4] == -1`?

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,44 @@ def _format_outputs(boxes, squeeze):
...
+        filtered_bounding_boxes = tf.boolean_mask(
+            bounding_boxes.to_tensor(), mask, axis=0
+        )
+        return tf.RaggedTensor.from_tensor(filtered_bounding_boxes)
+
+    return tf.map_fn(drop_padded_boxes, bounding_boxes)
+
+
+def pad_with_class_id(bounding_boxes):
```
LukeWood, line 113: lets write a docstring, lets also add a parameter for custom class_ids.

keras_cv/bounding_box/utils_test.py
```diff
@@ -38,3 +38,29 @@ def test_clip_to_image(self):
         )
         self.assertAllGreaterEqual(bboxes_out, 0)
         self.assertAllLessEqual(bboxes_out, 1)
+
+    def test_pad_with_sentinels(self):
+        bounding_boxes = tf.ragged.constant(
+            [[[1, 2, 3, 4, 5], [1, 2, 3, 4, 5]], [[1, 2, 3, 4, 5]]]
+        )
+        padded_bounding_boxes = bounding_box.pad_with_sentinels(bounding_boxes)
+        expected_output = bounding_boxes.to_tensor(-1)
```
LukeWood, line 47: instead of computing this we should actually code it out:

```
expected_output  = tf.constant(
            [[[1, 2, 3, 4, 5], [1, 2, 3, 4, 5]], [[1, 2, 3, 4, 5], [-1, -1, -1, -1, -1]]
)
```

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +292,61 @@ def _augment(self, inputs):
         return result
 
     def _batch_augment(self, inputs):
-        return self._map_fn(self._augment, inputs)
+        augmented_batch = self._map_fn(self._augment, inputs)
```
LukeWood, line 295: any reason to add this local variable here?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +292,61 @@ def _augment(self, inputs):
...
+        return augmented_batch
 
     def _format_inputs(self, inputs):
+        metadata = {"is_dict": False, "use_targets": False}
```
LukeWood, line 299: should we define all of our keys up top?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +292,61 @@ def _augment(self, inputs):
...
         if tf.is_tensor(inputs):
             # single image input tensor
-            return {IMAGES: inputs}, False, False
+            metadata = {"is_dict": False, "use_targets": False}
+            inputs = {IMAGES: inputs}
+            return inputs, metadata
```
LukeWood, line 304: reformulate the tf.is_tensor call as a guard clause.

All this means is make the next statement an `if`, and add a newline before it.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
LukeWood, line 299: is_dict should probably default to true so we don't have to rewrite it everywhere

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +292,61 @@ def _augment(self, inputs):
...
-        return self._map_fn(self._augment, inputs)
+        augmented_batch = self._map_fn(self._augment, inputs)
+        return augmented_batch
 
     def _format_inputs(self, inputs):
+        metadata = {"is_dict": False, "use_targets": False}
         if tf.is_tensor(inputs):
             # single image input tensor
-            return {IMAGES: inputs}, False, False
+            metadata = {"is_dict": False, "use_targets": False}
```
LukeWood, line 302: lets overwrite keys in metadata instead re-assigning to it.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +292,61 @@ def _augment(self, inputs):
...
+            inputs = {IMAGES: inputs}
+            return inputs, metadata
         elif isinstance(inputs, dict) and TARGETS in inputs:
             # TODO(scottzhu): Check if it only contains the valid keys
             inputs[LABELS] = inputs[TARGETS]
             del inputs[TARGETS]
-            return inputs, True, True
+            metadata = {"is_dict": True, "use_targets": True}
```
LukeWood, line 309: same here:

```
metadata["use_targets"] = True
```

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +292,61 @@ def _augment(self, inputs):
...
-            return inputs, True, False
+            metadata = {"is_dict": True, "use_targets": False}
         else:
             raise ValueError(
                 f"Expect the inputs to be image tensor or dict. Got {inputs}"
             )
+        if BOUNDING_BOXES in inputs:
+            if isinstance(inputs[BOUNDING_BOXES], tf.RaggedTensor):
+                metadata["is_bounding_boxes_ragged"] = True
+                metadata["ragged_row_lengths"] = inputs[
```
LukeWood, line 319: do we still need the row lengths?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +292,61 @@ def _augment(self, inputs):
...
+                    BOUNDING_BOXES
+                ].nested_row_lengths()
+                inputs[BOUNDING_BOXES] = bounding_box.pad_with_sentinels(
+                    inputs[BOUNDING_BOXES]
+                )
+            else:
+                metadata["is_bounding_boxes_ragged"] = False
+            if tf.shape(inputs[BOUNDING_BOXES])[-1] <5:
```
LukeWood, line 327: lets add this above the other checks

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -293,33 +292,61 @@ def _augment(self, inputs):
...
-            return inputs, True, True
+            metadata = {"is_dict": True, "use_targets": True}
         elif isinstance(inputs, dict):
-            return inputs, True, False
+            metadata = {"is_dict": True, "use_targets": False}
         else:
             raise ValueError(
                 f"Expect the inputs to be image tensor or dict. Got {inputs}"
             )
+        if BOUNDING_BOXES in inputs:
```
LukeWood, line 316: maybe bounding_boxes needs a helper method: `_format_bounding_boxes()`

```
bounding_boxes, updates = self._format_bounding_boxes(inputs["bounding_boxes"])
metadata.update(updates)
```

keras_cv/layers/preprocessing/base_image_augmentation_layer_test.py
```diff
@@ -146,7 +146,7 @@ def test_augment_leaves_batched_extra_dict_entries_unmodified(self):
     def test_augment_image_and_localization_data(self):
```
LukeWood, line 146: lets write a unit test to make sure it raises an error when the class axis is missing.

LukeWood (changes requested):
Thanks for the updates to the utilities!  Looking good functionality wise!  Left some minor code quality comments

keras_cv/bounding_box/utils.py
divyashreepathihalli: Done!

keras_cv/bounding_box/utils.py
divyashreepathihalli: Done!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: Done!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: I am still keeping it since it reduces computations in filtering. If we drop the padded boxes before hand, filtering will only take care of the boxes which were not dropped because of other bounding boxes being dropped during augmentataions.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: done!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: I am still keeping it since it reduces computations in filtering. If we drop the padded boxes before hand, filtering will only take care of the boxes which were not dropped because of other bounding boxes being dropped during augmentations.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: Done!

keras_cv/bounding_box/utils.py
divyashreepathihalli: Done

keras_cv/bounding_box/utils.py
divyashreepathihalli: done!

keras_cv/layers/preprocessing/base_image_augmentation_layer_test.py
divyashreepathihalli: done!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: Done!

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
         return tf.squeeze(boxes, axis=0)
     return boxes
+
+
+def pad_with_sentinels(bounding_boxes):
+    """Pads the given bounding boxes to convert it from a ragged tensor to
+    a regular tensor.
+
+    Args:
+      bounding_boxes: a ragged tensor of bounding boxes in 'corners' format.
```
LukeWood, line 97: does it have to be in any specific format?

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
     if squeeze:
         return tf.squeeze(boxes, axis=0)
     return boxes
+
+
+def pad_with_sentinels(bounding_boxes):
+    """Pads the given bounding boxes to convert it from a ragged tensor to
```
LukeWood, line 93: Docstrings should always start with a one line summary

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
+    a regular tensor.
```
LukeWood, line 94: Pads the given bounding box tensor with -1s.  This is done to convert RaggedTensors to be converted into standard Dense tensors, which have better performance and compatibility within the TensorFlow ecosystem.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
+      Can be batched or unbatched.
+
+    Returns:
+      bounding_boxes_tensor: a tensor containing the -1 padded bounding boxes.
+    """
+    return bounding_boxes.to_tensor(-1)
+
+
+def filter_sentinels(bounding_boxes):
+    """filters the -1 padded bounding boxes to convert it from a regular
```
LukeWood, line 107: converts a Dense padded bounding box `Tensor` to a `tf.RaggedTensor`.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
+
+    Args:
+      bounding_boxes: a tensor of bounding boxes in 'corners' format.
+      Can be batched or unbatched.
+
+    Returns:
+      bounding_boxes_tensor: a ragged tensor containing filtered bounding boxes.
+    """
+    if isinstance(bounding_boxes, tf.Tensor):
+        tf.RaggedTensor.from_tensor(bounding_boxes)
```
LukeWood, line 118: this is a no-op, there is no assignment here

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
+    return bounding_boxes.to_tensor(-1)
+
+
+def filter_sentinels(bounding_boxes):
+    """filters the -1 padded bounding boxes to convert it from a regular
+       tensor to a ragged tensor.
+
+    Args:
+      bounding_boxes: a tensor of bounding boxes in 'corners' format.
+      Can be batched or unbatched.
```
LukeWood, line 112: missing indentation

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
+    """
+    return bounding_boxes.to_tensor(-1)
+
+
+def filter_sentinels(bounding_boxes):
+    """filters the -1 padded bounding boxes to convert it from a regular
+       tensor to a ragged tensor.
+
+    Args:
+      bounding_boxes: a tensor of bounding boxes in 'corners' format.
```
LukeWood, line 111: is the format important here?

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
+
+    Returns:
+      bounding_boxes_tensor: a tensor containing the -1 padded bounding boxes.
+    """
+    return bounding_boxes.to_tensor(-1)
+
+
+def filter_sentinels(bounding_boxes):
+    """filters the -1 padded bounding boxes to convert it from a regular
+       tensor to a ragged tensor.
```
LukeWood, line 108: Provide some insight into how this process works (by using the class ID)

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
+    def drop_padded_boxes(bounding_boxes):
+        bounding_boxes = bounding_boxes.to_tensor()
+        mask = bounding_boxes[:, 4] != -1
+        filtered_bounding_boxes = tf.boolean_mask(bounding_boxes, mask, axis=0)
+        return tf.RaggedTensor.from_tensor(filtered_bounding_boxes)
+
+    return tf.map_fn(drop_padded_boxes, bounding_boxes)
+
+
+def pad_with_class_id(bounding_boxes, class_id=0):
```
LukeWood, line 129: maybe `add_class_id`.

This utility is less clear to users, so lets also include a sample `Usage:` section to the docstring.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
+    Args:
+      bounding_boxes: a tensor of bounding boxes in 'corners' format.
+        Can be batched or unbatched.
+      class_id: The value of class id that needs to be padded.
+        The default value is 0
+
+    Returns:
+      bounding_boxes_tensor: a tensor containing class id padded bounding boxes.
+    """
+    if isinstance(bounding_boxes, tf.RaggedTensor):
```
LukeWood, line 141: lets reformulate to not use if else, this can be handled by a single if:

```
is_ragged = isinstance(bounding_boxes, tf.RaggedTensor)

if is_ragged:
    row_lengths = ...
    bounding_boxes = bounding_boxes.to_tensor

... rest of the code.
```


As a general rule of thumb, if else statements are harder to read than if statements.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,73 @@ def _format_outputs(boxes, squeeze):
...
+        row_lengths = list(bounding_boxes.nested_row_lengths())
+        row_lengths[1] = row_lengths[1] + 1
+        dense_bounding_boxes = bounding_boxes.to_tensor()
+    else:
+        dense_bounding_boxes = bounding_boxes
```
LukeWood, line 146: Its a good habit to add newlines to indicate logical blocks.  In this case, one for the ragged input formatting, one for the actual padding then one for the output formatting.

LukeWood (changes requested):
Thanks for the updates, left some docstring fixes.

keras_cv/bounding_box/utils.py
divyashreepathihalli: yeah that is unnecessary. Removed it.

keras_cv/bounding_box/utils.py
divyashreepathihalli: Noted.

keras_cv/bounding_box/utils.py
divyashreepathihalli: updated!

keras_cv/bounding_box/utils.py
divyashreepathihalli: fixed it.

keras_cv/bounding_box/utils.py
divyashreepathihalli: removed it.

keras_cv/bounding_box/utils.py
divyashreepathihalli: added.

keras_cv/bounding_box/utils.py
divyashreepathihalli: noted.

keras_cv/bounding_box/utils.py
divyashreepathihalli: Noted!

LukeWood:
awesome job on this round of PR comments Divya!  This is looking great!

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,91 @@ def _format_outputs(boxes, squeeze):
     if squeeze:
         return tf.squeeze(boxes, axis=0)
     return boxes
+
+
+def pad_with_sentinels(bounding_boxes):
+    """Pads the given bounding box tensor with -1s.
+
+       This is done to convert RaggedTensors to be converted into
```
LukeWood, line 95: there are some indentation issues across all of the docstrings.  Make sure to match the python code indentation in your docstrings.

LukeWood, line 95: unindent this block or it will render as a quote

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,91 @@ def _format_outputs(boxes, squeeze):
...
+       standard Dense tensors, which have better performance and
+       compatibility within the TensorFlow ecosystem.
+
+    Args:
+      bounding_boxes: a ragged tensor of bounding boxes.
```
LukeWood, line 100: match python indentation here

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,91 @@ def _format_outputs(boxes, squeeze):
...
+      bounding_boxes_tensor: a tensor containing the -1 padded bounding boxes.
+    """
+    return bounding_boxes.to_tensor(-1)
+
+
+def filter_sentinels(bounding_boxes):
+    """converts a Dense padded bounding box `Tensor` to a `tf.RaggedTensor`.
+
+    Args:
+      bounding_boxes: a tensor of bounding boxes.
```
LukeWood, line 113: same indentation comment

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,91 @@ def _format_outputs(boxes, squeeze):
...
+      Can be batched or unbatched.
+
+    Returns:
+      bounding_boxes_tensor: a ragged tensor containing filtered bounding boxes.
```
LukeWood, line 117: you can drop the return name and just say: "a ragged Tensor containing the filtered bounding boxes"

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,91 @@ def _format_outputs(boxes, squeeze):
...
+        filtered_bounding_boxes = tf.boolean_mask(bounding_boxes, mask, axis=0)
+        return tf.RaggedTensor.from_tensor(filtered_bounding_boxes)
+
+    return tf.map_fn(drop_padded_boxes, bounding_boxes)
+
+
+def add_class_id(bounding_boxes, class_id=0):
+    """Add class ID to the innermost Tensor or RaggedTensor representing bounding boxes.
+
+       Bounding box utilities in Keras_CV expects bounding boxes to have class IDs
```
LukeWood, line 134: same indentation suggestions

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,91 @@ def _format_outputs(boxes, squeeze):
...
+       along with bounding box cordinates. This utility adds a class ID to the
+       innermost tensor representing the bounding boxes.
+
+       Usage example:
+        bounding_boxes = tf.random.uniform(shape=[2, 2, 4])
+        bounding_boxes_with_class_id = keras_cv.bounding_box.add_class_id(
```
LukeWood, line 140: Usage is a weird case of indentation, you have to unindent the code block, so it looks like:

Usage:
```python
...

File: keras_cv/bounding_box/utils.py [hunk 43]
```diff
@@ -87,3 +87,91 @@ def _format_outputs(boxes, squeeze):
...
+      bounding_boxes: a tensor of bounding boxes.
+        Can be batched or unbatched.
+      class_id: The value of class id that needs to be padded.
+        The default value is 0
+
+    Returns:
+      bounding_boxes_tensor: a tensor containing class id padded bounding boxes.
+    """
+    # format input bounding boxes
+    is_ragged = isinstance(bounding_boxes, tf.RaggedTensor)
```
LukeWood, line 154: theres no need for the variable `dense_bounding_boxes`, we can just override the `bounding_boxes` in the `if_ragged` check.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,91 @@ def _format_outputs(boxes, squeeze):
...
+    if is_ragged:
+        row_lengths = list(bounding_boxes.nested_row_lengths())
+        row_lengths[1] = row_lengths[1] + 1
+        dense_bounding_boxes = bounding_boxes.to_tensor()
+    else:
+        dense_bounding_boxes = bounding_boxes
+
+    # pad input bounding boxes
+    paddings = tf.constant([[0, 0], [0, 0], [0, 1]])
```
LukeWood, line 163: we actually need to dynamically compute paddings - because...

```
boxes = tf.random.uniform((8, 10, 5))
boxes = add_class_id(boxes)
```

should be a no-op in add_class_id

keras_cv/bounding_box/utils_test.py
```diff
@@ -38,3 +38,34 @@ def test_clip_to_image(self):
...
+
+    def test_filter_sentinels(self):
+        bounding_boxes = tf.ragged.constant(
+            [[[1, 2, 3, 4, 5], [1, 2, 3, 4, -1]], [[1, 2, 3, 4, 5]]]
+        )
+        filtered_bounding_boxes = bounding_box.filter_sentinels(bounding_boxes)
+        expected_output = tf.ragged.constant([[[1, 2, 3, 4, 5]], [[1, 2, 3, 4, 5]]])
+        self.assertAllEqual(filtered_bounding_boxes, expected_output)
+
+    def test_pad_with_class_id(self):
```
LukeWood, line 63: lets add a unit test covering the case where `class_id` is already present

LukeWood (changes requested):
Minor changes, great job on this!

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,90 @@ def _format_outputs(boxes, squeeze):
...
+
+    Returns:
+        `tf.Tensor` with an additional class id padded to the original bounding boxes.
+    """
+    # format input bounding boxes
+    is_ragged = isinstance(bounding_boxes, tf.RaggedTensor)
+
+    if is_ragged:
+        row_lengths = list(bounding_boxes.nested_row_lengths())
+        row_lengths[1] = row_lengths[1] + 1
```
LukeWood, line 158: lets add a comment for this.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,90 @@ def _format_outputs(boxes, squeeze):
...
+    paddings = tf.constant([[0, 0], [0, 0], [0, 1]])
+    bounding_boxes = tf.pad(
+        bounding_boxes,
+        paddings=paddings,
+        mode="CONSTANT",
+        constant_values=class_id,
+    )
+
+    # format output bounding boxes
+    if isinstance(bounding_boxes, tf.RaggedTensor):
```
LukeWood, line 171: if isinstance(bounding_boxes, tf.RaggedTensor):
should be

if is_ragged:

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,90 @@ def _format_outputs(boxes, squeeze):
...
+    # format input bounding boxes
+    is_ragged = isinstance(bounding_boxes, tf.RaggedTensor)
+
+    if is_ragged:
+        row_lengths = list(bounding_boxes.nested_row_lengths())
+        row_lengths[1] = row_lengths[1] + 1
+        bounding_boxes = bounding_boxes.to_tensor()
+
+    # pad input bounding boxes
+    paddings = tf.constant([[0, 0], [0, 0], [0, 1]])
```
LukeWood, line 162: we want to support unbatched inputs too

LukeWood, line 162: Lets add a test case for unbatched inputs

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,90 @@ def _format_outputs(boxes, squeeze):
...
+        `tf.Tensor` with an additional class id padded to the original bounding boxes.
+    """
+    # format input bounding boxes
+    is_ragged = isinstance(bounding_boxes, tf.RaggedTensor)
+
+    if is_ragged:
+        row_lengths = list(bounding_boxes.nested_row_lengths())
+        row_lengths[1] = row_lengths[1] + 1
+        bounding_boxes = bounding_boxes.to_tensor()
+
```
LukeWood, line 160: lets perform input validation checking that the last axis is of shape [4]

keras_cv/bounding_box/utils_test.py
LukeWood: (the new unit test should error out)

keras_cv/bounding_box/utils_test.py
```diff
@@ -38,3 +38,34 @@ def test_clip_to_image(self):
...
+        padded_bounding_boxes = bounding_box.pad_with_sentinels(bounding_boxes)
+        expected_output = tf.constant(
+            [
+                [[1, 2, 3, 4, 5], [1, 2, 3, 4, 5]],
+                [[1, 2, 3, 4, 5], [-1, -1, -1, -1, -1]],
+            ]
+        )
+        self.assertAllEqual(padded_bounding_boxes, expected_output)
+
+    def test_filter_sentinels(self):
```
LukeWood, line 55: can we add a few more test cases with more complex inputs/outputs for filter sentinels?  some more differing shapes would be good.

LukeWood (changes requested):

keras_cv/bounding_box/utils_test.py
divyashreepathihalli: added

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,108 @@ def _format_outputs(boxes, squeeze):
...
+    Returns:
+        `tf.RaggedTensor` containing the filtered bounding boxes.
+    """
+    if isinstance(bounding_boxes, tf.Tensor):
+        bounding_boxes = tf.RaggedTensor.from_tensor(bounding_boxes)
+
+    def drop_padded_boxes(bounding_boxes):
+        bounding_boxes = bounding_boxes.to_tensor()
+        mask = bounding_boxes[:, 4] != -1
```
tanzhenyu, line 123: Check the bounding_boxes shape before calling `bounding_boxes[:, 4]`?

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,108 @@ def _format_outputs(boxes, squeeze):
...
+        filtered_bounding_boxes = tf.boolean_mask(bounding_boxes, mask, axis=0)
+        return tf.RaggedTensor.from_tensor(filtered_bounding_boxes)
+
+    return tf.map_fn(drop_padded_boxes, bounding_boxes)
+
+
+def add_class_id(bounding_boxes, class_id=0):
+    """Add class ID to the innermost Tensor or RaggedTensor representing bounding boxes.
+
+    Bounding box utilities in Keras_CV expects bounding boxes to have class IDs
```
tanzhenyu, line 133: Just a general comment here: do we expect this happen if there's ragged tensor is not supported?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -296,30 +295,62 @@ def _batch_augment(self, inputs):
...
-        elif isinstance(inputs, dict):
-            return inputs, True, False
-        else:
+            metadata["use_targets"] = True
+            return inputs, metadata
+
+        return inputs, metadata
+
+    def _format_bounding_boxes(self, bounding_boxes):
+        metadata = {"ragged_bounding_boxes": False}
```
tanzhenyu, line 326: prefer using a global const like `RAGGED_BOUNDING_BOXES`

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -296,30 +295,62 @@ def _batch_augment(self, inputs):
...
+        if isinstance(bounding_boxes, tf.RaggedTensor):
+            metadata["ragged_row_lengths"] = bounding_boxes.nested_row_lengths()
```
tanzhenyu, line 328: same as above

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -156,6 +156,9 @@ def augment_bounding_boxes(
         image_shape = tf.shape(image)
         h = image_shape[H_AXIS]
         w = image_shape[W_AXIS]
+        _, _, _, _, rest = tf.split(
```
tanzhenyu, line 159: What is expected in `rest`?

keras_cv/layers/preprocessing/random_shear_test.py
```diff
@@ -73,7 +74,7 @@ def test_return_shapes(self):
     def test_single_image_input(self):
         """test for single image input"""
         xs = tf.ones((512, 512, 3))
-        ys = tf.ones(shape=(5, 4))
+        ys = tf.ones(shape=(5, 5))
```
tanzhenyu, line 77: So we always require users to call `add_class_id` before calling aug layers?

tanzhenyu (commented):
drive-by comments.
Thanks for the PR!

keras_cv/bounding_box/utils.py
divyashreepathihalli: done

keras_cv/bounding_box/utils.py
divyashreepathihalli: done

keras_cv/bounding_box/utils.py
divyashreepathihalli: added

keras_cv/bounding_box/utils_test.py
divyashreepathihalli: done

keras_cv/bounding_box/utils.py
divyashreepathihalli: added a simpler filtering logic and checks for batched and unbatched inputs

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
divyashreepathihalli: done

keras_cv/layers/preprocessing/random_rotation.py
divyashreepathihalli: class_id, etc. It is kept open in case we add more values to it.

keras_cv/bounding_box/utils.py
divyashreepathihalli: done

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,108 @@ def _format_outputs(boxes, squeeze):
...
+
+def pad_with_sentinels(bounding_boxes):
+    """Pads the given bounding box tensor with -1s.
+
+    This is done to convert RaggedTensors to be converted into
+    standard Dense tensors, which have better performance and
+    compatibility within the TensorFlow ecosystem.
+
+    Args:
+        bounding_boxes: a ragged tensor of bounding boxes.
```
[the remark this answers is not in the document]
LukeWood: perhaps:

```
        bounding_boxes: a `tf.RaggedTensor` of bounding boxes.  May be batched or unbatched.
```

This method should work on any dtype I believe, unless the -1 in `default_value` is not auto-casted, but I think it is.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,110 @@ def _format_outputs(boxes, squeeze):
...
+        Can be batched or unbatched.
+
+    Returns:
+        a Tensor containing the -1 padded bounding boxes.
+    """
+    return bounding_boxes.to_tensor(-1)
```
LukeWood, line 106: lets use the keyword argument here instead of positional.

default_value=-1.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,110 @@ def _format_outputs(boxes, squeeze):
...
+        bounding_boxes: a Tensor of bounding boxes.  May be batched, or unbatched.
+
+    Returns:
+        `tf.RaggedTensor` containing the filtered bounding boxes.
+    """
+    is_ragged = isinstance(bounding_boxes, tf.RaggedTensor)
+    is_batched = len(bounding_boxes.shape) == 3
+    if is_batched:
+        if is_ragged:
+            bounding_boxes = bounding_boxes.to_tensor(-1)
```
LukeWood, line 122: call the utility above for clarity:
```
    bounding_boxes = pad_with_sentinels(bounding_boxes)
```

Good to spread good patterns within our own code base!

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,110 @@ def _format_outputs(boxes, squeeze):
...
+
+    Args:
+        bounding_boxes: a Tensor of bounding boxes.  May be batched, or unbatched.
+
+    Returns:
+        `tf.RaggedTensor` containing the filtered bounding boxes.
+    """
+    is_ragged = isinstance(bounding_boxes, tf.RaggedTensor)
+    is_batched = len(bounding_boxes.shape) == 3
+    if is_batched:
```
LukeWood, line 120: can we reduce the nested conditionals in any way?  Do we need a special case for `is_batched`?

I believe that `mask = bounding_boxes[..., 4] != -1` should handle all cases. Correct me if that is incorrect!

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,110 @@ def _format_outputs(boxes, squeeze):
...
+    is_ragged = isinstance(bounding_boxes, tf.RaggedTensor)
+
+    if is_ragged:
+        row_lengths = list(bounding_boxes.nested_row_lengths())
+        # increase row length to account for clas-id addition
+        row_lengths[1] = row_lengths[1] + 1
+        bounding_boxes = bounding_boxes.to_tensor()
+
+    # pad input bounding boxes
+    if bounding_boxes.shape[-1] > 4:
```
LukeWood, line 165: any reason to not check that it is equal to 4 here?  > doesn't catch cases of [0, 1, 2 ,3]

LukeWood (changes requested):
Looks great!  Are all test cases passing?

keras_cv/bounding_box/utils.py
divyashreepathihalli: This would add a default class ID to the input bounding box. which can be a Tensor or a RaggedTensor.

keras_cv/layers/preprocessing/random_shear_test.py
divyashreepathihalli: if they don't have a class ID already, then yes!

keras_cv/bounding_box/utils.py
divyashreepathihalli: I checked with your regarding this. The bounding_boxes could get more than 5 values.

keras_cv/bounding_box/utils.py
divyashreepathihalli: Ahh thanks for the tip! this way I changed the output to return either tensor or ragged tensor.

keras_cv/bounding_box/utils.py
divyashreepathihalli: yess! updated.

keras_cv/bounding_box/utils.py
divyashreepathihalli: added.

keras_cv/bounding_box/utils.py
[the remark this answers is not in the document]
divyashreepathihalli: done

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+def pad_with_sentinels(bounding_boxes, default_value=-1):
+    """Pads the given bounding box tensor with -1s.
+
+    This is done to convert RaggedTensors to be converted into
+    standard Dense tensors, which have better performance and
+    compatibility within the TensorFlow ecosystem.
+
+    Args:
+        bounding_boxes: a ragged tensor of bounding boxes.
+        Can be batched or unbatched.
```
LukeWood, line 101: need to indent the second line or rendering will be broken

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+
+    Returns:
+        a Tensor containing the -1 padded bounding boxes.
+    """
+    return bounding_boxes.to_tensor(default_value)
```
LukeWood, line 106: what I mean't before was:

```
    return bounding_boxes.to_tensor(default_value=-1)
```

to_tensor has a named argument `default_value`.  

Instead of default_value, lets call our argument `sentinel_value=-1`.

```
def pad_with_sentinels(bounding_boxes, sentinel_value=-1):
    return bounding_boxes.to_tensor(default_value=sentinel_value)
```

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+
+
+def filter_sentinels(bounding_boxes, default_value=-1):
```
LukeWood, line 109: same here:
```
def filter_sentinels(bounding_boxes, sentinel_value=-1):
```

LukeWood: bounding_boxes, sentinel_value=sentinel_value

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+    Bounding box utilities in Keras_CV expects bounding boxes to have class IDs
+    along with bounding box cordinates. This utility adds a class ID to the
+    innermost tensor representing the bounding boxes.
+
+    Usage:
+    ```python
+    bounding_boxes = tf.random.uniform(shape=[2, 2, 4])
+    bounding_boxes_with_class_id = keras_cv.bounding_box.add_class_id(
+                                    bounding_boxes, class_id=1)
+    bounding_boxes_with_class_id is a Tensor of shape [2, 2, 5]
```
LukeWood, line 140: this renders as a code block, so put a `#` in front of this line to indicate a comment.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
     if squeeze:
         return tf.squeeze(boxes, axis=0)
     return boxes
+
+
+def pad_with_sentinels(bounding_boxes, default_value=-1):
+    """Pads the given bounding box tensor with -1s.
+
+    This is done to convert RaggedTensors to be converted into
```
LukeWood, line 95: Update to:

    This is done to convert RaggedTensors into
    standard Dense tensors, which have better performance and
    compatibility within the TensorFlow ecosystem.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+        Can be batched or unbatched.
+
+    Returns:
+        a Tensor containing the -1 padded bounding boxes.
+    """
+    return bounding_boxes.to_tensor(default_value)
+
+
+def filter_sentinels(bounding_boxes, default_value=-1):
+    """converts a Dense padded bounding box `Tensor` to a `tf.RaggedTensor`.
```
LukeWood, line 110: """converts a Dense padded bounding box `tf.Tensor` to a `tf.RaggedTensor`.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+
+    Args:
+        bounding_boxes: a Tensor of bounding boxes.  May be batched, or unbatched.
```
LukeWood, line 113: missing `sentinel_value` from Args list

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+
+def pad_with_sentinels(bounding_boxes, default_value=-1):
+    """Pads the given bounding box tensor with -1s.
+
+    This is done to convert RaggedTensors to be converted into
+    standard Dense tensors, which have better performance and
+    compatibility within the TensorFlow ecosystem.
+
+    Args:
+        bounding_boxes: a ragged tensor of bounding boxes.
```
LukeWood, line 100: missing default_value from args list, replace with sentinel_value and update

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+        bounding_boxes = bounding_box.pad_with_sentinels(
+            bounding_boxes, default_value=default_value
+        )
+    mask = bounding_boxes[..., 4] != default_value
+    filtered_bounding_boxes = tf.ragged.boolean_mask(bounding_boxes, mask)
+    return filtered_bounding_boxes
+
+
+def add_class_id(bounding_boxes, class_id=0):
+    """Add class ID to the innermost Tensor or RaggedTensor representing bounding boxes.
```
LukeWood, line 129: """Add class ID to a new dimension of the final axis of a bounding box Tensor.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+
+    Bounding box utilities in Keras_CV expects bounding boxes to have class IDs
```
LukeWood, line 131: Bounding box utilities in KerasCV expect bounding boxes to have class IDs. 
This utility adds a class ID to a new axis of the provided `tf.Tensor`.

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+    if is_ragged:
+        row_lengths = list(bounding_boxes.nested_row_lengths())
+        # increase row length to account for clas-id addition
+        row_lengths[1] = row_lengths[1] + 1
+        bounding_boxes = bounding_boxes.to_tensor()
+
+    # pad input bounding boxes
+    if bounding_boxes.shape[-1] != 4:
+        raise ValueError(
+            "The number of values along the bounding box axis is "
```
LukeWood, line 163: "The number of values along the final axis of `bounding_boxes` is "
            "expected to be 4. But got {}.".format(bounding_boxes.shape[-1])

keras_cv/bounding_box/utils.py
```diff
@@ -87,3 +87,104 @@ def _format_outputs(boxes, squeeze):
...
+            "expected to be 4. But got {}.".format(bounding_boxes.shape[-1])
+        )
+    bounding_box_rank = len(tf.shape(bounding_boxes))
+    if bounding_box_rank == 2:
+        paddings = tf.constant([[0, 0], [0, 1]])
+    elif bounding_box_rank == 3:
+        paddings = tf.constant([[0, 0], [0, 0], [0, 1]])
+    else:
+        raise ValueError(
```
LukeWood, line 172: prefer argument names in error messages:

```
        raise ValueError(
            "`bounding_boxes` should be of rank 2 or 3. However "
            f"add_class_id received `rank(bounding_boxes)={bounding_box_rank}`"
        )
```

LukeWood, line 172: use fstrings - suggestion has an fstring in place

LukeWood (changes requested):
changes to util API

keras_cv/bounding_box/utils.py
divyashreepathihalli: done

keras_cv/bounding_box/utils.py
divyashreepathihalli: updated

keras_cv/bounding_box/utils.py
divyashreepathihalli: updated

keras_cv/bounding_box/utils.py
divyashreepathihalli: updated

keras_cv/bounding_box/utils.py
divyashreepathihalli: updated

keras_cv/bounding_box/utils.py
divyashreepathihalli: done

keras_cv/bounding_box/utils.py
LukeWood: @divyashreepathihalli I see the review request is back but it looks like these 3 issues are still unresolved - GitHub can sometimes hide issues in big reviews even if un-resolved, so make sure to click "load hidden comments" before re-requesting!

keras_cv/bounding_box/utils.py
LukeWood: @divyashreepathihalli  and this one - thanks!

keras_cv/bounding_box/utils.py
LukeWood: @divyashreepathihalli this one

keras_cv/bounding_box/utils.py
divyashreepathihalli: ### sorry about that

keras_cv/bounding_box/utils.py
divyashreepathihalli: added

keras_cv/bounding_box/utils.py
divyashreepathihalli: done

keras_cv/bounding_box/utils.py
divyashreepathihalli: done

LukeWood:
Thanks @divyashreepathihalli!  Looks great and the utilities will certainly see a lot of use!

PR #679, opened 2022-08-09
Corrected RegNet implementation

AdityaKane2001:
@LukeWood @sayakpaul 

Corrected RegNets implementation which was causing tests to fail:
- Added input_tensor argument.
- `num_classes` -> `classes`

### DIFF
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -24,13 +24,13 @@ import sys
 import tensorflow as tf
 from absl import flags
 from tensorflow.keras import callbacks
-from tensorflow.keras import layers
 from tensorflow.keras import losses
 from tensorflow.keras import metrics
 from tensorflow.keras import optimizers
 
 import keras_cv
 from keras_cv import models
+from keras_cv.datasets import imagenet
 
 """
 ## Overview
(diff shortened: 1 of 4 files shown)


keras_cv/models/regnet.py
```diff
@@ -0,0 +1,1543 @@
...
+        )
+
+    img_input = utils.parse_model_inputs(input_shape, input_tensor)
+    x = img_input
+
+    if include_rescaling:
+        x = layers.Rescaling(scale=1.0 / 255.0)(x)
+    x = Stem(name=model_name)(x)
+
+    in_channels = 32  # Output from Stem
```
sayakpaul, line 763: Can't it be programmatically calculated?

keras_cv/models/regnet.py
```diff
@@ -0,0 +1,1543 @@
...
+
+    for stage_index in range(4):
```
sayakpaul, line 765: Prefer assigning the magic number to a meaningful variable.

keras_cv/models/regnet.py
```diff
@@ -0,0 +1,1543 @@
...
+            depth,
+            group_width,
+            in_channels,
+            out_channels,
+            name=model_name + "_Stage_" + str(stage_index),
+        )(x)
+        in_channels = out_channels
+
+    if include_top:
+        x = Head(classes=classes, activation=classifier_activation)(x)
```
sayakpaul, line 780: @LukeWood @ianstenbit isn't `num_classes` a better variable name? I know we have been following `classes` for models, but wanted to revisit it.

sayakpaul (changes requested):

keras_cv/models/regnet.py
AdityaKane2001: I agree here.

keras_cv/models/regnet.py
AdityaKane2001: `x.shape[-1]`?

keras_cv/models/regnet.py
LukeWood: Yeah... we have gone back and forth on this a lot of times.

@fchollet had some thoughts on this.

AdityaKane2001:
@sayakpaul @LukeWood 

Made changes from the review.

keras_cv/models/regnet.py
```diff
@@ -202,6 +202,8 @@
     },
 }
 
+NUM_STAGES = 4
```
LukeWood, line 205: Let’s  have this adjacent to the loop as a local variable.

LukeWood:
This is blocked on refactoring model tests to be written as a file per model.  Currently this will make our tests OOM.

PR #680, opened 2022-08-09
Fixes `label_smoothing` in `FocalLoss`

### LINKED ISSUE #672: Possibly wrong implementation of focal loss in regard of label smoothing
I think that there is a wrong implementation of focal loss in regard of label smoothing.

Here is the current implementation
https://github.com/keras-team/keras-cv/blob/35b9f9534fc193e00b4fa850963a05b331467065/keras_cv/losses/focal.py#L74-L90

as seen above, label smoothing occurs *before* the computation of `alpha` and `pt`. From the original paper, the value expected at this stage for `y_true` should be either `0` or `1`. However after label smoothing, `y_true` will be in `]0,1[`, therefore invalidating `alpha` and `pt` computation (at least not as written in the original paper).

Also to compute Multi-Label, I think the final reduction should be a sum to be correct. One alternative would be, as implemented in pytorch to add a `reduction` parameter, to either:

* 'none': perform no class reduction
* 'sum': perform a sum over classes
* 'mean': perform a mean over classes

What do you thin @LukeWood , @quantumalaviya ?

atuleu:
# What does this PR do?

Fixes label smoothing in `FocalLoss`.

Also as a code style, uses the same formulation for `alpha` and `pt` and computation

Fixes #672

## Before submitting
      Pull Request section?
      to it if that's the case.
Yes added a test for precomputed values from : https://docs.google.com/spreadsheets/d/1umGKfnSXzmz5NDeUO-s850ZDbJ_92kyD4MGbwEZGyLQ/edit?usp=sharing

## Who can review?

@LukeWood @quantumalaviya

### DIFF
--- a/keras_cv/losses/focal.py
+++ b/keras_cv/losses/focal.py
@@ -76,14 +76,16 @@ class FocalLoss(tf.keras.losses.Loss):
         y_true = tf.cast(y_true, y_pred.dtype)
 
         if self.label_smoothing:
-            y_true = self._smooth_labels(y_true)
+            y_smoothed = self._smooth_labels(y_true)
+        else:
+            y_smoothed = y_true
 
         if self.from_logits:
             y_pred = tf.nn.sigmoid(y_pred)
 
-        cross_entropy = K.binary_crossentropy(y_true, y_pred)
+        cross_entropy = K.binary_crossentropy(y_smoothed, y_pred)
 
-        alpha = tf.where(tf.equal(y_true, 1.0), self._alpha, (1.0 - self._alpha))
+        alpha = self._alpha * y_true + (1.0 - self._alpha) * (1.0 - y_true)
         pt = y_true * y_pred + (1.0 - y_true) * (1.0 - y_pred)
         loss = alpha * tf.pow(1.0 - pt, self._gamma) * cross_entropy
 
--- a/keras_cv/losses/focal_test.py
+++ b/keras_cv/losses/focal_test.py
@@ -86,3 +86,10 @@ class FocalTest(tf.test.TestCase):
         self.assertAllClose(
             focal_loss_on_logits(y_true, y_logits), focal_loss(y_true, y_pred)
         )
+
+    def test_smoothed_output(self):
+        y_true = [0.0, 1.0, 1.0]
+        y_pred = [0.1, 0.7, 0.9]
+
+        focal_loss = FocalLoss(label_smoothing=0.1)
+        self.assertAllClose(focal_loss(y_true, y_pred), 0.00371020)


keras_cv/losses/focal.py
```diff
@@ -76,14 +76,16 @@ def call(self, y_true, y_pred):
         y_true = tf.cast(y_true, y_pred.dtype)
 
         if self.label_smoothing:
-            y_true = self._smooth_labels(y_true)
+            y_smoothed = self._smooth_labels(y_true)
+        else:
+            y_smoothed = y_true
```
LukeWood, line 81: can we remove y_smoothed and just move smoothing to after we compute alpha?

keras_cv/losses/focal.py
```diff
@@ -76,14 +76,16 @@ def call(self, y_true, y_pred):
...
 
         if self.from_logits:
             y_pred = tf.nn.sigmoid(y_pred)
 
-        cross_entropy = K.binary_crossentropy(y_true, y_pred)
+        cross_entropy = K.binary_crossentropy(y_smoothed, y_pred)
 
-        alpha = tf.where(tf.equal(y_true, 1.0), self._alpha, (1.0 - self._alpha))
+        alpha = self._alpha * y_true + (1.0 - self._alpha) * (1.0 - y_true)
         pt = y_true * y_pred + (1.0 - y_true) * (1.0 - y_pred)
```
LukeWood, line 89: should pt use y_pred or y_smoothed?

LukeWood (changes requested):
Thanks for the PR!  Some clarification is needed

keras_cv/losses/focal.py
atuleu: From the original paper , `y_true`.

keras_cv/losses/focal.py
atuleu: This implementation comes from breaking down another implementation which include possible fixes for #681

In some case, you still want to leave untouched the `y_true` variable, in order to perform other computation at the reduction steps ( see FSAF's implementation of focal loss here https://github.com/xuannianz/FSAF/blob/791ca41187eec25918b2626322c43c9ab7c06f61/losses.py#L22-L60 ). Therefore you want two variable, one for computing cross-entropy, possibly smoothed, and the original `y_true` variable.

With a night to think over, I wonder if the label smoothing should maybe just be left out of this focal loss computation. I think that there is a reason why pytorch uses a formulation with mulitplications instead of a conditional statement. In that case the formula extends to the [0,1] domain. Therefore the label smoothing can be removed from the loss, and implemented outside it. Maybe the solution to clarify label smoothing, would be finally to remove it from the `FocalLoss` all together.

What do you think @LukeWood @quantumalaviya ?

keras_cv/losses/focal.py
LukeWood: you mean y_pred, right?

LukeWood: Oops, y_true - you are correct

keras_cv/losses/focal.py
LukeWood: Sorry, I just meant in terms of code structure:

```
    def call(self, y_true, y_pred):
        y_pred = tf.convert_to_tensor(y_pred)
        y_true = tf.cast(y_true, y_pred.dtype)

        if self.from_logits:
            y_pred = tf.nn.sigmoid(y_pred)

        alpha = self._alpha * y_true + (1.0 - self._alpha) * (1.0 - y_true)
        pt = y_true * y_pred + (1.0 - y_true) * (1.0 - y_pred)
        cross_entropy = K.binary_crossentropy(self._smooth_labels(y_true), y_pred)
        loss = alpha * tf.pow(1.0 - pt, self._gamma) * cross_entropy

        return tf.reduce_mean(loss, axis=-1)
```

If label_smoothing=0, its a no-op anyways so you can just call it directly.

LukeWood (changes requested):
Check out my code suggestion, thanks!

keras_cv/losses/focal.py
tanzhenyu: label smoothing makes sense for focal loss
Please go ahead and follow the CE loss implementation:
https://github.com/keras-team/keras/blob/b80dd12da9c0bc3f569eca3455e77762cf2ee8ef/keras/losses.py#L1978-L1988

keras_cv/losses/focal_test.py
```diff
@@ -86,3 +86,10 @@ def test_from_logits_argument(self):
         self.assertAllClose(
             focal_loss_on_logits(y_true, y_logits), focal_loss(y_true, y_pred)
         )
+
+    def test_smoothed_output(self):
+        y_true = [0.0, 1.0, 1.0]
+        y_pred = [0.1, 0.7, 0.9]
+
+        focal_loss = FocalLoss(label_smoothing=0.1)
+        self.assertAllClose(focal_loss(y_true, y_pred), 0.00371020)
```
tanzhenyu, line 95: please add comment on the magic number

tanzhenyu (commented):
Thanks for the PR!

LukeWood:
This would still be a great fix!  Let me know if you can continue pushing this PR forward !

PR #685, opened 2022-08-12
added random crop to keras_cv

divyashreepathihalli:
random_crop_demo

### DIFF
--- a/keras_cv/__init__.py
+++ b/keras_cv/__init__.py
@@ -30,4 +30,4 @@ from keras_cv.core import FactorSampler
 from keras_cv.core import NormalFactorSampler
 from keras_cv.core import UniformFactorSampler
 
-__version__ = "0.3.4"
+__version__ = "0.3.5"


keras_cv/layers/preprocessing/random_crop.py
```diff
@@ -0,0 +1,132 @@
...
+      seed: Integer. Used to create a random seed.
+    """
+
+    def __init__(self, height, width, seed=None, bounding_box_format=None, **kwargs):
+        super().__init__(**kwargs, autocast=False, seed=seed, force_generator=True)
+        self.height = height
+        self.width = width
+        self.seed = seed
+        self.auto_vectorize = False
+        self.image_shape = None
```
LukeWood, line 68: we shouldn't track image_shape on the class itself.  Instead, we should be passing this as a variable around between methods, or using `tf.shape(image)` to track this.

Using stateful-operations like this could cause some hard to follow behavior for future code maintainers

keras_cv/layers/preprocessing/random_crop.py
```diff
@@ -0,0 +1,132 @@
...
+        else:
+            inputs = self._ensure_inputs_are_compute_dtype(inputs)
+            inputs, metadata = self._format_inputs(inputs)
+            output = inputs
+            # self._resize() returns valid results for both batched and
+            # unbatched
+            output[IMAGES] = self._resize(inputs[IMAGES])
+            if BOUNDING_BOXES in inputs:
+                output[BOUNDING_BOXES] = self._map_fn(
+                    self._smart_resize_bounding_box, inputs[BOUNDING_BOXES]
```
LukeWood, line 84: looks like we still need to implement this method  self._smart_resize_bounding_box

LukeWood (changes requested):
Sorry for the long delay - remind me, are we shipping random crop in a separate PR and adding the bounding box behavior at a later date?

LukeWood:
If so, lets remove the bounding box behavior here so we can merge.

tanzhenyu:
> If so, lets remove the bounding box behavior here so we can merge.

Is this still being worked on?

PR #688, opened 2022-08-15
Add script version to training history script

ianstenbit:
@LukeWood

### DIFF
--- a/examples/training/classification/imagenet/training_history.json
+++ b/examples/training/classification/imagenet/training_history.json
@@ -6,7 +6,10 @@
             },
             "contributor": "ianjjohnson",
             "epochs_trained": 84,
-            "script": "basic_training.py",
+            "script": {
+                "name": "basic_training.py",
+                "version": "90d4c3548a2e989fe52d6cf7ae7439af794f0ae6",
+            }
             "tensorboard_logs": "https://tensorboard.dev/experiment/K5Q0gAk0RayXwP0WsLPpMA/",
             "validation_accuracy": "0.6771"
         }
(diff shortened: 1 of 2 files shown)


examples/training/classification/imagenet/training_history.json
```diff
@@ -18,7 +21,10 @@
             },
             "contributor": "ianjjohnson",
             "epochs_trained": 50,
-            "script": "basic_training.py",
+            "script": {
+                "name": "basic_training.py",
+                "version": "90d4c3548a2e989fe52d6cf7ae7439af794f0ae6",
```
LukeWood, line 26: Should this have a fully qualified path so we can track if it’s moved sometime in the git history?

ianstenbit: With a commit hash, we shouldn't need this, as the commit will point to the historical location of the script.

LukeWood: Perhaps I am missing a reason for the contrary, but what is the benefit of *not* including the path?  Is there a good reason not to include the full path?  One benefit of including it would be if anyone wanted to traverse the results and reproduce the results they could script the entire thing.

ianstenbit: I think the reason to the contrary is that this is the path relative to the location of the training_history.json file.

So if there were a separate model-specific training script for e.g. ResNet in a directory called resenet/, this would say resnet/other_training_script.py

Therefore it's pretty straightforward to script the whole thing without fully-qualified paths, and I think this is much cleaner than including the redundant prefix keras-cv/keras_cv/examples/training/classification/imagenet.

PR #697, opened 2022-08-17
Removed bias from Conv2D if followed by BatchNorm.

### LINKED ISSUE #629: Redundant bias paramter in ResNet's conv2d.
**Short description**
In `tf.keras` by default Conv2D has bias enabled. We should explicitly set `use_bias=False` if the convolution layer is followed by BatchNormalization.

This is the done in some existing Keras-CV implementations:
* DarkNet Conv Block - [link](https://github.com/keras-team/keras-cv/blob/f3e32b5d74d69419c5ec17b04f9ca96807474ed1/keras_cv/models/__internal__/darknet_utils.py#L28)
* DenseNet - [link](https://github.com/keras-team/keras-cv/blob/f3e32b5d74d69419c5ec17b04f9ca96807474ed1/keras_cv/models/densenet.py#L134)
* ResNet V2 - [link](https://github.com/keras-team/keras-cv/blob/f3e32b5d74d69419c5ec17b04f9ca96807474ed1/keras_cv/models/resnet_v2.py#L116)

But not in `resnet_v1` - all ResNet's have [bias enabled](https://github.com/keras-team/keras-cv/blob/f3e32b5d74d69419c5ec17b04f9ca96807474ed1/keras_cv/models/resnet_v1.py#L113), which causes redundant computation and is not aligned with other implementations.

I think this is rather a quick fix - add `use_bias=False` parameter to the Conv2d's in the `Block` function. If you approve I could submit a PR.

sebastian-sz:
# What does this PR do?
Remove bias parameter from ResNetV1 Conv2d if the next layer is batch normalization.
Batch Normalization eliminates the need for bias, as it is in Densenet or ResNetV2. 

Fixes #629 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood @ianstenbit

### DIFF
--- a/keras_cv/models/resnet_v1.py
+++ b/keras_cv/models/resnet_v1.py
@@ -105,7 +105,7 @@ def Block(filters, kernel_size=3, stride=1, conv_shortcut=True, name=None):
     def apply(x):
         if conv_shortcut:
             shortcut = layers.Conv2D(
-                4 * filters, 1, strides=stride, name=name + "_0_conv"
+                4 * filters, 1, strides=stride, use_bias=False, name=name + "_0_conv"
             )(x)
             shortcut = layers.BatchNormalization(
                 axis=BN_AXIS, epsilon=1.001e-5, name=name + "_0_bn"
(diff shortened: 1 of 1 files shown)


LukeWood:
So is this handled by the BatchNorm layers?  Do they have a bias built in?

sebastian-sz:
@LukeWood yes below is a summary from Nvidia's [performance tuning guide](https://nvlabs.github.io/eccv2020-mixed-precision-tutorial/files/szymon_migacz-pytorch-performance-tuning-guide.pdf) that eventually landed in [Pytorch docs](https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html#disable-bias-for-convolutions-directly-followed-by-a-batch-norm)

>Disable bias for convolutions directly followed by a batch norm
>
>If a nn.Conv2d layer is directly followed by a nn.BatchNorm2d layer, then the bias in the convolution is not needed, instead use nn.Conv2d(..., bias=False, ....). Bias is not needed because in the first step BatchNorm subtracts the mean, which effectively cancels out the effect of bias.

LukeWood:
> @LukeWood yes below is a summary from Nvidia's [performance tuning guide](https://nvlabs.github.io/eccv2020-mixed-precision-tutorial/files/szymon_migacz-pytorch-performance-tuning-guide.pdf) that eventually landed in [Pytorch docs](https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html#disable-bias-for-convolutions-directly-followed-by-a-batch-norm)
> 
> > Disable bias for convolutions directly followed by a batch norm
> > If a nn.Conv2d layer is directly followed by a nn.BatchNorm2d layer, then the bias in the convolution is not needed, instead use nn.Conv2d(..., bias=False, ....). Bias is not needed because in the first step BatchNorm subtracts the mean, which effectively cancels out the effect of bias.

@fchollet and compatibility concerns here?  Seems like the move to keras_cv.models is a good time to do this

LukeWood (approved):

LukeWood:
Thank you @sebastian-sz

PR #699, opened 2022-08-17
SimCLR implementation

ianstenbit:
### Example code using this API
```
import keras
from keras import layers

from keras_cv.losses import SimCLRLoss
from keras_cv.models import VGG19
from keras_cv.training import SimCLRAugmenter
from keras_cv.training import SimCLRTrainer

(x_train, y_train), (x_test, y_test) = keras.datasets.cifar10.load_data()
y_train = keras.utils.to_categorical(y_train, 10)

encoder = VGG19(
    include_rescaling=True, include_top=False, pooling="avg", input_shape=(32, 32, 3)
)

simclr = SimCLRTrainer(
    encoder,
    probe=layers.Dense(10),
    augmenter=SimCLRAugmenter(value_range=(0, 255), target_size=(32, 32)),
)

simclr.compile(
    optimizer=keras.optimizers.Adam(),
    loss=SimCLRLoss(temperature=1),
    probe_optimizer=keras.optimizers.Adam(),
    probe_loss=keras.losses.CategoricalCrossentropy(from_logits=True),
    probe_metrics=[keras.metrics.CategoricalAccuracy()],
)

simclr.fit(x_train, y_train)
```

One outstanding item to consider:
* Does the Augmenter class belong in `layers/preprocessing/` or in `training`?

PR #704, opened 2022-08-19
Add augmenter layer

ianstenbit:
This is a piece of the contrastive learning PR #699 which I'm breaking out for its own review.

The open question currently is:
- Does this belong in a `training` namespace, or just as a preprocessing layer?


LukeWood (approved):
Looks good to me.

LukeWood:
This is fine in `layers` for now too.

PR #735, opened 2022-08-25
Create ImageNet loader in keras_cv.datasets

ianstenbit:
@LukeWood

### DIFF
--- a/keras_cv/datasets/imagenet/load.py
+++ b/keras_cv/datasets/imagenet/load.py
@@ -1,0 +1,89 @@
+def parse_imagenet_example(example, image_size):
+    """Function to parse a TFRecord example into an image and label"""
+    # Read example
+    image_key = "image/encoded"
+    label_key = "image/class/label"
+    keys_to_features = {
+        image_key: tf.io.FixedLenFeature((), tf.string, ""),
+        label_key: tf.io.FixedLenFeature([], tf.int64, -1),
+    }
+    parsed = tf.io.parse_single_example(example, keys_to_features)
+
+    # Decode and resize image
+    image_bytes = tf.reshape(parsed[image_key], shape=[])
+    image = tf.io.decode_jpeg(image_bytes, channels=3)
+    image = layers.Resizing(
+        width=image_size[0], height=image_size[1], crop_to_aspect_ratio=True
+    )(image)
+
+    # Decode label
+    label = tf.cast(tf.reshape(parsed[label_key], shape=()), dtype=tf.int32) - 1
+    label = tf.one_hot(label, 1000)
+
+    return {"images": image, "labels": label}
+
+
+def load(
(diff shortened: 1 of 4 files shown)


bhack:
Why we are not using TFDS in this case?

ianstenbit:
> Why we are not using TFDS in this case?

The preprocessed TFRecords are formatted and TPU-ready, which makes them a bit easier to plug-and-play with. TFDS ImageNet is still available as an API for those that want to use it.

@LukeWood

keras_cv/datasets/imagenet/load.py
```diff
@@ -0,0 +1,107 @@
...
+    # Decode label
+    label = tf.cast(tf.reshape(parsed[label_key], shape=()), dtype=tf.int32) - 1
+    label = tf.one_hot(label, 1000)
+
+    return image, label
+
+
+def load(
+    split,
+    tfrecords_path,
```
LukeWood, line 46: nit: tfrecord_path

I think tfrecord is already plural implied (maybe I'm wrong:?)

LukeWood (changes requested):
Very minor nits, thanks for putting this together

bhack:
> > Why we are not using TFDS in this case?
> 
> The preprocessed TFRecords are formatted and TPU-ready, which makes them a bit easier to plug-and-play with. TFDS ImageNet is still available as an API for those that want to use it.
> 
> @LukeWood

Have you seen @sayakpaul gist https://gist.github.com/sayakpaul/9544d3ba935805bd47d71fd8596e7bc0#get-the-imagenet2012-data-prepared ?

LukeWood:
Looks good!

keras_cv/datasets/imagenet/load.py
```diff
@@ -0,0 +1,107 @@
...
+    keys_to_features = {
+        image_key: tf.io.FixedLenFeature((), tf.string, ""),
+        label_key: tf.io.FixedLenFeature([], tf.int64, -1),
+    }
+    parsed = tf.io.parse_single_example(example, keys_to_features)
+
+    # Decode and resize image
+    image_bytes = tf.reshape(parsed[image_key], shape=[])
+    image = tf.io.decode_jpeg(image_bytes, channels=3)
+    image = layers.Resizing(
```
LukeWood, line 33: Ah just noticed that!

constructing a Resizing every call here will cause retracing.

you need to curry a function to get around this unfortunately.

LukeWood (changes requested):
Just spotted a major performance issue.  This may be bottlenecking your train script too

keras_cv/datasets/imagenet/load.py
ianstenbit: Ahhhh nice catch -- thanks!

PR #736, opened 2022-08-25
Restructure model tests to try to reduce OOM

ianstenbit:
My opinion is that if these are too slow overall, we might want to consider just making all of the tests under /models run separately as integration tests only.

What are your thoughts? @LukeWood


ianstenbit:
Tests are all green locally -- let's see what GitHub CI thinks.

LukeWood (approved):
Certainly not the "right" solution, but at least it unblocks us.

LukeWood:
Thanks @ianstenbit for the fix

PR #738, opened 2022-08-29
Split RandomlyResizedCrop into two API surfaces (RandomlyZoomedCrop, RandomCropAndResize)

### LINKED ISSUE #676: RandomResizedCrop API Change
Currently, RandomResizedCrop is tuned by taking a crop_area_factor and aspect_ratio_factor.

The API should be updated to take a target_size, zoom_factor, and an aspect ratio factor.  At augmentation time, a value is drawn randomly from the Zoom factor and aspect ratio factor distributions.  a crop size is computed by multiplying each dimension of target size with the value drawn from zoom_factor.  Next, the width and height of the crop size are distorted accordingly by the value drawn from the aspect ratio factor.  A crop of this size is taken from the image, and finally is resized to the target size.

Some edge cases:
- when the crop size is larger than the image, we want to still respect aspect ratio (random crop uses smart resize)

AdityaKane2001:
@sayakpaul @LukeWood @martin-gorner

As discussed in #676 I have updated the RRC code to incorporate `zoom_factor`. I will be adding a Colab gist demonstrating the layer and to verify that it works as expected.

/auto Closes #676


keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -91,32 +85,52 @@ def __init__(
...
     ):
-        crop_area_factor = self.crop_area_factor()
+        zoom_factor = self.zoom_factor()
         aspect_ratio = self.aspect_ratio_factor()
 
-        new_height = tf.clip_by_value(
-            tf.sqrt(crop_area_factor / aspect_ratio), 0.0, 1.0
-        )  # to avoid unwanted/unintuitive effects
-        new_width = tf.clip_by_value(tf.sqrt(crop_area_factor * aspect_ratio), 0.0, 1.0)
+        if image.shape.rank == 3:
```
LukeWood, line 91: you can achieve this behavior using negative indices:

original_heights = image.shape[-2] I believe should work!  examples of this are available in RandomCrop/RandomHeight

martin-gorner:
Hi

I tried to test the new functionality but I'm getting an error every time.
Here is a Colab reproducing the error:
https://colab.research.google.com/drive/1VvE8PzoYQFvCEjfjDg6z_Wi0fPhPnH3B?usp=sharing

bhack:
@martin-gorner In your case it is an autograph issue cause we are receiving `None` values  in [`_get_corrected_crop_dimensions`](https://github.com/keras-team/keras-cv/pull/738/files#diff-6834fc0ba79a728ed2b63f05a521251d5edc3c057ad1320022cfb83f2449bd2aR166):
`ValueError: Tried to convert 'y' to a tensor and failed. Error: None values not supported.`

Some samplses are `None`
```python
print("original_height ", original_height)
print("original_width", original_width)
```

AdityaKane2001:
@bhack @martin-gorner 

Got it, will take a look.

AdityaKane2001:
@martin-gorner 

According to https://github.com/keras-team/keras-cv/issues/676#issuecomment-1216742375 we would need to know the dimensions of the input image. Is there any workaround to this?

martin-gorner:
Great, did you test with RaggedTensors ?
The primary use case for RandomResizeCrop is to go from a Ragged batch of images of various sizes to a batch of images of the same size ready to be fed into a convolutional backbone.

AdityaKane2001:
@martin-gorner 

Yes, I tested it using your notebook itself. I was wondering if there is any way to display the older images as well so we can directly compare the effects of the layer. Here is the [gist](https://colab.research.google.com/gist/AdityaKane2001/9af03e7966f38a2f80b37057f74723b5/test-new-randomresizecrop.ipynb).

LukeWood:
@AdityaKane2001 looks like the tests fail!

martin-gorner:
Thank you @AdityaKane2001. The gist is looking good.
The bounding boxes are off on the cropped images but I guess bounding box support is not implemented yet in RandomResizeCrop ?

martin-gorner:
I just noticed that RandomCrop and RandomResizeCrop do not use the same arguments for the target size.  It's width, height vs. target_size (a tuple made of (height, width)). Could this be aligned ?
```
tf.keras.layers.RandomCrop(height, width, seed=None, **kwargs)
tf.keras.layers.RandomResizeCrop(target_size, zoom_factor, aspect_ratio_factor, interpolation="bilinear", seed=None, **kwargs)
```

martin-gorner:
I tested some more and I see black bars in the output:
https://colab.research.google.com/drive/1VvE8PzoYQFvCEjfjDg6z_Wi0fPhPnH3B#scrollTo=E4X1J5Tiw0W6

AdityaKane2001:
@LukeWood 

I will fix the tests after we have finalized the API.

@martin-gorner 

I'll take a look and get back to you. Thanks for pointing out the same.

LukeWood:
@AdityaKane2001 any updates?

AdityaKane2001:
@LukeWood 

Sorry for the delay, but I am not able to pinpoint why the resultant images contain black bars.

martin-gorner:
Hi Aditya,
Have you been able to find the cause or do you need help ?

AdityaKane2001:
@martin-gorner @LukeWood @sayakpaul 

Please take a look at this [gist](https://colab.research.google.com/gist/AdityaKane2001/d7d42b175af69a2fce7b4bbf74e93fc2/-aditya-s-copy-test-new-randomresizecrop.ipynb#scrollTo=oEGIBwju5Kvk). I have tried to incorporate all cases of output dimensions mentioned in the table below. I did not observe any distortion or black bars. Please let me know if there are any inconsistencies which I might have missed.

Shorthands used for brevity:
nh: new_height
oh: original_height

nw: new_width
ow: original_width

|     | Subcase                   | Height | Width |
|-----|---------------------------|--------|-------|
| Old |                           | 372    | 512   |
| New | nh > oh; nw > ow; nh < nw | 400    | 600   |
| New | nh > oh; nw > ow; nh > nw | 700    | 600   |
| New | nh < oh; nw > ow; nh < nw | 250    | 600   |
| New | nh < oh; nw > ow; nh > nw | NA     | NA    |
| New | nh > oh; nw < ow; nh < nw | 400    | 500   |
| New | nh > oh; nw < ow; nh > nw | 400    | 200   |

AdityaKane2001:
I will clean up the code and make sure that it is graph mode compatible (we cannot use comparison operators directly). Please let me know if the functionality is as expected, I'll make the rest of the changes all at once after that is confirmed. 

Sorry for the delay on my side so far 😓

LukeWood:
> @martin-gorner @LukeWood @sayakpaul
> 
> Please take a look at this [gist](https://colab.research.google.com/gist/AdityaKane2001/d7d42b175af69a2fce7b4bbf74e93fc2/-aditya-s-copy-test-new-randomresizecrop.ipynb#scrollTo=oEGIBwju5Kvk). I have tried to incorporate all cases of output dimensions mentioned in the table below. I did not observe any distortion or black bars. Please let me know if there are any inconsistencies which I might have missed.
> 
> Shorthands used for brevity: nh: new_height oh: original_height
> 
> nw: new_width ow: original_width
> 
> Subcase	Height	Width
> Old		372	512
> New	nh > oh; nw > ow; nh < nw	400	600
> New	nh > oh; nw > ow; nh > nw	700	600
> New	nh < oh; nw > ow; nh < nw	250	600
> New	nh < oh; nw > ow; nh > nw	NA	NA
> New	nh > oh; nw < ow; nh < nw	400	500
> New	nh > oh; nw < ow; nh > nw	400	200

We can unit test for this, right?  Using tf.ones(), rrc, and then assert no values are 0

AdityaKane2001:
@LukeWood Yes we can. Just wanted to get the functionality correct first, compatibility and tests can be handled easily after that.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -149,14 +159,41 @@ def augment_image(self, image, transformation, **kwargs):
 
         return tf.squeeze(augmented_image, axis=0)
 
+    def _get_corrected_crop_dimensions(
+        self, new_height, new_width, original_height, original_width
+    ):
+        if new_height > original_height and new_width > original_width:
```
LukeWood, line 165: can we add some comments to justify each case?  Just to improve maintainability

LukeWood:
@AdityaKane2001 looks good, but add comments to justify the cases, add test cases to ensure no black bars, and please update to be graph compatible so I can play with it locally in the demo.

AdityaKane2001:
> @AdityaKane2001 looks good, but add comments to justify the cases, add test cases to ensure no black bars, and please update to be graph compatible so I can play with it locally in the demo.

Sure, will do.

martin-gorner:
Hi Aditya

We discussed internally what the behavior should be in training=False mode and concluded it should be the same as RandomCrop, i.e. cropping the largest possible part of the image of the target aspect ratio and then resizing it to the target size. This ensures there are no image distorsions in training=False mode.

The rationale is that, training=False should completely turn off all Random* layers but in this case, the resizing is necessary for the model to work so instead training=False should be as close to a no-op as possible, which means no distorsions, possibly some cropping, necessary resizing.

The current code does not seem to implement that but you can get this behavior easily by calling tf.keras.layers.Resizing(
crop_to_aspect_ratio=True)

martin-gorner:
And could you please align the API with the rest of Keras layers:
 - tf.keras.layers.Resizing(height, width)
 - tf.keras.layers.RandomCrop(height, width)

Can you please change the API from:
RandomResizeCrop(target_size=(640, 480))
to:
RandomResizeCrop(height, width)

Thank you!

AdityaKane2001:
@martin-gorner 

> The current code does not seem to implement that but you can get this behavior easily by calling tf.keras.layers.Resizing(
crop_to_aspect_ratio=True)

I agree with the conclusion. The current implementation uses `tf.image.crop_and_resize`, which makes it harder for me (or for anyone who handles the code later on) to make modifications. Thus, I will change the code to *not* use this function. Instead, I will do things manually (taking a crop is essentially slicing the tensor). I believe it will make the code more readable and future-proof as well.

I will make the other API changes as well.

/cc @LukeWood @sayakpaul

LukeWood:
Thank you @AdityaKane2001 for your hard work!

LukeWood:
hey @AdityaKane2001 !  Can we ensure we reflect overflow properly like we have in https://github.com/keras-team/keras-cv/pull/457

This might be a good reference as you push forward with a custom op

AdityaKane2001:
> hey @AdityaKane2001 ! Can we ensure we reflect overflow properly like we have in https://github.com/keras-team/keras-cv/pull/457

@LukeWood Could you please elaborate on this? Not sure I understand what "overflow" means in this case.

LukeWood:
Sorry for the lack of clarity.  You'll notice that when the zoom out overflows the borders (like in image 2) the elephant is mirrored from the top down.  This is what I mean - you can achieve this by using ImageProjectiveTransformV3 or the keras transform helper.  I seem to remember we didn't use transform for some reason, but maybe we should look at this again?  https://github.com/keras-team/keras-cv/pull/457/files has some code for this

AdityaKane2001:
@martin-gorner @LukeWood 

I have updated the implementation to use `preprocessing.transform` i.e. `ImageProjectiveTransformV3`. Please feel free to play around with it [here](https://colab.research.google.com/gist/AdityaKane2001/77d728a639d8d16e552f88f2c0b8ada5/test_rrc.ipynb) (it is a minimal example, but I think will be sufficient to test out all cases).

I will add the test cases first thing tomorrow morning.

martin-gorner:
I wanted to test but I'm hitting an error on the first run:
```
random_resize_crop = kcv.layers.RandomResizedCrop(height=128, width=512, zoom_factor=(1.0,1.0), aspect_ratio_factor=(1.0,1.0))

TypeError: Exception encountered when calling layer 'random_resized_crop_4' (type RandomResizedCrop).
TypeError: Input 'y' of 'Minimum' Op has type float32 that does not match type int32 of argument 'x'.
```

Repro Colab [here](https://colab.research.google.com/drive/1VvE8PzoYQFvCEjfjDg6z_Wi0fPhPnH3B#scrollTo=MOTNVUhd5rL0)

AdityaKane2001:
@martin-gorner 

Could you please try now? It seems to work fine now.

LukeWood:
@AdityaKane2001 can you rebase?

AdityaKane2001:
@LukeWood 

Resolved conflicts but still adding tests.

AdityaKane2001:
@LukeWood added unit tests. Please take a look when you get a moment.

LukeWood:
@AdityaKane2001 please fix the failing tests!  thanks

LukeWood (changes requested):
Some tests are failing, once those get fixed, then LGTM!

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -59,69 +52,78 @@ class RandomResizedCrop(BaseImageAugmentationLayer):
...
-        x2 = width_offset + new_width
+        new_height = new_height / original_height
+        new_width = new_width / original_width
 
-        return [[y1, x1, y2, x2]]
+        height_offset = height_offset / original_height
+        width_offset = width_offset / original_width
 
-    def call(self, inputs, training=True):
+        return (new_height, new_width, height_offset, width_offset)
```
LukeWood, line 124: lets actually make these a dictionary.

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -25,32 +25,25 @@
...
-    This implementation takes an intuitive approach, where we crop the images to a
-    random height and width, and then resize them. To do this, we first sample a
-    random value for area using `crop_area_factor` and a value for aspect ratio using
-    `aspect_ratio_factor`. Further we get the new height and width by
-    dividing and multiplying the old height and width by the random area
-    respectively. We then sample offsets for height and width and clip them such
-    that the cropped area does not exceed image boundaries. Finally we do the
-    actual cropping operation and resize the image to `target_size`.
+    This implementation takes a distortion-oriented approach, which means the
+    amount of distortion in the image is proportional to the `zoom_factor`
```
martin-gorner, line 29: Probable error in the docstring:
The amount of distorsion is proportional to aspect_ratio_factor, not the zoom_factor. Distorsion means changing the aspect ratio, not zooming.

martin-gorner:
I did a new round of testing. Repro Colab [here](https://colab.research.google.com/drive/1VvE8PzoYQFvCEjfjDg6z_Wi0fPhPnH3B#scrollTo=NSwTYV6eeKaF).

1. distorsions still occur even when aspect_ratio_factor=(1,1) is set
Screen Shot 2022-09-21 at 12 03 47 PM
2. When the target size is larger than the image, the image is now mirrored. This is better than black borders but I think we agreed above that it should be zoomed. Mirroring means that bounding boxes will also have to be replicated which might be hard. Also, replication can become a false clue for a neural network just as black bars. Yes, something must be sacrificed when the target size is larger than the image. If you don't think it should be zoom_factor, then please chime in.
Screen Shot 2022-09-21 at 12 09 09 PM
3. bounding boxes do not seem to be supported:
`in augment_bounding_boxes raise NotImplementedError()`
4. operation on batches does not seem to be supported. Yes, I was trying on a ragged batch. Converting images of varying dimensions to a batch of equal-sized images is the main purpose of this layer.
```
in get_random_transformation
            original_width = tf.cast(tf.shape(image)[-2], tf.float32)
        ValueError: Index 1 is not uniform
```

LukeWood:
@martin-gorner I actually think the behavior is correct here, I think we just need to rename the layer for clarity and keep both APIs.  I think they both actually serve a good purpose:  RandomResizedCrop and RandomlyZoomedCrop perhaps?

AdityaKane2001:
@martin-gorner 

> distorsions still occur even when aspect_ratio_factor=(1,1) is set

I am not sure why the distortions are still arising. I will check the code again, but honestly I am not really sure what to look for.

> operation on batches does not seem to be supported. 

That is strange to me as well. AFAIK `get_random_transformation` runs for each image, so it should be dealing with one image at a time, and not the tensor.

> I think they both actually serve a good purpose: RandomResizedCrop and RandomlyZoomedCrop perhaps?

Bounding box support for the previous RRC implementation is fairly straightforward.

martin-gorner:
Mirroring is one possible choice when dealing with "crops" larger than the original image but it is not the only choice. "Cropping" to a size larger than the original image is a pathological case so something has to give. My suggestion was to disrespect the zoom_factor in that case. Mostly because with mirroring we would have to duplicate bounding boxes and that's really complicated and offer no benefits. Also because a zoomed image is a less obvious outlier than an image with black borders or a mirrored image. Anyway, it's a pathological use case, we just need it to work in the most unsurprising way possible, without cliff-edge effects and in a way that is simple to implement.

LukeWood:
Hello everyone!  Thank you for all of your design inputs on this.  I've discussed this with @fchollet, @martin-gorner, and many others over the last few weeks.

Here is what I have landed on design wise:

1.) the name RandomResizedCrop unclear and a bad API surface.  This is super unclear to people what exactly is going on under the hood.  As such, we have renamed RandomResizedCrop to RandomCropAndResize.  This much more clearly conveys that the crop is random and the resize is deterministic

2.) the zoom_factor API is much nicer for tuning a nob up and down.  How do you turn RRC off?  You really can't is the answer. zoom_factor solves this

3.) so why keep RandomCropAndResize?  In short, people rely on this layer.  We will really discourage people from adopting KerasCV and migrating if we don't have a compatible operation.  This operation IS distinct, even if the final goal is similar to RandomlyZoomedCrop.

As such, it makes sense to keep both of them.

Additionally, RandomCropAndResize bounding box support is trivial - time to market matter and we should prioritize it.

LukeWood (approved):

LukeWood:
@fchollet added you here per our discussion.

LukeWood:
@AdityaKane2001 lets actually do bounding box support in a follow up, this PR is already big enough and the previous implementation was bugged.

LukeWood:
Accompanying PR:

- https://github.com/tensorflow/similarity/compare/development...LukeWood:similarity:patch-1

I checked KerasCV's keras.io docs - no references so we are safe to break this API for now.

LukeWood:
All tests are passing, we have https://github.com/keras-team/keras-cv/issues/823 to migrate to the new API.

LukeWood:
Thank you @AdityaKane2001 for your hard work on these APIs.  I feel confident in both to be useful - and now we can quickly ship OD workflows, contrastive workflows, etc.

PR #742, opened 2022-08-29
Add support for model kwargs to basic training script

ianstenbit:
This is needed for training ConvNeXt because it has hyperparameters that are specified at model construction time (e.g. `drop_path_rate`)

@LukeWood

### DIFF
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -69,6 +69,11 @@ flags.DEFINE_float(
     0.1,
     "Initial learning rate which will reduce on plateau.",
 )
+flags.DEFINE_string(
+    "model_kwargs",
+    "{}",
+    "Keyword argument dictionary to pass to the constructor of the model being trained",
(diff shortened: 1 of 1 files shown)


examples/training/classification/imagenet/basic_training.py
```diff
@@ -148,6 +153,7 @@ def augment(img, label):
         include_top=True,
         classes=CLASSES,
         input_shape=IMAGE_SIZE + (3,),
+        **eval(FLAGS.model_kwargs),
```
LukeWood, line 156: can we json.loads this instead?

ianstenbit: Since JSON requires double-quote tags, I think this is pretty unintuitive from the CLI because you'd have to use single-quote outer quotes with double-quote inner quotes, whereas this way it works with either approach

LukeWood: ok, fair enough.  Great point

PR #747, opened 2022-08-29
Make loss reduction consistent with model garden

LukeWood:
some questions:

- is this how we want to scale reductions for losses with multiple axes?


LukeWood:
This is correct Keras behavior - the more I look

PR #748, opened 2022-08-30
Add support for segmentation mask in BaseImageAugmentationLayer

ianstenbit:
This doesn't add support for any actual layers yet -- I'll add that once we're sure this looks good.

@LukeWood @qlzh727

### DIFF
--- a/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
@@ -28,6 +28,7 @@ TARGETS = "targets"
 BOUNDING_BOXES = "bounding_boxes"
 KEYPOINTS = "keypoints"
 RAGGED_BOUNDING_BOXES = "ragged_bounding_boxes"
+SEGMENTATION_MASK = "segmentation_mask"
 IS_DICT = "is_dict"
 USE_TARGETS = "use_targets"
 
@@ -210,8 +211,28 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
         """
         raise NotImplementedError()
 
+    def augment_segmentation_mask(self, segmentation_mask, transformation, **kwargs):
+        """Augment a single image's segmentation mask during training.
+
+        Args:
+          segmentation_mask: 3D segmentation mask input tensor to the layer.
+            Forwarded from `layer.call()`.
+          transformation: The transformation object produced by
+            `get_random_transformation`. Used to coordinate the randomness
+            between image, label, and segmentation mask.
+
+        Returns:
+          output 3D tensor, which will be forward to `layer.call()`.
+        """
+        raise NotImplementedError()
+
     def get_random_transformation(
-        self, image=None, label=None, bounding_boxes=None, keypoints=None
+        self,
+        image=None,
+        label=None,
+        bounding_boxes=None,
+        keypoints=None,
+        segmentation_mask=None,
     ):
         """Produce random transformation config for one single input.
 
@@ -222,6 +243,7 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
           image: 3D image tensor from inputs.
           label: optional 1D label tensor from inputs.
           bounding_box: optional 2D bounding boxes tensor from inputs.
+          segmentation_mask: optional 3D segmentation mask tensor from inputs.
 
         Returns:
           Any type of object, which will be forwarded to `augment_image`,
@@ -253,8 +275,13 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
         label = inputs.get(LABELS, None)
         bounding_boxes = inputs.get(BOUNDING_BOXES, None)
         keypoints = inputs.get(KEYPOINTS, None)
+        segmentation_mask = inputs.get(SEGMENTATION_MASK, None)
         transformation = self.get_random_transformation(
-            image=image, label=label, bounding_boxes=bounding_boxes, keypoints=keypoints
+            image=image,
+            label=label,
+            bounding_boxes=bounding_boxes,
+            keypoints=keypoints,
+            segmentation_mask=segmentation_mask,
         )
         image = self.augment_image(
             image,
@@ -288,6 +315,12 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
                 image=image,
             )
             result[KEYPOINTS] = keypoints
+        if segmentation_mask is not None:
+            segmentation_mask = self.augment_segmentation_mask(
+                segmentation_mask,
+                transformation=transformation,
+            )
+            result[SEGMENTATION_MASK] = segmentation_mask
 
         # preserve any additional inputs unmodified by this layer.
         for key in inputs.keys() - result.keys():
--- a/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
@@ -144,7 +144,7 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
             `layer.call()`.
           transformation: The transformation object produced by
             `get_random_transformation`. Used to coordinate the randomness
-            between image, label and bounding box.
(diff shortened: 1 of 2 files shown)


keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -210,8 +211,28 @@ def augment_keypoints(self, keypoints, transformation, **kwargs):
         """
         raise NotImplementedError()
 
+    def augment_segmentation_mask(self, segmentation_mask, transformation, **kwargs):
+        """Augment a single image's segmentation mask during training.
+
+        Args:
+          segmentation_mask: 3D segmentation mask input tensor to the layer.
```
qlzh727, line 218: For the segmentation_mask, what's the rank we expect for it? Is it [H, W, 1] or just [H, W]?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -210,8 +211,28 @@ def augment_keypoints(self, keypoints, transformation, **kwargs):
...
+            Forwarded from `layer.call()`.
+          transformation: The transformation object produced by
+            `get_random_transformation`. Used to coordinate the randomness
+            between image, label, and segmentation mask.
```
qlzh727, line 222: "and segmentation mask" should be applied to all the other augment_x methods.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -222,6 +243,7 @@ def get_random_transformation(
           image: 3D image tensor from inputs.
           label: optional 1D label tensor from inputs.
           bounding_box: optional 2D bounding boxes tensor from inputs.
+          segmentation_mask: optional 3D segmentation mask tensor from inputs.
 
         Returns:
           Any type of object, which will be forwarded to `augment_image`,
```
qlzh727, line 249: Please update the return docstring for augment_segmentation_mask.

keras_cv/layers/preprocessing/base_image_augmentation_layer_test.py
```diff
@@ -148,15 +151,22 @@ def test_augment_image_and_localization_data(self):
         images = np.random.random(size=(8, 8, 3)).astype("float32")
         bounding_boxes = np.random.random(size=(3, 5)).astype("float32")
         keypoints = np.random.random(size=(3, 5, 2)).astype("float32")
+        segmentation_mask = np.random.random(size=(8, 8, 3)).astype("float32")
```
qlzh727, line 154: The shape of the mask seems wrong. The last dim should be 1 (the class). If you get the mask from the raw data, then it should be processed first to convert the RGB value to the class index.

qlzh727 (changes requested):
Thanks for the PR.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
ianstenbit: The rank is variable based on the segmentation task. E.g. for binary segmentation this would be 2 (assuming one-hot encoding), and for multi-class segmentation this would be `classes`

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
ianstenbit: Done -- also added keypoints, since that wasn't in the docstring.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
ianstenbit: Done

keras_cv/layers/preprocessing/base_image_augmentation_layer_test.py
ianstenbit: The last dim is arbitrary, since it can be any number representing e.g `classes`.

But I think it was confusing having it as 3 because it somewhat implies that it represents color channels, so I've updated it to 2.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
qlzh727: Using one-hot seems to be inconsistent with other values in the API, eg label shape in the test is [1] (and doesn't seems to be one-hot from the docstring). The class for the bbox is also attached as the 5th value in the list, and not using one-hot.

ianstenbit: The segmentation map doesn't *have* to be one-hot encoded. It could just as well be an integer class encoding, and the last dimension would be `1`.

Is there a reason we should force users to choose one or the other?

LukeWood: we should probably start out segmentation map support with a literature review on the different segmentation map formats.  TFDS is a good first place to look, I believe there are some polygon based ones - it would be good to add some utilities for converting these as a starting point

ianstenbit: The two most popular datasets I'm seeing anecdotally in papers (as well as the top-2 on [papers with code](https://paperswithcode.com/datasets?task=semantic-segmentation)) are MS COCO and Cityscapes, which both use pixel-level segmaps.

FWIW, Cityscapes does also provide polygon-based segmentation information, but they call it "weakly annotated."

We could certainly add utilities for transforming common non-pixelmap formats to pixel maps, but it does seem like a pixelmap is the standard for segmentation masks, and these utilities may end up being dataset specific (and perhaps therefore belong in `keras_cv/datasets/`

ianstenbit: @qlzh727 I think the non-onehot approach is clearly reasonable for binary classification, but I think for multi-class segmentation it doesn't make sense.

But that's somewhat tangential here, because it shouldn't affect the API, as we should support any number of channels in the segmentation mask.

qlzh727: For MS coco, I think they choose to use a RGB value to encode the mask, so that the mask can be saved just images (for easy distribution), but user will still have to decode the class ids by `ids=R+G*256+B*256^2`. The extra step of decode is kind of annoying and other datasets didn't follow that. 

Do you have an example dataset of multi-class segmentation mask? I was always assuming that the mask is single classed (not binary necessarily), but multi-class is doesn't work with [H, W, 1].

ianstenbit: It looks like this is how the [cityscapes](https://www.cityscapes-dataset.com/dataset-overview/#features) dataset is encoded, but I haven't confirmed that by building my own segmentation map for it.

the examples I've seen which use this dataset do use a [H, W, Classes] format for training (e.g. [this one](https://github.com/srihari-humbarwadi/cityscapes-segmentation-with-Unet))

qlzh727: Btw, same for https://www.tensorflow.org/datasets/catalog/cityscapes, which has `'segmentation_label': Image(shape=(1024, 2048, 1), dtype=tf.uint8)`

qlzh727: Also, if you check the latest implementation from the same author, the mask seems to be a 1-channel PNG image. https://github.com/srihari-humbarwadi/DeepLabV3_Plus-Tensorflow2.0/blob/f06df320aebe4aeeeccfa789340b14c0df163b09/train.py#L50

The [H, W, 1] is more aligned with user's intuition (for label), and [H, W, C] is bit unclear whether the data is one-hot, or actually multi-labeled. 

Having said that, I think we should document this in the docstring, that the segmentation map should be rank 3, and [H, W, 1] (for single class/labeled) or [H, W, C] in a very uncommon case.

qlzh727: Btw, the one-hot encoding does have some implications for the KPLs, eg mixup, which will mix up the labels as well. The layer will perform quite differently if the mask is one-hot encoded.

ianstenbit: Sounds good -- thank you for the explanation! I'm updating the docstrings and will update the tests to use [H, W, 1]

qlzh727 (approved):
Thanks for the PR.

PR #751, opened 2022-08-30
Add all of the no-op segmentation mask augmentations

ianstenbit:
This adds "segmentation map support" to all the layers that shouldn't change a segmentation map (mostly because they're just color adjustments)

### DIFF
--- a/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
@@ -28,6 +28,7 @@ TARGETS = "targets"
 BOUNDING_BOXES = "bounding_boxes"
 KEYPOINTS = "keypoints"
 RAGGED_BOUNDING_BOXES = "ragged_bounding_boxes"
+SEGMENTATION_MASK = "segmentation_mask"
 IS_DICT = "is_dict"
 USE_TARGETS = "use_targets"
 
@@ -210,8 +211,28 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
         """
         raise NotImplementedError()
 
+    def augment_segmentation_mask(self, segmentation_mask, transformation, **kwargs):
+        """Augment a single image's segmentation mask during training.
+
+        Args:
+          segmentation_mask: 3D segmentation mask input tensor to the layer.
+            Forwarded from `layer.call()`.
+          transformation: The transformation object produced by
+            `get_random_transformation`. Used to coordinate the randomness
+            between image, label, and segmentation mask.
+
+        Returns:
+          output 3D tensor, which will be forward to `layer.call()`.
+        """
+        raise NotImplementedError()
+
     def get_random_transformation(
-        self, image=None, label=None, bounding_boxes=None, keypoints=None
+        self,
+        image=None,
+        label=None,
+        bounding_boxes=None,
+        keypoints=None,
+        segmentation_mask=None,
     ):
         """Produce random transformation config for one single input.
 
@@ -222,6 +243,7 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
           image: 3D image tensor from inputs.
           label: optional 1D label tensor from inputs.
           bounding_box: optional 2D bounding boxes tensor from inputs.
+          segmentation_mask: optional 3D segmentation mask tensor from inputs.
 
         Returns:
           Any type of object, which will be forwarded to `augment_image`,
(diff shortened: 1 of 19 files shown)


fchollet (commented):
Thanks for the PR!

For features like this, we want to make sure that we don't forget to implement support for the feature in new layers that we ship. This requires having a unit test that automatically checks the presence of the feature (similar to `get_config()` checks in the standard layer test).

qlzh727:
Agreed. We probably want to have a test like https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/preprocessing/with_labels_test.py for segmentation_map. Basically a layer_test for all the common functionality of image KPLs.

qlzh727 (changes requested):
Please add a test case for common no-op layers.

keras_cv/layers/preprocessing/with_segmentation_masks_test.py
```diff
@@ -0,0 +1,101 @@
...
+class WithSegmentationMasksTest(tf.test.TestCase, parameterized.TestCase):
+    @parameterized.named_parameters(*TEST_CONFIGURATIONS)
+    def test_can_run_with_segmentation_masks(self, layer_cls, init_args):
+        layer = layer_cls(**init_args)
+
+        img = tf.random.uniform(
+            shape=(3, 512, 512, 3), minval=0, maxval=1, dtype=tf.float32
+        )
+        segmentation_masks = tf.random.uniform(
+            shape=(3, 512, 512, 3), minval=0, maxval=1, dtype=tf.int32
```
qlzh727, line 82: The shape of the mask should probably be 3, 512, 512, 1

ianstenbit:
I've added a with_segmentation_mask test, but I'm not sure that this fully solves the problem -- we still need to manually make sure these tests include all of the relevant layers (which the with_labels test currently did not).

These tests also don't assert that the segmentation mask and/or label are actually passed through in the output, so I've updated the with_labels test to do that and done the same in with with_segmentation_mask test.

keras_cv/layers/preprocessing/with_segmentation_masks_test.py
qlzh727, line 82: The value range should be 0 ~ classes, and 1 seems to be a bad choice here.

keras_cv/layers/preprocessing/with_segmentation_masks_test.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+
+        inputs = {"images": img, "segmentation_masks": segmentation_masks}
+        _ = layer(inputs)
```
qlzh727: We should validate that the output mask is not changed (no op).

keras_cv/layers/preprocessing/with_segmentation_masks_test.py
```diff
@@ -0,0 +1,101 @@
...
+    # This has to be a separate test case to exclude CutMix and MixUp
+    # (which are not yet supported for segmentation mask augmentation)
+    @parameterized.named_parameters(*TEST_CONFIGURATIONS)
+    def test_can_run_with_segmentation_mask_single_image(self, layer_cls, init_args):
+        layer = layer_cls(**init_args)
+        img = tf.random.uniform(
+            shape=(512, 512, 3), minval=0, maxval=1, dtype=tf.float32
+        )
+        segmentation_mask = tf.random.uniform(
+            shape=(512, 512, 3), minval=0, maxval=1, dtype=tf.int32
```
qlzh727, line 99: same here.

qlzh727 (changes requested):

qlzh727:
I think we probably  

> I've added a with_segmentation_mask test, but I'm not sure that this fully solves the problem -- we still need to manually make sure these tests include all of the relevant layers (which the with_labels test currently did not).
> 
> These tests also don't assert that the segmentation mask and/or label are actually passed through in the output, so I've updated the with_labels test to do that and done the same in with with_segmentation_mask test.

keras_cv/layers/preprocessing/with_segmentation_masks_test.py
ianstenbit: Done

ianstenbit: Done

keras_cv/layers/preprocessing/with_segmentation_masks_test.py
ianstenbit: Done, although this won't always be true (once we add non-noop KPLs for seg masks). I've left a comment about this, and we can add a parameter for this to the tests in the future.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -28,7 +28,7 @@
 BOUNDING_BOXES = "bounding_boxes"
 KEYPOINTS = "keypoints"
 RAGGED_BOUNDING_BOXES = "ragged_bounding_boxes"
-SEGMENTATION_MASK = "segmentation_mask"
+SEGMENTATION_MASK = "segmentation_masks"
```
qlzh727, line 31: SEGMENTATION_MASK[S]?

ianstenbit: Done. Made it plural for consistency with the other labels

qlzh727 (approved):

LukeWood (approved):
Looks good to me, thanks Ian!

PR #753, opened 2022-08-31
Add MaskRCNN model as a test for the wrapper approach with TF model garden code.

qlzh727:
# Note that this PR is more for a demo and not intend to be submitted at the moment. 

# What does this PR do?
Add MaskRCNN model as a test for the wrapper approach with TF model garden code.

1. The current class only contains the model building logic.
2. Most of the tunable params are currently backed in the function, and not exposed to use yet.
3. Verify the model building logic in the test. (will need tfm pip deps to work)

Next step:
1. Configure the input parsing logic for coco, and also anchor generation.
2. Mimic the training loop logic as model.train_step/eval_step
3. Train model e2e.

## Before submitting
      Pull Request section?
      to it if that's the case.


keras_cv/models/object_detection/mask_rcnn/mask_rcnn.py
```diff
@@ -0,0 +1,307 @@
...
+except ImportError:
+    # Didn't have tensorflow model garden package installed, and the model can't
+    # be instantiated.
+    tfm = None
+
+
+def _assert_tensorflow_model_installed():
+    if tfm is None:
+        raise ImportError(
+            'Tensor model package is not installed in the '
```
LukeWood, line 38: Should we provide a model name here in the error message.

"To use {MODEL}... you must install tf-models-official"

keras_cv/models/object_detection/mask_rcnn/mask_rcnn.py
```diff
@@ -0,0 +1,307 @@
...
+            'environment. Please use `pip install tf-models-official` '
+            'or `pip install tf-models-nightly` to install the required '
+            'dependency package')
+
+
+def mask_rcnn(
```
LukeWood, line 44: do we plan to support bounding_box_format and the other KerasCV APIs (include_rescaling, etc)?

LukeWood (commented):
some preliminary thoughts (i.e. do we plan to support bounding_box_format(), etc)

keras_cv/models/object_detection/mask_rcnn/mask_rcnn.py
qlzh727: Ack, will do.

keras_cv/models/object_detection/mask_rcnn/mask_rcnn.py
qlzh727: Yea, we should add those common params to make the interface more look like the existing keras cv APIs.

LukeWood, line 44: is `mask_rcnn` our intended API surface?

I think that by wrapping in a sub-model similar to the RetinaNet we will be able to better manage the bounding_box_format, include_rescaling, and metric evaluation during fit().

I think you'll also need to do this to get label encoding and prediction decoding working w/ model.fit(), model.evaluate(), and model.predict().

qlzh727:
Updated the PR description. This PR is more for a demo, and not intended to be submitted at the moment.

LukeWood:
> Updated the PR description. This PR is more for a demo, and not intended to be submitted at the moment.

Thanks for the heads up!  cool, sounds good!

keras_cv/models/object_detection/mask_rcnn/mask_rcnn.py
qlzh727: yea, I think we probably should take the delegate approach, that make MaskRCNN to include a tfm MaskRCNN model, and delegate all the public methods to it.

My original thought was to directly subclass the tfm MaskRCNN model, but then this requires a hard dependency to tfm, otherwise the code will fail at the package load time when tfm is not installed.

qlzh727:
I have moved my PR to https://github.com/keras-team/keras-cv/pull/754 (since the current one was created on Master). Let's review that one. Thanks.

PR #755, opened 2022-09-01
Promote anchor generator default to a public api

LukeWood:
This is needed to customize the NMS layer easily


keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -187,6 +187,22 @@ def __init__(
                 f"`label_encoder.box_variance={label_encoder.box_variance}`."
             )
 
+    @static
```
ianstenbit, line 190: This decorator is giving a name error, breaking tests and linting

ianstenbit (approved):

PR #760, opened 2022-09-02
Fix FocalLoss reduction, SmoothL1 reduction, and add extra loss validation to RetinaNet model

LukeWood:
This fixes convergence issues in the tutorial and prevents others from falling into the same issue.

If you reduce your results on too many axes, you cannot mask out the ignore mask in your samples.


ianstenbit (approved):
Is there a losses_test that we can write which would enforce that all losses are reducing along the last axis?

LukeWood:
> Is there a losses_test that we can write which would enforce that all losses are reducing along the last axis?

I think so, for now lets merge this and do a follow up PR.

PR #761, opened 2022-09-02
Add integration tests to GCB

LukeWood:


ianstenbit:
@LukeWood did you verify that the environment variable was set successfully and the additional tests were run? I don't see anything in the CI logs about it.

PR #772, opened 2022-09-07
Remove the tensorflow_datasets as installation dep

qlzh727:
It should be used as a test dependency.


ianstenbit (approved):
LGTM as long as the tests are happy.

PR #775, opened 2022-09-08
Add segmask support for random_flip

ianstenbit:
@LukeWood @qlzh727

### DIFF
--- a/keras_cv/layers/preprocessing/random_flip.py
+++ b/keras_cv/layers/preprocessing/random_flip.py
@@ -85,18 +85,7 @@ class RandomFlip(BaseImageAugmentationLayer):
         return label
 
     def augment_image(self, image, transformation, **kwargs):
-        flipped_output = tf.cond(
-            transformation["flip_horizontal"],
-            lambda: tf.image.flip_left_right(image),
-            lambda: image,
-        )
-        flipped_output = tf.cond(
-            transformation["flip_vertical"],
-            lambda: tf.image.flip_up_down(flipped_output),
-            lambda: flipped_output,
-        )
-        flipped_output.set_shape(image.shape)
-        return flipped_output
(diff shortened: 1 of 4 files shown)


ianstenbit:
A handful of KPLs will have a similar pattern to this, where the `augment_segmentation_mask` method will perform either the exact same or a very similar operation on the `segmentation_mask `as the one applied to the `image`.

We could potentially therefore override `augment_segmentation_mask` to simply call `augment_image` with the `segmentation_mask` as a parameter.

The benefit of this is that we'd reduce code duplication, but we'd also make the implementation a bit indirected, and in some cases we may have to update `augment_image` to support the segmask augmentation.

What are your thoughts @LukeWood @qlzh727

qlzh727:
> A handful of KPLs will have a similar pattern to this, where the `augment_segmentation_mask` method will perform either the exact same or a very similar operation on the `segmentation_mask `as the one applied to the `image`.
> 
> We could potentially therefore override `augment_segmentation_mask` to simply call `augment_image` with the `segmentation_mask` as a parameter.
> 
> The benefit of this is that we'd reduce code duplication, but we'd also make the implementation a bit indirected, and in some cases we may have to update `augment_image` to support the segmask augmentation.
> 
> What are your thoughts @LukeWood @qlzh727

SGTM, you can extra a method as `_flip_image` and point both `augment_image` and `augment_segmentation_map` to it. Similarly for other KPLs.

qlzh727 (changes requested):
Please try to extract the image flip into a private method, and make both augment_image and augment_segmentation_map to it.

ianstenbit:
> Please try to extract the image flip into a private method, and make both augment_image and augment_segmentation_map to it.

Done -- thanks!

qlzh727 (approved):
Thanks.

PR #776, opened 2022-09-08
Add accelerator count to training history

ianstenbit:
This is relevant now because we're scaling learning rate and batch size based on the number of accelerators in #757


LukeWood:
> /gcbrun

Gcb run is globally broken, but let’s merge as the tests are unaffected

PR #777, opened 2022-09-08
Add segmask support for RandomResizedCrop

ianstenbit:


LukeWood:
@ianstenbit we will be changing the API for this in https://github.com/keras-team/keras-cv/pull/738

LukeWood:
Looks like this PR will be (unchanged though)

ianstenbit:
> Looks like this PR will be (unchanged though)

Ack -- Want to wait to merge this until after that PR?

LukeWood (approved):

keras_cv/layers/preprocessing/random_resized_crop.py
```diff
@@ -194,6 +187,9 @@ def _check_class_arguments(
     def augment_target(self, augment_target, **kwargs):
         return augment_target
 
+    def augment_segmentation_mask(self, segmentation_mask, transformation, **kwargs):
+        return self._crop_and_resize(segmentation_mask, transformation)
```
qlzh727, line 191: Will this have the same issue for mixing the pixel value when resize? eg if the [[0, 2], [0, 2]] is resize to [1, 1,]?

qlzh727 (changes requested):

keras_cv/layers/preprocessing/random_resized_crop.py
ianstenbit: Yes -- I overlooked the interpolation issue for this layer. I will update this and send for re-review later today.

ianstenbit: I've updated this implementation to use the build-in nearest-neighbor approach for interpolation, which handles the discrete class issue and avoids bad interpolation + rounding.

keras_cv/layers/preprocessing/random_resized_crop_test.py
```diff
@@ -123,3 +123,28 @@ def test_crop_area_factor_errors(self, crop_area_factor):
...
+            target_size=self.target_size,
+            aspect_ratio_factor=(3 / 4, 4 / 3),
+            crop_area_factor=(0.8, 1.0),
+            seed=self.seed,
+        )
+
+        input_mask_resized = tf.image.resize(mask, self.target_size)
+
+        output = layer(input, training=True)
+        self.assertNotAllClose(output["segmentation_masks"], input_mask_resized)
```
qlzh727: Can we do more strict assertion, eg make sure we don't mix the class of the mask?

qlzh727 (changes requested):

keras_cv/layers/preprocessing/random_resized_crop_test.py
ianstenbit: I think this comment is based on an old diff -- the most recent version of this code does verify that we only output classes `0` and `7` using `self.assertAllInSet(output["segmentation_masks"], [0, 7])`

qlzh727 (approved):

PR #778, opened 2022-09-08
Add visualization tools for object detection models

LukeWood:


LukeWood:
@fchollet lets review the API for these here, then lets add custom box plotting to handle bounding box thickness in a follow up.  After adding `line_width` as a parameter, I can update the keras.io tutorials

LukeWood:
> Can you provide a TL;DR for this PR? e.g., what methods are provided, how we do expect users to use them, etc.

Yeah absolutely. Happy to do so shortly.

LukeWood:
Lets use @martin-gorner 's tools instead, they look great!

PR #779, opened 2022-09-09
Make contrastive API private

ianstenbit:
As requested by @tanzhenyu (since API has not been formally approved yet)

@tanzhenyu @fchollet PTAL at the contrastive learning design doc so that we can finalize this API and make it public in KerasCV

@LukeWood

### DIFF
--- a/keras_cv/losses/simclr_loss_test.py
+++ b/keras_cv/losses/simclr_loss_test.py
@@ -14,7 +14,7 @@
 
 import tensorflow as tf
 
-from keras_cv.losses.simclr_loss import SimCLRLoss
+from keras_cv.losses.simclr_loss import _SimCLRLoss
 
 
 class SimCLRLossTest(tf.test.TestCase):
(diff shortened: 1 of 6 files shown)


LukeWood:
What is the motivation to make this private?  Isn't the API finalized?

tanzhenyu (approved):

ianstenbit:
> What is the motivation to make this private? Are there outstanding API changes? Ideally I’d like to have a tutorial for this launched by mid October.

@tanzhenyu are you able to take a look at the design for this? I am going to merge this and make this private for now, but this will be important for Luke's conference talk so we should make sure we have a path forward to make this public.

tanzhenyu:
> > What is the motivation to make this private? Are there outstanding API changes? Ideally I’d like to have a tutorial for this launched by mid October.
> 
> @tanzhenyu are you able to take a look at the design for this? I am going to merge this and make this private for now, but this will be important for Luke's conference talk so we should make sure we have a path forward to make this public.

Approved. Closing this PR.

PR #780, opened 2022-09-09
Add segmask support for RandomRotate

ianstenbit:
@LukeWood @qlzh727 

As discussed, I'm transforming to one-hot in order to perform rotation while preserving the mask properly for multi-class masks.


keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -78,6 +78,8 @@ class RandomRotation(BaseImageAugmentationLayer):
       bounding_box_format: The format of bounding boxes of input dataset. Refer
         https://github.com/keras-team/keras-cv/blob/master/keras_cv/bounding_box/converters.py
         for more details on supported bounding box formats.
+      segmentation_classes: The number of classes in the input segmentation mask.
```
qlzh727, line 81: Optional integer.

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -78,6 +78,8 @@ class RandomRotation(BaseImageAugmentationLayer):
...
+        Required iff augmenting data with sparse (non one-hot) segmentation masks.
```
qlzh727, line 82: We should also make it clear whether the number of classes contain the 0 (background).

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -226,6 +233,28 @@ def augment_bounding_boxes(
     def augment_label(self, label, transformation, **kwargs):
         return label
 
+    def augment_segmentation_mask(self, segmentation_mask, transformation, **kwargs):
+        # If segmentation_classes is specified, we have a dense segmentation mask.
+        # We therefore one-hot encode before rotation to avoid bad interpolation
+        # during the rotation transformation. Later, we make the mask sparse
+        # again using tf.argmax.
+        if self.segmentation_classes:
```
qlzh727, line 241: I think we should do a check for the last column of the input mask, and raise error if it is 1. This will prevent user from accidentally miss the mask class.

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -226,6 +233,28 @@ def augment_bounding_boxes(
...
+            one_hot_mask = tf.one_hot(
+                tf.squeeze(segmentation_mask, axis=-1), self.segmentation_classes
+            )
+        else:
+            one_hot_mask = segmentation_mask
```
qlzh727, line 246: it is bit confusing call the original mask to be one hot.

keras_cv/layers/preprocessing/random_rotation.py
```diff
@@ -226,6 +233,28 @@ def augment_bounding_boxes(
...
+
+        rotated_one_hot_mask = self._rotate_image(one_hot_mask, transformation)
+
+        if self.segmentation_classes:
```
qlzh727, line 250: readability wise, I think it will be more clear to have:
```
if self.segmentation_classes:
  # change to one hot
  # rotate the one hot
  # argmax
  # return result
else:
  # rotate mask
  # return tf.round(result)

File: keras_cv/layers/preprocessing/random_rotation_test.py [hunk 6]
```diff
@@ -99,3 +99,33 @@ def test_ragged_bounding_boxes(self):
...
+        self.assertAllClose(expected_masks, outputs["segmentation_masks"])
+
+    def test_augment_one_hot_segmentation_mask(self):
+        classes = 8
+
+        input_images = np.random.random((2, 20, 20, 3)).astype(np.float32)
+        masks = tf.one_hot(np.random.randint(classes, size=(2, 20, 20)), classes)
+        inputs = {"images": input_images, "segmentation_masks": masks}
+
+        # 90 rotation.
```
qlzh727, line 125: We should do strict validation about non 90' rotation, since that will trigger the mix of pixel and argmax.

qlzh727 (changes requested):

keras_cv/layers/preprocessing/random_rotation.py
LukeWood, line 82: have we determined that we will only be supporting masks in the form of one-hot images?  (i.e. no polygon formats)

I think this is a pretty reasonable choice, mainly just curious.

keras_cv/layers/preprocessing/random_rotation.py
ianstenbit: Agreed -- done.

keras_cv/layers/preprocessing/random_rotation.py
ianstenbit: Yes -- given that this is how predictions are made in segmentation, I think they should live in the training pipeline like this. For datasets with polygonal masks, we can offer preprocessing code to transform those into pixel masks.

keras_cv/layers/preprocessing/random_rotation.py
ianstenbit: Done -- and added a test to verify checking.

keras_cv/layers/preprocessing/random_rotation_test.py
ianstenbit: Good call -- I've added some assertions to this test case which verify that, for a 45-degree rotation, no interpolation takes place.

keras_cv/layers/preprocessing/random_rotation_test.py
```diff
@@ -99,3 +99,43 @@ def test_ragged_bounding_boxes(self):
...
+
+        input_images = np.random.random((2, 20, 20, 3)).astype(np.float32)
+        # Masks are all 0s or 8s, to verify that when we rotate we don't do bad
+        # mask interpolation to either a 0 or a 7
+        masks = np.random.randint(2, size=(2, 20, 20, 1)) * (classes - 1)
+        inputs = {"images": input_images, "segmentation_masks": masks}
+
+        # Attempting to rotate a sparse mask without specifying classes fails.
+        bad_layer = RandomRotation(factor=(0.25, 0.25))
+        with self.assertRaises(ValueError):
```
qlzh727, line 114: Let's also assert the error message here.

qlzh727 (approved):
Thanks for the PR.

keras_cv/layers/preprocessing/random_rotation_test.py
ianstenbit: Done -- thank you for the review!

PR #781, opened 2022-09-09
Convergence tests for ResNetV2 (experimental)

ianstenbit:
@LukeWood This is a proof-of-concept for convergence testing using Google Cloud Build.

I've created a separate cloudbuild YAML, and a separate JSONNet configuration.
I've also set up a separate cloudbuild trigger set up to run these tests on a schedule.

If this approach LG, we can extend it afterwards to cover all of the classification models, and also add some assertions about step_time.

Some discussion points:
- It's possible that upstream framework changes (e.g. in Keras) could slightly change the results produced by these tests. So we may want to use a custom higher tolerance for the assertions to allow some wiggle room.
- We may want to host the dataset on a connected GCS bucket. This would give us some more flexibility which TFDS doesn't offer, for example we could only load a small fraction of the dataset. (alternatively, we could generate data in the tests)

We can also explore an entirely different avenue, and make a separate folder in this repo for convergence testing, storing all the necessary code in that repo. This would let us skip the test decorators for the `CONVERGENCE` environment variable, and would separate convergence testing from all other tests (which has the benefit of making convergence testing self-contained and the downside of sharding our testing)

What do you think?

### DIFF
--- a/keras_cv/models/models_test.py
+++ b/keras_cv/models/models_test.py
@@ -15,6 +15,7 @@
 
 import pytest
 import tensorflow as tf
+import tensorflow_datasets as tfds
 from tensorflow import keras
 from tensorflow.keras import backend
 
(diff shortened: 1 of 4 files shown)


bhack:
It could be also nice to have some example/doc for users to reuse/customize these for a more standard Kubernetes/GKE deployment using:

https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/gke-deploy

keras_cv/models/models_test.py
```diff
@@ -120,6 +121,39 @@ def _test_model_can_be_used_as_backbone(self, app, last_dim, args):
...
+                input_shape=(28, 28, 1),
+                **args
+            )
+            model.compile(
+                optimizer="adam", loss="categorical_crossentropy", metrics=["accuracy"]
+            )
+
+        results = model.fit(train_ds)
+
+        self.assertAlmostEqual(results.history["accuracy"][0], accuracy)
```
LukeWood, line 155: I think this should be assertGreaterThan and we should provide a floor for convergence, not a guarantee

ianstenbit: Ack -- I've updated this to be a greater than assertion

ianstenbit:
Closing this PR in favor of a `tf.test.Benchmark`-based approach

PR #782, opened 2022-09-10
Migrate to rely on GCB from GitHub actions

LukeWood:
we unfortunately have hit the point where we cannot keep using GitHub actions due to scaling concerns.  GCB allows us to configure out own CI cluster and timeouts - as such we will need to rely on that going forward.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -6,37 +6,6 @@ on:
   release:
     types: [created]
 jobs:
-  test:
-    name: Test the code
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [windows-latest, ubuntu-latest]
-    steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python 3.7
-      uses: actions/setup-python@v1
-      with:
-        python-version: 3.7
-    - name: Get pip cache dir
-      id: pip-cache
-      run: |
-        python -m pip install --upgrade pip setuptools
-        echo "::set-output name=dir::$(pip cache dir)"
-    - name: pip cache
-      uses: actions/cache@v2
-      with:
-        path: ${{ steps.pip-cache.outputs.dir }}
-        key: ${{ runner.os }}-pip-${{ hashFiles('setup.py') }}
-        restore-keys: |
-          ${{ runner.os }}-pip-
-    - name: Install dependencies
-      run: |
-        pip install tensorflow==2.9.1
-        pip install -e ".[tests]" --progress-bar off --upgrade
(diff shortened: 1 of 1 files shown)


bhack:
We could still use Action well known syntax but using GCP resources

https://github.com/terraform-google-modules/terraform-google-github-actions-runners

bhack:
As a side note switching to the cloudbuild syntax will make contributing or co-maintaining the CI with the community harder.

(Just extending or make a change in the CI it will be not testable locally or in a fork).

bhack:
See also https://github.com/tensorflow/build/issues/61#issuecomment-991244641

LukeWood:
> As a side note switching to the cloudbuild syntax will make contributing or co-maintaining the CI with the community harder.
> 
> (Just extending or make a change in the CI it will be not testable anymore [locally](https://github.com/nektos/act) or in a fork).

Good point, yeah this would make it much harder for community members to update it.  Ideally we could still use GH-actions, so maybe self hosted is the best option.

LukeWood:
> > As a side note switching to the cloudbuild syntax will make contributing or co-maintaining the CI with the community harder.
> > (Just extending or make a change in the CI it will be not testable anymore [locally](https://github.com/nektos/act) or in a fork).
> 
> Good point, yeah this would make it much harder for community members to update it. Ideally we could still use GH-actions, so maybe self hosted is the best option.

As of now though… the repo is fully blocked by failing unit tests

bhack:
> As of now though… the repo is fully blocked by failing unit tests

Generally models tests are quite heavy and I don't think that we could accumulate models and run these test on every PR (it could be ok on the PR to check the specific when we modify a specific network  or probably test all models before the nightly cut) 

In any case I think that currently the main problem with the free hosted Action is that this test is exhausting 7GB max ram in the runner (Github free tier):
`keras_cv/models/object_detection/retina_net/retina_net_inference_test.py`

qlzh727:
If the build is failing due to heavy ram consumption, how about config some of the test to be integration test and not run them in the unit test suite? 

Since the code base is expending, we should find a more scalable solution (probably discuss in the team meeting). For now, let's run the GPU test which should have the same test coverage.

qlzh727 (approved):
Let's check with the team before submission.

LukeWood:
> Let's check with the team before submission.

Sounds good, I also like @bhack ’s idea to disable model tests on GitHub actions and only run those before cuts

LukeWood:
Dropping this PR in favor of @bhack 's solution to just not run model tests on GH actions.

Thanks for the suggestion @bhack this is a much better suggestion!

PR #784, opened 2022-09-11
Some final touches to the RetinaNet API

LukeWood:


LukeWood:
Looks like the layer names change in v2.  I’ll have to figure out which ones we want to use.

LukeWood:
@fchollet @qlzh727 let me know if these changes look OK, I've confirmed these work

PR #786, opened 2022-09-12
Add pre-trained weights for MobileNetV3Small

ianstenbit:
@LukeWood 

This is the first model for which our basic_training script did not yield a result close to SOTA for a specific model. In this case, it's about 10% worse on top-1 validation accuracy than the paper-claimed results for this model.

Based on the look of the tensorboard logs, I think this may be due to having a too-high initial learning rate, so we could re-run this with a lower initial LR.

I think I'm indifferent about whether we add these weights -- they're a decent first set for MobileNetV3Small but they're not great. Feel free to close this PR if you'd prefer to wait for a re-run with better results.

### DIFF
--- a/examples/training/classification/imagenet/training_history.json
+++ b/examples/training/classification/imagenet/training_history.json
@@ -31,6 +31,22 @@
             "validation_accuracy": "0.6613"
         }
     },
+    "mobilenetv3small": {
+        "v0": {
+            "accelerators": 8,
+            "args": {
+                "batch_size": "512"
+            },
+            "contributor": "ianstenbit",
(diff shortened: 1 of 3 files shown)


LukeWood:
> I think I'm indifferent about whether we add these weights -- they're a decent first set for MobileNetV3Small but they're not great. Feel free to close this PR if you'd prefer to wait for a re-run with better results.

Lets try another run with a lower LR.  I'm actually also having some trouble training my MobileNetV3Large so perhaps there is a bug somewhere in the code.  First things first: verify the LR is not the issue

PR #788, opened 2022-09-12
Promote model tests to integration

LukeWood:
The goal is to prevent timeouts, while still supporting GitHub actions.  This is a suggestion by @bhack to unblock development, while keeping the model tests running on GCB.

### DIFF
--- a/keras_cv/models/convnext_test.py
+++ b/keras_cv/models/convnext_test.py
@@ -51,6 +51,12 @@ MODEL_LIST = [
 ]
 
 
+@pytest.mark.skipif(
+    "INTEGRATION" not in os.environ or os.environ["INTEGRATION"] != "true",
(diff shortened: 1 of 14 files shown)


keras_cv/models/convnext_test.py
```diff
@@ -51,6 +51,12 @@
 ]
 
 
+@pytest.mark.skipif(
+    "INTEGRATION" not in os.environ or os.environ["INTEGRATION"] != "true",
```
qlzh727, line 55: This annotation is bit long, and is kind of pain to copy/paste to every integration test. Is it possible to consolidate integration tests based on file/directory? so that we can skip them altogether.

qlzh727 (approved):
Might want to add a TODO to run all the models test somewhere else.

PR #791, opened 2022-09-13
Step 1/2: refactor kerascv

tanzhenyu:
# What does this PR do?

Refactor KerasCV so that keypoints will live under ops
Next we will make bounding_box live under ops as well


keras_cv/ops/__init__.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from keras_cv.keypoint.converters import convert_format
-from keras_cv.keypoint.formats import REL_XY
-from keras_cv.keypoint.formats import XY
-from keras_cv.keypoint.utils import filter_out_of_image
+from keras_cv.ops.keypoint_converters import convert_format
```
qlzh727, line 15: Should we keep a sub-package as ops.key_point.xxx? The current API like `keras_cv.ops.convert_format` lost the context of keypoints.

qlzh727 (changes requested):

keras_cv/ops/__init__.py
ianstenbit: +1 -- or perhaps the method should have context in-name (e.g. convert_keypoint_format)

ianstenbit:
What's the motivation for this? My guess would be that we don't think `bounding_box` or `keypoints` are worthy of level-1 namespaces but I just want to make sure I understand. Thanks!

keras_cv/ops/__init__.py
tanzhenyu: "convert_keypoint_format" sounds good to me
We also need to consolidate "REL_XY" in keypoint and "REL_XY" in bbox, maybe next PR then...

tanzhenyu:
> What's the motivation for this? My guess would be that we don't think `bounding_box` or `keypoints` are worthy of level-1 namespaces but I just want to make sure I understand. Thanks!

`bbox` and `keypoints` are all ops, and we do plan to support more ops like these, and we want to be consistent with what we propose: kerascv provides APIs in layers/losses/metrics/ops

LukeWood:
> > What's the motivation for this? My guess would be that we don't think `bounding_box` or `keypoints` are worthy of level-1 namespaces but I just want to make sure I understand. Thanks!
> 
> `bbox` and `keypoints` are all ops, and we do plan to support more ops like these, and we want to be consistent with what we propose: kerascv provides APIs in layers/losses/metrics/ops

what is the definition of an OP here?  I'm not sure this actually produces more readable code to users.

The API reads as follows:
```
bounding_box.convert_format()
```

vs

```
ops.convert_bounding_box_format()
```

Can we provide more context in the PR & also include what step 2 is?

tanzhenyu:
> > > What's the motivation for this? My guess would be that we don't think `bounding_box` or `keypoints` are worthy of level-1 namespaces but I just want to make sure I understand. Thanks!
> > 
> > 
> > `bbox` and `keypoints` are all ops, and we do plan to support more ops like these, and we want to be consistent with what we propose: kerascv provides APIs in layers/losses/metrics/ops
> 
> what is the definition of an OP here? I'm not sure this actually produces more readable code to users.
> 
> The API reads as follows:
> 
> ```
> bounding_box.convert_format()
> ```
> 
> vs
> 
> ```
> ops.convert_bounding_box_format()
> ```
> 
> Can we provide more context in the PR & also include what step 2 is?

Ops are anything that goes under here:
https://github.com/tensorflow/tensorflow/tree/master/tensorflow/python/ops

Basically, having a bounding_box namespace will become messy when we start to support more things. Instead, `bounding_box_ops` , `keypoint_ops`, and `bounding_box_3d_ops`, `keypoint_3d_ops` creates better structure.

The next step is to move bounding_box under ops

qlzh727:
> ops

If we are aligning with tf.ops, then I think will make sense to keep the subpackage name so that the ops are properly grouped, eg tf has math_ops, which then map to tf.math.xxx. The there is something common between subgroups, we can lift them out and put directly under keras_cv.ops.

tanzhenyu:
> > ops
> 
> If we are aligning with tf.ops, then I think will make sense to keep the subpackage name so that the ops are properly grouped, eg tf has math_ops, which then map to tf.math.xxx. The there is something common between subgroups, we can lift them out and put directly under keras_cv.ops.
> 
> My proposal will be
> 
> `keras_cv.ops.xxx` for ops common `keras_cv.ops.bbox.xxx` for bbox `keras_cv.ops.keypoint.xxx` for keypoint `keras_cv.ops.more_subpackge.xxx` for any future subpackages.

My main question / concern is, how many we have and plan to have in `common`, vs `keypoint`? Today we only have conversion functions. What makes the structure more distributed is we have a `core` group which only contains some sampling function.

LukeWood:
> > > ops
> > 
> > 
> > If we are aligning with tf.ops, then I think will make sense to keep the subpackage name so that the ops are properly grouped, eg tf has math_ops, which then map to tf.math.xxx. The there is something common between subgroups, we can lift them out and put directly under keras_cv.ops.
> > My proposal will be
> > `keras_cv.ops.xxx` for ops common `keras_cv.ops.bbox.xxx` for bbox `keras_cv.ops.keypoint.xxx` for keypoint `keras_cv.ops.more_subpackge.xxx` for any future subpackages.
> 
> My main question / concern is, how many we have and plan to have in `common`, vs `keypoint`? Today we only have conversion functions. What makes the structure more distributed is we have a `core` group which only contains some sampling function.

I'd be more in favor of removing `core` and moving Samplers => preprocessing or something similar than I should condensing label namespaces into a singular namespace.  There are only a few task types, and we only need one namespace per task.  Would that handle this concern?

tanzhenyu:
> > > > ops
> > > 
> > > 
> > > If we are aligning with tf.ops, then I think will make sense to keep the subpackage name so that the ops are properly grouped, eg tf has math_ops, which then map to tf.math.xxx. The there is something common between subgroups, we can lift them out and put directly under keras_cv.ops.
> > > My proposal will be
> > > `keras_cv.ops.xxx` for ops common `keras_cv.ops.bbox.xxx` for bbox `keras_cv.ops.keypoint.xxx` for keypoint `keras_cv.ops.more_subpackge.xxx` for any future subpackages.
> > 
> > 
> > My main question / concern is, how many we have and plan to have in `common`, vs `keypoint`? Today we only have conversion functions. What makes the structure more distributed is we have a `core` group which only contains some sampling function.
> 
> I'd be more in favor of removing `core` and moving Samplers => preprocessing or something similar than I should condensing label namespaces into a singular namespace. There are only a few task types, and we only need one namespace per task. Would that handle this concern?

I'd prefer ops/bounding_box_ops.py + ops/bounding_box_3d_ops.py + iou + many other things, instead of bounding_box/convert_format.py & bounding_box/convert_3d_format.py, WDYT?

LukeWood:
> > > > > ops
> > > > 
> > > > 
> > > > If we are aligning with tf.ops, then I think will make sense to keep the subpackage name so that the ops are properly grouped, eg tf has math_ops, which then map to tf.math.xxx. The there is something common between subgroups, we can lift them out and put directly under keras_cv.ops.
> > > > My proposal will be
> > > > `keras_cv.ops.xxx` for ops common `keras_cv.ops.bbox.xxx` for bbox `keras_cv.ops.keypoint.xxx` for keypoint `keras_cv.ops.more_subpackge.xxx` for any future subpackages.
> > > 
> > > 
> > > My main question / concern is, how many we have and plan to have in `common`, vs `keypoint`? Today we only have conversion functions. What makes the structure more distributed is we have a `core` group which only contains some sampling function.
> > 
> > 
> > I'd be more in favor of removing `core` and moving Samplers => preprocessing or something similar than I should condensing label namespaces into a singular namespace. There are only a few task types, and we only need one namespace per task. Would that handle this concern?
> 
> I'd prefer ops/bounding_box_ops.py + ops/bounding_box_3d_ops.py + iou + many other things, instead of bounding_box/convert_format.py & bounding_box/convert_3d_format.py, WDYT?

3d is definitely a weird case, I agree that bounding_box/3d.. is a bit weird namespace wise.

Should bounding_box_3d just be its own namespace at that point?  That is more comparable to ops/bounding_box_ops & ops/bounding_box_3d_ops.  Do we expect to have many cases like this where we have name collisions?

When I read `ops` to me it reads as if the API is being organized based on the implementation details; not by how we expect users to use the API.  User facing names matter a lot in terms of user experience.

keras_cv/ops/__init__.py
```diff
@@ -12,7 +12,7 @@
...
 # limitations under the License.
 
-from keras_cv.keypoint.converters import convert_format
-from keras_cv.keypoint.formats import REL_XY
-from keras_cv.keypoint.formats import XY
-from keras_cv.keypoint.utils import filter_out_of_image
+from keras_cv.ops.keypoint_converters import convert_keypoint_format
+from keras_cv.ops.keypoint_formats import REL_XY
+from keras_cv.ops.keypoint_formats import XY
+from keras_cv.ops.keypoint_utils import filter_out_of_image
```
qlzh727, line 18: Should we have keypoints in the filter_out_of_image?

qlzh727: `keras_cv.ops.filter_out_of_image` is bit out of the context of keypoint (assuming its keypoints specific.)

Also `out_of` is bit weird as well, how about `filter_image`?

tanzhenyu: Done.
filter_keypoints is more natural (so we can do filter_boxes if we need)

qlzh727 (approved):

ianstenbit:
@tanzhenyu this PR breaks at least one of our [examples](https://github.com/keras-team/keras-cv/blob/master/examples/models/object_detection/retina_net/basic/pascal_voc/train.py) in KerasCV, and at least one [KerasIO tutorial](https://github.com/keras-team/keras-io/commit/b3806be26bf139b9f07be6dff42c6886556994d3)

tanzhenyu:
> @tanzhenyu this PR breaks at least one of our [examples](https://github.com/keras-team/keras-cv/blob/master/examples/models/object_detection/retina_net/basic/pascal_voc/train.py) in KerasCV, and at least one [KerasIO tutorial](https://github.com/keras-team/keras-io/commit/b3806be26bf139b9f07be6dff42c6886556994d3).
> 
> If we're confident that this is the right approach, then I think we should urgently update both of those.

Any specific breakage? This only touches the keypoint which is not being used, not bounding boxes

ianstenbit:
> > @tanzhenyu this PR breaks at least one of our [examples](https://github.com/keras-team/keras-cv/blob/master/examples/models/object_detection/retina_net/basic/pascal_voc/train.py) in KerasCV, and at least one [KerasIO tutorial](https://github.com/keras-team/keras-io/commit/b3806be26bf139b9f07be6dff42c6886556994d3).
> > If we're confident that this is the right approach, then I think we should urgently update both of those.
> 
> Any specific breakage? This only touches the keypoint which is not being used, not bounding boxes

😅 sorry I totally misread the PR description -- I thought this was updating bounding boxes. I'm not aware of any examples or tutorials using the Keypoint code

PR #792, opened 2022-09-13
Add API to disable geometric transformations in RandAugment

LukeWood:


LukeWood:
@qlzh727 @fchollet we can use this in the OD tutorial to boost performance once merged

keras_cv/layers/preprocessing/rand_augment.py
```diff
@@ -63,7 +63,8 @@ class RandAugment(RandomAugmentationPipeline):
             the rate to 10/11 in our implementation, the behavior is identical to
             sampling an Identity augmentation 10/11th of the time.
             Defaults to `1.0`.
-
+        geometric: whether or not to include geometric augmentations.  This should be
```
qlzh727, line 66: Please state the type and default value in the docstring.

keras_cv/layers/preprocessing/rand_augment.py
```diff
@@ -63,7 +63,8 @@ class RandAugment(RandomAugmentationPipeline):
...
+            set to False when performing object detection.
```
qlzh727: Also what if user forget to change this to False when they have bbox? Are we raising any meaningful error?

LukeWood: They'll get an error that Augment_bounding_box is not implemented - once it is implemented for Shear and translate setting this to true will be valid.

LukeWood:
Thanks @qlzh727, comments addressed

qlzh727 (approved):

PR #794, opened 2022-09-13
Add box matcher.

tanzhenyu:
# What does this PR do?
Add ArgmaxBoxMatcher to be used in RetinaNet and RCNN.

Ack:
KerasCV and OD API authors co-developed and trying to merge back to the repo:
https://github.com/tensorflow/models/blob/master/official/vision/ops/box_matcher.py

### DIFF
--- a/keras_cv/ops/box_matcher.py
+++ b/keras_cv/ops/box_matcher.py
@@ -14,12 +14,15 @@
 
 "Argmax-based box matching"
 
-from typing import List, Tuple
+from typing import List
+from typing import Tuple
+
 import tensorflow as tf
+
 
 class ArgmaxBoxMatcher:
     """Box matching logic based on argmax of highest value (iou).
-    
+
     This class computes matches from a similarity matrix. Each row will be
     matched to at least one column, the matched result can either be positive
      / negative, or simply ignored depending on the setting.
@@ -39,25 +42,27 @@ class ArgmaxBoxMatcher:
        argmax result is below negative_threshold, and the rows will be assigned
        to negative_value if its argmax result is between negative_threshold and
        positive_threshold.
-    
-    Note: If 
-    
+
+    Note: If
+
     TODO(tanzhenyu): document when to use which mode.
 
     """
 
-    def __init__(self,
-                 thresholds: List[float],
-                 indicators: List[int],
-                 force_match_for_each_col: bool = False):
+    def __init__(
+        self,
+        thresholds: List[float],
+        indicators: List[int],
+        force_match_for_each_col: bool = False,
+    ):
         """Constructs ArgmaxBoxMatcher.
-        
+
         Args:
           thresholds: A sorted list of floats to classify the matches into
             different results (e.g. positive or negative or ignored match). The
             list will be prepended with -Inf and and appended with +Inf.
           indicators: A list of integers representing matched results (e.g.
-            positive or negative or ignored match). len(`indicators`) must 
+            positive or negative or ignored match). len(`indicators`) must
             equal to len(`thresholds`) + 1.
           force_match_for_each_col: each row will be argmax matched to at
             least one column. This means some columns will be matched to
@@ -65,33 +70,32 @@ class ArgmaxBoxMatcher:
             Filtering by `thresholds` will make less columns match to positive
             result. Setting this to True guarantees that each column will be
             matched to positive result to at least one row.
-        
+
         Raises:
-          ValueError: if `thresholds` not sorted or 
+          ValueError: if `thresholds` not sorted or
             len(`indicators`) != len(`thresholds`) + 1
         """
         if not all([lo <= hi for (lo, hi) in zip(thresholds[:-1], thresholds[1:])]):
-            raise ValueError('`threshold` must be sorted, got {}'.format(thresholds))
+            raise ValueError("`threshold` must be sorted, got {}".format(thresholds))
         self.indicators = indicators
         if len(indicators) != len(thresholds) + 1:
-            raise ValueError('len(`indicators`) must be len(`thresholds`) + 1, got '
-                            'indicators {}, thresholds {}'.format(
-                                indicators, thresholds))
+            raise ValueError(
+                "len(`indicators`) must be len(`thresholds`) + 1, got "
+                "indicators {}, thresholds {}".format(indicators, thresholds)
+            )
         thresholds = thresholds[:]
-        thresholds.insert(0, -float('inf'))
-        thresholds.append(float('inf'))
+        thresholds.insert(0, -float("inf"))
+        thresholds.append(float("inf"))
         self.thresholds = thresholds
         self._force_match_for_each_col = force_match_for_each_col
-    
 
-    def __call__(self,
-                 similarity_matrix: tf.Tensor) -> Tuple[tf.Tensor, tf.Tensor]:
+    def __call__(self, similarity_matrix: tf.Tensor) -> Tuple[tf.Tensor, tf.Tensor]:
         """Matches each row to a column based on argmax
-        
+
         Args:
           similarity_matrix: A float Tensor of shape [num_rows, num_cols] or
             [batch_size, num_rows, num_cols] representing any similarity metric.
-        
+
         Returns:
           matched_columns: An integer tensor of shape [num_rows] or [batch_size,
             num_rows] storing the index of the matched colum for each row.
@@ -118,11 +122,11 @@ class ArgmaxBoxMatcher:
                 num_rows] storing the match type indicator (e.g. positive or negative
                 or ignored match).
             """
-            with tf.name_scope('empty_gt_boxes'):
+            with tf.name_scope("empty_gt_boxes"):
                 matched_columns = tf.zeros([batch_size, num_rows], dtype=tf.int32)
                 match_indicators = -tf.ones([batch_size, num_rows], dtype=tf.int32)
                 return matched_columns, match_indicators
-        
+
         def _match_when_cols_are_non_empty():
             """Performs matching when the rows of similarity matrix are non empty.
             Returns:
@@ -132,63 +136,79 @@ class ArgmaxBoxMatcher:
                 num_rows] storing the match type indicator (e.g. positive or negative
                 or ignored match).
             """
-            with tf.name_scope('non_empty_gt_boxes'):
+            with tf.name_scope("non_empty_gt_boxes"):
                 matched_columns = tf.argmax(
-                    similarity_matrix, axis=-1, output_type=tf.int32)
+                    similarity_matrix, axis=-1, output_type=tf.int32
+                )
 
                 # Get logical indices of ignored and unmatched columns as tf.int64
                 matched_vals = tf.reduce_max(similarity_matrix, axis=-1)
                 match_indicators = tf.zeros([batch_size, num_rows], tf.int32)
 
                 match_dtype = matched_vals.dtype
-                for (ind, low, high) in zip(self.indicators, self.thresholds[:-1],
-                                            self.thresholds[1:]):
+                for (ind, low, high) in zip(
+                    self.indicators, self.thresholds[:-1], self.thresholds[1:]
+                ):
                     low_threshold = tf.cast(low, match_dtype)
                     high_threshold = tf.cast(high, match_dtype)
                     mask = tf.logical_and(
                         tf.greater_equal(matched_vals, low_threshold),
-                        tf.less(matched_vals, high_threshold))
+                        tf.less(matched_vals, high_threshold),
+                    )
                     match_indicators = self._set_values_using_indicator(
-                        match_indicators, mask, ind)
+                        match_indicators, mask, ind
+                    )
 
                 if self._force_match_for_each_col:
                     # [batch_size, num_cols], for each column (groundtruth_box), find the
                     # best matching row (anchor).
                     matching_rows = tf.argmax(
-                        input=similarity_matrix, axis=1, output_type=tf.int32)
+                        input=similarity_matrix, axis=1, output_type=tf.int32
+                    )
                     # [batch_size, num_cols, num_rows], a transposed 0-1 mapping matrix M,
                     # where M[j, i] = 1 means column j is matched to row i.
                     column_to_row_match_mapping = tf.one_hot(
-                        matching_rows, depth=num_rows)
+                        matching_rows, depth=num_rows
+                    )
                     # [batch_size, num_rows], for each row (anchor), find the matched
                     # column (groundtruth_box).
                     force_matched_columns = tf.argmax(
-                        input=column_to_row_match_mapping, axis=1, output_type=tf.int32)
+                        input=column_to_row_match_mapping, axis=1, output_type=tf.int32
+                    )
                     # [batch_size, num_rows]
                     force_matched_column_mask = tf.cast(
-                        tf.reduce_max(column_to_row_match_mapping, axis=1), tf.bool)
+                        tf.reduce_max(column_to_row_match_mapping, axis=1), tf.bool
+                    )
                     # [batch_size, num_rows]
-                    matched_columns = tf.where(force_matched_column_mask,
-                                                force_matched_columns, matched_columns)
+                    matched_columns = tf.where(
+                        force_matched_column_mask,
+                        force_matched_columns,
+                        matched_columns,
+                    )
                     match_indicators = tf.where(
-                        force_matched_column_mask, self.indicators[-1] *
-                        tf.ones([batch_size, num_rows], dtype=tf.int32), match_indicators)
+                        force_matched_column_mask,
+                        self.indicators[-1]
+                        * tf.ones([batch_size, num_rows], dtype=tf.int32),
+                        match_indicators,
+                    )
 
                 return matched_columns, match_indicators
 
-        num_gt_boxes = similarity_matrix.shape.as_list()[-1] or tf.shape(
-            similarity_matrix)[-1]
+        num_gt_boxes = (
+            similarity_matrix.shape.as_list()[-1] or tf.shape(similarity_matrix)[-1]
+        )
         matched_columns, match_indicators = tf.cond(
             pred=tf.greater(num_gt_boxes, 0),
             true_fn=_match_when_cols_are_non_empty,
-            false_fn=_match_when_cols_are_empty)
+            false_fn=_match_when_cols_are_empty,
+        )
 
         if squeeze_result:
             matched_columns = tf.squeeze(matched_columns, axis=0)
             match_indicators = tf.squeeze(match_indicators, axis=0)
 
         return matched_columns, match_indicators
-    
+
     def _set_values_using_indicator(self, x, indicator, val):
         """Set the indicated fields of x to val.
 
--- a/keras_cv/ops/box_matcher.py
+++ b/keras_cv/ops/box_matcher.py
@@ -21,7 +21,7 @@ import tensorflow as tf
 
 
 class ArgmaxBoxMatcher:
-    """Box matching logic based on argmax of highest value (e.g., iou).
+    """Box matching logic based on argmax of highest value (e.g., IOU).
 
     This class computes matches from a similarity matrix. Each row will be
     matched to at least one column, the matched result can either be positive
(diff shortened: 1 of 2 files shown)


LukeWood (commented):
Thanks for adding the BoxMatcher and sending the PR.  Overall the logic and flow look great, very minor API comments (mainly around the indicator values being a list)

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,223 @@
...
+    """Box matching logic based on argmax of highest value (iou).
+
+    This class computes matches from a similarity matrix. Each row will be
+    matched to at least one column, the matched result can either be positive
+     / negative, or simply ignored depending on the setting.
+
+    The settings include `thresholds` and `indicators`, for example if:
+    1) thresholds=[negative_threshold, positive_threshold], and
+       indicators=[negative_value, ignore_value, positive_value]: the rows will
+       be assigned to positive_value if its argmax result is above
```
ianstenbit, line 33: Isn't this it's `max` result, not its `argmax`?

related: is `ArgmaxBoxMatcher` the right name for this class? It seems like, while the `argmax` is used to compute the result, the thresholding is based on the max value, not its index.

(I may be totally off base here as OD is new to me -- just curious)

ianstenbit (commented):
Thanks! Some of my comments are more like questions -- just trying to make sure I understand the API

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,223 @@
...
+       positive_threshold.
+    2) thresholds=[negative_threshold, positive_threshold], and
+       indicators=[ignore_value, negative_value, positive_value]: the rows will
+       be assigned to positive_value if its argmax result is above
+       positive_threshold; the rows will be assigned to ignore_value if its
+       argmax result is below negative_threshold, and the rows will be assigned
+       to negative_value if its argmax result is between negative_threshold and
+       positive_threshold.
+
+    Note: If
```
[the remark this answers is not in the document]
tanzhenyu: Sounds good. Done.

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,223 @@
...
+    The settings include `thresholds` and `indicators`, for example if:
+    1) thresholds=[negative_threshold, positive_threshold], and
+       indicators=[negative_value, ignore_value, positive_value]: the rows will
+       be assigned to positive_value if its argmax result is above
+       positive_threshold; the rows will be assigned to negative_value if its
+       argmax result is below negative_threshold, and the rows will be assigned
+       to ignore_value if its argmax result is between negative_threshold and
+       positive_threshold.
+    2) thresholds=[negative_threshold, positive_threshold], and
+       indicators=[ignore_value, negative_value, positive_value]: the rows will
```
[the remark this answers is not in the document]
tanzhenyu: This class is actually more generic than that, it supports list of any length. Instead of splitting the arguments, I add another example where there's only [positive_threshold] and [negative_value, positive_value] -- so there's no `ignore_value` here.
In the end, we aim to providing default values to the constructor when using it within RetinaNetLabelEncoder and RPNLabelEncoder, acting as examples of how it can be used, and users can customize their own matching logic on top of these two.

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,223 @@
...
+    Note: If
+
+    TODO(tanzhenyu): document when to use which mode.
+
+    """
+
+    def __init__(
+        self,
+        thresholds: List[float],
+        indicators: List[int],
```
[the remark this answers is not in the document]
tanzhenyu: We should make type hints as default in KerasCV. We don't intend to support Python 2.x anyway.

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,223 @@
...
+"Argmax-based box matching"
+
+from typing import List
+from typing import Tuple
+
+import tensorflow as tf
+
+
+class ArgmaxBoxMatcher:
+    """Box matching logic based on argmax of highest value (iou).
```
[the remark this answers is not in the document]
tanzhenyu: Maybe we should say "e.g." then to avoid confusion

keras_cv/ops/box_matcher.py
tanzhenyu: It's argmax result for the first return

keras_cv/ops/box_matcher.py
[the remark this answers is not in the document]
LukeWood: > This class is actually more generic than that, it supports list of any length. Instead of splitting the arguments, I add another example where there's only [positive_threshold] and [negative_value, positive_value] -- so there's no ignore_value here.


ok, I understand now.  Thanks for explaining this - I misread the api.

keras_cv/ops/box_matcher.py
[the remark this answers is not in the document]
LukeWood: We've discussed this a lot in keras-team.  Thus far we haven't reached a consensus that they're helpful given the polymorphic nature of the Keras API.  We landed on optional for simple types (these fall in that category) and then free to opt-out for more complex unions and such (i.e. string|model|callable)

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+
+    """
+
+    def __init__(
+        self,
+        thresholds: List[float],
+        match_values: List[int],
+        force_match_for_each_col: bool = False,
+    ):
+        """Constructs ArgmaxBoxMatcher.
```
LukeWood, line 70: nit: Lets move the __init__ docstring to the class docstring to be consistent with Keras/Cv/NLP

LukeWood (approved):
Minor nits but overall LGTM

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+"Argmax-based box matching"
+
+from typing import List
+from typing import Tuple
+
+import tensorflow as tf
+
+
+class ArgmaxBoxMatcher:
+    """Box matching logic based on argmax of highest value (e.g., iou).
```
qlzh727, line 24: We might want to change all the iou to IoU.

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+
+    This class computes matches from a similarity matrix. Each row will be
```
qlzh727, line 26: what's the row and column here? Is each row a ground truth box? Does it mean the user input (IoU matrix) need to follow this format?

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+
+    The settings include `thresholds` and `match_values`, for example if:
+    1) thresholds=[negative_threshold, positive_threshold], and
+       match_values=[negative_value, ignore_value, positive_value]: the rows will
+       be assigned to positive_value if its argmax result is above
+       positive_threshold; the rows will be assigned to negative_value if its
+       argmax result is below negative_threshold, and the rows will be assigned
+       to ignore_value if its argmax result is between negative_threshold and
+       positive_threshold.
+    2) thresholds=[negative_threshold, positive_threshold], and
```
qlzh727, line 38: I have little difficulty to parse this example, comparing to example 1. The only difference I can see is that the order of the  ignore_value and negative_value are switched, but its unclear to me why they are different. 

To me, a more clear description is that:
thresholds is buckets, so for [t1, t2], there will be three buckets: (-inf, t1], (t1, t2], (t2, +inf), and user will provide a list of 3 values for each of the bucket. You can have arbitrary number of buckets (skip the logic of positive and negative).

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+       match_values=[negative_values, positive_value]: the rows will be assigned to
+       positive value if its argmax result is above positive_threshold; the rows
+       will be assigned to negative_value if its argmax result is below
+       negative_threshold.
+
+    Usage:
+
+    ```python
+    box_matcher = keras_cv.ops.ArgmaxBoxMatcher([0.3, 0.7], [-1, 0, 1])
+    iou_metric = keras_cv.bounding_box.compute_iou(anchors, gt_boxes)
```
qlzh727, line 55: what if user give a wrong order of the y_true and anchor here? (when we do metric/loss, usually the y_true comes before y_pred, which is different from this example.)

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+        match_values: List[int],
+        force_match_for_each_col: bool = False,
+    ):
+        """Constructs ArgmaxBoxMatcher.
+
+        Args:
+          thresholds: A sorted list of floats to classify the matches into
+            different results (e.g. positive or negative or ignored match). The
+            list will be prepended with -Inf and and appended with +Inf.
+          match_values: A list of integers representing matched results (e.g.
```
qlzh727, line 76: Any reason this need to be a int?

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+            positive or negative or ignored match). len(`match_values`) must
+            equal to len(`thresholds`) + 1.
+          force_match_for_each_col: each row will be argmax matched to at
```
qlzh727, line 79: let's change row/column to ground truth and generated anchor.

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+            multiple rows while some columns will not be matched to any rows.
+            Filtering by `thresholds` will make less columns match to positive
+            result. Setting this to True guarantees that each column will be
+            matched to positive result to at least one row.
+
+        Raises:
+          ValueError: if `thresholds` not sorted or
+            len(`match_values`) != len(`thresholds`) + 1
+        """
+        if not all([lo <= hi for (lo, hi) in zip(thresholds[:-1], thresholds[1:])]):
```
qlzh727, line 90: how about just `sorted(thresholds) !=  thresholds`

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+            raise ValueError("`threshold` must be sorted, got {}".format(thresholds))
```
qlzh727, line 91: let's use f"`threshold` must be sorted, got {thresholds}", same for the other error messages.

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+        self.match_values = match_values
+        if len(match_values) != len(thresholds) + 1:
+            raise ValueError(
+                "len(`match_values`) must be len(`thresholds`) + 1, got "
+                "match_values {}, thresholds {}".format(match_values, thresholds)
+            )
+        thresholds = thresholds[:]
```
qlzh727, line 98: what does this line do?

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+        """
+        squeeze_result = False
+        if len(similarity_matrix.shape) == 2:
+            squeeze_result = True
+            similarity_matrix = tf.expand_dims(similarity_matrix, axis=0)
+        static_shape = similarity_matrix.shape.as_list()
+        num_rows = static_shape[1] or tf.shape(similarity_matrix)[1]
+        batch_size = static_shape[0] or tf.shape(similarity_matrix)[0]
+
+        def _match_when_cols_are_empty():
```
qlzh727, line 126: should this be _math_when_rows_are_empty()?

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+            a tensor of -1's to indicate that the rows do not match to any columns.
+            Returns:
+                matched_columns: An integer tensor of shape [num_rows] or [batch_size,
+                num_rows] storing the index of the matched column for each row.
+                matched_values: An integer tensor of shape [num_rows] or [batch_size,
+                num_rows] storing the match type indicator (e.g. positive or negative
+                or ignored match).
+            """
+            with tf.name_scope("empty_gt_boxes"):
+                matched_columns = tf.zeros([batch_size, num_rows], dtype=tf.int32)
```
qlzh727, line 138: [batch_size, num_rows] might have issue when one of them are tensor and the other are int, which I think might happen when batch size is not static.

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+                matched_values = -tf.ones([batch_size, num_rows], dtype=tf.int32)
+                return matched_columns, matched_values
+
+        def _match_when_cols_are_non_empty():
+            """Performs matching when the rows of similarity matrix are non empty.
+            Returns:
+                matched_columns: An integer tensor of shape [num_rows] or [batch_size,
```
qlzh727, line 145: Since you expand the batch dim, lets just stay with [batch_size, num_rows] here.

keras_cv/ops/box_matcher_test.py
```diff
@@ -0,0 +1,59 @@
...
+            match_values=[-3, -2, -1, 1],
+        )
+        match_indices, matched_values = matcher(sim_matrix)
+        positive_matches = tf.greater_equal(matched_values, 0)
+        negative_matches = tf.equal(matched_values, -2)
+
+        self.assertAllEqual(positive_matches.numpy(), [[False, True]])
+        self.assertAllEqual(negative_matches.numpy(), [[True, False]])
+        self.assertAllEqual(match_indices.numpy(), [[0, 2]])
+        self.assertAllEqual(matched_values.numpy(), [[-2, 1]])
```
qlzh727, line 59: Can we add test coverage for:

1. number of row = 0
2. force_match_for_each_col = True
3. input validation for sort/length.

qlzh727 (changes requested):

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"Argmax-based box matching"
+
+from typing import List
```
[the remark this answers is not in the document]
tanzhenyu: Right, Ian asked the same question. I think we should just add type hints for ALL modules, this seems better practice to me (given we don't support python 2.x anyway due to TF version)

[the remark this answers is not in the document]
LukeWood: We've discussed type hints somewhat thoroughly as a group and generally opted against them - we should be consistent with Keras team and KerasNLP

keras_cv/ops/box_matcher.py
tanzhenyu: Done.

keras_cv/ops/box_matcher.py
tanzhenyu: responding to the questions below as well: row is always anchors, and col is always gt_boxes. We could evaluate swapping these two, but since this is the existing implementation for many models, I'd like to get this tested within KerasCV first, and then try to swap them and evaluate the performance difference.

keras_cv/ops/box_matcher.py
tanzhenyu: positive / ignore is a OD specific concept, and our example should reflect that.
Though this does present some confusion, so I updated the examples to reflect how 1) and 2) are different.

keras_cv/ops/box_matcher.py
tanzhenyu: Same response as above -- we can evaluate which has better performance once we have models using it. If there's no diff, then I can swap it.

keras_cv/ops/box_matcher.py
tanzhenyu: int being used for downstream logic, I think it should be handled fine in GPUs

keras_cv/ops/box_matcher.py
tanzhenyu: same respond as above -- evaluate the performance once we have models

keras_cv/ops/box_matcher.py
tanzhenyu: Done.

keras_cv/ops/box_matcher.py
tanzhenyu: Removed.

keras_cv/ops/box_matcher.py
tanzhenyu: it should be cols -- when gt_boxes are empty.

keras_cv/ops/box_matcher.py
tanzhenyu: Yes, that's why `static_shape` is being used here. Extra effort will be required to support dynamic shape.

keras_cv/ops/box_matcher.py
tanzhenyu: Done.

keras_cv/ops/box_matcher_test.py
tanzhenyu: Good point. Done.

keras_cv/ops/box_matcher.py
qlzh727: I think there will be numerical difference, eg the dim for arg max will be different (we always want to argmax for the a true box against all the generated anchor), so if user specify the wrong order, the result will be totally different (each of the achor is argmax match to all the true box.)

keras_cv/ops/box_matcher.py
tanzhenyu: Done.

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,235 @@
...
+    Usage:
+
+    ```python
+    box_matcher = keras_cv.ops.ArgmaxBoxMatcher([0.3, 0.7], [-1, 0, 1])
+    iou_metric = keras_cv.bounding_box.compute_iou(anchors, gt_boxes)
+    matched_columns, matched_match_values = box_matcher(iou_metric)
+    cls_mask = tf.less_equal(matched_match_values, 0)
+    ```
+
+    TODO(tanzhenyu): document when to use which mode.
```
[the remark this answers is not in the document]
tanzhenyu: Yes

keras_cv/ops/box_matcher.py
tanzhenyu: Yes the user will have to change it. That's why this is not exposed as public API yet (through __init__.py). We will need to test it with RetinaNet and RCNN

keras_cv/ops/box_matcher.py
qlzh727: I think my point is that static_shape could be None if you have dynamic batch size, and you end of concating a Tensor with a int, which will result into error. A unit test will reveal this.

keras_cv/ops/box_matcher_test.py
```diff
@@ -0,0 +1,127 @@
...
+from keras_cv.ops.box_matcher import ArgmaxBoxMatcher
+
+
+class ArgmaxBoxMatcherTest(tf.test.TestCase):
+    def test_box_matcher_invalid_length(self):
+        fg_threshold = 0.5
+        bg_thresh_hi = 0.2
+        bg_thresh_lo = 0.0
+
+        with self.assertRaises(ValueError):
```
qlzh727, line 26: let's also assert the error message as well.

keras_cv/ops/box_matcher_test.py
```diff
@@ -0,0 +1,127 @@
...
+                thresholds=[bg_thresh_lo, bg_thresh_hi, fg_threshold],
+                match_values=[-3, -2, -1],
+            )
+
+    def test_box_matcher_unsorted_thresholds(self):
+        fg_threshold = 0.5
+        bg_thresh_hi = 0.2
+        bg_thresh_lo = 0.0
+
+        with self.assertRaises(ValueError):
```
qlzh727, line 37: same here.

keras_cv/ops/box_matcher_test.py
```diff
@@ -0,0 +1,127 @@
...
+            _ = ArgmaxBoxMatcher(
+                thresholds=[bg_thresh_hi, bg_thresh_lo, fg_threshold],
+                match_values=[-3, -2, -1],
```
qlzh727, line 40: since you want to  assert the wrong order here, maybe provide a match_value with correct length.

keras_cv/ops/box_matcher.py
tanzhenyu: it has to be static_shape -- actual integers for now to work.

keras_cv/ops/box_matcher_test.py
```diff
@@ -0,0 +1,127 @@
...
+        match_indices, matched_values = matcher(sim_matrix)
+        positive_matches = tf.greater_equal(matched_values, 0)
+        negative_matches = tf.equal(matched_values, -2)
+
+        self.assertAllEqual(positive_matches.numpy(), [True, True, True, True])
+        self.assertAllEqual(negative_matches.numpy(), [False, False, False, False])
+        # the first anchor cannot be matched to 4th gt box given that is matched to
+        # the last anchor.
+        self.assertAllEqual(match_indices.numpy(), [1, 2, 0, 3])
+        self.assertAllEqual(matched_values.numpy(), [1, 1, 1, 1])
```
qlzh727, line 107: since we pick 0.04 for first row and 0.1 for the third row, shouldn't the match_values be [-2, 1, -2, 1] (based on the bucket)?

tanzhenyu: No, this is when each col needs a match, so it is all 1

keras_cv/ops/box_matcher.py
qlzh727: Ack.

keras_cv/ops/box_matcher_test.py
tanzhenyu: Done

keras_cv/ops/box_matcher_test.py
tanzhenyu: Done.

keras_cv/ops/box_matcher_test.py
tanzhenyu: Done.

keras_cv/ops/box_matcher.py
qlzh727: Ack, let's add a TODO here, since this will be hard to detect this kind of error, and result into a silent failure.

keras_cv/ops/box_matcher.py
```diff
@@ -0,0 +1,234 @@
...
+       values.
+    3) thresholds=[positive_threshold], and
+       match_values=[negative_values, positive_value]: the rows will be assigned to
+       positive value if its argmax result is above positive_threshold; the rows
+       will be assigned to negative_value if its argmax result is below
+       negative_threshold.
+
+    Args:
+        thresholds: A sorted list of floats to classify the matches into
+        different results (e.g. positive or negative or ignored match). The
```
qlzh727, line 54: we should indent here to make the start of each param more noticable.

keras_cv/ops/box_matcher.py
tanzhenyu: I will add a TODO before merging

qlzh727 (approved):

keras_cv/ops/box_matcher.py
tanzhenyu: I wonder why black or flake does not format this automatically.
Will indent before merging

keras_cv/ops/box_matcher.py
tanzhenyu: Isn't the TODO already added in previous commit?

PR #804, opened 2022-09-16
Add ROIGenerator.

tanzhenyu:
# What does this PR do?
Add ROIGenerator for RPN network.
This replaces the ROIFilter proposed by [RFC](https://github.com/keras-team/governance/blob/master/rfcs/20220804-keras-cv-two-stage-2d-object-detection.md)

### DIFF
--- a/keras_cv/layers/object_detection/roi_generator.py
+++ b/keras_cv/layers/object_detection/roi_generator.py
@@ -1,0 +1,183 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Mapping
+from typing import Optional
+from typing import Tuple
+
+import tensorflow as tf
+
+from keras_cv import bounding_box
+
+
+class ROIGenerator(tf.keras.layers.Layer):
+    """
+    Generates region of interests (ROI, or proposal) from scores.
+
+    This works for a multi-level input, both boxes and scores are dictionary
+    inputs with the same set of keys.
+
+    Users can configure topk and threshold differently in train and inference.
+
+    Users can choose to combine all levels if nms across all levels are desired.
+
+    The following steps are applied to pair of (boxes, scores):
+    1) pre_nms_topk scores and boxes sorted and selected per level
+    2) nms applied and selected post_nms_topk scores and rois per level
+    3) combined scores and rois across all levels
+    4) post_nms_topk scores and rois sorted and selected
+
+    Args:
+        bounding_box_format: a case-insensitive string which is one of `"xyxy"`,
+            `"rel_xyxy"`, `"xyWH"`, `"center_xyWH"`, `"yxyx"`, `"rel_yxyx"`. The
+            position and shape of the bounding box will be followed by the class and
+            confidence values (in that order). This is required for proper ranking of
+            the bounding boxes. Therefore, each bounding box is defined by 6 values.
+            For detailed information on the supported format, see the
+            [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+        pre_nms_topk_train: int. number of top k scoring proposals to keep before applying NMS in training mode.
+            When RPN is run on multiple feature maps / levels (as in FPN) this number is per
+            feature map / level.
+        nms_score_threshold_train: float. score threshold to use for NMS in training mode.
+        nms_iou_threshold_train: float. IOU threshold to use for NMS in training mode.
+        post_nms_topk_train: int. number of top k scoring proposals to keep after applying NMS in training mode.
+            When RPN is run on multiple feature maps / levels (as in FPN) this number is per
+            feature map / level.
+        pre_nms_topk_test: int. number of top k scoring proposals to keep before applying NMS in inference mode.
+            When RPN is run on multiple feature maps / levels (as in FPN) this number is per
+            feature map / level.
+        nms_score_threshold_test: float. score threshold to use for NMS in inference mode.
+        nms_iou_threshold_test: float. IOU threshold to use for NMS in inference mode.
+        post_nms_topk_test: int. number of top k scoring proposals to keep after applying NMS in inference mode.
+            When RPN is run on multiple feature maps / levels (as in FPN) this number is per
+            feature map / level.
+
+    """
+
+    def __init__(
+        self,
+        bounding_box_format,
+        pre_nms_topk_train: int = 2000,
+        nms_score_threshold_train: float = 0.0,
+        nms_iou_threshold_train: float = 0.7,
+        post_nms_topk_train: int = 1000,
+        pre_nms_topk_test: int = 1000,
+        nms_score_threshold_test: float = 0.0,
+        nms_iou_threshold_test: float = 0.7,
+        post_nms_topk_test: int = 1000,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.bounding_box_format = bounding_box_format
+        self.pre_nms_topk_train = pre_nms_topk_train
+        self.nms_score_threshold_train = nms_score_threshold_train
+        self.nms_iou_threshold_train = nms_iou_threshold_train
+        self.post_nms_topk_train = post_nms_topk_train
+        self.pre_nms_topk_test = pre_nms_topk_test
+        self.nms_score_threshold_test = nms_score_threshold_test
+        self.nms_iou_threshold_test = nms_iou_threshold_test
+        self.post_nms_topk_test = post_nms_topk_test
+        self.built = True
+
+    def call(
+        self,
+        multi_level_boxes: Mapping[str, tf.Tensor],
+        multi_level_scores: Mapping[str, tf.Tensor],
+        training: Optional[bool] = None,
+    ) -> Tuple[tf.Tensor, tf.Tensor]:
+        """
+        Args:
+          multi_level_boxes: float Tensor. A dictionary of boxes, one per level. shape is
+            [batch_size, num_boxes, 4] each level, in `bounding_box_format`.
+            The boxes from RPNs are usually encoded as deltas w.r.t to anchors,
+            they need to be decoded before passing in here.
+          multi_level_scores: float Tensor. A dictionary of scores, usually confidence scores,
+            one per level. shape is [batch_size, num_boxes] each level.
+
+        Returns:
+          rois: float Tensor of [batch_size, post_nms_topk, 4]
+          roi_scores: float Tensor of [batch_size, post_nms_topk]
+        """
+
+        if training:
+            pre_nms_topk = self.pre_nms_topk_train
+            post_nms_topk = self.post_nms_topk_train
+            nms_score_threshold = self.nms_score_threshold_train
+            nms_iou_threshold = self.nms_iou_threshold_train
+        else:
+            pre_nms_topk = self.pre_nms_topk_test
+            post_nms_topk = self.post_nms_topk_test
+            nms_score_threshold = self.nms_score_threshold_test
+            nms_iou_threshold = self.nms_iou_threshold_test
+
+        rois = []
+        roi_scores = []
+        for level in sorted(multi_level_scores.keys()):
+            scores = multi_level_scores[level]
+            boxes = multi_level_boxes[level]
+            _, num_boxes = scores.get_shape().as_list()
+            level_pre_nms_topk = min(num_boxes, pre_nms_topk)
+            level_post_nms_topk = min(num_boxes, post_nms_topk)
+            scores, sorted_indices = tf.nn.top_k(
+                scores, k=level_pre_nms_topk, sorted=True
+            )
+            boxes = tf.gather(boxes, sorted_indices, batch_dims=1)
+            # convert from input format to yxyx for the TF NMS operation
+            boxes = bounding_box.convert_format(
+                boxes,
+                source=self.bounding_box_format,
+                target="yxyx",
+            )
+            # TODO(tanzhenyu): consider supporting soft / batched nms for accl
+            selected_indices, num_valid = tf.image.non_max_suppression_padded(
+                boxes,
+                scores,
+                max_output_size=level_post_nms_topk,
+                iou_threshold=nms_iou_threshold,
+                score_threshold=nms_score_threshold,
+                pad_to_max_output_size=True,
+                sorted_input=True,
+                canonicalized_coordinates=True,
+            )
+            # convert back to input format
+            boxes = bounding_box.convert_format(
+                boxes,
+                source="yxyx",
+                target=self.bounding_box_format,
+            )
+            level_rois = tf.gather(boxes, selected_indices, batch_dims=1)
+            level_roi_scores = tf.gather(scores, selected_indices, batch_dims=1)
+            level_rois = level_rois * tf.cast(
+                tf.reshape(tf.range(level_post_nms_topk), [1, -1, 1])
+                < tf.reshape(num_valid, [-1, 1, 1]),
+                level_rois.dtype,
+            )
+            level_roi_scores = level_roi_scores * tf.cast(
+                tf.reshape(tf.range(level_post_nms_topk), [1, -1])
+                < tf.reshape(num_valid, [-1, 1]),
+                level_roi_scores.dtype,
+            )
+            rois.append(level_rois)
+            roi_scores.append(level_roi_scores)
+
+        rois = tf.concat(rois, axis=1)
+        roi_scores = tf.concat(roi_scores, axis=1)
+        _, num_valid_rois = roi_scores.get_shape().as_list()
+        overall_top_k = min(num_valid_rois, post_nms_topk)
+        roi_scores, sorted_indices = tf.nn.top_k(
+            roi_scores, k=overall_top_k, sorted=True
+        )
+        rois = tf.gather(rois, sorted_indices, batch_dims=1)
+
+        return rois, roi_scores
--- a/keras_cv/layers/object_detection/roi_generator.py
+++ b/keras_cv/layers/object_detection/roi_generator.py
@@ -25,6 +25,8 @@ class ROIGenerator(tf.keras.layers.Layer):
     """
     Generates region of interests (ROI, or proposal) from scores.
 
+    Mainly used in Region CNN (RCNN) networks.
+
     This works for a multi-level input, both boxes and scores are dictionary
     inputs with the same set of keys.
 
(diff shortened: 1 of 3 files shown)


keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,183 @@
...
+from typing import Tuple
+
+import tensorflow as tf
+
+from keras_cv import bounding_box
+
+
+class ROIGenerator(tf.keras.layers.Layer):
+    """
+    Generates region of interests (ROI, or proposal) from scores.
```
LukeWood, line 26: mainly out of my own curiosity - can we put references here covering which types of models use this?

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,183 @@
...
+    Users can choose to combine all levels if nms across all levels are desired.
+
+    The following steps are applied to pair of (boxes, scores):
+    1) pre_nms_topk scores and boxes sorted and selected per level
+    2) nms applied and selected post_nms_topk scores and rois per level
+    3) combined scores and rois across all levels
+    4) post_nms_topk scores and rois sorted and selected
+
+    Args:
+        bounding_box_format: a case-insensitive string which is one of `"xyxy"`,
```
LukeWood, line 42: I think linking back to the documentation is the best approach here - mainly because the format list is expected to grow going forward.

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,183 @@
...
+            `"rel_xyxy"`, `"xyWH"`, `"center_xyWH"`, `"yxyx"`, `"rel_yxyx"`. The
+            position and shape of the bounding box will be followed by the class and
+            confidence values (in that order). This is required for proper ranking of
+            the bounding boxes. Therefore, each bounding box is defined by 6 values.
+            For detailed information on the supported format, see the
+            [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+        pre_nms_topk_train: int. number of top k scoring proposals to keep before applying NMS in training mode.
+            When RPN is run on multiple feature maps / levels (as in FPN) this number is per
+            feature map / level.
+        nms_score_threshold_train: float. score threshold to use for NMS in training mode.
```
LukeWood, line 52: so almost all of these parameters are passed into a NMS layer.

My intuition is that wrapping the underlying op (in this case tf.image.combine_non_max_suppression_padded) in a layer and accepting a layer to the constructor instead will produce the best API.

We did this for: https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/object_detection/non_max_suppression.py and it still supports full customization, but it eliminates these length constructor calls - and separates the customization into a logical block.

So in this case, you'd take two layers: a pre-nms-layer and a test-nms layer.

WDYT?

LukeWood (changes requested):
Looks good! 

One big comment: the constructor has a lot of parameters that are basically passed to a single OP.  

Perhaps we should wrap these in a layer and then take these layers in the constructor like we do with the decoder.  Then we can expose full customization, while still keeping the API readable

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Hmmm...not sure if I understand the comment -- the object_detection/non_max_suppression also takes arguments that control nms: `confidence_threshold`, `iou_threshold`, `max_detections`, `max_detections_per_class`, isn't it the same here?

If we do think this is too much, we can follow an established pattern which is:
`nms_iou_threshold`: List[float] -- a list of size 2 floats, the first for train, the second for test.
So we have 4, instead of 8 arguments.

I choose the existing approach mainly that I want to allow a simpler tuner integration, i.e., users can tune train and test configs separately. But I'm ok if we combine them into list.

WDYT?

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Done. 
(The name should be self-explanatory for applied ML users though)

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Sounds good, Done.
Then we need to change that across the repo as well

LukeWood: yeah, i agree.  Let’s do this during the preprocessing layer fixit.  I’ll make sure to list checking bounding box format docstrings  as one of the fixit tasks.

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,181 @@
...
+class ROIGenerator(tf.keras.layers.Layer):
+    """
+    Generates region of interests (ROI, or proposal) from scores.
+
+    Mainly used in Region CNN (RCNN) networks.
+
+    This works for a multi-level input, both boxes and scores are dictionary
+    inputs with the same set of keys.
+
+    Users can configure topk and threshold differently in train and inference.
```
qlzh727, line 33: nit: maybe top_k?

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Users can choose to combine all levels if nms across all levels are desired.
```
qlzh727, line 35: NMS, or non-max surpression

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,181 @@
...
+
+    The following steps are applied to pair of (boxes, scores):
+    1) pre_nms_topk scores and boxes sorted and selected per level
+    2) nms applied and selected post_nms_topk scores and rois per level
```
qlzh727, line 39: nit: ROIs

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,181 @@
...
+    4) post_nms_topk scores and rois sorted and selected
+
+    Args:
+        bounding_box_format: a case-insensitive string.
+            For detailed information on the supported format, see the
+            [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+        pre_nms_topk_train: int. number of top k scoring proposals to keep before applying NMS in training mode.
+            When RPN is run on multiple feature maps / levels (as in FPN) this number is per
+            feature map / level.
+        nms_score_threshold_train: float. score threshold to use for NMS in training mode.
```
qlzh727, line 50: Param order wise, should we pair all the train/test param together? eg pre_nms_topk_train followed by pre_nms_topk_test

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,181 @@
...
+    2) nms applied and selected post_nms_topk scores and rois per level
+    3) combined scores and rois across all levels
+    4) post_nms_topk scores and rois sorted and selected
+
+    Args:
+        bounding_box_format: a case-insensitive string.
+            For detailed information on the supported format, see the
+            [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+        pre_nms_topk_train: int. number of top k scoring proposals to keep before applying NMS in training mode.
+            When RPN is run on multiple feature maps / levels (as in FPN) this number is per
```
qlzh727, line 48: Is it a map of a int when in multiple feature level? or it is still a int? (same for all the other params.)

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,181 @@
...
+            feature map / level.
+        pre_nms_topk_test: int. number of top k scoring proposals to keep before applying NMS in inference mode.
+            When RPN is run on multiple feature maps / levels (as in FPN) this number is per
+            feature map / level.
+        nms_score_threshold_test: float. score threshold to use for NMS in inference mode.
+        nms_iou_threshold_test: float. IOU threshold to use for NMS in inference mode.
+        post_nms_topk_test: int. number of top k scoring proposals to keep after applying NMS in inference mode.
+            When RPN is run on multiple feature maps / levels (as in FPN) this number is per
+            feature map / level.
+
```
qlzh727, line 63: Do u have a sample code to create and call the ROI?

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,181 @@
...
+        self.post_nms_topk_train = post_nms_topk_train
+        self.pre_nms_topk_test = pre_nms_topk_test
+        self.nms_score_threshold_test = nms_score_threshold_test
+        self.nms_iou_threshold_test = nms_iou_threshold_test
+        self.post_nms_topk_test = post_nms_topk_test
+        self.built = True
+
+    def call(
+        self,
+        multi_level_boxes: Mapping[str, tf.Tensor],
```
qlzh727, line 93: Should we unify the design here for the key of the mapping with the one in my PR for FPN? eg i was using the int key as the level, so that we can sort this properly (string might have the issue for '10' comes before '2'), and also using the key for loops.

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,181 @@
...
+        rois = tf.concat(rois, axis=1)
+        roi_scores = tf.concat(roi_scores, axis=1)
+        _, num_valid_rois = roi_scores.get_shape().as_list()
+        overall_top_k = min(num_valid_rois, post_nms_topk)
+        roi_scores, sorted_indices = tf.nn.top_k(
+            roi_scores, k=overall_top_k, sorted=True
+        )
+        rois = tf.gather(rois, sorted_indices, batch_dims=1)
+
+        return rois, roi_scores
```
qlzh727, line 181: do u need a get_config() for serialization?

qlzh727 (changes requested):
Took a quick pass first.

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Ok, in that case, let's unify with what you have, i.e., int instead of str. Will update.

keras_cv/layers/object_detection/roi_generator.py
LukeWood: > Hmmm...not sure if I understand the comment -- the object_detection/non_max_suppression also takes arguments that control nms: `confidence_threshold`, `iou_threshold`, `max_detections`, `max_detections_per_class`, isn't it the same here?
> 
> If we do think this is too much, we can follow an established pattern which is: `nms_iou_threshold`: List[float] -- a list of size 2 floats, the first for train, the second for test. So we have 4, instead of 8 arguments.
> 
> I choose the existing approach mainly that I want to allow a simpler tuner integration, i.e., users can tune train and test configs separately. But I'm ok if we combine them into list.
> 
> WDYT?

> Hmmm...not sure if I understand the comment -- the object_detection/non_max_suppression also takes arguments that control nms: confidence_threshold, iou_threshold, max_detections, max_detections_per_class, isn't it the same here?


it is the same, I'm proposing we accept an actual NMS layer as a parameter and produce a default once instead of having all of these prefixed layers.

LukeWood: (so you'd take two layers instead of a list of two floats)

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: should be int as this applied both per level and across all levels.

keras_cv/layers/object_detection/roi_generator.py
LukeWood: Here is an example of a comparable API:

https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/object_detection/nms_prediction_decoder.py#L35

we take a suppression_layer instead of a list of all of the parameters to pass to the layer.

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Two options that I see:
1. group all train args, then all test args, as users mostly care about one thing or another at one time
2. let all args be list of [train_arg, test_arg]

Personally I think 1 is better for KerasTuner.

qlzh727: 2 will be more confusing, let's stay with 1 for now.

keras_cv/layers/object_detection/roi_generator_test.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        # selecting the 1st, then 3rd, then 2nd as they don't overlap
+        # 0th box overlaps with 1st box
+        expected_roi_scores = tf.gather(rpn_scores, [[1, 3, 2]], batch_dims=1)
+        expected_roi_scores = tf.concat([expected_roi_scores, tf.zeros([1, 1])], axis=1)
+        rpn_scores = {"2": rpn_scores}
+        rois, roi_scores = roi_generator(rpn_boxes, rpn_scores, training=True)
+        self.assertAllClose(expected_rois, rois)
+        self.assertAllClose(expected_roi_scores, roi_scores)
+        # print(rois)
```
qlzh727, line 44: let's remove the print here (I assume for debug.)

keras_cv/layers/object_detection/roi_generator_test.py
```diff
@@ -0,0 +1,178 @@
...
+        # print(expected_rois)
+        # print(roi_scores)
+        # print(expected_roi_scores)
+
+    def test_singlelevel_singlebatch_roi_all_box(self):
+        roi_generator = ROIGenerator("xyxy", nms_iou_threshold_train=0.97)
```
qlzh727, line 50: for the 0.97/0.96 iou score, we might need some comment for them, since it need some calculation to come up the IOU score between [0, 0, 10, 10] and [0.1, 0.1, 9.9, 9.9]

keras_cv/layers/object_detection/roi_generator_test.py
```diff
@@ -0,0 +1,178 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv.layers.object_detection.roi_generator import ROIGenerator
+
+
+class ROIGeneratorTest(tf.test.TestCase):
+    def test_singlelevel_singlebatch_roi_ignore_box(self):
```
qlzh727, line 21: maybe "single_level" and "single_batch"?

qlzh727 (changes requested):

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Done

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Done

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_generator_test.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_generator_test.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_generator_test.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Good catch. Done.

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: OIC. This makes sense. Though for this particular layer, I'd rather not accept *any* nms layer as I want to make sure this works on GPU/TPU -- NMS is very tricky to work across platforms.

keras_cv/layers/object_detection/roi_generator.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def get_config(self):
+        config = {
+            "bounding_box_format": self.bounding_box_format,
+            "pre_nms_topk_train": self.pre_nms_topk_train,
+            "nms_score_threshold_train": self.nms_score_threshold_train,
+            "nms_iou_threshold_train": self.nms_iou_threshold_train,
+            "post_nms_topk_train": self.post_nms_topk_train,
+            "pre_nms_topk_train": self.pre_nms_topk_test,
```
qlzh727, line 198: copy-paste typo, should be "pre_nms_topk_test". Probably need a test coverage for this.

qlzh727 (changes requested):
The format test is also failing.

keras_cv/layers/object_detection/roi_generator.py
tanzhenyu: Corrected here -- maybe we should have a keras.util_test kind of serialization test then?

qlzh727 (approved):

keras_cv/layers/object_detection/roi_generator.py
qlzh727: we do have one in https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/serialization_test.py

keras_cv/layers/object_detection/roi_generator.py
LukeWood: I see.  Maybe we add a TODO to accept a padded NMS layer then?

My main concern with this thread is the API surface.  Accepting this huge list of nms- prefixed parameters is a bit hard to read - but shouldn't be considered blocking.

LukeWood:
so if padded works on TPU, can we still expose a PaddedNMS layer like we do here?   https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/object_detection/non_max_suppression.py#L86

I still believe this to be a much nicer API for ROIGenerator.  If we want to ship the model w/o vetting the API for ROIGenerator, can we do so from an __internal__ namespace and simply construct the ROIGenerator in the model?

LukeWood (changes requested):
I think we should either create a PaddedNMSLayer that performs NMS in the padded setting, and accept two of these in the constructor - or ship this as an __internal__ API because I'm fairly sure that we will ship a PaddedNMSLayer some point, and we should design around this accordingly.

tanzhenyu:
> https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/object_detection/non_max_suppression.py#L86

Being able to work on TPU is more than just padding things. Most of our users don't know which tf nms ops work with which device, and it's likely many experienced TF developers don't know that either. What makes it even more challenging is that this list keeps changing, and it's KerasCV's job to keep aligning with TF version to handle that, not users' job. 
Another reason why I don't want to provide a layer argument for ROIGenerator is, 4 out of the 8 arguments will be required for the layer even if we add the layer argument. So we're really just shrinking it from 8 arguments to 6 arguments (need layer for training and layer for testing).

The example [provided above](https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/object_detection/non_max_suppression.py#L86) does not work with TPU, IIUC. So as mentioned above, we should help users avoid these pitfalls if possible.

LukeWood (approved):
Alright, after going back and forth a bit I’m convinced that an exception is fine here. I think api divergence is a bit of an issue, but I’m happy to move forward with this API as I think it’s just a design choice.

LGTM

tanzhenyu:
#https://github.com/keras-team/keras-cv/pull/811

tanzhenyu:
> Alright, after going back and forth a bit I’m convinced that an exception is fine here. I think api divergence is a bit of an issue, but I’m happy to move forward with this API as I think it’s just a design choice.
> 
> LGTM

I'm cross referencing this issue so that we don't forget to update here when the 811 PR proposal is passed.

PR #805, opened 2022-09-16
added the implementations of renset 18 and 34

zaccharieramzi:
# What does this PR do?

Fixes [# (issue)](https://github.com/keras-team/keras-applications/issues/151)

## Before submitting
      Pull Request section?
- [x] Was this discussed/approved via a Github issue? Not via a GitHub issue per se, but rather on a PR in keras : https://github.com/keras-team/keras/pull/16363#issuecomment-1227552253
- [ ] Did you write any new necessary tests? WIP
- [ ] If this adds a new model, can you run a few training steps on TPU in Colab to ensure that no XLA incompatible OP are used? WIP

## Who can review?

### DIFF
--- a/keras_cv/models/resnet_v1.py
+++ b/keras_cv/models/resnet_v1.py
@@ -24,6 +24,16 @@ from tensorflow.keras import layers
 from keras_cv.models import utils
 
 MODEL_CONFIGS = {
+    "ResNet18": {
+        "stackwise_filters": [64, 128, 256, 512],
+        "stackwise_blocks": [2, 2, 2, 2],
+        "stackwise_strides": [1, 2, 2, 2],
+    },
+    "ResNet34": {
+        "stackwise_filters": [64, 128, 256, 512],
+        "stackwise_blocks": [3, 4, 6, 3],
+        "stackwise_strides": [1, 2, 2, 2],
+    },
     "ResNet50": {
         "stackwise_filters": [64, 128, 256, 512],
         "stackwise_blocks": [3, 4, 6, 3],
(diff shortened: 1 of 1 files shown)


LukeWood:
Thanks for the pull request!! Any chance we can do this for ResNetV2 instead of v1?

zaccharieramzi:
Sure no prob, let me give it a shot after this weekend

keras_cv/models/resnet_v2.py
```diff
@@ -86,6 +96,62 @@
...
+        x = layers.BatchNormalization(
+            axis=BN_AXIS, epsilon=1.001e-5, name=name + "_1_bn"
+        )(x)
+        x = layers.Activation("relu", name=name + "_1_relu")(x)
+
+        x = layers.ZeroPadding2D(padding=((1, 1), (1, 1)), name=name + "_2_pad")(x)
+        x = layers.Conv2D(
+            filters,
+            kernel_size,
+            strides=stride,
```
zaccharieramzi, line 144: @LukeWood here I am not sure about this design choice.
Indeed in V1, I know that the strides are supposed to be used in the first convolution in both the basic block and the bottleneck one.
However, here in the bottleneck block, the strides are used in the 2nd (middle) convolution. I didn't see anywhere in [the corresponding paper](https://arxiv.org/abs/1603.05027) where such a change was mentioned, so I didn't really know how to adapt to the basic block case.

Tell me if all makes sense

keras_cv/models/resnet_v2.py
```diff
@@ -86,6 +96,62 @@
...
+            filters, kernel_size, padding="SAME", strides=1, use_bias=False, name=name + "_1_conv"
+        )(
+            use_preactivation
+        )
+        x = layers.BatchNormalization(
+            axis=BN_AXIS, epsilon=1.001e-5, name=name + "_1_bn"
+        )(x)
+        x = layers.Activation("relu", name=name + "_1_relu")(x)
+
+        x = layers.ZeroPadding2D(padding=((1, 1), (1, 1)), name=name + "_2_pad")(x)
```
zaccharieramzi, line 140: Here I am not sure I understand why we want manual padding, instead of specifying the padding type in the conv right after.

keras_cv/models/resnet_v1.py
```diff
@@ -141,23 +200,32 @@ def apply(x):
...
+      block_type: string, "block" or "basic_block".
+      first_shortcut: default True, use convolution shortcut if True,
+          otherwise identity shortcut.
     Returns:
       Output tensor for the stacked blocks.
     """
     if name is None:
         name = f"v1_stack_{backend.get_uid('v1_stack')}"
 
+    block_fn = Block if block_type == "block" else BasicBlock
```
LukeWood, line 221: lets change this to just accept block_fn as an argument instead of block_type.

Then you can write Stack(block_fn=BasicBlock)

LukeWood: alternatively explicitly iterate the acceptable parameters for block_type

LukeWood (changes requested):
one minor change but overall this looks pretty good!  Thanks!

keras_cv/models/resnet_v1.py
zaccharieramzi: sure, corrected that!

LukeWood (approved):
LGTM

LukeWood:
Hey @zaccharieramzi thanks for the great PR!  I'm going to merge this, feel free to revisit the zero padding in the conv in a follow up PR.  this is in fact a performance hit to do this, and so we will be more performant if you fix it.

In the meantime I'm actually going to play with ResNet18 a bit using TensorFlow similarity for contrastive training and verify these can converge on great scores (and hopefully upload some weights)!  thanks!

zaccharieramzi:
@LukeWood thanks!! Really happy this went through!

Sure, I will open the follow-up PR on the zero-padding.
What about the question on the strides?

CRosero:
Any updates regarding the ImageNet weights?

PR #808, opened 2022-09-20
Add ROIPooler layer for RCNN.

tanzhenyu:
# What does this PR do?
Add ROIPooler layer for FasterRCNN

### DIFF
--- a/keras_cv/layers/object_detection/roi_pool.py
+++ b/keras_cv/layers/object_detection/roi_pool.py
@@ -32,12 +32,13 @@ class ROIPooler(tf.keras.layers.Layer):
             For detailed information on the supported format, see the
             [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
         target_size: List or Tuple of 2 integers of the pooled shape
-        image_shape: List of Tuple of 3 integers, or `TensorShape` of the image shape.
+        image_shape: List of Tuple of 3 integers, or `TensorShape` of the input image shape.
 
     Usage:
     ```python
     feature_map = tf.random.normal([2, 16, 16, 512])
-    roi_pooler = ROIPooler("yxyx", [7, 7], [224, 224, 3])
+    roi_pooler = ROIPooler(bounding_box_format="yxyx", target_size=[7, 7],
+      image_shape=[224, 224, 3])
     rois = tf.constant([[[15., 30., 25., 45.]], [[22., 1., 30., 32.]]])
     pooled_feature_map = roi_pooler(feature_map, rois)
     ```
@@ -96,6 +97,7 @@ class ROIPooler(tf.keras.layers.Layer):
         feature_map, rois = args
         num_rois = rois.get_shape().as_list()[0]
         height, width, channel = feature_map.get_shape().as_list()
+        # TODO (consider vectorize it for better performance)
         for n in range(num_rois):
             # [4]
             roi = rois[n, :]
--- a/keras_cv/layers/object_detection/roi_pool.py
+++ b/keras_cv/layers/object_detection/roi_pool.py
@@ -1,0 +1,143 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from email.mime import image
+
+import tensorflow as tf
+
+from keras_cv import bounding_box
+
+
+class ROIPooler(tf.keras.layers.Layer):
+    """
+    Pooling feature map of dynamic shape into region of interest (ROI) of fixed shape.
+
+    Mainly used in Region CNN (RCNN) networks. This works for a single-level
+    input feature map.
+
+    This layer splits the feature map into [target_size[0], target_size[1]] areas,
+    and performs max pooling for each area. The area coordinates will be quantized.
+
+    Args:
+        bounding_box_format: a case-insensitive string.
+            For detailed information on the supported format, see the
+            [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+        target_size: List or Tuple of 2 integers of the pooled shape
+        image_shape: List of Tuple of 3 integers, or `TensorShape` of the image shape.
+
+    Usage:
+    ```python
+    feature_map = tf.random.normal([2, 16, 16, 512])
+    roi_pooler = ROIPooler("yxyx", [7, 7], [224, 224, 3])
+    rois = tf.constant([[[15., 30., 25., 45.]], [[22., 1., 30., 32.]]])
+    pooled_feature_map = roi_pooler(feature_map, rois)
+    ```
+    """
+
+    def __init__(
+        self,
+        bounding_box_format,
+        target_size,
+        image_shape,
+        **kwargs,
+    ):
+        if not isinstance(target_size, (tuple, list)):
+            raise ValueError(
+                f"Expected `target_size` to be tuple or list, got {type(target_size)}"
+            )
+        if len(target_size) != 2:
+            raise ValueError(
+                f"Expected `target_size` to be size 2, got {len(target_size)}"
+            )
+        super().__init__(**kwargs)
+        self.bounding_box_format = bounding_box_format
+        self.target_height = target_size[0]
+        self.target_width = target_size[1]
+        self.image_shape = image_shape
+        self.built = True
+
+    def call(self, feature_map, rois):
+        """
+        Args:
+          feature_map: [batch_size, H, W, C] float Tensor, the feature map extracted from image.
+          rois: [batch_size, N, 4] float Tensor, the region of interests to be pooled.
+        Returns:
+          pooled_feature_map: [batch_size, N, target_size, C] float Tensor
+        """
+        # convert to relative format given feature map shape != image shape
+        rois = bounding_box.convert_format(
+            rois,
+            source=self.bounding_box_format,
+            target="rel_yxyx",
+            image_shape=self.image_shape,
+        )
+        pooled_feature_map = tf.vectorized_map(
+            self._pool_single_sample, (feature_map, rois)
+        )
+        return pooled_feature_map
+
+    def _pool_single_sample(self, args):
+        """
+        Args: tuple of
+          feature_map: [H, W, C] float Tensor
+          rois: [N, 4] float Tensor
+        Returns:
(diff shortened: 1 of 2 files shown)


keras_cv/layers/object_detection/roi_pool.py
```diff
@@ -0,0 +1,141 @@
...
+    def _pool_single_sample(self, args):
+        """
+        Args: tuple of
+          feature_map: [H, W, C] float Tensor
+          rois: [N, 4] float Tensor
+        Returns:
+          pooled_feature_map: [target_size, C] float Tensor
+        """
+        feature_map, rois = args
+        N = rois.get_shape().as_list()[0]
```
LukeWood, line 97: throughout the codebase we've been preferring name like:

```
N => num_rois
H => height
W => width
```

This is in-line with the style of Keras core.  No particularly strong opinion but this has been the case so far.

keras_cv/layers/object_detection/roi_pool_test.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+        self.assertAllClose(expected_feature_map, pooled_feature_map)
+
+    def test_roi_quantize_w(self):
+        roi_pooler = ROIPooler("yxyx", target_size=[2, 3], image_shape=[224, 224, 3])
+        feature_map = tf.expand_dims(tf.reshape(tf.range(64), [8, 8, 1]), axis=0)
+        rois = tf.reshape(tf.constant([0.0, 0.0, 224, 224]), [1, 1, 4])
+        pooled_feature_map = roi_pooler(feature_map, rois)
+        # the maximum value would be at bottom-right at each block, roi sharded into 2x3 blocks
+        # | 0, 1        | 2, 3, 4           | 5, 6, 7           |
```
LukeWood, line 107: I really like the tables here to visualize the expected outputs.  Really helpful diagrams - thanks

LukeWood (changes requested):
minor stylistic changes but LGTM.  Nothing blocking on this PR - I really like the diagrams.

keras_cv/layers/object_detection/roi_pool.py
tanzhenyu: Sticking to that style then.
Done.

keras_cv/layers/object_detection/roi_pool.py
```diff
@@ -0,0 +1,141 @@
...
+        bounding_box_format: a case-insensitive string.
+            For detailed information on the supported format, see the
+            [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+        target_size: List or Tuple of 2 integers of the pooled shape
+        image_shape: List of Tuple of 3 integers, or `TensorShape` of the image shape.
+
+    Usage:
+    ```python
+    feature_map = tf.random.normal([2, 16, 16, 512])
+    roi_pooler = ROIPooler("yxyx", [7, 7], [224, 224, 3])
```
qlzh727, line 40: let's add all the param names to provide more readability.

keras_cv/layers/object_detection/roi_pool.py
```diff
@@ -0,0 +1,141 @@
...
+    rois = tf.constant([[[15., 30., 25., 45.]], [[22., 1., 30., 32.]]])
+    pooled_feature_map = roi_pooler(feature_map, rois)
+    ```
+    """
+
+    def __init__(
+        self,
+        bounding_box_format,
+        target_size,
```
qlzh727, line 49: naming wise, I think we should consolidate "size" and "shape".

keras_cv/layers/object_detection/roi_pool.py
```diff
@@ -0,0 +1,141 @@
...
+
+    This layer splits the feature map into [target_size[0], target_size[1]] areas,
+    and performs max pooling for each area. The area coordinates will be quantized.
+
+    Args:
+        bounding_box_format: a case-insensitive string.
+            For detailed information on the supported format, see the
+            [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+        target_size: List or Tuple of 2 integers of the pooled shape
+        image_shape: List of Tuple of 3 integers, or `TensorShape` of the image shape.
```
qlzh727, line 35: From the docstring, its unclear to me what's the "image" in this context.

keras_cv/layers/object_detection/roi_pool.py
```diff
@@ -0,0 +1,141 @@
...
+        num_rois = rois.get_shape().as_list()[0]
+        height, width, channel = feature_map.get_shape().as_list()
+        for n in range(num_rois):
+            # [4]
+            roi = rois[n, :]
+            y_start = height * roi[0]
+            x_start = width * roi[1]
+            region_height = height * (roi[2] - roi[0])
+            region_width = width * (roi[3] - roi[1])
+            h_step = region_height / self.target_height
```
qlzh727, line 106: The step here can be floats I think. do u want to cast them to ints?

keras_cv/layers/object_detection/roi_pool.py
```diff
@@ -0,0 +1,141 @@
...
+            w_step = region_width / self.target_width
+            regions = []
+            for i in range(self.target_height):
```
qlzh727, line 109: The double for loop here might be a performance bottle neck, and we might need a TODO to vectorize this computation.

keras_cv/layers/object_detection/roi_pool.py
```diff
@@ -0,0 +1,141 @@
...
+          rois: [batch_size, N, 4] float Tensor, the region of interests to be pooled.
+        Returns:
+          pooled_feature_map: [batch_size, N, target_size, C] float Tensor
+        """
+        # convert to relative format given feature map shape != image shape
+        rois = bounding_box.convert_format(
+            rois,
+            source=self.bounding_box_format,
+            target="rel_yxyx",
+            image_shape=self.image_shape,
```
qlzh727, line 81: I guess this means that we need all the dims to be populated, maybe we should do a validation beforehand.

keras_cv/layers/object_detection/roi_pool_test.py
```diff
@@ -0,0 +1,161 @@
...
+        expected_feature_map = tf.reshape(tf.constant([27, 31, 59, 63]), [1, 2, 2, 1])
+        self.assertAllClose(expected_feature_map, pooled_feature_map)
+
+    def test_roi_quantize_y(self):
+        roi_pooler = ROIPooler("yxyx", target_size=[2, 2], image_shape=[224, 224, 3])
+        feature_map = tf.expand_dims(tf.reshape(tf.range(64), [8, 8, 1]), axis=0)
+        rois = tf.reshape(tf.constant([0.0, 0.0, 224, 220]), [1, 1, 4])
+        pooled_feature_map = roi_pooler(feature_map, rois)
+        # the maximum value would be at bottom-right at each block, roi sharded into 2x2 blocks
+        # | 0, 1, 2             | 3, 4, 5, 6            |
```
qlzh727, line 48: To complete the whole graph, I guess there is another column (7, 15, 23....) on the right most side, that is cut off as the round off value.

keras_cv/layers/object_detection/roi_pool_test.py
```diff
@@ -0,0 +1,161 @@
...
+        self.assertAllClose(expected_feature_map, pooled_feature_map)
+
+    def test_roi_quantize_h(self):
+        roi_pooler = ROIPooler("yxyx", target_size=[3, 2], image_shape=[224, 224, 3])
+        feature_map = tf.expand_dims(tf.reshape(tf.range(64), [8, 8, 1]), axis=0)
+        rois = tf.reshape(tf.constant([0.0, 0.0, 224, 224]), [1, 1, 4])
+        pooled_feature_map = roi_pooler(feature_map, rois)
+        # the maximum value would be at bottom-right at each block, roi sharded into 3x2 blocks
+        # | 0, 1, 2, 3          | 4, 5, 6, 7            |
+        # | 8, 9, 10, 11        | 12, 13, 14, 15        |
```
qlzh727, line 86: missing max annotation for 11 and 15 here.

qlzh727 (changes requested):
THanks for the PR.

keras_cv/layers/object_detection/roi_pool.py
tanzhenyu: Hmm, I'd actually like to have size and shape here because one is a tuple, the other one can be a tuple as well as TensorShape
If we don't like this, I can make it as `target_height`, `target_width`, image_shape

keras_cv/layers/object_detection/roi_pool.py
tanzhenyu: input image shape then.

keras_cv/layers/object_detection/roi_pool.py
tanzhenyu: they can be and they are float here.

keras_cv/layers/object_detection/roi_pool.py
tanzhenyu: Done

keras_cv/layers/object_detection/roi_pool.py
tanzhenyu: a validation of what?

keras_cv/layers/object_detection/roi_pool.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_pool_test.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_pool_test.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_pool.py
qlzh727: eg, the image shape can be [None, None, 3] which will fail the conversion in this case.

keras_cv/layers/object_detection/roi_pool.py
qlzh727: i see. there is a int32 cast down below.

keras_cv/layers/object_detection/roi_pool.py
qlzh727: I think there was also another discussion about whether kpl should accept "size" vs "height + width", and we should consolidate with that as well.

keras_cv/layers/object_detection/roi_pool.py
qlzh727: There isn't a input image directly related this ROI layer, the inputs this layer get are feature map, and bbox. The image shape here is more for the conversion of bbox to required format, and we might want to state this clearly.

keras_cv/layers/object_detection/roi_pool.py
tanzhenyu: Done -- added a check and test

keras_cv/layers/object_detection/roi_pool.py
tanzhenyu: Added a TODO here

qlzh727 (approved):

qlzh727 (approved):

PR #810, opened 2022-09-20
Setup initial benchmarks for model training

ianstenbit:
This adds benchmarks for only two models for now because I want to make sure parameterizing the benchmarks works without adding dozens of benchmarks at once.

This is a simple training benchmark which loads MNIST via TFDS (works on Borg) and trains for one epoch, recording training time, train accuracy, and compilation time.

Once this is merged, I will update our internal configs to automatically pick up these benchmarks daily.

### DIFF
--- a/keras_cv/benchmarks/classification_training_benchmark_test.py
+++ b/keras_cv/benchmarks/classification_training_benchmark_test.py
@@ -50,9 +50,9 @@ class ClassificationTrainingBenchmark(
                 num_parallel_calls=tf.data.AUTOTUNE,
             )
             .batch(self.batch_size)
-        ).take(10)
+        )
         self.epochs = 1
-        self.strategy = tf.distribute.OneDeviceStrategy("device:GPU:0")
+        self.strategy = tf.distribute.MirroredStrategy()
 
     def benchmark_classification_training(self, app):
         with self.strategy.scope():
--- a/keras_cv/benchmarks/classification_training_benchmark_test.py
+++ b/keras_cv/benchmarks/classification_training_benchmark_test.py
@@ -1,0 +1,94 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


keras_cv/benchmarks/classification_training_benchmark_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+        metrics = []
+        metrics.append({"name": "compile_time", "value": compile_time})
+        metrics.append({"name": "avg_epoch_time", "value": training_time / self.epochs})
+        metrics.append({"name": "epochs", "value": self.epochs})
+        metrics.append(
+            {"name": "accuracy", "value": training_results.history["accuracy"][0]}
+        )
+
+        self.report_benchmark(wall_time=total_time, metrics=metrics)
```
tanzhenyu, line 93: out of curiosity -- will this fail if new benchmark is `worse` than previous benchmark? How do we set the tolerance diff?

keras_cv/benchmarks/classification_training_benchmark_test.py
```diff
@@ -0,0 +1,97 @@
...
+        self.strategy = tf.distribute.MirroredStrategy()
+
+    def benchmark_classification_training(self, app):
+        with self.strategy.scope():
+            t0 = time.time()
+
+            model = app(
+                include_top=True,
+                classes=self.num_classes,
+                input_shape=(56, 56, 1),
```
LukeWood, line 64: Am I wrong in thinking MNIST is 28x28?

LukeWood (commented):
All in all looks good to me structurally.

keras_cv/benchmarks/classification_training_benchmark_test.py
ianstenbit: The tolerances are all configured internally with our benchmark running config. I pinged you a link

keras_cv/benchmarks/classification_training_benchmark_test.py
LukeWood: Gotcha, good to know.  Thanks

LukeWood (approved):

tanzhenyu (approved):

PR #813, opened 2022-09-20
Add target gather as private API.

tanzhenyu:
# What does this PR do?
Added target_gather as private API for it to be used in both OD workflows.

### DIFF
--- a/keras_cv/ops/target_gather.py
+++ b/keras_cv/ops/target_gather.py
@@ -25,7 +25,8 @@ def _target_gather(
 ):
     """A utility function wrapping tf.gather, which deals with:
      1) both batched and unbatched `targets`
-     2) when batched `indices` have empty rows
+     2) when unbatched `targets` have empty rows, the result will be filled
+        with `mask_val`
      3) target masking.
 
     Args:
--- a/keras_cv/ops/target_gather.py
+++ b/keras_cv/ops/target_gather.py
@@ -1,0 +1,123 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+
+import tensorflow as tf
+
+
+def _target_gather(
+    targets: tf.Tensor,
+    indices: tf.Tensor,
+    mask: Optional[tf.Tensor] = None,
+    mask_val: Optional[float] = 0.0,
+):
+    """A utility function wrapping tf.gather, which deals with:
+     1) both batched and unbatched `targets`
+     2) when batched `indices` have empty rows
+     3) target masking.
+
+    Args:
+     targets: [N, ...] or [batch_size, N, ...] Tensor representing targets such
+       as boxes, keypoints, etc.
+     indices: [M] or [batch_size, M] int32 Tensor representing indices within
+       `targets` to gather.
+     mask: optional [M, ...] or [batch_size, M, ...] boolean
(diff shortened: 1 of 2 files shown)


LukeWood:
Looks like formatting is broken, but this looks good to me.

Can we just create a `keras_cv/models/object_detection/maskrcnn/` directory for all of these instead of having `_{component}` scattered across the repo?  That way we know exactly what pieces of MaskRCNN are still private

tanzhenyu:
> Looks like formatting is broken, but this looks good to me.
> 
> Can we just create a `keras_cv/models/object_detection/maskrcnn/` directory for all of these instead of having `_{component}` scattered across the repo? That way we know exactly what pieces of MaskRCNN are still private

Hmm...I'd like to use this in RetinaNet too

the formatting is somehow not deterministic...I will re-run it

keras_cv/ops/target_gather.py
```diff
@@ -0,0 +1,123 @@
...
+            else:
+                result = _gather_unbatched(
+                    tf.squeeze(labels, axis=0),
+                    tf.squeeze(match_indices, axis=0),
+                    None,
+                    mask_val,
+                )
+            return tf.expand_dims(result, axis=0)
+        else:
+            indices_shape = tf.shape(match_indices)
```
qlzh727, line 105: It seems that the batched implementation is much simpler than the unbatched implementation. Shouldn't we just tf.expand_dim(inputs, 0) and let it always go through the batched implementation?

LukeWood:
> > Looks like formatting is broken, but this looks good to me.
> > Can we just create a `keras_cv/models/object_detection/maskrcnn/` directory for all of these instead of having `_{component}` scattered across the repo? That way we know exactly what pieces of MaskRCNN are still private
> 
> Hmm...I'd like to use this in RetinaNet too
> 
> the formatting is somehow not deterministic...I will re-run it

Sounds good to me, fine by me.

LukeWood (approved):

keras_cv/ops/target_gather.py
tanzhenyu: I think the batched version is slightly more complicated? Otherwise the unbatched version is simple, but deals with one thing: empty rows

qlzh727: readability width, I think the batched version is more concise, and the unbatched version as a lot of redirection and wrapped functions. And I would prefer to have single code path to handle both batch and unbatched case.

keras_cv/ops/target_gather_test.py
```diff
@@ -0,0 +1,63 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv.ops.target_gather import _target_gather
+
+
+class TargetGatherTest(tf.test.TestCase):
```
ianstenbit, line 20: Should we add test coverage for the mask functionality as well?

ianstenbit (approved):
LGTM, but I think we might need a new test case for the `mask` behavior

keras_cv/ops/target_gather_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_target_gather_classes_unbatched(self):
+        target_classes = tf.constant([1, 2, 3, 4])
+        target_classes = target_classes[..., tf.newaxis]
+        indices = tf.constant([0, 2], dtype=tf.int32)
+        expected_classes = tf.constant([1, 3])
+        expected_classes = expected_classes[..., tf.newaxis]
+        res = _target_gather(target_classes, indices)
+        self.assertAllClose(expected_classes, res)
+
```
qlzh727, line 58: let's also add a test for N = 0 for targets.

keras_cv/ops/target_gather_test.py
tanzhenyu: Great catch!

keras_cv/ops/target_gather_test.py
tanzhenyu: Done.

qlzh727 (approved):

keras_cv/ops/target_gather.py
tanzhenyu: discussed offline, the unbatched version needs to deal with empty rows, where the batched version do not, so keeping the code path as is.

PR #814, opened 2022-09-21
Update docs for contrastive projector and probe

ianstenbit:
@tanzhenyu (based on your doc comments)


LukeWood (approved):
Good clarification, nice catch!  Thanks both

tanzhenyu (approved):

PR #815, opened 2022-09-21
Update current FPN for the extra level support. This is an extension that is used by the Retina-net application

qlzh727:
# What does this PR do?

Update current FPN for the extra level support. This is an extension that is used by the Retina-net application.

We could use this as the proper replacement for https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/object_detection/retina_net/__internal__/layers/feature_pyramid.py

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.


keras_cv/layers/feature_pyramid.py
```diff
@@ -66,6 +66,12 @@ class FeaturePyramid(tf.keras.layers.Layer):
             with feature inputs and merged result from upstream levels. Default to None,
             and a `keras.Conv2D` layer with kernel 3x3 will be created for each pyramid
             level.
+        num_extra_level: int, extra level of pyramid on top of existing `max_level`
```
ianstenbit, line 69: I've noticed we have some inconsistency in our API about how we handle parameters that represent a count of something

For example, here (and with `num_channels` above), we use the `num_` prefix, but for `classes` we exclude the `num_` prefix (and in the Grayscale preprocessing layer we use `output_channels`, not `num_output_channels`).

How would you feel about standardizing this to always exclude the `num_` prefix? Then we could call this `extra_levels` and rename `num_channels` to `channels`

qlzh727: Good question. I think the "num_" prefix is indicating that the param is a number (usually a positive int). If we are fine with using docstring/types to provide the type information, then I think its totally make sense to remove this prefix. @tanzhenyu  and @LukeWood

tanzhenyu: IMHO I think `num_classes` is better...I do imagine that there will be some APIs -- especially visualization related APIs that would require `classes` to be a dict between index and the class name.

keras_cv/layers/feature_pyramid.py
```diff
@@ -66,6 +66,12 @@ class FeaturePyramid(tf.keras.layers.Layer):
...
+            features. This is used in case like RetinaNet, described in
+            (https://arxiv.org/pdf/1708.02002). Those extra levels will use features
+            from the top most level as inputs.
```
tanzhenyu, line 72: Do we have cases where for non-consecutive levels, i.e., [2, 3, 5, 7]? It doesn't sounds useful, just want to make sure consecutive levels is our assumption for all our workflows.

keras_cv/layers/feature_pyramid.py
```diff
@@ -137,6 +147,19 @@ def __init__(
...
         self.merge_op = tf.keras.layers.Add()
 
+        if self.num_extra_level:
+            self.extra_level_layers = {}
+            # Note that the extra levels starts with self.max_level + 1
+            for i in range(self.num_extra_level):
+                level = i + self.max_level + 1
+                self.extra_level_layers[level] = tf.keras.layers.Conv2D(
+                    self.num_channels,
+                    kernel_size=3,
```
tanzhenyu, line 157: from RetinaNet, this seems to be the case. Should we allow configurable parameters for kernels size, conv type, etc??

keras_cv/layers/feature_pyramid.py
```diff
@@ -191,6 +219,25 @@ def build_feature_pyramid(self, input_features):
...
+            for i in range(self.num_extra_level):
+                level = i + self.max_level + 1
+                # For the first extra level, it will use the feature map input from the
+                # top most pyramid level
+                if level == self.max_level + 1:
+                    inputs = input_features[self.max_level]
+                else:
+                    inputs = output_features[level - 1]
+
+                # Apply activations to all the levels except for the first one
```
tanzhenyu, line 234: Hmmm...this seems quite specific, I wonder if it's possible to have a generic-fpn, and then a RetinaNetFPN (maybe even for a particular backbone) that deals with this?

tanzhenyu (commented):
Thanks for the PR!

keras_cv/layers/feature_pyramid.py
LukeWood: I agree with @tanzhenyu that num_classes is better, but @fchollet and I think a few others agreed classes is better.  The group is about 50/50 on that one

LukeWood: Zhenyu’s point regarding visualization tools needing a dictionary for classes  is a really good point.

keras_cv/layers/feature_pyramid.py
qlzh727: I didn't see any use case like that. Note that the size diff between each levels are 2x, which means if user config skip levels, the connections between those levels need to be customized as well.

keras_cv/layers/feature_pyramid.py
qlzh727: So far I didn't see any customization for this case, we could expose more params when we hit those use cases. Note that the `num_extra_level` is only used by the retina-net at the moment.

keras_cv/layers/feature_pyramid.py
qlzh727: We could choose to fit all the logic for `num_extra_level` to a separate class, as long as we have a good name for it. 

Note that the generic FPN is having `num_extra_level` = 0, which totally skip this logic.

keras_cv/layers/feature_pyramid.py
tanzhenyu: Ok sounds good -- maybe we should document that somewhere (also as a general principle maybe we find sometime to figure all the hidden implementation assumptions to put that in README).
For this PR, let's leave as-is.

keras_cv/layers/feature_pyramid.py
qlzh727: We could take that approach, or have a RetinatNetFPN subclass the FPN. This will make the extra logic more separate from the base FPN.

qlzh727:
Chatted with @tanzhenyu offline for this. We will park this PR for now since this is only going to be used by the Retina-net and not blocking the segmentation model. We will revisit this when we clean up the Retina net model code.

LukeWood:
> Chatted with @tanzhenyu offline for this. We will park this PR for now since this is only going to be used by the Retina-net and not blocking the segmentation model. We will revisit this when we clean up the Retina net model code.

Sounds good to me.

bhack:
> Chatted with @tanzhenyu offline for this. We will park this PR for now since this is only going to be used by the Retina-net and not blocking the segmentation model. We will revisit this when we clean up the Retina net model code.

Have you chatted about RetinMask for the segmentation to reuse the retinanet backbone?

bhack:
> Have you chatted about RetinaMask for the segmentation to reuse the same RetinaNet backbones?

See also the related thread we had with @tanzhenyu at https://github.com/keras-team/keras-cv/issues/920#issuecomment-1278279734

qlzh727:
SGTM.

PR #817, opened 2022-09-21
Add classification head for segmentation model

qlzh727:
# What does this PR do?
Add classification head for segmentation model

This will be the last piece before we ship the deeplab v3 with KCV components only.

## Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/keras_cv/models/segmentation/__internal__/segmentation_head.py
+++ b/keras_cv/models/segmentation/__internal__/segmentation_head.py
@@ -1,0 +1,30 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class SegmentationHead(layers.Layer):
+
+    def __init__(self, classes, level):
+        pass
+
+    def build(self, input_shape):
+        pass
+
+    def call(self, inputs):
+
+
--- a/keras_cv/models/segmentation/__internal__/segmentation_head.py
+++ b/keras_cv/models/segmentation/__internal__/segmentation_head.py
@@ -20,15 +20,15 @@ from tensorflow.keras import layers
 class SegmentationHead(layers.Layer):
     """Prediction head for the segmentation model
 
-    The head will take the output from decoder (eg FPN or ASPP), and produce production
-    (pixel level classifications) as the output for the model.
+    The head will take the output from decoder (eg FPN or ASPP), and produce a
+    segmentation mask (pixel level classifications) as the output for the model.
 
     Args:
         classes: int, the number of output classes for the prediction. This should
             include all the classes (eg background) for the model to predict.
         convs: int, the number of conv2D layers that are stacked before the final
             classification layer. Default to 2.
-        channels: int, the number of filter/channels for the the conv2D layers. Default
+        filters: int, the number of filter/channels for the the conv2D layers. Default
             to 256.
         activations: str or 'tf.keras.activations', activation functions between the
             conv2D layers and the final classification layer. Default to 'relu'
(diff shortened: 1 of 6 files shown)


keras_cv/models/segmentation/__internal__/segmentation_head.py
```diff
@@ -0,0 +1,109 @@
...
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class SegmentationHead(layers.Layer):
+    """Prediction head for the segmentation model
+
+    The head will take the output from decoder (eg FPN or ASPP), and produce production
+    (pixel level classifications) as the output for the model.
+
+    Args:
+        classes: int, the number of output classes for the prediction.
```
tanzhenyu, line 27: Can we help users by clarifying whether this will include background class? It's likely we require 1+num_foreground_classes here, but user pass num_foreground_classes?
(assuming we will make this as public API)

keras_cv/models/segmentation/__internal__/segmentation_head.py
```diff
@@ -0,0 +1,109 @@
...
+                tf.keras.layers.Conv2D(
+                    name=conv_name,
+                    filters=self.channels,
+                    kernel_size=3,
+                    padding="same",
+                    use_bias=False,
+                )
+            )
+            norm_name = "segmentation_head_norm_{}".format(i)
+            self._bn_layers.append(tf.keras.layers.BatchNormalization(name=norm_name))
```
tanzhenyu, line 72: should BN be after each conv layer?

keras_cv/models/segmentation/__internal__/segmentation_head.py
```diff
@@ -0,0 +1,109 @@
...
+            name="segmentation_output",
+            filters=self.classes,
+            kernel_size=1,
+            padding="same",
+        )
+
+    def call(self, inputs):
+        """Forward path for the segmentation head.
+
+        For now, it accepts the output from the decoder only, which is a dict with int
```
tanzhenyu, line 84: specify the shape of decoder output here?

tanzhenyu (commented):
Some minor comments.

keras_cv/models/segmentation/__internal__/segmentation_head.py
LukeWood: Good point - if this is only to be used in segmentation asking for foreground classes is a nice way to keep it clear what the user is to pass.  Another approach would be to call this output_dim instead of classes

qlzh727: Done.

keras_cv/models/segmentation/__internal__/segmentation_head.py
qlzh727: There is one BN per conv2D the stacking, the final classification conv doesn't have a BN.

keras_cv/models/segmentation/__internal__/segmentation_head.py
qlzh727: Oh, it is included in the docstring above (as the code example.)

keras_cv/models/segmentation/__internal__/segmentation_head.py
```diff
@@ -0,0 +1,110 @@
...
+
+    The head will take the output from decoder (eg FPN or ASPP), and produce production
+    (pixel level classifications) as the output for the model.
+
+    Args:
+        classes: int, the number of output classes for the prediction. This should
+            include all the classes (eg background) for the model to predict.
+        convs: int, the number of conv2D layers that are stacked before the final
+            classification layer. Default to 2.
+        channels: int, the number of filter/channels for the the conv2D layers. Default
```
LukeWood, line 31: Nit: Let’s rename to filters because that’s how it is named in conv2d

keras_cv/models/segmentation/__internal__/segmentation_head.py
```diff
@@ -0,0 +1,110 @@
...
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class SegmentationHead(layers.Layer):
+    """Prediction head for the segmentation model
+
+    The head will take the output from decoder (eg FPN or ASPP), and produce production
```
ianstenbit, line 23: nit: should this say "produce a segmentation mask"?

ianstenbit (approved):

keras_cv/models/segmentation/__internal__/segmentation_head.py
qlzh727: Done.

keras_cv/models/segmentation/__internal__/segmentation_head.py
qlzh727: Done.

tanzhenyu (approved):
LGTM.
Thanks for the PR!

LukeWood (approved):
Looks good, thanks Scott

i wonder if we should export these at some point.  Users customizing their models with regularization and such may want to use these components.  Another approach is to just write tutorials on it…

qlzh727:
> Looks good, thanks Scott
> 
> i wonder if we should export these at some point. Users customizing their models with regularization and such may want to use these components. Another approach is to just write tutorials on it…

yes, once we verify this with more configs/use case, this will become public.

PR #818, opened 2022-09-21
Add ROISampler.

tanzhenyu:
# What does this PR do?
This PR adds ROISampler as a private API.
This shows why we should have `target_gather` and `box_matcher`
Some slight API changes according to the design doc:
https://github.com/keras-team/governance/blob/master/rfcs/20220804-keras-cv-two-stage-2d-object-detection.md
The ProposalSampler is renamed to ROISampler.

### DIFF
--- a/keras_cv/layers/object_detection/roi_sampler.py
+++ b/keras_cv/layers/object_detection/roi_sampler.py
@@ -1,0 +1,157 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv import bounding_box
+from keras_cv.bounding_box.iou import compute_iou
+from keras_cv.ops.box_matcher import ArgmaxBoxMatcher
+from keras_cv.ops.sampling import balanced_sample
+from keras_cv.ops.target_gather import _target_gather
+
+
+class _ROISampler(tf.keras.layers.Layer):
+    """
+    Sample ROIs for loss related calucation.
+
+    With proposals (ROIs) and ground truth, tt performs the following:
+    1) compute IOU similarity matrix
+    2) match each proposal to ground truth box based on IOU
+    3) samples positive matches and negative matches and return
+
+    `append_gt_boxes` augments proposals with ground truth boxes. This is
+    useful in 2 stage detection networks during initialization where the
+    1st stage often cannot produce good proposals for 2nd stage. Setting it
+    to True will allow it to generate more reasonable proposals at the begining.
+
+    `background_class` allow users to set the labels for background proposals. Default
+    is 0, where users need to manually shift the incoming `gt_classes` if its range is
+    [0, num_classes).
+
+    Args:
+      bounding_box_format: The format of bounding boxes to generate. Refer
+        [to the keras.io docs](https://keras.io/api/keras_cv/bounding_box/formats/)
+        for more details on supported bounding box formats.
+      roi_matcher: a `ArgmaxBoxMatcher` object that matches proposals
+        with ground truth boxes. the positive match must be 1 and negative match must be -1.
+        Such assumption is not being validated here.
+      positive_fraction: the positive ratio w.r.t `num_sampled_rois`. Defaults to 0.25.
+      background_class: the background class which is used to map returned the sampled
+        ground truth which is classified as background.
+      num_sampled_rois: the number of sampled proposals per image for
+        further (loss) calculation. Defaults to 256.
+      append_gt_boxes: boolean, whether gt_boxes will be appended to rois
+        before sample the rois. Defaults to True.
+    """
+
+    def __init__(
+        self,
+        bounding_box_format: str,
+        roi_matcher: ArgmaxBoxMatcher,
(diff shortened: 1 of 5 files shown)


keras_cv/layers/object_detection/roi_sampler.py
```diff
@@ -0,0 +1,157 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv import bounding_box
+from keras_cv.bounding_box.iou import compute_iou
```
qlzh727, line 18: style wise, its better to have `from keras.packge import module` and then `module.function` in the code, which is more readable.

keras_cv/layers/object_detection/roi_sampler_test.py
```diff
@@ -0,0 +1,214 @@
...
+
+import tensorflow as tf
+
+from keras_cv.layers.object_detection.roi_sampler import _ROISampler
+from keras_cv.ops.box_matcher import ArgmaxBoxMatcher
+
+
+class ROISamplerTest(tf.test.TestCase):
+    def test_roi_sampler(self):
+        box_matcher = ArgmaxBoxMatcher([0.3], [-1, 1])
```
qlzh727, line 23: let's add the param here to make it more readable. Same for _ROISampler below.

keras_cv/layers/object_detection/roi_sampler_test.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+        rois = rois[tf.newaxis, ...]
+        # the 3rd box will generate 0 IOUs and not sampled.
+        gt_boxes = tf.constant(
+            [[10, 10, 15, 15], [2.5, 2.5, 7.5, 7.5], [-1, -1, -1, -1]]
+        )
+        gt_boxes = gt_boxes[tf.newaxis, ...]
+        gt_classes = tf.constant([[2, 10, -1]], dtype=tf.int32)
+        gt_classes = gt_classes[..., tf.newaxis]
+        a, sampled_gt_boxes, sampled_gt_classes = roi_sampler(
```
qlzh727, line 42: The `a` isn't get used down below, maybe make it _.

keras_cv/ops/sampling.py
```diff
@@ -0,0 +1,85 @@
...
+        # pick the top k values as random positive indices
+        _, positive_indices = tf.math.top_k(random_pos, k=num_pos)
+
+        # randomly generate positive values for negative match, and 0 for positive match
+        random_neg = tf.random.uniform(tf.shape(negative_match), minval=0.0, maxval=1.0)
+        random_neg = tf.where(negative_match, random_neg, zeros)
+        # pick the top k values as random negative indices
+        _, negative_indices = tf.math.top_k(random_neg, k=num_neg)
+
+        selected_indices = tf.concat([positive_indices, negative_indices], axis=0)
```
qlzh727, line 69: I am bit confused about the concat of pos and neg index here. Is the index going to be applied on the same matrix (eg the postive/negative result in the same matrix). If that the case, maybe its better to just intake the original matrix and produce positive/negative match within this function.

keras_cv/ops/sampling.py
```diff
@@ -0,0 +1,85 @@
...
+      positive_fraction: float. 0.5 means positive samples should be half
+        of all collected samples.
+
+    Returns:
+      selected_indicators: [`num_samples`] or [batch_size, `num_samples`]
+        integer Tensor, 1 for indicating the index is sampled, 0 for
+        indicating the index is not sampled.
+    """
+
+    def balanced_single_sample(positive_match, negative_match):
```
qlzh727, line 43: Readability wise, maybe its better to lift this one out and have it as `_balanced_single_sample` in this file.

keras_cv/ops/sampling.py
tanzhenyu: We need the indices as it can be applied to both boxes and classes. Otherwise we'd have to call this twice which might not get fused for optimal performance.

keras_cv/ops/sampling.py
tanzhenyu: it uses some other args within the function, such as `positive_fraction`, which cannot be unbatched during tf.vectorized_map

keras_cv/layers/object_detection/roi_sampler.py
```diff
@@ -0,0 +1,157 @@
...
+from keras_cv.ops.box_matcher import ArgmaxBoxMatcher
+from keras_cv.ops.sampling import balanced_sample
+from keras_cv.ops.target_gather import _target_gather
+
+
+class _ROISampler(tf.keras.layers.Layer):
+    """
+    Sample ROIs for loss related calucation.
+
+    With proposals (ROIs) and ground truth, tt performs the following:
```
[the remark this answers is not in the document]
tanzhenyu: Fixed.

keras_cv/layers/object_detection/roi_sampler.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_sampler_test.py
tanzhenyu: Done.

keras_cv/layers/object_detection/roi_sampler.py
```diff
@@ -0,0 +1,157 @@
...
+    1) compute IOU similarity matrix
+    2) match each proposal to ground truth box based on IOU
+    3) samples positive matches and negative matches and return
+
+    `append_gt_boxes` augments proposals with ground truth boxes. This is
```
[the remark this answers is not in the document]
tanzhenyu: Fixed.

qlzh727 (approved):

LukeWood (approved):
Looks good to me, thanks for all of the Rcnn components they look great!

PR #820, opened 2022-09-22
Introduce a new function to convert an application model to backbone model.

qlzh727:
# What does this PR do?
WIP, this PR is more for a demo and not finalized yet.

The current PR will attach the output on each feature level to the model, so that we could receate the application model to a backbone model, so that features on each level could be consumed by decoder like FPN or RPN.

## Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/keras_cv/models/resnet_v2.py
+++ b/keras_cv/models/resnet_v2.py
@@ -385,7 +385,7 @@ def ResNetV2(
 
         The backbone model will usually take same inputs as the original application
         model, but produce multiple outputs, one for each feature level. Those outputs
-        can be feed to network downstream, like FPN and PRN.
+        can be feed to network downstream, like FPN and RPN.
 
         The output of the backbone model will be a dict with int as key and tensor as
         value. The int key represent the level of the feature output.
--- a/keras_cv/models/resnet_v2.py
+++ b/keras_cv/models/resnet_v2.py
@@ -17,6 +17,8 @@ Reference:
   - [Based on the original keras.applications ResNet](https://github.com/keras-team/keras/blob/master/keras/applications/resnet_v2.py)
 """
 
+import types
+
 import tensorflow as tf
 from tensorflow.keras import backend
 from tensorflow.keras import layers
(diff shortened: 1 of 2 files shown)


LukeWood:
Hey Scott are you looking for feedback on the approach, the API surface, or both?

I really like the approach - maybe we could promote backbone outputs to a public API even.

API wise maybe we name the function models.as_backbone(model) or maybe we make a BaseCvModel class.  If we go with the latter there are some other features we could add.  One that comes to mind is automatically unpacking dictionary inputs in the format {images, labels}.  This would allow us to drop the code to unpackage tf dataset dictionaries all over

keras_cv/models/resnet_v2.py
```diff
@@ -355,10 +374,21 @@ def ResNetV2(
...
     if weights is not None:
         model.load_weights(weights)
+    # Set this private attribute for recreate backbone model with outputs at each of the
+    # resolution level.
+    model._backbone_level_outputs = stack_level_outputs
 
     return model
 
 
+def create_backbone_model(model):
```
tanzhenyu, line 384: maybe add some documentation here, how we imagine users to use this?

qlzh727:
> 

Both. The ideal API interface to me is like:

```
application_model = resnet_v2.Resnet50(....)
backbone_model = application_model.to_backbone_model()
```

In the application model building logic, it will create a new method `to_backbone_model` for the model, and populate the logic.

LukeWood:
> Both. The ideal API interface to me is like:
> 
> ```
> application_model = resnet_v2.Resnet50(....)
> backbone_model = application_model.to_backbone_model()
> ```
> 
> In the application model building logic, it will create a new method `to_backbone_model` for the model, and populate the logic.

Yep, I agree this is the ideal interface.

what do you think about a CVModel base class that offers this method (raises error if no backbone outputs are set), and unpackages images/labels from the dict format in fit() for us?

keras_cv/models/resnet_v2.py
qlzh727: Done. Changed to a new interface.

ianstenbit:
> what do you think about a CVModel base class that offers this method (raises error if no backbone outputs are set), and unpackages images/labels from the dict format in fit() for us?

I personally like this idea, but it's also a bit more frameworky than just building a raw Keras model. One benefit is that the structure could be easily extended for e.g. DiffusionModels, which will likely share a lot of functionality and likely a custom train_step.

FWIW, I personally like `as_backbone` more than `to_backbone_model` for the API

qlzh727:
> 

The new base class approach might be more complicated, since we need to handle all 3 models types (Functional, subclassing, Sequential). We could have a try for this approach, but I feel this is bit orthogonal to the current change.

LukeWood (approved):
I agree that the class approach is out of scope for this.  All in all, looks great and I really like this API.

keras_cv/models/resnet_v2.py
```diff
@@ -355,6 +376,37 @@ def ResNetV2(
...
+    # Set this private attribute for recreate backbone model with outputs at each of the
+    # resolution level.
+    model._backbone_level_outputs = stack_level_outputs
+
+    def as_backbone(self):
+        """Convert the Resnet application model into a model backbone for other tasks.
+
+        The backbone model will usually take same inputs as the original application
+        model, but produce multiple outputs, one for each feature level. Those outputs
+        can be feed to network downstream, like FPN and PRN.
```
tanzhenyu, line 388: I think you mean RPN?

tanzhenyu (approved):

keras_cv/models/resnet_v2.py
```diff
@@ -355,6 +376,37 @@ def ResNetV2(
...
+
+    def as_backbone(self):
+        """Convert the Resnet application model into a model backbone for other tasks.
+
+        The backbone model will usually take same inputs as the original application
+        model, but produce multiple outputs, one for each feature level. Those outputs
+        can be feed to network downstream, like FPN and RPN.
+
+        The output of the backbone model will be a dict with int as key and tensor as
+        value. The int key represent the level of the feature output.
```
martin-gorner, line 391: Please define "level"
From "Practical ML for CV": "A typical feature pyramid has five levels corresponding to scales P3, P4, P5, P6, P7 in the backbone. Scale Pn represents a feature map 2n times smaller in width and height than the input image."
What does "level" stand for in the docstring ?

Issue #833, opened 2022-09-25
StableDiffusion does not work with `mixed_precision.set_global_policy('mixed_float16')` policy


LukeWood:
mixed_precision.set_global_policy('mixed_float16')
stable_diffusion_mixed_precision = keras_cv.models.StableDiffusion()

```

    TypeError: Exception encountered when calling layer "clip_attention_26" (type CLIPAttention).
    
    in user code:
    
        File "/home/lukewood/workspace/keras-cv/keras_cv/models/generative/stable_diffusion/text_encoder.py", line 105, in call  *
            attn_weights = attn_weights + attention_mask
    
        TypeError: Input 'y' of 'AddV2' Op has type float32 that does not match type float16 of argument 'x'.
    
    
    Call arguments received by layer "clip_attention_26" (type CLIPAttention):
      • inputs=tf.Tensor(shape=(None, 77, 768), dtype=float16)
      • attention_mask=None


Call arguments received by layer "clip_encoder_layer_26" (type CLIPEncoderLayer):
  • inputs=tf.Tensor(shape=(None, 77, 768), dtype=float16)
```

LukeWood:
I'll be adding a unit test to reproduce this, and I'll go ahead and fix the bug.  

Just an FYI @fchollet - I think at some point you had mixed precision working  - perhaps you know what broke it.  this seems like a simple weight initialization bug though.

innat:
If mixed precision is on, in some cases, it's needed to cast the output layer's dtype to float32. I wonder if it's consider for the [segmentation head](https://github.com/keras-team/keras-cv/blob/e0c405350d5e70ae30ed1cab6b9219d911712c19/keras_cv/models/segmentation/__internal__/segmentation_head.py#L75) layer.

PR #834, opened 2022-09-25
Add mixed precision support

LukeWood:
Fixes mixed precision support in `keras_cv.models.StableDiffusion`.

I have added a unit test to confirm the fix, and have fixed the behavior.  it turns out `tnp` does not by-default know about compute_dtype.

/auto closes https://github.com/keras-team/keras-cv/issues/833


fchollet (approved):
Thanks!

PR #846, opened 2022-09-26
Fix label encoder when all -1s are provided

LukeWood:
In short; this is a bug that happens due to the support of -1 padded tensors to represent bounding boxes. 

This PR makes the label encoder function properly when -1 padded bounding boxes are passed.  There is effectively an edge-case where you can sometimes match with a bounding box that has all -1s in its representation in some circumstances resulting in a NaN encoding.

I tried to solve this by masking out the -1 boxes; but this messed with the performance in unexpected ways.  I've re-run the RetinaNet guide using these numerics and have confirmed that the results are unchanged.

/auto closes https://github.com/keras-team/keras-cv/issues/845

### DIFF
--- a/examples/layers/preprocessing/bounding_box/mosaic_demo.py
+++ b/examples/layers/preprocessing/bounding_box/mosaic_demo.py
@@ -26,7 +26,7 @@ BATCH_SIZE = 9
 
 def main():
     dataset = demo_utils.load_voc_dataset(bounding_box_format="rel_xyxy")
-    mosaic = preprocessing.Mosaic(bounding_box_format="rel_xyxy")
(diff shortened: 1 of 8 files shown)


LukeWood:
Maybe a better solution here would be to make IOU “0” for all entries for any GT box with class ID -1

ianstenbit (approved):

LukeWood:
> Maybe a better solution here would be to make IOU “0” for all entries for any GT box with class ID -1.
> 
> The result is the same though; and I have verified that this combined with Mosaic scores a 0.16 MaP~ on some epochs. Waiting for final convergence now.

Its not a 1:1 transition to https://github.com/keras-team/keras-cv/pull/852, so In the meantime, this is a better solution than NaN-ing on training when an empty image shows up.

PR #847, opened 2022-09-26
Readability improvement to SD

LukeWood:
Adds a getter for start of text and end of text.

I've had a few people ask what these hard coded constants are in the tokenizer and I think this improves that a great deal.

### DIFF
--- a/keras_cv/models/generative/stable_diffusion/clip_tokenizer.py
+++ b/keras_cv/models/generative/stable_diffusion/clip_tokenizer.py
@@ -101,6 +101,14 @@ class SimpleTokenizer:
             re.IGNORECASE,
         )
 
+    @property
+    def end_of_text(self):
+        return self.encoder["<|endoftext|>"]
+
+    @property
+    def start_of_text(self):
+        return self.encoder["<|startoftext|>"]
+
     def bpe(self, token):
         if token in self.cache:
             return self.cache[token]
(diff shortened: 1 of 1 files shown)


keras_cv/models/generative/stable_diffusion/clip_tokenizer.py
```diff
@@ -101,6 +101,14 @@ def __init__(self, bpe_path=None):
             re.IGNORECASE,
         )
 
+    @property
+    def end_of_text(self):
+        return self.encoder["<|endoftext|>"]
+
+    @property
+    def start_of_text(self):
```
ianstenbit, line 109: This token is also used as a magic number in `stable_diffusion.py` as:
```
phrase = inputs + [49407] * (MAX_PROMPT_LENGTH - len(inputs))
```

Maybe we could also replace it there with self.tokenizer.end_of_text?

(It might also be nice to add a comment to `constants.py` explaining this for the `UNCONDITIONAL_TOKENS`

fchollet (approved):
LGTM

keras_cv/models/generative/stable_diffusion/clip_tokenizer.py
LukeWood: Will do in a separate branch when I update `add_tokens()`.

PR #851, opened 2022-09-27
Small performance optimization for StableDiffusion

ianstenbit:
Maybe this is totally pointless, in which case feel free to close the PR.

But I noticed while working with the code yesterday that we're unnecessarily passing many copies of the input phrase (and the unconditional phrase) into the text encoder. With this change, we pass just one copy into the encoder and repeat it to match the batch dimension _after_ encoding.

(This also guarantees that the output of the encoder is _exactly_ equal for each image in the batch, whereas before there was some minor variation due to fuzzy GPU math. I don't think that actually matters but I figured I'd note it)

Edit: I also manually ran the tests for stable diffusion and used the code in this state for my experiments with latent space walks


fchollet (approved):
LGTM, thanks!

PR #852, opened 2022-09-27
Add IOUSimilarity layer to handle masking.

tanzhenyu:
# What does this PR do?
IOU with masking

### DIFF
--- a/keras_cv/layers/object_detection/iou_similarity.py
+++ b/keras_cv/layers/object_detection/iou_similarity.py
@@ -85,3 +85,14 @@ class IouSimilarity(tf.keras.layers.Layer):
             )
         iou_lookup_table = tf.where(background_mask, mask_val_t, result)
         return iou_lookup_table
+
+    def get_config(self):
+        config = {
+            "box1_format": self.box1_format,
+            "box2_format": self.box2_format,
+            "box1_mask": self.box1_mask,
+            "box2_mask": self.box2_mask,
+            "mask_val": self.mask_val,
+        }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
--- a/keras_cv/layers/object_detection/iou_similarity.py
+++ b/keras_cv/layers/object_detection/iou_similarity.py
@@ -1,0 +1,87 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv import bounding_box
+
+
+class IouSimilarity(tf.keras.layers.Layer):
+    """Computes a lookup table matrix containing the ious for a given pair of boxes.
+
+    The lookup vector is to be indexed by [`boxes1_index`,`boxes2_index`] if boxes
+    are unbatched and by [`batch`, `boxes1_index`,`boxes2_index`] if the boxes are
+    batched.
+
+    Args:
+      boxes1: a list of bounding boxes in 'corners' format. Can be batched or unbatched.
+      boxes2: a list of bounding boxes in 'corners' format. This should match the rank and
+        shape of boxes1.
+      bounding_box_format: a case-insensitive string.
+        For detailed information on the supported format, see the
+        [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+
+    Returns:
+      iou_lookup_table: a vector containing the pairwise ious of boxes1 and
+        boxes2.
+    """
+
+    def __init__(
+        self,
+        box1_format,
+        box2_format,
+        box1_mask=False,
+        box2_mask=False,
+        mask_val=-1,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.box1_format = box1_format
+        self.box2_format = box2_format
+        self.mask_val = mask_val
+        self.box1_mask = box1_mask
+        self.box2_mask = box2_mask
+        self.built = True
+
+    def call(self, boxes1, boxes2, boxes1_mask=None, boxes2_mask=None):
+        boxes1 = bounding_box.convert_format(
+            boxes1, source=self.box1_format, target="xyxy"
+        )
+        boxes2 = bounding_box.convert_format(
+            boxes2, source=self.box2_format, target="xyxy"
+        )
+        result = bounding_box.compute_iou(boxes1, boxes2, bounding_box_format="xyxy")
+        if not self.box1_mask and not self.box2_mask:
+            return result
+        background_mask = None
+        boxes2_rank = len(boxes2.shape)
+        mask_val_t = tf.cast(self.mask_val, result.dtype) * tf.ones_like(result)
+        perm = [1, 0] if boxes2_rank == 2 else [0, 2, 1]
+        if self.box1_mask and self.box2_mask:
+            boxes1_mask = tf.less(tf.reduce_max(boxes1, axis=-1, keepdims=True), 0.0)
+            boxes2_mask = tf.less(tf.reduce_max(boxes2, axis=-1, keepdims=True), 0.0)
+            background_mask = tf.logical_or(
+                boxes1_mask, tf.transpose(boxes2_mask, perm)
+            )
+        elif self.box1_mask:
+            boxes1_mask = tf.less(tf.reduce_max(boxes1, axis=-1, keepdims=True), 0.0)
+            background_mask = boxes1_mask
+        else:
+            boxes2_mask = tf.less(tf.reduce_max(boxes2, axis=-1, keepdims=True), 0.0)
+            background_mask = tf.logical_or(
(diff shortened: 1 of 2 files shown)


keras_cv/layers/object_detection/iou_similarity.py
```diff
@@ -0,0 +1,103 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv import bounding_box
+
+
+class IouSimilarity(tf.keras.layers.Layer):
+    """Computes a lookup table matrix containing the ious for a given pair of boxes.
```
ianstenbit, line 21: I think "pair of boxes" should be "pair of lists of boxes"

ianstenbit, line 21: Maybe /s/lookup table matrix/similarity matrix

I think it would be good if the docstring refers to what's returned with the same phrase throughout (currently it calls it both a lookup table matrix and a lookup vector and an iou_lookup_table

keras_cv/layers/object_detection/iou_similarity.py
```diff
@@ -0,0 +1,103 @@
...
+    out of the boundary, i.e., all their coodinates are less than 0. Users can pad ground truth
+    boxes with -1. in order for it to be masked here.
+
+    Args:
+      boxes1: a list of bounding boxes in 'corners' format. Can be batched or unbatched.
+      boxes2: a list of bounding boxes in 'corners' format. This should match the rank and
+        shape of boxes1.
+      bounding_box_format: a case-insensitive string.
+        For detailed information on the supported format, see the
+        [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
```
ianstenbit, line 38: Should we add arg descriptors for box1_mask and box2_mask?

keras_cv/layers/object_detection/iou_similarity.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Returns:
+      iou_lookup_table: a vector containing the pairwise ious of boxes1 and
+        boxes2.
+    """
+
+    def __init__(
+        self,
+        box1_format,
```
ianstenbit, line 47: (Just curious) -- in what sort of use cases would we expect the two input boxes to be in different formats?

keras_cv/layers/object_detection/iou_similarity.py
```diff
@@ -0,0 +1,103 @@
...
+      boxes2: a list of bounding boxes in 'corners' format. This should match the rank and
+        shape of boxes1.
+      bounding_box_format: a case-insensitive string.
+        For detailed information on the supported format, see the
+        [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+
+    Returns:
+      iou_lookup_table: a vector containing the pairwise ious of boxes1 and
+        boxes2.
+    """
```
ianstenbit, line 43: Should we have an example usage for this layer?

I think that might be useful in explaining the meaning of the masking behavior, but maybe that's overkill

keras_cv/layers/object_detection/iou_similarity.py
```diff
@@ -0,0 +1,103 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv import bounding_box
+
+
+class IouSimilarity(tf.keras.layers.Layer):
```
LukeWood, line 20: Lets remove references to "Similarity" here as thats a somewhat overloaded term.  IoU is more precise from a naming perspective.

keras_cv/layers/object_detection/iou_similarity.py
```diff
@@ -0,0 +1,103 @@
...
+    ):
+        super().__init__(**kwargs)
+        self.box1_format = box1_format
+        self.box2_format = box2_format
+        self.mask_val = mask_val
+        self.box1_mask = box1_mask
+        self.box2_mask = box2_mask
+        self.built = True
+
+    def call(self, boxes1, boxes2, boxes1_mask=None, boxes2_mask=None):
```
LukeWood, line 62: Whats the specific use case of accepting explicit masks into call() instead of masking with -1s?

keras_cv/layers/object_detection/iou_similarity.py
```diff
@@ -0,0 +1,103 @@
...
+        elif self.box1_mask:
+            boxes1_mask = tf.less(tf.reduce_max(boxes1, axis=-1, keepdims=True), 0.0)
+            background_mask = boxes1_mask
+        else:
+            boxes2_mask = tf.less(tf.reduce_max(boxes2, axis=-1, keepdims=True), 0.0)
+            background_mask = tf.logical_or(
+                tf.zeros(tf.shape(boxes2)[:-1], dtype=tf.bool),
+                tf.transpose(boxes2_mask, perm),
+            )
+        iou_lookup_table = tf.where(background_mask, mask_val_t, result)
```
LukeWood, line 91: Any reason for this to be a layer instead of a utility?  

I wouldn't expect this to be plugged anywhere else into the API in a configurable way (i.e. as a constructor parameter for another component), so I think it makes a lot of sense to keep it as a function.

LukeWood (changes requested):
Personally I think `compute_iou()` fits the construct of a function better than a class.  Here is my reasoning:

- users will never configure this themselves; there is one way and one way only to compute IoU
- any parameters we DO need to configure are on a component-level


As such, I think we should upstream returning -1 for background class boxes to:
https://github.com/keras-team/keras-cv/blob/master/keras_cv/bounding_box/iou.py#L20

instead of creating a new component.

LukeWood:
Thanks for the PR!  Returning -1 IoU for background classes is a much better pattern!

I think IoU is a better fit for a function rather than a component

tanzhenyu:
> Thanks for the PR! Returning -1 IoU for background classes is a much better pattern!
> 
> I think IoU is a better fit for a function rather than a component

I originally wrote it as function, but then calling this in several places carrying information such as "mask_val", "box_mask", "box_format" seems pretty heavy to user, so I switched to layer, users can create it once, use it everywhere.
I'm fine with a function as well. WDYT?

LukeWood:
> > Thanks for the PR! Returning -1 IoU for background classes is a much better pattern!
> > I think IoU is a better fit for a function rather than a component
> 
> I originally wrote it as function, but then calling this in several places carrying information such as "mask_val", "box_mask", "box_format" seems pretty heavy to user, so I switched to layer, users can create it once, use it everywhere. I'm fine with a function as well. WDYT?

are all of the arguments present required?  I think this influences the answer to the questions.

I don't see a ton of use cases for dis-joint bounding box formats, and do we need to provide masks manually?  Or can we craft them based on class_id=background_class?

tanzhenyu:
> > > Thanks for the PR! Returning -1 IoU for background classes is a much better pattern!
> > > I think IoU is a better fit for a function rather than a component
> > 
> > 
> > I originally wrote it as function, but then calling this in several places carrying information such as "mask_val", "box_mask", "box_format" seems pretty heavy to user, so I switched to layer, users can create it once, use it everywhere. I'm fine with a function as well. WDYT?
> 
> are all of the arguments present required? I think this influences the answer to the questions.
> 
> I don't see a ton of use cases for dis-joint bounding box formats, and do we need to provide masks manually? Or can we craft them based on class_id=background_class?

So the behavior is turn on/off the masking. The layer will automatically compute mask for users, with the assumption that padded boxes are with default values less than 0 -- that seems to be a good assumption so far.
Otherwise, we'd have to ask users to compute mask everytime, which is more flexible, but bloats the model code

LukeWood:
> So the behavior is turn on/off the masking. The layer will automatically compute mask for users, with the assumption that padded boxes are with default values less than 0 -- that seems to be a good assumption so far.


If its just one parameter then we can support:

```
compute_iou(boxes1, boxes2, bounding_box_format, use_masking=True):
```

any make masking the default.  If this is the case I think its a better fit as a utility than a layer.  Let me know if there are cases missing in this API.

tanzhenyu:
> > So the behavior is turn on/off the masking. The layer will automatically compute mask for users, with the assumption that padded boxes are with default values less than 0 -- that seems to be a good assumption so far.
> 
> If its just one parameter then we can support:
> 
> ```
> compute_iou(boxes1, boxes2, bounding_box_format, use_masking=True):
> ```
> 
> any make masking the default. If this is the case I think its a better fit as a utility than a layer. Let me know if there are cases missing in this API.

That's fine with me. (Though we might need to support masking for boxes1 only, or for boxes2 only, or both. I'm not seeing that pattern yet)

LukeWood:
> > > So the behavior is turn on/off the masking. The layer will automatically compute mask for users, with the assumption that padded boxes are with default values less than 0 -- that seems to be a good assumption so far.
> > 
> > 
> > If its just one parameter then we can support:
> > ```
> > compute_iou(boxes1, boxes2, bounding_box_format, use_masking=True):
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > any make masking the default. If this is the case I think its a better fit as a utility than a layer. Let me know if there are cases missing in this API.
> 
> That's fine with me. (Though we might need to support masking for boxes1 only, or for boxes2 only, or both. I'm not seeing that pattern yet)

I'm fairly sure we'll always mask out the -1s, so lets try it with a utility and we can always add a parameter like:

```
`masking='boxes1'` or `masking='boxes2'` down the line if a need arises
```

tanzhenyu:
#870

tanzhenyu:
Closing this PR as the FasterRCNN PR already deals with all the comments above.

LukeWood:
Greatl thanks!

PR #853, opened 2022-09-27
Add encode_text and generate_image to StableDiffusion API

ianstenbit:
This provides an API surface for tinkering with / walking through the text encoding space as well as the initial noise space for diffusion, both of which are popular use cases.

Let me know what your thoughts are on this API change. If we think this is a reasonable approach, I'll add docstrings and examples for the new methods.

The thing that I dislike the most about this PR in its current form is the way that diffusion noise is provided to the `generate_image` method. I haven't thought of a cleaner way to do this in a way that makes it easy to arbitrarily (and sequentially) mutate the noise for a sequence of images. Critically, we can't perform single-batch inference for large walks, because even large accelerators will OOM, so users need to be able to run consecutive batches and control a walk themselves.

### DIFF
--- a/keras_cv/models/generative/stable_diffusion/stable_diffusion.py
+++ b/keras_cv/models/generative/stable_diffusion/stable_diffusion.py
@@ -164,7 +164,7 @@ class StableDiffusion:
             self._get_unconditional_context(), batch_size, axis=0
         )
 
-        if diffusion_noise:
+        if diffusion_noise is not None:
             latent = diffusion_noise
         else:
             latent = self._get_initial_diffusion_noise(batch_size, seed)
--- a/keras_cv/models/generative/stable_diffusion/stable_diffusion.py
+++ b/keras_cv/models/generative/stable_diffusion/stable_diffusion.py
@@ -231,7 +231,7 @@ class StableDiffusion:
 
         if diffusion_noise is not None:
             diffusion_noise = tf.squeeze(diffusion_noise)
-            if diffusion_noise.shape.rank == 2:
+            if diffusion_noise.shape.rank == 3:
                 diffusion_noise = tf.tile(diffusion_noise, batch_size)
             latent = diffusion_noise
         else:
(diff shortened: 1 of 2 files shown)


keras_cv/models/generative/stable_diffusion/stable_diffusion.py
```diff
@@ -127,27 +138,33 @@ def text_to_image(
...
+        seed=None,
+    ):
+        context = tf.repeat(encoded_text, batch_size, axis=0)
+        unconditional_context = tf.repeat(
+            self._get_unconditional_context(), batch_size, axis=0
         )
 
+        if diffusion_noise:
+            latent = diffusion_noise
+        else:
```
LukeWood, line 162: should we error if a user passes both seed and diffusion noise?  If I understand correctly you pas one or the other.

LukeWood (changes requested):
looks very good to me; improves readability.  one API question I left

keras_cv/models/generative/stable_diffusion/stable_diffusion.py
ianstenbit: Yes, good call. I've updated this to throw an error and added a test (which I've manually run but subsequently added the DISABLED_ prefix to so that it doesn't get run)

LukeWood (approved):
LGTM, but lets wait for @fchollet 's LGTM too.

fchollet (approved):
LGTM -- this is a nicer factoring

keras_cv/models/generative/stable_diffusion/stable_diffusion.py
```diff
@@ -127,27 +138,47 @@ def text_to_image(
...
-        # Encode unconditional tokens + positions as "unconditional context"
-        unconditional_tokens = tf.convert_to_tensor(
-            [_UNCONDITIONAL_TOKENS], dtype=tf.int32
-        )
-        self.unconditional_tokens = tf.repeat(unconditional_tokens, batch_size, axis=0)
-        unconditional_context = self.text_encoder.predict_on_batch(
-            [self.unconditional_tokens, pos_ids]
+        return context
+
+    def generate_image(
```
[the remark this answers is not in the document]
ianstenbit: I think they should be exposed to users so that users can easily do walks through the encoder latent manifold and diffusion noise space (will send out tutorial on this soon)

ianstenbit:
I've manually run the `DISABLED` tests locally, and they pass. If GCBRun is happy, I am going to merge

PR #854, opened 2022-09-27
keras_cv.bounding_box => keras_cv.data.bounding_box

LukeWood:
# What does this PR do?

Per internal design review:

_The problem this addresses_:  KerasCV will eventually have a ton of top level namespaces with the current structure.  This fixes that by nesting under `data`.  While in general, Keras APIs strive to be flat this is similar to `datasets`.

Migrates `keras_cv.bounding_box` to `keras_cv.data.bounding_box`.  The goal of this change is to prevent the pollution of our top level directory and namespacing.  Right now, we only have one format so `keras_cv.bounding_box` is fine - but  when we have 10-20 formats our top level namespaces will all be polluted.

Instead, we will prefer sub-namespaces like this.  This preserves a nice user facing API;

```
from keras_cv.data import bounding_box

bounding_box.convert_format(boxes)
bounding_box.clip_to_image(boxes, images)
```

## Who can review?

# An important note

Do not merge w/o merging https://github.com/keras-team/keras-io/pull/1094/files


tanzhenyu:
Seems that this touches a lot of files (way more than I originally thought) -- in this case can we wait until we have both RetinaNet and FasterRCNN?

LukeWood:
> Seems that this touches a lot of files (way more than I originally thought) -- in this case can we wait until we have both RetinaNet and FasterRCNN?

I'd rather break the API now than later.  I don't think doing it later will simplify the move.

tanzhenyu:
> > Seems that this touches a lot of files (way more than I originally thought) -- in this case can we wait until we have both RetinaNet and FasterRCNN?
> 
> I'd rather break the API now than later. I don't think doing it later will simplify the move.
> 
> Any reason to hold off?

As discussed, I'd wait until we all agreed keras_cv.data is the way to go, otherwise we need to break users twice

LukeWood:
> > > Seems that this touches a lot of files (way more than I originally thought) -- in this case can we wait until we have both RetinaNet and FasterRCNN?
> > 
> > 
> > I'd rather break the API now than later. I don't think doing it later will simplify the move.
> > Any reason to hold off?
> 
> As discussed, I'd wait until we all agreed keras_cv.data is the way to go, otherwise we need to break users twice

did anyone disagree?  i thought we got consensus?

tanzhenyu:
> > > > Seems that this touches a lot of files (way more than I originally thought) -- in this case can we wait until we have both RetinaNet and FasterRCNN?
> > > 
> > > 
> > > I'd rather break the API now than later. I don't think doing it later will simplify the move.
> > > Any reason to hold off?
> > 
> > 
> > As discussed, I'd wait until we all agreed keras_cv.data is the way to go, otherwise we need to break users twice
> 
> Did I misunderstand? I thought we got consensus?

I believe the current consensus is keras_cv.data seems the top choice here, but:
1) we give some time to think this through and see if there are any better options
2) we have to document what needs to be moved to the new namespace and what stays

In summary, can we wait before next Keras SIG meeting, which is next Friday? By that we should have a consensus, bump the version, and update all our documents

LukeWood:
> > > > > Seems that this touches a lot of files (way more than I originally thought) -- in this case can we wait until we have both RetinaNet and FasterRCNN?
> > > > 
> > > > 
> > > > I'd rather break the API now than later. I don't think doing it later will simplify the move.
> > > > Any reason to hold off?
> > > 
> > > 
> > > As discussed, I'd wait until we all agreed keras_cv.data is the way to go, otherwise we need to break users twice
> > 
> > 
> > Did I misunderstand? I thought we got consensus?
> 
> I believe the current consensus is keras_cv.data seems the top choice here, but:
> 
> 1. we give some time to think this through and see if there are any better options
> 2. we have to document what needs to be moved to the new namespace and what stays
> 
> In summary, can we wait before next Keras SIG meeting, which is next Friday? By that we should have a consensus, bump the version, and update all our documents

Personally I'd rather take the incremental step towards what we think is the right API.  The longer we wait, the more things that will need to be migrated.  We don't really have that many users other than the tutorials; 

as I work on more tutorials I'd rather go with what we *think* the final API will be (`from keras_cv.data import bounding_box`) even if we don't land on that, than use what we KNOW will not be the final API (`keras_cv.bounding_box`.

tanzhenyu:
> > > > > > Seems that this touches a lot of files (way more than I originally thought) -- in this case can we wait until we have both RetinaNet and FasterRCNN?
> > > > > 
> > > > > 
> > > > > I'd rather break the API now than later. I don't think doing it later will simplify the move.
> > > > > Any reason to hold off?
> > > > 
> > > > 
> > > > As discussed, I'd wait until we all agreed keras_cv.data is the way to go, otherwise we need to break users twice
> > > 
> > > 
> > > Did I misunderstand? I thought we got consensus?
> > 
> > 
> > I believe the current consensus is keras_cv.data seems the top choice here, but:
> > 
> > 1. we give some time to think this through and see if there are any better options
> > 2. we have to document what needs to be moved to the new namespace and what stays
> > 
> > In summary, can we wait before next Keras SIG meeting, which is next Friday? By that we should have a consensus, bump the version, and update all our documents
> 
> Personally I'd rather take the incremental step towards what we think is the right API. The longer we wait, the more things that will need to be migrated. We don't really have that many users other than the tutorials;
> 
> as I work on more tutorials I'd rather go with what we _think_ the final API will be (`from keras_cv.data import bounding_box`) even if we don't land on that, than use what we KNOW will not be the final API (`keras_cv.bounding_box`.
> 
> Definitely still open to feedback, but I'd like to have a strong reason as to why we wouldn't do this (i.e. an alternative API we are still considering, from what I understand we don't believe in either of the alternatives proposed)

I strongly believe we should wait until we have a consensus

LukeWood:
> > > > > > > Seems that this touches a lot of files (way more than I originally thought) -- in this case can we wait until we have both RetinaNet and FasterRCNN?
> > > > > > 
> > > > > > 
> > > > > > I'd rather break the API now than later. I don't think doing it later will simplify the move.
> > > > > > Any reason to hold off?
> > > > > 
> > > > > 
> > > > > As discussed, I'd wait until we all agreed keras_cv.data is the way to go, otherwise we need to break users twice
> > > > 
> > > > 
> > > > Did I misunderstand? I thought we got consensus?
> > > 
> > > 
> > > I believe the current consensus is keras_cv.data seems the top choice here, but:
> > > 
> > > 1. we give some time to think this through and see if there are any better options
> > > 2. we have to document what needs to be moved to the new namespace and what stays
> > > 
> > > In summary, can we wait before next Keras SIG meeting, which is next Friday? By that we should have a consensus, bump the version, and update all our documents
> > 
> > 
> > Personally I'd rather take the incremental step towards what we think is the right API. The longer we wait, the more things that will need to be migrated. We don't really have that many users other than the tutorials;
> > as I work on more tutorials I'd rather go with what we _think_ the final API will be (`from keras_cv.data import bounding_box`) even if we don't land on that, than use what we KNOW will not be the final API (`keras_cv.bounding_box`.
> > Definitely still open to feedback, but I'd like to have a strong reason as to why we wouldn't do this (i.e. an alternative API we are still considering, from what I understand we don't believe in either of the alternatives proposed)
> 
> I strongly believe we should wait until we have a consensus

Sounds good.

@CommunityMembers please feel free to comment on this namespacing, we're going to give it a few days to see if we can come up with any alternatives.

bhack:
> @CommunityMembers please feel free to comment on this namespacing, we're going to give it a few days to see if we can come up with any alternatives.

I am a little bit confused by the semantic nature of the `ops` folder and de-facto `ops` that are moved under `data`.

LukeWood:
Closed in favor of keeping top level namespaces.  Instead, @tanzhenyu will revert the PR moving keypoints to `ops` and we will create new top level namespaces.

This decision was made to optimize for users in the API:

`keras_cv.bounding_box` is short, nice to write, and cleaner than `keras_cv.data.bounding_box`

PR #859, opened 2022-09-28
Add martins comment

LukeWood:
@martin-gorner

### DIFF
--- a/keras_cv/models/resnet_v2.py
+++ b/keras_cv/models/resnet_v2.py
@@ -389,8 +389,8 @@ def ResNetV2(
 
         The output of the backbone model will be a dict with int as key and tensor as
         value. The int key represent the level of the feature output.
-        A typical feature pyramid has five levels corresponding to scales P3, P4, P5, 
-        P6, P7 in the backbone. Scale Pn represents a feature map 2n times smaller in 
+        A typical feature pyramid has five levels corresponding to scales P3, P4, P5,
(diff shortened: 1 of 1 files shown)


qlzh727 (approved):

keras_cv/models/resnet_v2.py
```diff
@@ -389,6 +389,9 @@ def as_backbone(self):
 
         The output of the backbone model will be a dict with int as key and tensor as
```
martin-gorner, line 390: Are you sure it's a dict ? I see a list when I test ([repro colab](https://colab.research.google.com/drive/1-Fvf4sd0X0R14WNeC5899hho9B6hVHEG#scrollTo=Ezw0mHdT3tmy))

keras_cv/models/resnet_v2.py
```diff
@@ -389,6 +389,9 @@ def as_backbone(self):
...
         value. The int key represent the level of the feature output.
+        A typical feature pyramid has five levels corresponding to scales P3, P4, P5,
+        P6, P7 in the backbone. Scale Pn represents a feature map 2n times smaller in
```
martin-gorner, line 393: It's not 2n but 2^n (two to the power of n)

keras_cv/models/resnet_v2.py
qlzh727: It is a dict, the model.outputs flatten any structure into a dict. If you check model.output, then you will see the dict. You can also check the raw output if you give it a input tensor.

Issue #702, opened 2022-08-18
Make `keras_cv.layers.Resizing` support bounding_boxes


LukeWood:

DB11051998:
Hi! is anyone working on this issue?  if noone, i would like to work on this.

LukeWood:
No, please feel free to contribute this.  You should port the Keras layer to the repo, update to use the KerasCV style, and add bounding box support.  thank you!

DB11051998:
Thanks @LukeWood for assigning me this.

DB11051998:
@LukeWood 

![image](https://user-images.githubusercontent.com/43936231/192819388-eded1c3b-4c43-4e3f-b9ea-085951b52aea.png)

when u say "port  the keras layer to the repo" do you mean importing the resizing class present in the keras in the __init__.py code of the boundingbox  package(I have attached the screenshot of it). Could you please correct me here. If i am wrong, could you please explain me the objective?

LukeWood:
This is what I mean:

This PR is incomplete, but you can use it as a starter
https://github.com/keras-team/keras-cv/pull/861

DB11051998:
Thanks @LukeWood !! it's a good starter code.

DB11051998:
![image](https://user-images.githubusercontent.com/43936231/193401964-9076b6cf-8ad4-45e2-8baf-93be4922c1a6.png)


 whenever i am calling the Resizing class, it is calling from the keras package, which is ultimately throwing an error, to it. why it is going to that package??....whereas in the __init__.py, i am importing this Resizing class

![image](https://user-images.githubusercontent.com/43936231/193402116-d9fe6fed-69ef-4f24-8083-4255a182885b.png)
this is the Resizing class, by taking the refernce from your code.

tre3x:
Hi, can i work on this issue?

LukeWood:
Yes, feel free to open a PR.  You’ll need to port the layer as I have done in the linked PR, then add bounding box support to it.

Snimm:
Hi, I want to work on this. Or is it done?

LukeWood:
> Hi, I want to work on this. Or is it done?

It works in pad to aspect ratio mode.  Support for other modes is more than welcome!

arcAman07:
Hey is this issue still open? I was looking forward to contribute to keras_cv

LukeWood:
I've actually completed this! Sorry, forgot to mark complete.

PR #861, opened 2022-09-28
Add bounding box support to resizing

LukeWood:
Fixes https://github.com/keras-team/keras-cv/issues/936, https://github.com/keras-team/keras-cv/issues/702

This allows us to resize with a pad which is the optimal resize mode for object detection.

Sample out from examples:
Screen Shot 2022-11-03 at 2 24 06 PM

---

### DIFF
--- a/examples/layers/preprocessing/bounding_box/resizing_demo.py
+++ b/examples/layers/preprocessing/bounding_box/resizing_demo.py
@@ -1,0 +1,40 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+mosaic_demo.py shows how to use the Mosaic preprocessing layer for
+object detection.
+"""
+import demo_utils
+import tensorflow as tf
+
+from keras_cv import layers
+import keras_cv
+
+def main():
+    dataset, _ = keras_cv.datasets.pascal_voc.load(
+        split='train',
+        bounding_box_format="xyxy",
+        batch_size=9,
+        img_size=(512, 512)
+    )
+    # ragged tensor of images
+    # sample = next(iter(dataset))
+    # resizing = layers.Resizing(height=512, width=512, bounding_box_format='xyxy')
+    # outputs = resizing(sample)
+    # result = dataset.map(resizing, num_parallel_calls=tf.data.AUTOTUNE)
+    demo_utils.visualize_data(dataset, bounding_box_format="xyxy")
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/layers/preprocessing/bounding_box/resizing_demo.py
+++ b/examples/layers/preprocessing/bounding_box/resizing_demo.py
@@ -22,10 +22,6 @@ import tensorflow_datasets as tfds
 import keras_cv
 from keras_cv import layers
 
-train_ds = tfds.load(
-    "voc/2007", split="train+test", with_info=False, shuffle_files=True
-)
-
 
 def preproc(inputs):
     image = inputs["image"]
@@ -46,15 +42,42 @@ def preproc(inputs):
     return {"images": image, "bounding_boxes": bounding_boxes}
 
 
+# pad to aspect ratio height > width
+train_ds = tfds.load(
+    "voc/2007", split="train+test", with_info=False, shuffle_files=True
+)
 train_ds = train_ds.map(preproc)
 train_ds = train_ds.apply(
     tf.data.experimental.dense_to_ragged_batch(16, drop_remainder=True)
 )
-print(train_ds)
 resizing = layers.Resizing(
     height=600, width=400, pad_to_aspect_ratio=True, bounding_box_format="xywh"
 )
-
 train_ds = train_ds.map(resizing)
-next(iter(train_ds))
+demo_utils.visualize_data(train_ds, bounding_box_format="xywh")
+
+# pad to aspect ratio width > height
+train_ds = tfds.load(
+    "voc/2007", split="train+test", with_info=False, shuffle_files=True
+)
+train_ds = train_ds.map(preproc)
+train_ds = train_ds.apply(
+    tf.data.experimental.dense_to_ragged_batch(16, drop_remainder=True)
+)
+resizing = layers.Resizing(
+    height=400, width=600, pad_to_aspect_ratio=True, bounding_box_format="xywh"
+)
+train_ds = train_ds.map(resizing)
+demo_utils.visualize_data(train_ds, bounding_box_format="xywh")
+
+# resize with distortion
+train_ds = tfds.load(
+    "voc/2007", split="train+test", with_info=False, shuffle_files=True
+)
+train_ds = train_ds.map(preproc)
+train_ds = train_ds.apply(
+    tf.data.experimental.dense_to_ragged_batch(16, drop_remainder=True)
+)
+resizing = layers.Resizing(height=400, width=600, bounding_box_format="xywh")
+train_ds = train_ds.map(resizing)
 demo_utils.visualize_data(train_ds, bounding_box_format="xywh")
--- a/examples/layers/preprocessing/bounding_box/resizing_demo.py
+++ b/examples/layers/preprocessing/bounding_box/resizing_demo.py
@@ -18,12 +18,14 @@ object detection.
 import demo_utils
 import tensorflow as tf
 import tensorflow_datasets as tfds
+
 import keras_cv
 from keras_cv import layers
 
 train_ds = tfds.load(
     "voc/2007", split="train+test", with_info=False, shuffle_files=True
 )
+
 
 def preproc(inputs):
     image = inputs["image"]
@@ -39,9 +41,10 @@ def preproc(inputs):
     gt_classes = tf.expand_dims(gt_classes, axis=-1)
     bounding_boxes = tf.concat([gt_boxes, gt_classes], axis=-1)
     bounding_boxes = keras_cv.bounding_box.convert_format(
-        bounding_boxes, images=image, source="yxyx", target='xywh'
+        bounding_boxes, images=image, source="yxyx", target="xywh"
     )
     return {"images": image, "bounding_boxes": bounding_boxes}
+
 
 train_ds = train_ds.map(preproc)
 train_ds = train_ds.apply(
@@ -49,8 +52,7 @@ train_ds = train_ds.apply(
 )
 print(train_ds)
 resizing = layers.Resizing(
-    height=600, width=400,
-    pad_to_aspect_ratio=True, bounding_box_format='xywh'
+    height=600, width=400, pad_to_aspect_ratio=True, bounding_box_format="xywh"
 )
 
 train_ds = train_ds.map(resizing)
--- a/examples/layers/preprocessing/bounding_box/resizing_demo.py
+++ b/examples/layers/preprocessing/bounding_box/resizing_demo.py
@@ -69,15 +69,3 @@ resizing = layers.Resizing(
 )
 train_ds = train_ds.map(resizing)
 demo_utils.visualize_data(train_ds, bounding_box_format="xywh")
-
-# resize with distortion
-train_ds = tfds.load(
-    "voc/2007", split="train+test", with_info=False, shuffle_files=True
-)
-train_ds = train_ds.map(preproc)
-train_ds = train_ds.apply(
-    tf.data.experimental.dense_to_ragged_batch(16, drop_remainder=True)
-)
-resizing = layers.Resizing(height=400, width=600, bounding_box_format="xywh")
-train_ds = train_ds.map(resizing)
-demo_utils.visualize_data(train_ds, bounding_box_format="xywh")
--- a/examples/layers/preprocessing/bounding_box/resizing_demo.py
+++ b/examples/layers/preprocessing/bounding_box/resizing_demo.py
@@ -18,15 +18,13 @@ object detection.
 import demo_utils
 import tensorflow as tf
 
-from keras_cv import layers
 import keras_cv
+from keras_cv import layers
+
 
 def main():
     dataset, _ = keras_cv.datasets.pascal_voc.load(
-        split='train',
-        bounding_box_format="xyxy",
-        batch_size=9,
-        img_size=(512, 512)
+        split="train", bounding_box_format="xyxy", batch_size=9, img_size=(512, 512)
     )
     # ragged tensor of images
     # sample = next(iter(dataset))
--- a/examples/layers/preprocessing/bounding_box/resizing_demo.py
+++ b/examples/layers/preprocessing/bounding_box/resizing_demo.py
@@ -17,22 +17,42 @@ object detection.
 """
 import demo_utils
 import tensorflow as tf
-
+import tensorflow_datasets as tfds
 import keras_cv
 from keras_cv import layers
 
+train_ds = tfds.load(
+    "voc/2007", split="train+test", with_info=False, shuffle_files=True
+)
 
-def main():
-    dataset, _ = keras_cv.datasets.pascal_voc.load(
-        split="train", bounding_box_format="xyxy", batch_size=9, img_size=(512, 512)
+def preproc(inputs):
+    image = inputs["image"]
+    image = tf.cast(image, tf.float32)
+    gt_boxes = inputs["objects"]["bbox"]
+    gt_boxes = keras_cv.bounding_box.convert_format(
+        gt_boxes,
+        images=image,
+        source="rel_yxyx",
+        target="yxyx",
     )
-    # ragged tensor of images
-    # sample = next(iter(dataset))
-    # resizing = layers.Resizing(height=512, width=512, bounding_box_format='xyxy')
-    # outputs = resizing(sample)
-    # result = dataset.map(resizing, num_parallel_calls=tf.data.AUTOTUNE)
-    demo_utils.visualize_data(dataset, bounding_box_format="xyxy")
+    gt_classes = tf.cast(inputs["objects"]["label"], tf.float32)
+    gt_classes = tf.expand_dims(gt_classes, axis=-1)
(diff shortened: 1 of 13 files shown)


LukeWood:
ok @tanzhenyu @qlzh727 @ianstenbit, I think this is ready for review.  This one was pretty tricky - but very critical to the OD story, so extra rigour on code review would be great.

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,287 @@
...
+W_AXIS = -2
+
+
+class Resizing(BaseImageAugmentationLayer):
+    """A preprocessing layer which resizes images.
+
+    This layer resizes an image input to a target height and width. The input
+    should be a 4D (batched) or 3D (unbatched) tensor in `"channels_last"`
+    format.  Input pixel values can be of any range (e.g. `[0., 1.)` or `[0,
+    255]`) and of interger or floating point dtype. By default, the layer will
```
ianstenbit, line 32: nit: integer

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,287 @@
...
+    For an overview and full list of preprocessing layers, see the preprocessing
+    [guide](https://www.tensorflow.org/guide/keras/preprocessing_layers).
+
+    Args:
+      height: Integer, the height of the output shape.
+      width: Integer, the width of the output shape.
+      interpolation: String, the interpolation method. Defaults to `"bilinear"`.
+        Supports `"bilinear"`, `"nearest"`, `"bicubic"`, `"area"`, `"lanczos3"`,
+        `"lanczos5"`, `"gaussian"`, `"mitchellcubic"`.
+      crop_to_aspect_ratio: If True, resize the images without aspect
```
ianstenbit, line 48: Need docs for the pad_to_aspect_ratio param (probably explaining that it's mutually exclusive with crop_to_aspect_ratio?)

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,287 @@
...
+        self.width = width
+        self.interpolation = interpolation
+        self.crop_to_aspect_ratio = crop_to_aspect_ratio
+        self.pad_to_aspect_ratio = pad_to_aspect_ratio
+        self._interpolation_method = keras_cv.utils.get_interpolation(interpolation)
+        self.bounding_box_format = bounding_box_format
+
+        if pad_to_aspect_ratio and crop_to_aspect_ratio:
+            raise ValueError(
+                "`Resizing()` expects either `crop_to_aspect_ratio` or "
```
ianstenbit, line 76: This comment to me implies that one of the two is required -- maybe instead it should be

"Resizing expects at most one of `crop_to_aspect_ratio` or `pad_to_aspect_ratio` to be True"?

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,287 @@
...
+                "`pad_to_aspect_ratio`, but not both."
+            )
+        if crop_to_aspect_ratio and bounding_box_format:
+            # TODO(lukewood): support `bounding_box.smart_resize()`
+            raise ValueError(
+                "Resizing() does not support `crop_to_aspect_ratio=True` "
```
ianstenbit, line 82: Nit: maybe just "Resizing() does not support `crop_to_aspect_ratio` with bounding boxes"?

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,287 @@
...
+                "and `bounding_box_format` at the same time.  In order to resize with "
+                "bounding boxes, please pass `crop_to_aspect_ratio=False`."
+            )
+        super().__init__(**kwargs)
+
+    def _augment(self, inputs):
```
ianstenbit, line 88: Is it really necessary to override _augment and _batch_augment for this layer?

Can't we just do the normal _augment_image, _augment_bounding_boxes, etc?

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,287 @@
...
+
+        if bounding_boxes is not None:
+            bounding_boxes = tf.squeeze(outputs["bounding_boxes"], axis=0)
+            inputs["bounding_boxes"] = bounding_boxes
+        return inputs
+
+    def _resize_with_distortion(self, inputs):
+        images = inputs.get("images", None)
+        bounding_boxes = inputs.get("bounding_boxes", None)
+        if bounding_boxes is not None:
```
ianstenbit, line 114: Seems to me like this guard clause should live in `init` with the one about `crop_to_aspect_ratio`

keras_cv/layers/preprocessing/resizing.py
ianstenbit: Then instead of having these big methods e.g. resize_with_pad, individual methods like augment_image could delegate to image-specific resizing functionality based on crop/pad settings

keras_cv/layers/preprocessing/resizing_test.py
```diff
@@ -0,0 +1,246 @@
...
+# limitations under the License.
+import numpy as np
+import tensorflow as tf
+from absl.testing import parameterized
+
+from keras_cv import layers as cv_layers
+
+
+class ResizingTest(tf.test.TestCase, parameterized.TestCase):
+    def _run_test(self, kwargs, height, width):
```
ianstenbit, line 22: nit: maybe _run_output_shape_test?

keras_cv/layers/preprocessing/resizing_test.py
```diff
@@ -0,0 +1,246 @@
...
+        images = tf.ragged.constant(
+            [
+                np.ones((8, 8, 3)),
+                np.ones((8, 4, 3)),
+                np.ones((4, 8, 3)),
+                np.ones((2, 2, 3)),
+            ],
+            dtype="float32",
+        )
+        boxes = tf.ragged.stack(
```
ianstenbit, line 229: I think given the complexity of the code, it would be good to have a bounding box test case where we assert the numerics of the resulting boxes

ianstenbit (changes requested):

keras_cv/layers/preprocessing/resizing.py
LukeWood: Good point; the guard class in batch augment will catch the case where no format is specified too so we have full coverage.

keras_cv/layers/preprocessing/resizing_test.py
LukeWood: Great point; I added some assertions on the numerics here.

keras_cv/layers/preprocessing/resizing.py
LukeWood: For now lets keep this given that it matches `keras.layers.Resizing` and that this layer is used in many inference time pipelines.  We want to make sure we preserve the performance of these inference pipelines, so by avoiding unexpected vmaps and map_fns we can ensure that we don't regress on performance.

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,282 @@
...
+        if pad_to_aspect_ratio and crop_to_aspect_ratio:
+            raise ValueError(
+                "`Resizing()` expects at most one of `crop_to_aspect_ratio` or "
+                "`pad_to_aspect_ratio` to be True."
+            )
+
+        if not pad_to_aspect_ratio and bounding_box_format:
+            raise ValueError(
+                "Resizing() only supports bounding boxes when in "
+                "`pad_to_aspect_ratio=True` model.  "
```
ianstenbit, line 85: nit: s/model/mode

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,282 @@
...
+import keras_cv.utils
+from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
+    BaseImageAugmentationLayer,
+)
+
+H_AXIS = -3
+W_AXIS = -2
+
+
+class Resizing(BaseImageAugmentationLayer):
```
ianstenbit, line 26: Let's add this to `with_labels_test.py` and `with_mixed_precision_test.py` (I'm not 100% certain the latter will pass due to the hardcoded fp32 in this file)

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,282 @@
...
+        outputs = self._batch_augment(inputs)
+
+        if images is not None:
+            images = tf.squeeze(outputs["images"], axis=0)
+            inputs["images"] = images
+
+        if bounding_boxes is not None:
+            bounding_boxes = tf.squeeze(outputs["bounding_boxes"], axis=0)
+            inputs["bounding_boxes"] = bounding_boxes
+        return inputs
```
ianstenbit, line 112: I think in its current form this will pass through any keys included in `inputs` that aren't augmented (e.g. segmentation masks).

We should either:
- Throw if segmasks (or other unsupported data) are passed in
- Drop unsupported data from the output
- Add support for segmasks (and any other required data keys for our KPLs)

This is one benefit that we lose by overriding the BaseImageAugmentationLayer's `augment` method, but if it's necessary let's at least make sure we don't silently pass through broken segmentation masks.

ianstenbit (changes requested):

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,282 @@
...
+        interpolation: String, the interpolation method. Defaults to `"bilinear"`.
+            Supports `"bilinear"`, `"nearest"`, `"bicubic"`, `"area"`, `"lanczos3"`,
+            `"lanczos5"`, `"gaussian"`, `"mitchellcubic"`.
+        crop_to_aspect_ratio: If True, resize the images without aspect
+            ratio distortion. When the original aspect ratio differs from the target
+            aspect ratio, the output image will be cropped so as to return the
+            largest possible window in the image (of size `(height, width)`) that
+            matches the target aspect ratio. By default
+            (`crop_to_aspect_ratio=False`), aspect ratio may not be preserved.
+        pad_to_aspect_ratio:
```
tanzhenyu, line 54: no description here?

keras_cv/layers/preprocessing/resizing.py
tanzhenyu: Do we want to support segmasks before 0.4? This would be helpful for DeepLab

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,282 @@
...
+            resize_scale = tf.math.minimum(height_scale, width_scale)
+            target_height = img_height * resize_scale
+            target_width = img_width * resize_scale
+
+            image = tf.image.resize(
+                image,
+                size=(target_height, target_width),
+                method=self._interpolation_method,
+            )
+            if boxes is not None:
```
tanzhenyu, line 158: combine line 137 with the logic here?

tanzhenyu (commented):
Thanks for the PR!

keras_cv/layers/preprocessing/resizing.py
LukeWood: Actually does pass!  Great !

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,282 @@
...
+
+    def _augment(self, inputs):
+        images = inputs.get("images", None)
+        bounding_boxes = inputs.get("bounding_boxes", None)
+
+        if images is not None:
+            images = tf.expand_dims(images, axis=0)
+            inputs["images"] = images
+
+        if bounding_boxes is not None:
```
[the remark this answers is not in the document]
LukeWood: This will be caught in batch_augment called 5 lines below.

keras_cv/layers/preprocessing/resizing.py
LukeWood: That would be great but lets keep this PR localized.  I'll raise an error if any key other than `['images', 'labels', 'targets', 'bounding_boxes']` is passed for now.

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,282 @@
...
+            raise ValueError(
+                "Resizing(crop_to_aspect_ratio=True) does not support "
+                "bounding box inputs.  Please use `pad_to_aspect_ratio=True` when "
+                "processing bounding boxes with Resizing()."
+            )
+        inputs["images"] = images
+        size = [self.height, self.width]
+
+        if self.interpolation == "nearest":
```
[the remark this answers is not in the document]
LukeWood: Done!  

        # tf.image.resize will always output float32 and operate more
        # efficiently on float32 unless interpolation is nearest, in which case
        # ouput type matches input type.

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -0,0 +1,282 @@
...
+                resize_with_crop_to_aspect, images, fn_output_signature=spec
+            )
+        else:
+            images = resize_with_crop_to_aspect(images)
+
+        inputs["images"] = images
+        return inputs
+
+    def call(self, inputs, training=True):
+        # inputs = self._ensure_inputs_are_compute_dtype(inputs)
```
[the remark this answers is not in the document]
LukeWood: Uncommented

LukeWood:
All review comments addressed!

keras_cv/layers/preprocessing/resizing.py
tanzhenyu: raising error in this PR sounds good -- can you also add a TODO here as well?

keras_cv/utils/preprocessing.py
```diff
@@ -16,6 +16,38 @@
 
 from keras_cv import core
 
+ResizeMethod = tf.image.ResizeMethod
```
tanzhenyu, line 19: do we really need this since tf.image.ResizeMethod is already string based?
https://github.com/tensorflow/tensorflow/blob/v2.10.0/tensorflow/python/ops/image_ops_impl.py#L1419-L1429
Having less indirection would be probably more preferrable.

examples/layers/preprocessing/bounding_box/resizing_demo.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2022 The KerasCV Authors
```
tanzhenyu, line 1: Can we test this with the RetinaNet training script?

keras_cv/utils/preprocessing.py
ianstenbit: +1 -- if possible to remove this I think that would be easier to follow

ianstenbit (approved):

examples/layers/preprocessing/bounding_box/resizing_demo.py
LukeWood: Yeah; Once we have RandomCropAndResize we'll have a much clearer value add here I think.  Added to the proc eval function portion.

keras_cv/datasets/pascal_voc/load.py
```diff
@@ -73,12 +75,11 @@ def load(
             For a list of supported formats, please  Refer
             [to the keras.io docs](https://keras.io/api/keras_cv/bounding_box/formats/)
             for more details on supported bounding box formats.
-        batch_size: (Optional) how many instances to include in batches after loading. If
-            not provided, no batching will occur.
-        shuffle_buffer: (Optional) the size of the buffer to use in shuffling.
-        shuffle_files: (Optional) whether or not to shuffle files, defaults to True.
```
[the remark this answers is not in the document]
LukeWood: I just re-ordered them to match the argument orders

keras_cv/utils/preprocessing.py
LukeWood: Done

examples/layers/preprocessing/bounding_box/resizing_demo.py
tanzhenyu: Can you clarify? Given this has crop and pad flags, how do we expect OD pipelines to use Resizing and RandomCropAndResize?

keras_cv/datasets/pascal_voc/load.py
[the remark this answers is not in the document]
tanzhenyu: `shuffle_files` seem to be removed, no?

keras_cv/utils/__init__.py
```diff
@@ -14,6 +14,8 @@
 
 from keras_cv.utils.fill_utils import fill_rectangle
 from keras_cv.utils.preprocessing import blend
+from keras_cv.utils.preprocessing import ensure_tensor
```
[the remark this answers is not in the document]
LukeWood: Just an attempt to standardize our namespacing and export it from the same place as the other utilities

keras_cv/datasets/pascal_voc/load.py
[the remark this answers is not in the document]
tanzhenyu: I'd suggest adding `shuffle_files` back

keras_cv/utils/__init__.py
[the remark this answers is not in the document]
tanzhenyu: Ok, `ensure_tensor` is a weird name though, I wonder what is this being used for. But unrelated to the PR.

keras_cv/utils/preprocessing.py
```diff
@@ -16,6 +16,37 @@
...
+    "bicubic": tf.image.ResizeMethod.BICUBIC,
+    "area": tf.image.ResizeMethod.AREA,
+    "lanczos3": tf.image.ResizeMethod.LANCZOS3,
+    "lanczos5": tf.image.ResizeMethod.LANCZOS5,
+    "gaussian": tf.image.ResizeMethod.GAUSSIAN,
+    "mitchellcubic": tf.image.ResizeMethod.MITCHELLCUBIC,
+}
+
+
+def get_interpolation(interpolation):
```
tanzhenyu, line 31: just remove this file?

keras_cv/datasets/pascal_voc/load.py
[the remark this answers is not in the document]
LukeWood: Oops - yes it was.  Sorry about that, fixed.

keras_cv/utils/preprocessing.py
LukeWood: All of preprocessing.py?  It has a lot of utilities in it right now.

LukeWood: Actually, I'll need the same logic in CropAndResize - lets just keep it.  I think it is actually somewhat useful for consistency across all layers.

LukeWood:
@tanzhenyu let me know if anything else needs to be updated!

PR #862, opened 2022-09-28
Support no image size in PascalVOC loader

LukeWood:
This will allow us to support Ragged pipelines where the batch sizes are randomly set each batch.

### DIFF
--- a/keras_cv/datasets/pascal_voc/load.py
+++ b/keras_cv/datasets/pascal_voc/load.py
@@ -22,13 +22,17 @@ from keras_cv import bounding_box
 def curry_map_function(bounding_box_format, img_size):
     """Mapping function to create batched image and bbox coordinates"""
 
-    resizing = keras.layers.Resizing(
-        height=img_size[0], width=img_size[1], crop_to_aspect_ratio=False
-    )
+    if img_size is not None:
+        resizing = keras.layers.Resizing(
+            height=img_size[0], width=img_size[1], crop_to_aspect_ratio=False
+        )
 
     # TODO(lukewood): update `keras.layers.Resizing` to support bounding boxes.
     def apply(inputs):
-        inputs["image"] = resizing(inputs["image"])
+        # Support image size none.
+        if img_size is not None:
+            inputs["image"] = resizing(inputs["image"])
+
         inputs["objects"]["bbox"] = bounding_box.convert_format(
             inputs["objects"]["bbox"],
             images=inputs["image"],
--- a/keras_cv/datasets/pascal_voc/load.py
+++ b/keras_cv/datasets/pascal_voc/load.py
@@ -55,7 +55,7 @@ def load(
     batch_size=None,
     shuffle=True,
     shuffle_buffer=None,
-    img_size=(512, 512),
+    img_size=None,
 ):
     """Loads the PascalVOC 2007 dataset.
 
--- a/keras_cv/datasets/pascal_voc/load.py
+++ b/keras_cv/datasets/pascal_voc/load.py
@@ -53,7 +53,6 @@ def load(
     split,
     bounding_box_format,
     batch_size=None,
-    shuffle=None,
     shuffle_buffer=None,
     img_size=None,
 ):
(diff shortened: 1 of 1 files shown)


tanzhenyu (commented):
According to the description, "where the batch sizes" -- is batch sizes related to the image sizes in this PR?

LukeWood:
TODO:

- update keras.io guides to specify image size
- make sure other places using the data loader don't rely on img_size

tanzhenyu:
> TODO:
> 
> * update keras.io guides to specify image size
> * make sure other places using the data loader don't rely on img_size

Related question -- any chance we don't make this as a public API? This seems to be a very thin wrapper for now and also not very stable.

LukeWood:
You can't batch non-resized images, so I've removed the default batch size to make the default constructor work.

tanzhenyu:
> You can't batch non-resized images, so I've removed the default batch size to make the default constructor work.

Oh you submitted another commit after my initial review.

keras_cv/datasets/pascal_voc/load.py
```diff
@@ -69,10 +73,13 @@ def load(
...
-        batch_size: how many instances to include in batches after loading
-        shuffle: whether or not to shuffle the dataset.  Defaults to True.
+        batch_size: (Optional) how many instances to include in batches after loading. If
+            not provided, no batching will occur.
+        shuffle: whether or not to shuffle the dataset.  Defaults to None, which indicates
+            shuffling will only occur if a `shuffle_buffer` is provided.
         shuffle_buffer: the size of the buffer to use in shuffling.
-        img_size: the size to resize the images to.
+        img_size: (Optional) size to resize the images to.  By default, images are not
+            resized and ragged batches are produced if batching occurs.
```
tanzhenyu, line 82: "ragged batch" is confusing, consider "tf.RaggedTensor"

keras_cv/datasets/pascal_voc/load.py
```diff
@@ -69,10 +73,13 @@ def load(
             For a list of supported formats, please  Refer
             [to the keras.io docs](https://keras.io/api/keras_cv/bounding_box/formats/)
             for more details on supported bounding box formats.
-        batch_size: how many instances to include in batches after loading
-        shuffle: whether or not to shuffle the dataset.  Defaults to True.
+        batch_size: (Optional) how many instances to include in batches after loading. If
+            not provided, no batching will occur.
+        shuffle: whether or not to shuffle the dataset.  Defaults to None, which indicates
```
tanzhenyu, line 78: If that's the case, then we don't need a `shuffle` argument. Just `shuffle_buffer` being None would mean no shuffling, and being int would mean shuffling.

keras_cv/datasets/pascal_voc/load.py
```diff
@@ -49,9 +53,9 @@ def load(
     split,
     bounding_box_format,
     batch_size=None,
-    shuffle=True,
+    shuffle=None,
     shuffle_buffer=None,
-    img_size=(512, 512),
+    img_size=None,
```
tanzhenyu, line 58: IMHO, providing a `img_size` around `Resizing` is not a very useful argument, we might instead want to provide arguments around `RandomCropResize` (or whatever the name is)

tanzhenyu (commented):
Thanks for the PR!

keras_cv/datasets/pascal_voc/load.py
LukeWood: To be honest, I'm considering just dropping the img_size argument altogether

LukeWood: It feels like we mght want to just augment images indiviudally and use `tf.data.padded_batch`

LukeWood: Although; it is nice for visualization.

tanzhenyu: I think we should probably allow full customization for KPLs and probably don't even need a public data loader.

tanzhenyu (approved):

PR #863, opened 2022-09-28
Add licensing notice and information about model safety to model card

LukeWood:


ianstenbit (approved):
Maybe at the top of the README we should reference that this model card is adapted from HuggingFace?

Probably worth waiting for an LGTM from Francois

fchollet (approved):

PR #869, opened 2022-09-29
revert keypoint refactoring.

tanzhenyu:
# What does this PR do?

Documenting the reasoning:

Throughout many discussions, there are three options:
1. keras_cv.data that touches the conversion between data formats
2. using ops as a clustered namespace
3. keras_cv.keypoint

We feel like option 3 introduces issues for overlapping namespaces such as sampling that touches both boxes and keypoints, and introduce way too many namespaces as we're working on point clouds, but this gives us the best compromise


LukeWood:
looks like format fails; but LGTM as soon as that is fixed

LukeWood (approved):

PR #878, opened 2022-10-04
Support mixed-precision training in basic_training script

ianstenbit:
Some test runs have shown that we have some KPLs which don't support this yet, but the script still can.

### DIFF
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -71,7 +71,7 @@ flags.DEFINE_boolean(
 )
 flags.DEFINE_boolean(
     "use_mixed_precision",
-    True,
+    False,
     "Whether or not to use FP16 mixed precision for training.",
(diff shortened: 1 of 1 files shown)


examples/training/classification/imagenet/basic_training.py
```diff
@@ -68,6 +69,11 @@
 flags.DEFINE_boolean(
     "use_xla", True, "Whether or not to use XLA (jit_compile) for training."
 )
+flags.DEFINE_boolean(
+    "use_mixed_precision",
+    True,
```
tanzhenyu, line 74: Default to False until this has been verified to produce sota weights?

tanzhenyu (approved):

examples/training/classification/imagenet/basic_training.py
ianstenbit: sg -- Done.

PR #875, opened 2022-10-03
Address review comments from #874

qlzh727:
# What does this PR do?

1. Add a param to scale up the output for segmentation head so that the output mask
can have the same shape as the input image.
2. Enforce the output layer dtype to float32, to avoid the numerical issue when running
with mixed precision.

Fixes #874

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/keras-team/keras-cv/blob/master/.github/CONTRIBUTING.md),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue? Please add a link
      to it if that's the case.
- [x] Did you write any new necessary tests?
- [ ] If this adds a new model, can you run a few training steps on TPU in Colab to ensure that no XLA incompatible OP are used?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #882, opened 2022-10-05
Fix the unit test for segmentation head serialization.

qlzh727:
# What does this PR do?
Not sure why the unit test didn't catch it in https://github.com/keras-team/keras-cv/pull/875. All the test check was passing.

## Before submitting
      Pull Request section?
      to it if that's the case.


ianstenbit (approved):
It might be worth digging into why our CI didn't catch this. From looking at the logs, it seems like the serialization test is run, so I'm not certain what happened.

PR #883, opened 2022-10-05
Modify len with shape and index.

pashu123:
This makes it easy for the downstream compilers like MLIR to trace the code.

# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/keras_cv/models/generative/stable_diffusion/__internal__/layers/group_normalization.py
+++ b/keras_cv/models/generative/stable_diffusion/__internal__/layers/group_normalization.py
@@ -84,7 +84,7 @@ class GroupNormalization(tf.keras.layers.Layer):
         return gamma, beta
 
     def _create_broadcast_shape(self, input_shape):
-        broadcast_shape = [1] * len(input_shape)
+        broadcast_shape = [1] * input_shape.shape[0]
         broadcast_shape[self.axis] = input_shape[self.axis] // self.groups
         broadcast_shape.insert(self.axis, self.groups)
         return broadcast_shape


pashu123:
@LukeWood @fchollet PTAL.

LukeWood:
Please review the CLA!

Also, can you provide some more info on this?

pashu123:
> Please review the CLA!
> 
> Also, can you provide some more info on this?

Thanks, I have reviewed the CLA. So, we were trying to use the runtime  https://github.com/iree-org/iree which in turn calls https://github.com/tensorflow/tensorflow/tree/master/tensorflow/compiler/mlir for tracing the tensorflow computation graph. The problem is `len()` is not supported under tracing and hence it can be replaced by `.shape[0]`.

pashu123:
@LukeWood PTAL.

LukeWood:
does this not change the behavior?  Is the original implementation bugged?  

```
broadcast_shape = [1] * input_shape.shape[0]
```

is NOT the same as len(input_shape)!

pashu123:
> does this not change the behavior? Is the original implementation bugged?
> 
> ```
> broadcast_shape = [1] * input_shape.shape[0]
> ```
> 
> is NOT the same as len(input_shape)!

Do you mean `input_shape.shape[0]` is not same as `len(input_shape)`? Well if input_shape is a 1d tensor (input_shape is indeed a 1d tensor) they are the same?

LukeWood:
> > does this not change the behavior? Is the original implementation bugged?
> > ```
> > broadcast_shape = [1] * input_shape.shape[0]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is NOT the same as len(input_shape)!
> 
> Do you mean `input_shape.shape[0]` is not same as `len(input_shape)`? Well if input_shape is a 1d tensor (input_shape is indeed a 1d tensor) they are the same?

True, is our GroupNormalization only called on 1D tensors?  If so, then this PR is fine.

how about we have an if statement validating that the input is a 1D tensor.  This way, we can raise a value error if it’s not and know we’re using the layer incorrectly.  If you add a guard clause like that I can then pull the PR, run the tests, and merge this if it all passes.

tanzhenyu:
Looks like there's merge conflict

LukeWood:
> @LukeWood Why we have still a custom class here? Is this different from the Group normalization we have in Keras now? https://github.com/keras-team/keras/blob/master/keras/layers/normalization/group_normalization.py

Verisoning

tanzhenyu:
This is fixed per https://github.com/keras-team/keras-cv/pull/1035

PR #897, opened 2022-10-10
[draft] Initial noise scheduler for StableDiffusion

ianstenbit:
@LukeWood


LukeWood:
Added:  feel free to push commits to https://github.com/keras-team/keras-io/pull/1088, and do anything else to the guide you think makes sense.  It doesn't run right now, I have a few more bugs to iron out.  Happy to have some help on that

Issue #108, opened 2022-02-07
Define the best API for users to apply augmentations in a probabilistic manner.


LukeWood:
Maybe this will look like:
```
RandomChance(layer, rate=0.5)
```

or maybe:
```
MaybeApply(layer, rate=0.5)
```

Either way, this is important for the composition of image augmentation pipelines.

bhack:
I think that we Need to think about this evaluating we want to achieve in:
 https://github.com/keras-team/keras-cv/issues/41

innat:
@LukeWood @bhack @qlzh727 @kartik4949 

I think two types of functions need to be addressed regarding the processing layers.  For examples:

1. Randomly apply the layer 

```python

random_mixup  = RandomApply(Mixup(),  prob = 0.5)
random_cutmix = RandomApply(CutMix(), prob = 0.5)
random_equliz = RandomApply(Equalization(), prob = 0.5)
```

2. Randomly pick a layers

```python

random_mixup  = RandomApply(Mixup(),  prob = 0.5)
random_cutmix = RandomApply(CutMix(), prob = 0.5)

random_blur   = RandomPick( 
          [
              RandomApply(GaussianBlur(), prob = 0.5),
              RandomApply(MotionBlur(),   prob = 0.5),
              RandomApply(MediamBlur(),   prob = 0.5)
          ], 
     prob = 0.8
)
```

kartik4949:
@innat 
I think by RandomPick you mean RandAug in defined space!

innat:
It's like the `OneOf` method that is provided in [argumentation (cell 7)](https://albumentations.ai/docs/examples/example/) or img_aug libraries.

bhack:
I think that we need to reason a little bit more generally about what kind of access to the augmentation layers an [auto augmentation](https://www.ilfattoquotidiano.it/2022/02/15/dopo-due-anni-di-emergenza-covid-i-medici-sono-costretti-a-scioperare-turni-insostenibili-poche-tutele-carenza-di-personale-poi-lo-schiaffo-dei-mancati-ristori/6494438/) policy need to achieve. 
Then I suppose we could think at `RandomApply` as just one o the simplest policy.

I don't know if @haifeng-jin has any feedbaqck.

bhack:
Just a reminder, we had also https://github.com/keras-team/autokeras/issues/879

LukeWood:
We've decided to include a `rate` argument in a base augmentation layer.  This provides the best user API imo:

[...]

PR #898, opened 2022-10-10
Improve FasterRCNN model quality.

tanzhenyu:
This PR add mores paper specific training logic and improves FasterRCNN model quality / training by:
1) add padding
2) add variance during box encoding and decoding
3) less anchor gen
4) larger image size
5) more aggressive resizing

It makes the training script a little more verbose, which is intentional given our data augmentation layers set do not support native box format and paper specific resizing logic.

### DIFF
--- a/examples/models/object_detection/faster_rcnn/train.py
+++ b/examples/models/object_detection/faster_rcnn/train.py
@@ -38,7 +38,7 @@ eval_ds = tfds.load("voc/2007", split="test", with_info=False)
 
 model = keras_cv.models.FasterRCNN(classes=20, bounding_box_format="yxyx")
 
-
+# TODO: migrate to KPL.
 def resize_and_crop_image(
     image,
     desired_size,
--- a/examples/models/object_detection/faster_rcnn/train.py
+++ b/examples/models/object_detection/faster_rcnn/train.py
@@ -38,6 +38,7 @@ eval_ds = tfds.load("voc/2007", split="test", with_info=False)
 
 model = keras_cv.models.FasterRCNN(classes=20, bounding_box_format="yxyx")
 
+
 # TODO: migrate to KPL.
 def resize_and_crop_image(
     image,
--- a/examples/models/object_detection/faster_rcnn/train.py
+++ b/examples/models/object_detection/faster_rcnn/train.py
@@ -25,23 +25,143 @@ import tensorflow_datasets as tfds
 
 import keras_cv
 
-physical_devices = tf.config.list_physical_devices("GPU")
-tf.config.set_visible_devices(physical_devices[:1], "GPU")
-
 # parameters from FasterRCNN [paper](https://arxiv.org/pdf/1506.01497.pdf)
 
 global_batch = 4
-image_size = [256, 256, 3]
+image_size = [640, 640, 3]
 train_ds = tfds.load(
-    "voc/2007", split="train+test", with_info=False, shuffle_files=True
+    "voc/2007", split="train+validation", with_info=False, shuffle_files=True
 ).concatenate(
-    tfds.load("voc/2012", split="train+test", with_info=False, shuffle_files=True)
+    tfds.load("voc/2012", split="train+validation", with_info=False, shuffle_files=True)
 )
-eval_ds = tfds.load("voc/2007", split="validation", with_info=False).concatenate(
-    tfds.load("voc/2012", split="validation", with_info=False)
-)
+eval_ds = tfds.load("voc/2007", split="test", with_info=False)
 
 model = keras_cv.models.FasterRCNN(classes=20, bounding_box_format="yxyx")
+
+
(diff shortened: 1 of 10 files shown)


keras_cv/bounding_box/converters.py
```diff
@@ -25,6 +25,25 @@ class RequiresImagesException(Exception):
     pass
 
 
+def clip_boxes(boxes, box_format, image_shape):
```
ianstenbit, line 28: keras_cv.bounding_box.clip_to_image should do this, right?
(from bounding_box/utils.py)

ianstenbit: (If not, and this needs to be a separate method, maybe it should also live in `utils.py` instead of `converters.py`?)

examples/models/object_detection/faster_rcnn/train.py
```diff
@@ -25,25 +25,145 @@
...
+
+def get_non_empty_box_indices(boxes):
+    # Selects indices if box height or width is 0.
+    height = boxes[:, 2] - boxes[:, 0]
+    width = boxes[:, 3] - boxes[:, 1]
+    indices = tf.where(tf.logical_and(tf.greater(height, 0), tf.greater(width, 0)))
+    return indices[:, 0]
+
+
+def resize_fn(image, gt_boxes, gt_classes):
```
ianstenbit, line 154: Can this functionality be achieved with RandomCropAndResize? If not, should we either update that layer or create a new KPL which provides this functionality?

keras_cv/bounding_box/converters.py
tanzhenyu: Ideally it does, but that would probably error out given:
"`bounding_boxes` must include a class_id index on the final "
-- as discussed last time, this is a huge tech debt we should get rid of, but I don't want it to block the current development as well.

examples/models/object_detection/faster_rcnn/train.py
tanzhenyu: After training many classification and detection models, I have came to conclusion that a RandomResizeAndCrop-like behavior is required, but it cannot be achieved with the layers we have today.
What I would really like to see is, once I get to train some segmentation models and achieve SOTA, we can try to discuss what the best RandomResizeAndCrop layer should look like.
Note: Also the layer we have today does not support native box format, the same issue with the above comment (as mentioned in the PR description)

bhack: I think that we could re-evaluate if some layers need to have an attached policy:
- https://github.com/keras-team/keras-cv/issues/108#issuecomment-1040626481
- https://github.com/keras-team/keras-cv/issues/108#issuecomment-1043598240

tanzhenyu:
@innat Once you patch this PR, you should get better results.

examples/models/object_detection/faster_rcnn/train.py
tanzhenyu: The resize is somewhat different than other augmentation layers we have, especially for auto/rand-aug, colorized ops are used extensively, and only 1-2 geometric ops are popular choices

keras_cv/bounding_box/converters.py
ianstenbit: Gotcha. If that method doesn't work as intended and it's not simple to update it, I think we should keep this version of it internal to the training script.

Otherwise we have two separate but nearly identical utilities.

examples/models/object_detection/faster_rcnn/train.py
ianstenbit: sgtm. Maybe let's put a TODO in the code here for now to migrate this to a KPL?

keras_cv/bounding_box/converters.py
tanzhenyu: Done.

examples/models/object_detection/faster_rcnn/train.py
tanzhenyu: Done.

bhack: In the case we will want to re-evaluate some policy in the future, at least for the classification task there was a nice survey out recently:
https://arxiv.org/abs/2206.06544

keras_cv/bounding_box/converters.py
```diff
@@ -25,6 +25,25 @@ class RequiresImagesException(Exception):
     pass
 
 
+def _clip_boxes(boxes, box_format, image_shape):
```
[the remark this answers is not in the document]
tanzhenyu: Good point. Done.

examples/models/object_detection/faster_rcnn/train.py
bhack: I don't have the permission level to assign it to you but it is here:
https://github.com/keras-team/keras-cv/issues/900

PR #902, opened 2022-10-11
Minor performance improvements to basic ImageNet training

ianstenbit:
- Switches to horizontal-only RandomFlip (performed very slightly better in my experiments on MobileNet and EfficientNet)
- Adds global clipnorm
- Increases max epochs to 1000 (training on large TPU clusters with larger batch size is taking more epochs to converge in some cases, setting 1000 as a conservative upper bound)/

### DIFF
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -97,6 +97,7 @@ if FLAGS.use_mixed_precision:
 
 CLASSES = 1000
 IMAGE_SIZE = (224, 224)
+# An upper bound for number of epochs (this script uses EarlyStopping).
 EPOCHS = 1000
 
 """
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -97,7 +97,7 @@ if FLAGS.use_mixed_precision:
 
 CLASSES = 1000
 IMAGE_SIZE = (224, 224)
-EPOCHS = 250
+EPOCHS = 1000
 
 """
 We start by detecting the type of accelerators we have available and picking an
@@ -147,7 +147,7 @@ We define a set of augmentation layers and then apply them to our input dataset.
 
 
 AUGMENT_LAYERS = [
-    keras_cv.layers.RandomFlip(),
+    keras_cv.layers.RandomFlip(mode="horizontal"),
     keras_cv.layers.RandAugment(value_range=(0, 255), magnitude=0.3),
     keras_cv.layers.CutMix(),
 ]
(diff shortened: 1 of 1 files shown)


examples/training/classification/imagenet/basic_training.py
```diff
@@ -97,7 +97,7 @@
 
 CLASSES = 1000
 IMAGE_SIZE = (224, 224)
-EPOCHS = 250
+EPOCHS = 1000
```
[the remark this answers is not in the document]
ianstenbit: This script does use EarlyStopping -- the value here is an upper bound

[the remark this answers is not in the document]
tanzhenyu: Maybe put a comment here saying this is an upper bound

examples/training/classification/imagenet/basic_training.py
```diff
@@ -147,7 +147,7 @@
 
 
 AUGMENT_LAYERS = [
-    keras_cv.layers.RandomFlip(),
+    keras_cv.layers.RandomFlip(mode="horizontal"),
```
tanzhenyu, line 150: Instead, can we change the default mode for RandomFlip? vertical flip doesn't seem right to me in the first place

examples/training/classification/imagenet/basic_training.py
```diff
@@ -188,8 +188,9 @@ def augment(img, label):
 """
 
 
-optimizer = optimizers.SGD(learning_rate=INITIAL_LEARNING_RATE, momentum=0.9)
-
+optimizer = optimizers.SGD(
+    learning_rate=INITIAL_LEARNING_RATE, momentum=0.9, global_clipnorm=10
```
tanzhenyu, line 192: Does this help you to stabilize the training?

ianstenbit: Yes, for a few models I've seen some wild swings in validation accuracy which were resolved with this clipping.

examples/training/classification/imagenet/basic_training.py
ianstenbit: Yes, I'll update the default in a separate PR. I think it's nice to make it explicit here anyway.

examples/training/classification/imagenet/basic_training.py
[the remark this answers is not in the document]
ianstenbit: Done

tanzhenyu (approved):

examples/training/classification/imagenet/basic_training.py
tanzhenyu: Maybe this would be where the warmup lr schedule I mentioned earlier would help

DavidLandup0:
Shouldn't the mixed float policy use `bfloat16` instead of `float16` for TPUs? Maybe add that in the accelerator check conditional?

tanzhenyu:
> Shouldn't the mixed float policy use `bfloat16` instead of `float16` for TPUs? Maybe add that in the accelerator check conditional?

Yeah we should do that

DavidLandup0:
Created #908 with change to `float16`, a typo fix and a warmup scheduler and a flag for turning it on and off

PR #903, opened 2022-10-11
Make RandomFlip perform horizontal flipping only by default

ianstenbit:
Based on training experiments on ImageNet classification using our `basic_training.py` script, using horizontal-only flipping produces minor performance improvements for at least MobileNet and EfficientNet.


tanzhenyu:
Maybe add some comments as "why make this change" as well

LukeWood:
One consideration we should make:

when we decided to port some layers from Keras core to KerasCV we decided that the ported layers would strictly be a superset of functionality (i.e. augment bounding boxes too).  The thought here was to minimize API drift between two identically named symbols.

Honestly I still support this change because vertical flips aren't very useful, but a thought regarding future similar changes.

PR #906, opened 2022-10-11
Add pre-trained weights for EfficientNetV2B0

ianstenbit:
Scores 75.3% top-1 accuracy, compared to paper-claimed result of 78.7%


bhack:
Are these weights from https://storage.googleapis.com/cloud-tpu-checkpoints/efficientnet/v2/efficientnetv2-b0.tgz ?

ianstenbit:
> Are these weights from https://storage.googleapis.com/cloud-tpu-checkpoints/efficientnet/v2/efficientnetv2-b0.tgz ?

No -- all of our pretrained weights are trained using code from KerasCV (for reproducibility)

bhack:
Will the training script that produced these weights land later in the repo?

ianstenbit:
> Will the training script that produced these weights land later in the repo?

It's in the repo already: https://github.com/keras-team/keras-cv/blob/master/examples/training/classification/imagenet/basic_training.py

tanzhenyu (approved):
Thanks for the PR!

bhack:
> It's in the repo already: https://github.com/keras-team/keras-cv/blob/master/examples/training/classification/imagenet/basic_training.py

Ok, so we are using the same augmentation as the original paper/check-point right?

ianstenbit:
> > It's in the repo already: https://github.com/keras-team/keras-cv/blob/master/examples/training/classification/imagenet/basic_training.py
> 
> Ok, so we are using the same augmentation as the original paper/checkpoint right?

No -- we're using a standard Flip + RandAugment + CutMix augmentation pipeline for training all backbone models.
The current goal is not to exactly match SOTA results for all models, but to produce reproducible training results for all models which are 90+% of SOTA using a modular training script. 

Model specific training scripts can be added (see [contributing guide](https://github.com/keras-team/keras-cv/blob/master/examples/training/classification/imagenet/basic_training.py)) to get these models to exactly match SOTA results in the future.

bhack:
> No -- we're using a standard Flip + RandAugment + CutMix augmentation pipeline for training all backbone models.
The current goal is not to exactly match SOTA results for all models, but to produce reproducible training results for all models which are 90+% of SOTA using a modular training script.

Is this documented/advised somewhere?
Are you applying eval preprocessing to these models as the reference impl?

ianstenbit:
> Is this documented/advised somewhere?
> Are you applying eval preprocessing to these models as the reference impl?

The augmentation pipeline is visible in our training script. The goal here is to verify that our models are correct and offer good (if not perfect) weights for them.

Preprocessing is standard 1/255 scaling using the KerasCV model API.

bhack:
> The augmentation pipeline is visible in our training script. The goal here is to verify that our models are correct and offer good (if not perfect) weights for them.

Yes no problem with this. But I think it is better to advertise the users about this cause It is normal to associate a network name with a pretrained checkpoint with the well known reference performances instead of looking at the script and compare it with the original impl/paper.

Probably when we will prepare a backbone Markdown table with the expected performance we could specify this, or we could add few metatada in the API/model itself.

> Preprocessing is standard 1/255 scaling using the KerasCV model API.

Ok so in this case is different from the reference impl.

ianstenbit:
> But I think it is better to advertise the users about this

Agreed. I am working on a README about model weights which will address this for users.

tanzhenyu:
> > The augmentation pipeline is visible in our training script. The goal here is to verify that our models are correct and offer good (if not perfect) weights for them.
> 
> Yes no problem with this. But I think it is better to advertise the users about this cause It is normal to associate a network name with a pretrained checkpoint with the well known reference performances instead of looking at the script and compare it with the original impl/paper.
> 
> Probably when we will prepare a backbone Markdown table with the expected performance we could specify this, or we could add few metatada in the API/model itself.
> 
> > Preprocessing is standard 1/255 scaling using the KerasCV model API.
> 
> Ok so in this case is different from the reference impl.

IMHO I think we should start with the "standard" preprocessing without any CutMix logic, reach the paper-claimed result, and then build on top of it. Any difference will bury modeling bugs

bhack:
> MHO I think we should start with the "standard" preprocessing 
What do you mean with "standard"? If you meant the original paper/reference impl.. we don't use the same number of random augmentation layers, we don't use the same magnitude etc...

tanzhenyu:
> > MHO I think we should start with the "standard" preprocessing
> 
> What do you mean with "standard"? If you meant the original paper/reference impl.. we don't use the same number of random augmentation layers, we don't use the same magnitude etc...

Ideally we should only use random resizing and horizontal flip with the same (227, 227) at this point

bhack:
Ah ok so it is a preliminar baseline step

PR #908, opened 2022-10-11
Training script improvement proposal

DavidLandup0:
# What does this PR do?

Sorry for not communicating it before making a pull request, so feel free to totally reject it if it doesn't fit.

- Fixed typo (Adam -> SGD)
- Fixed? wrong dtype for TPU (float16 -> bfloat16)
- Added warmup with cosine decay as an optional scheduler + flag for it

Thoughts?
@ianstenbit @tanzhenyu @bhack


examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,17 +202,72 @@ def augment(img, label):
         **eval(FLAGS.model_kwargs),
     )
 
-
 """
-Next, we pick an optimizer. Here we use Adam with a constant learning rate.
-Note that learning rate will decrease over time due to the ReduceLROnPlateau callback.
+Optional LR schedule with cosine decay instead of ReduceLROnPlateau
 """
+def lr_warmup_cosine_decay(global_step,
```
bhack, line 208: Is this different from https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/schedules/CosineDecayRestarts ?

DavidLandup0: Yes - it doesn't restart. It goes globally from 0 to the initial rate, and then slowly lowers it from there (after an optional hold)

DavidLandup0: scheduler

DavidLandup0: @ianstenbit mentioned swings in valid accuracy, which were resolved with a `clipnorm` - though, a warmup LR could also help IMO. I generally prefer using `ReduceLROnPlateau`, since most networks I trained seemed to do better with it than with a scheduler like this _but_ it might be worth testing

tanzhenyu: I'm bringing warm up lr to Keras Core, so hopefully this will be temporary. Can you add a TODO here?

examples/training/classification/imagenet/basic_training.py
```diff
@@ -208,23 +283,22 @@ def augment(img, label):
...
 As a last piece of configuration, we configure callbacks for the method.
 We use EarlyStopping, BackupAndRestore, and a model checkpointing callback.
 """
 
 
 callbacks = [
-    callbacks.ReduceLROnPlateau(
-        monitor="val_loss", factor=0.1, patience=10, min_delta=0.001, min_lr=0.0001
-    ),
     callbacks.EarlyStopping(patience=20),
```
innat, line 293: These days I hardly see people use early-stop. Like, don't give up yet. 🤕

examples/training/classification/imagenet/basic_training.py
innat: fyi, a reference. of `WarmupLearningRateSchedule`, [script [line 34]](https://www.kaggle.com/code/ipythonx/tf-hybrid-efficientnet-swin-transformer-gradcam?scriptVersionId=107559761&cellId=54). It would be great to have it in core keras.

examples/training/classification/imagenet/basic_training.py
```diff
@@ -110,6 +128,8 @@
 try:
     tpu = tf.distribute.cluster_resolver.TPUClusterResolver.connect()
     strategy = tf.distribute.TPUStrategy(tpu)
+    if FLAGS.use_mixed_precision:
+        keras.mixed_precision.set_global_policy("mixed_bfloat16")
```
sebastian-sz, line 132: In the case of mixed precision training, should the model also be modified to produce float32 output?

From the [docs](https://www.tensorflow.org/guide/mixed_precision#summary):
> (...) regardless of what your model ends in, make sure the output is float32.

Or is this a no big deal?

examples/training/classification/imagenet/basic_training.py
DavidLandup0: @tanzhenyu Glad to hear you're bringing it to core Keras - it really felt like it was needed. I'll add a note that this is a temporary scheduler that'll be replaced later once the core Keras update goes out.

@innat Looks good! Do you want me to replace mine with this due to other decay schedules being supported as well?

examples/training/classification/imagenet/basic_training.py
DavidLandup0: Mixed precision doesn't turn all operations into float16. It keeps float32 for operations that require it for numerical stability, like `softmax` does. So I don't think it would actually make a difference here?

examples/training/classification/imagenet/basic_training.py
DavidLandup0: Added the TODO

ianstenbit (changes requested):
Thank you for the PR!

Just one nit for you. 

Also, please add yourself to the `script_authors` list in `training_history.json` for this script. (In the future we will include attribution for all our pretrained weights and we want to credit script contributors in that process)

examples/training/classification/imagenet/basic_training.py
```diff
@@ -85,6 +85,24 @@
     "Keyword argument dictionary to pass to the constructor of the model being trained",
 )
 
+flags.DEFINE_string(
+    "use_warmup_schedule",
```
ianstenbit, line 89: What do you think about naming this `learning_rate_schedule` and making it a string that supports "ReduceOnPlateau" or "CosineDecayWithWarmup"?

examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,17 +202,73 @@ def augment(img, label):
...
+                                    start_lr=self.start_lr,
+                                    target_lr=self.target_lr,
+                                    hold=self.hold)
+
+        return tf.where(
+            step > self.total_steps, 0.0, lr, name="learning_rate"
+        )
+
+# If batched
+total_steps = len(train_ds) * EPOCHS
```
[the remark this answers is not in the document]
DavidLandup0: Sure! Doing that in a second

examples/training/classification/imagenet/basic_training.py
DavidLandup0: Good catch. Much better than having conditionals like I originally did. Updating!

DavidLandup0: Done!

examples/training/classification/imagenet/basic_training.py
bhack: > Yes - it doesn't restart. It goes globally from 0 to the initial rate, and then slowly lowers it from there (after an optional hold)

I meant if there was some specific improvement in your experiment against that.

examples/training/classification/imagenet/basic_training.py
bhack: > Yes - it doesn't restart. It goes globally from 0 to the initial rate, and then slowly lowers it from there (after an optional hold)

I meant if there was some specific improvement in your experiment against that.

examples/training/classification/imagenet/basic_training.py
```diff
@@ -85,6 +85,30 @@
     "Keyword argument dictionary to pass to the constructor of the model being trained",
 )
 
+flags.DEFINE_string(
+    "learning_rate_schedule",
+    "ReduceLROnPlateau",
```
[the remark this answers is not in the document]
DavidLandup0: Sure, updating! I went with the class name out of inertia haha
You mean, create `REDUCE_ON_PLATEAU = "ReduceOnPlateau"` (and one for cosine) to avoid writing the string out or?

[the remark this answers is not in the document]
ianstenbit: yeah that's what I was thinking. Helps make sure we avoid typos :)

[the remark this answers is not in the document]
DavidLandup0: Done :)

examples/training/classification/imagenet/basic_training.py
DavidLandup0: Oh - I haven't benchmarked them against each other yet. I didn't have much luck with cyclical LR schedules before, and by extension, cosine with repeats. I think that this implementation is almost verbatim the same as the official one.
I really like @innat's implementation since it lets you switch between decay schedules instead of having only a fixed one. Thinking about maybe porting it into the script but adding the optional `hold` as well, since IMHO, it usually makes sense to hold at the initial LR at least for a bit

DavidLandup0:
Another note - I don't know what the device utilization is on the GC environment this is running on. If there's space for more compute, we might want to try increasing the `steps_per_execution` as well. I've seen ~10% speedups on some of my own models with it, which amounts to a lot when it comes to pre-training a bunch of models on ImageNet

examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,49 +206,98 @@ def augment(img, label):
...
+TODO: Replace with Core Keras LRWarmup when it's released. This is a temporary solution.
 """
+def lr_warmup_cosine_decay(global_step,
+                           warmup_steps,
+                           hold=0,
+                           total_steps=0,
+                           start_lr=0.0,
+                           target_lr=1e-3):
+    # Cosine decay
+    # There is no tf.pi :(
```
tanzhenyu, line 220: you can use np.pi?

DavidLandup0: Yup, originally did, but then I saw no non-TF imports and thought that it might be discouraged 😬
Thinking back, pretty naive on my end. Pushing an update for this

DavidLandup0: Done

bhack: > Yup, originally did, but then I saw no non-TF imports and thought that it might be discouraged grimacing
You can also have
```python
import math as m
pi = tf.constant(m.pi)
```

examples/training/classification/imagenet/basic_training.py
sebastian-sz: Got it, thanks :)

DavidLandup0:
Should I run `format.sh` on this script as well?

ianstenbit:
> Should I run `format.sh` on this script as well?

Yes, the formatter should cover these scripts.

ianstenbit:
I am doing a test run of this script against a KerasCV EfficientNet

DavidLandup0:
Formatting in a second, thanks!
Excited to hear how EffNet does with this script! Can you post the results here?

ianstenbit:
> Formatting in a second, thanks! Excited to hear how EffNet does with this script! Can you post the results here?

Will do!

examples/training/classification/imagenet/basic_training.py
```diff
@@ -21,6 +21,7 @@
 
 import sys
 
+import numpy as np
```
bhack, line 24: NIT:
without side effects we could remove numpy with
```python
import math as m
pi = tf.constant(m.pi)
```

DavidLandup0: Sure! I can update it :)
Are there any benefits of `m.pi` vs `np.pi`? Lower overhead perhaps?

bhack: I don't think we have a specific overhead.. it was just nit to not import the library just for a single call.

examples/training/classification/imagenet/basic_training.py
```diff
@@ -85,21 +85,41 @@
     "Keyword argument dictionary to pass to the constructor of the model being trained",
 )
 
+flags.DEFINE_string(
+    "learning_rate_schedule",
+    "ReduceOnPlateau",
+    "String denoting the type of learning rate schedule to be used",
+)
+
+flags.DEFINE_string(
```
ianstenbit, line 94: For this and the percentage below (and epochs), should be DEFINE_integer

examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,49 +202,114 @@ def augment(img, label):
...
+            start_lr=self.start_lr,
+            target_lr=self.target_lr,
+            hold=self.hold,
+        )
+
+        return tf.where(step > self.total_steps, 0.0, lr, name="learning_rate")
+
+
+# If batched
+total_steps = len(train_ds) * FLAGS.epochs
```
ianstenbit, line 262: The DS isn't loaded at this point so getting its `len` errors.

I am testing the script by using 
`total_steps = (1281167 // BATCH_SIZE) * FLAGS.epochs`

examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,49 +202,114 @@ def augment(img, label):
...
     training_metrics = [metrics.CategoricalAccuracy()]
 
-
 """
 As a last piece of configuration, we configure callbacks for the method.
 We use EarlyStopping, BackupAndRestore, and a model checkpointing callback.
 """
-
-
 callbacks = [
```
ianstenbit, line 299: Also it looks like this is overwriting the name of the callbacks import, which is causing a failure  when adding the ReduceLROnPlateau callback. So let's rename this.

examples/training/classification/imagenet/basic_training.py
DavidLandup0: How come it's not loaded there yet? Mapping and prefetching is way before that line

examples/training/classification/imagenet/basic_training.py
DavidLandup0: Done

examples/training/classification/imagenet/basic_training.py
DavidLandup0: Done

examples/training/classification/imagenet/basic_training.py
DavidLandup0: Done

examples/training/classification/imagenet/basic_training.py
ianstenbit: IIRC the tf.data pipeline is lazy loaded, so it's not loaded until it gets used during training

examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,49 +202,114 @@ def augment(img, label):
...
 
 
-optimizer = optimizers.SGD(
-    learning_rate=INITIAL_LEARNING_RATE, momentum=0.9, global_clipnorm=10
+def lr_warmup_cosine_decay(
+    global_step, warmup_steps, hold=0, total_steps=0, start_lr=0.0, target_lr=1e-3
+):
+    # Cosine decay
+    # There is no tf.pi :(
+    learning_rate = (
```
tanzhenyu, line 216: Can this part be done by directly calling tf.keras.optimizers.schedules.CosineDecay?
`https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/schedules/CosineDecay`

examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,49 +202,114 @@ def augment(img, label):
...
+        learning_rate = tf.where(
+            global_step > warmup_steps + hold, learning_rate, target_lr
+        )
+
+    learning_rate = tf.where(global_step < warmup_steps, warmup_lr, learning_rate)
+    return learning_rate
+
+
+class WarmUpCosineDecay(keras.optimizers.schedules.LearningRateSchedule):
+    def __init__(self, start_lr, target_lr, warmup_steps, total_steps, hold):
```
tanzhenyu, line 240: Needs a docstring section to :
1) describe what this lr schedule does
2) preferrably have a math formula
3) describes what each arguments means, its type, and default value if there's any.

examples/training/classification/imagenet/basic_training.py
tanzhenyu: It's not loaded, but it might give us the cardinality

examples/training/classification/imagenet/basic_training.py
```diff
@@ -85,21 +85,41 @@
...
+
+flags.DEFINE_integer(
+    "warmup_steps_percentage",
+    10,
+    "For how many steps expressed in percentage of total steps should the schedule warm up if we're using the warmup schedule",
+)
+
+flags.DEFINE_integer(
+    "warmup_hold_steps_percentage",
+    10,
```
tanzhenyu, line 102: is there any particular reason that this is the same as default warmup_steps_percentage?

examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,49 +202,114 @@ def augment(img, label):
...
 """
 
 
-optimizer = optimizers.SGD(
-    learning_rate=INITIAL_LEARNING_RATE, momentum=0.9, global_clipnorm=10
+def lr_warmup_cosine_decay(
+    global_step, warmup_steps, hold=0, total_steps=0, start_lr=0.0, target_lr=1e-3
+):
+    # Cosine decay
+    # There is no tf.pi :(
```
tanzhenyu, line 215: You can remove this comment :-)

examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,49 +202,114 @@ def augment(img, label):
...
 
 """
-Next, we pick a loss function. We use CategoricalCrossentropy with label smoothing.
+Next, we pick an optimizer. Here we use SGD.
+Note that learning rate will decrease over time due to the ReduceLROnPlateau callback or with the LRWarmup scheduler.
 """
 
-
+if FLAGS.learning_rate_schedule == COSINE_DECAY_WITH_WARMUP:
+    optimizer = optimizers.SGD(learning_rate=schedule, momentum=0.9, global_clipnorm=10)
```
tanzhenyu, line 278: do you still need the gradient clipping given warm up lr exists?

tanzhenyu (commented):
Thanks for the PR! Some minor comments.

examples/training/classification/imagenet/basic_training.py
DavidLandup0: Good call. I'll update it to do the direct call. Might very well update it to reflect Innat's implementation which has switchable decay schedules

examples/training/classification/imagenet/basic_training.py
DavidLandup0: Other than it worked for me empirically before - no. I've set it as the default to fall back to, with the intention of the user manually setting it to any number between `0..total_steps`

examples/training/classification/imagenet/basic_training.py
DavidLandup0: There _shouldn't_ be a need, but I've had SGD perform more unstably than Adam in personal Imagenette tests, so I've left it as Ian had better results with it.

DavidLandup0:
@ianstenbit Did the warmup help? :)

ianstenbit:
> @ianstenbit Did the warmup help? :)

Sorry for the delay -- since I did the first test run (when I sent my last set of comments), I've been focusing on some other training runs. I am planning to kick off a training run of an EfficientNet using this script some time this week.

ianstenbit:
Starting a training run with this change now. In the meantime, could you merge from master to get #935 included? @DavidLandup0

examples/training/classification/imagenet/basic_training.py
ianstenbit: This is still not working correctly, and I am testing it using the manual length from above.

I think we can specify NUM_RECORDS and calculate it manually, or we can figure out how to correctly read this from `train_ds`

ianstenbit (changes requested):
I'm now running a test run of the updated script using CosineDecay strategy to train an EfficientNetV2B3. If that looks reasonable, I'll approve and merge (pending my other comments too).

Thank you for this PR! My apologies for the slowness in getting it pushed through

DavidLandup0:
> Starting a training run with this change now. In the meantime, could you merge from master to get #935 included? @DavidLandup0

Will do! I'm getting to my PC in a couple of hours, so I'll get that and update it according to your comments then. Thanks! Excited to see whether the schedule helps :)
No problem in taking the time to test it out!

examples/training/classification/imagenet/basic_training.py
```diff
@@ -182,49 +202,114 @@ def augment(img, label):
...
+
+# If batched
+total_steps = len(train_ds) * FLAGS.epochs
+warmup_steps = int(FLAGS.warmup_steps_percentage * total_steps)
+schedule = WarmUpCosineDecay(
+    start_lr=0.0,
+    target_lr=INITIAL_LEARNING_RATE,
+    warmup_steps=warmup_steps,
+    total_steps=total_steps,
+    hold=FLAGS.warmup_hold_steps_percentage,
```
ianstenbit, line 269: I think this needs to be `hold=hold_steps`, and `hold_steps` needs to be computed above the same way as `warmup_steps`

examples/training/classification/imagenet/basic_training.py
```diff
@@ -85,21 +85,41 @@
...
+    "For how many steps expressed in percentage of total steps should the schedule warm up if we're using the warmup schedule",
+)
+
+flags.DEFINE_integer(
+    "warmup_hold_steps_percentage",
+    10,
+    "For how many steps expressed in percentage of total steps should the schedule hold the initial learning rate if we're using the warmup schedule",
+)
+
+# An upper bound for number of epochs (this script uses EarlyStopping).
```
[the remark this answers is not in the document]
DavidLandup0: Done

examples/training/classification/imagenet/basic_training.py
DavidLandup0: What was the error's name? I can't reproduce it with other datasets (maybe it's only applicable to TFRecords?)
Since it might actually be impossible to get it working, a `NUM_RECORDS` is probably the best option

examples/training/classification/imagenet/basic_training.py
```diff
@@ -197,49 +217,115 @@ def augment(img, label):
...
+    # Cosine decay
+    # There is no tf.pi :(
+    learning_rate = (
+        0.5
+        * target_lr
+        * (
+            1
+            + tf.cos(
+                tf.constant(np.pi)
+                * (global_step - warmup_steps - hold)
```
ianstenbit, line 238: This needs to be cased to a float 32:
`tf.cast(global_step - warmup_steps - hold, tf.float32)`

examples/training/classification/imagenet/basic_training.py
```diff
@@ -197,49 +217,115 @@ def augment(img, label):
...
+                / float(total_steps - warmup_steps - hold)
+            )
+        )
+    )
+    warmup_lr = target_lr * (global_step / warmup_steps)
+
```
ianstenbit, line 244: I'm also casting warmup and target lr to float32 here in my run, because otherwise I'm getting errors about incompatible types:

```
warmup_lr = tf.cast(warmup_lr, tf.float32)
target_lr = tf.cast(target_lr, tf.float32)
```

examples/training/classification/imagenet/basic_training.py
ianstenbit: I think it is TFRecords related -- let's just add `NUM_RECORDS`

ianstenbit (approved):
I had to make a few minor changes to the LR schedule to make the script run (it failed after I started it up last night due to some mismatched FP types)

Thank you for iterating! I think it's ready now. Just gotta see how the EfficientNet trains!

DavidLandup0:
Sounds great, thanks! I want to add some docstrings. Do I add the casts and push again or do you want to add the casts in and I push the docstrings?

examples/training/classification/imagenet/basic_training.py
```diff
@@ -197,49 +217,115 @@ def augment(img, label):
...
+            target_lr=self.target_lr,
+            hold=self.hold,
+        )
+
+        return tf.where(step > self.total_steps, 0.0, lr, name="learning_rate")
+
+
+# If batched
+total_steps = len(train_ds) * FLAGS.epochs
+warmup_steps = int(FLAGS.warmup_steps_percentage * total_steps)
```
ianstenbit, line 278: Both of these need to be multiplied by 1/100 (10% == 0.1)
Re-started the training run :-)

ianstenbit:
> Sounds great, thanks! I want to add some docstrings. Do I add the casts and push again or do you want to add the casts in and I push the docstrings?

Sorry I missed this before. If I push this it'll add me to the authors on this PR, so I think you ought to just add the casts and push :)

examples/training/classification/imagenet/basic_training.py
DavidLandup0: 🤦‍♂️ I usually write them as `0..1` floats instead of integers, and it seemed like most other do integers instead, so I used them in this proposal, totally mentally forgetting that in the call below. My bad!

ianstenbit: No worries at all!

DavidLandup0:
- Changed warmup percentages to be floats between 0 and 1 instead of integers to avoid further multiplication later
- Added docstrings
- Replaced NumPy with math because of a single constant call
- Added `NUM_IMAGES` used later instead of the dataset's length
- Added `tf.cast` calls
- Updated default target_lr to 1e-2 (we're using SGD)
- Removed gradient clipnorm from optimizer when using scheduler (shouldn't be a need for it)


It's odd that the casts were required. In my own runs, dtypes weren't an issue. Might be a TPU thing?

DavidLandup0:
@ianstenbit How's the training going so far? :D

ianstenbit:
> @ianstenbit How's the training going so far? :D

Good! Running with default settings and 300 epochs. After 44 epochs we're at 66% top-1 accuracy.

DavidLandup0:
How did the run end up doing? Tried different parameters?

ianstenbit:
> How did the run end up doing? Tried different parameters?

After 250/300 epochs, EfficientNetV2B3 is at ~77% top-1 accuracy. This isn't good enough to offer the weights (unless it gets at least 1% more in the next 50 epochs, which is possible), but it's close enough that I think this LR schedule code should be merged. We can tweak runs and try different parameters for future runs.

ianstenbit (approved):
Thank you again for this contribution!

ianstenbit:
Looks like this needs to be formatted. Can you please run the linter using `sh shell/format.sh`

DavidLandup0:
Thanks for the update! I'll try out a couple of variations on Imagenette to see if there's some parameter that obviously works better, because it might translate into a smaller increase on ImageNet. Keeping you updated with a table of params/results :)

The target lr sounds good. Thanks!

Hmm, I've ran the format script before pushing and it's currently also showing no files changed when run 🤔

DavidLandup0:
This is for the last commit

image

DavidLandup0:
Oh okay, the action shows what the issue was. There was a trailing whitespace in line 261 in the comment before the newline which the linter doesn't see apparently

DavidLandup0:
Thanks for the merge!
I'll keep you updated on some of the experiments with the script on Imagenette :)

DavidLandup0:
Okay, so I did some test runs with this script, substituted ImageNet with Imagenette, and here's a finding for LR warmup. If we only hold for ~10%, it starts decaying too fast and prevents the later epochs to be utilized well enough. I've upped the hold period to 35% in my test script, and the accuracy jumped from ~80% to ~85%. Now, that's Imagenette, with fewer training epochs and a faster reduction on plateau _but_ here's the guiding principle:

In the tensorboard runs you've had so far, the first reduction happens at ~80% of the training in. If we try to match the hold period to get near that point (or at the start of the plateau, because reduction happens only after a plateau), we can slowly reduce the training instead of step-reduce it at that point. In the case of Imagenette, the first plateau/reduction hit at ~66% of the training, and a 35% holding period boosted the accuracy significantly.

Maybe we could get the Imagenet accuracy up by some amount by having the `warmup_hold_percentage` increased to anywhere between `0.3` and `0.4`? @ianstenbit

DavidLandup0:
We could customize this to "CosineDecayOnPlateau()", to make it automated instead of trying to aim for the approximate epoch at which reduction occurs

DavidLandup0:
ReduceLROnPlateau:

image

Warmup + Cosine with high holding period:

image

DavidLandup0:
Re-running the experiments with longer training for ReduceLROnPlateau, with a longer plateau period and longer early stopping

DavidLandup0:
Also, we seem to be missing a `ModelCheckpoint` with `save_best`, _just in case_ there's a better epoch before the last one

ianstenbit:
> Also, we seem to be missing a `ModelCheckpoint` with `save_best`, _just in case_ there's a better epoch before the last one

This is a big issue. I think this means many of our weights don't get the exact scores listed in the training history, because the training history shows the best validation accuracy, not the final weights. I'll send a PR to fix this.

In the meantime, I'm interested in this cutmix / mixup policy. If you're interested in sending a PR for that, I can take a look. (In the same PR maybe we can update the default hold percentage to 0.45?)

DavidLandup0:
Sending a PR in an hour or so. Want me to include the checkpoint too?

DavidLandup0:
Oh, you already included it, nvm

PR #910, opened 2022-10-12
Begin adding CLIP to KerasCV

LukeWood:
Ported from https://github.com/RobertBiehl/CLIP-tf2

Remaining tasks:

- [ ] clean up code
- [ ] port & host weights
- [ ] support mixed precision computation
- [ ] determine if we can host a generic transformer in KerasCV to fit these needs or not
- [ ] unit testing
- [ ] contribute an example showing how to use KerasCV CLIP

---

Not ready for review, just a first iteration. (draft)


bhack:
I think it could be a good practice to notify the original author on porting/refactoring PRs
/cc @RobertBiehl

bhack:
>  that is why this is marked a draft.
Ok probably you edited  the draft comment at the same time of my comment. I think we could open these as Github draft PR and then live switch it.

tanzhenyu:
> > > that is why this is marked a draft.
> > 
> > 
> > Ok probably you edited the draft comment at the same time of my comment. I think we could open these as Github draft PR and then live switch it.
> 
> Yes, edited the description for additional clarity. Originally this message was meant to indicate this, but it was not clear enough:
> 
> > “ Not ready for review, just a first iteration”
> 
> Apologies for that hopefully the updated description is clearer.
> 
> I'm personally not convinced yet that CLIP belongs in KerasCV at all, mainly just wanted to begin cleaning up the code a bit for my own purposes.

I'm convinced that CLIP should be in KerasCV, but not as public API, just internal implementation to power our other models, from both CLIP-text and CLIP-image. If we can get input from original authors, that'd be really great.

bhack:
> I'm convinced that CLIP should be in KerasCV, but not as public API, just internal implementation to power our other models, from both CLIP-text and CLIP-image. If we can get input from original authors, that'd be really great.

Are we are talking only about the pre-trained model?

If not we need to consider the huge efforts already done with open_clip...
https://github.com/mlfoundations/open_clip

tanzhenyu:
> > I'm convinced that CLIP should be in KerasCV, but not as public API, just internal implementation to power our other models, from both CLIP-text and CLIP-image. If we can get input from original authors, that'd be really great.
> 
> Are we are talking only about the pre-trained model?
> 
> If not we need to consider the huge efforts already done with open_clip... https://github.com/mlfoundations/open_clip

Yes, only models with pre-trained weights will be provided. This will help us with more downstream tasks, and we don't want to provide it as public API.

bhack:
> Yes, only models with pre-trained weights will be provided. This will help us with more downstream tasks, and we don't want to provide it as public API.

Ok but please double check as there is an emerging literature of fine-tuning clip/openclip on other domains.

https://huggingface.co/blog/fine-tune-clip-rsicd
https://github.com/jina-ai/finetuner
etc...

bhack:
https://github.com/mlfoundations/open_clip#fine-tuning-on-classification-tasks

Issue #912, opened 2022-10-12
Add MaxViT model


innat:
**Short Description**

![FeuQq4CUoAkGkMe](https://user-images.githubusercontent.com/17668390/195369211-8c2979d7-fb44-4f92-bf16-15b8f1a2d09a.jpg)

**Multi-Axis Vision Transformer**: [MaxViT](https://arxiv.org/abs/2204.01697) is a family of **hybrid** (CNN + ViT) image classification models, that achieves better performances across the board for both parameter and FLOPs efficiency than both SoTA ConvNets and Transformers.

**Papers**

https://arxiv.org/abs/2204.01697

**Existing Implementations**

Official Implementation:
 Goolge, **TensorFlow 2 (Keras)**. https://github.com/google-research/maxvit

cc. @Yinxiaoli @vztu

bhack:
Quite related to https://github.com/keras-team/keras-cv/issues/911

ayulockin:
Hey all I can work on this. :)

DavidLandup0:
I'd gladly also port this from the official repo to here. :)
If someone could assign me to it, I'd get to it as soon as the Dice and Jaccard coefficients are done.

tanzhenyu:
> Hey all I can work on this. :)

@ayulockin @innat Ideally we would like to have SwinTransformer first:
https://github.com/keras-team/keras-cv/issues/671

innat:
@tanzhenyu 
I think vanila [ViT](https://pytorch.org/vision/main/_modules/torchvision/models/vision_transformer.html) should be first, it's like VGG for transformer 😄 

In the mean time, I think it's also ok to start working on the basic component like window partition, grid attention, trail-dense etc. cc @ayulockin @DavidLandup0

tanzhenyu:
> @tanzhenyu I think vanila [ViT](https://pytorch.org/vision/main/_modules/torchvision/models/vision_transformer.html) should be first, it's like VGG for transformer 😄
> 
> In the mean time, I think it's also ok to start working on the basic component like window partition, grid attention, trail-dense etc. cc @ayulockin @DavidLandup0

https://github.com/keras-team/keras-cv/issues/668

DavidLandup0:
[...]

PR #916, opened 2022-10-13
Rough draft for Transformer-related components and ViT

DavidLandup0:
# What does this PR do?

KerasCV currently lacks transformer-based models, and for them, we'll want to have modular components to build models rapidly. These include, and are not limited to:

- MLP heads
- Transformer Blocks
- Window Partition
- Patching
- Grid Attention
- etc...

I've started porting over some of these components from @sayakpaul's: https://github.com/sayakpaul/swin-transformers-tf
And incorporated some code from the Keras.io blog into a *rough* draft of what the modularization might look like.

Please note that the code currently doesn't work nor is formatted and lacks documentation - this PR is more to check the direction of where we want to take some of these components, before taking them further. If it's okay, we can make them work and then make them pretty ;)

With what's done, we should be able to build a simple ViT with as little as:

```
x = inputs
 if include_rescaling:
     x = layers.Rescaling(1 / 255.0)(x)

# Patching and positional encoding
 patches = Patching(patch_size)(x)
 encoded_patches = PatchEncoder(num_patches, project_dim)(patches)

    # Create multiple layers of the Transformer block.
    for _ in range(transformer_layer_num):
        x = TransformerEncoder(project_dim=project_dim,
                               num_heads=num_heads,
                               dropout=dropout,
                               activation=activation)(encoded_patches)

 representation = layers.LayerNormalization(epsilon=1e-6)(x)
 representation = layers.Flatten()(representation)
 representation = layers.Dropout(0.5)(representation)
 features =  mlp_head(representation, hidden_units=head_units, dropout_rate=0.5)
 logits = layers.Dense(classes)(features)

 model = keras.Model(inputs=inputs, outputs=logits)
 return model
```

For this, I've added a few layers:

- `keras_cv.layers.Patching()` (to patch images)
- `keras_cv.layers.PatchEncoder()` (to positionally embed patches, might be worth thinking of a better name)
- `keras_cv.layers.TransformerEncoder()` (generic transformer block)
- `keras_cv.layers.PatchMerging()` - from Sayak, not ported fully yet.
- `keras_cv.layers.StochasticDepth()` - from Sayak, not ported fully yet.
- `keras_cv.layers.WindowAttention()` - from Sayak, not ported fully yet.

And a couple of building blocks such as:

- `keras_cv.transformers.mlp_head`

There's a lot more work to be done to make this both functional and conform to the repo's styles, but again - this PR is mainly to get the dialogue going on how we want to structure this. Some outstanding questions are:

- Transformer Encoder blocks: To avoid similar-but-different implementations with KerasNLP and core Keras, we should probably aim to standardize transformer encoders and have it in core Keras, and import into KerasCV. It might not be feasible to have KerasNLP implement them since they've already got their own transformer blocks, though. Currently, the one that's implemented is a pretty regular transformer skeleton. We can additionally create a generic transformer block and then override it by other components like `SwinTransformerBlock` instead.
- Module directory: Where should we keep transformer blocks? I've made a tentative `transformers` directory at the same level as `object_detection`.
- What should the components be: For example, do we turn the `representation -> features -> logits` step from above into a `ViTHead` or keep them separate? This is possibly the largest question here. There will be many internal modules and some external ones for end-users to create their own models. If nothing else, we might want to categorize components by that criteria.

@bhack mentioned possibly taking a look at how HuggingFace modularized their blocks: https://github.com/huggingface/transformers/blob/main/src/transformers/models/vit/modeling_tf_vit.py

I'm definitely for this. Let's try to brainstorm the general outline and we can create separate issues for separate components. This is meant to (possibly) be a starter :)

Since this is a high-level discussion on the direction, it might be best if everyone joins in, so I'm tagging several people.

/cc @LukeWood @tanzhenyu @qlzh727 @innat @bhack @ianstenbit @sayakpaul @fchollet 

With some of the foundation work out of the way, we should be able to implement and ship a new series of models, including MaxViT, CoAtNet, etc.

## Before submitting
      Pull Request section?
      to it if that's the case. (https://github.com/keras-team/keras-cv/issues/912)


keras_cv/layers/patch_encoder.py
```diff
@@ -0,0 +1,21 @@
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+"""
+Linear projection of patches made with `Patching` layer
```
LukeWood, line 6: lets put docstrings inside of the class

keras_cv/layers/patch_encoder.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+#@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class PatchEncoder(layers.Layer):
+    def __init__(self, num_patches, projection_dim):
+        super(PatchEncoder, self).__init__()
```
LukeWood, line 11: in python 3 you can rewrite `super(PatchEncoder, self).__init__()` to just `super().__init__()`

keras_cv/layers/patch_encoder.py
```diff
@@ -0,0 +1,21 @@
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+"""
+Linear projection of patches made with `Patching` layer
+"""
+#@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class PatchEncoder(layers.Layer):
+    def __init__(self, num_patches, projection_dim):
```
LukeWood, line 10: for Keras layers its always good to take a **kwargs argument and propogate it to super().__init__()

keras_cv/layers/patch_merging.py
```diff
@@ -0,0 +1,66 @@
+"""
+Code copied and modified from
```
LukeWood, line 2: you can add these in a references section in the docstring:

```
References:
    - [pytorch-image-models](link)
    - [Swin Transformers TF (Sayak Paul)](link)
```

keras_cv/layers/patch_merging.py
```diff
@@ -0,0 +1,66 @@
...
+        input_resolution (tuple[int]): Resolution of input feature.
+        dim (int): Number of input channels.
+    """
+
+    def __init__(
+        self,
+        input_resolution,
+        dim,
+        out_dim=None,
+        norm_layer=partial(layers.LayerNormalization, epsilon=1e-5),
```
LukeWood, line 24: instead just default this to None, and describe the default value in the docstring.

```
norm_layer=None,
...
norm_layer = norm_layer or partial(layers.LayerNormalization, epsilon=1e-5)
```

keras_cv/layers/patch_merging.py
```diff
@@ -0,0 +1,66 @@
...
+        self.dim = dim
+        self.out_dim = out_dim or 2 * dim
+        self.norm = norm_layer()
+        self.reduction = layers.Dense(self.out_dim, use_bias=False)
+
+    def call(self, x):
+        """
+        x: B, H*W, C
+        """
+        H, W = self.input_resolution
```
LukeWood, line 39: try to give these fully expanded names:

```
height, width = self.input_resolution
```

sayakpaul:
Thank you for your hardwork. 

I like your approach. Really nice direction. 

I would really insist on working on separate PRs for each individual component and test them in isolation after we settle on a particular direction. This reduces the cognitive burden of reviewing a large PR like this and also reduces the chances of reviewing errors. 

Some quick points: 

* If you used any code snippets from my Swin port, I would just humbly request you to add me as a co-author of the commits to respect the contributions :) 
* We're still missing one major component: Positional embeddings (learned: both single and relative position).

keras_cv/layers/patch_merging.py
```diff
@@ -0,0 +1,66 @@
...
+        B, L, C = tf.shape(x)[0], tf.shape(x)[1], tf.shape(x)[2]
```
LukeWood, line 40: replace these single letter upper case names with more descriptive names

LukeWood, line 40: refactor this to use a single `tf.shape(x)` call.  `tf.shape()` gets encoded in the graph as an op, so its more performant to assign to an intermediate `input_shape` variable.

LukeWood (changes requested):
Thanks for the PR!

Gave some initial feedback.  Perhaps we should refactor this into a few PRs so we can focus on one component at a time.

sayakpaul:
Also, instead of `representation = layers.Flatten()(representation)` ViT family of models either takes the representations from the CLS token (which also seems to be missing in the plan) or aggregates the representations from the rest of the patches. 

I know this is still quite early but I wanted to call out a few important pieces early on so that we don't miss out on them.

DavidLandup0:
> Thank you for your hardwork.
I like your approach. Really nice direction.
I would really insist on working on separate PRs for each individual component and test them in isolation after we settle on a particular direction. This reduces the cognitive burden of reviewing a large PR like this and also reduces the chances of reviewing errors.

Thank you for the feedback!
This PR is meant to be trashed and broken down into smaller PRs for each component once we settle for some general directions. Figured that if we segment the dialogue, it might make it harder to keep track of the high-level decisions.

For the co-authorship - of course! I wanted to ask how I can properly attribute those to you besides a comment in the code, since I think a lot of the implementation can be re-used when tweaked.

DavidLandup0:
> Also, instead of `representation = layers.Flatten()(representation)` ViT family of models either takes the representations from the CLS token (which also seems to be missing in the plan) or aggregates the representations from the rest of the patches.
> 
> I know this is still quite early but I wanted to call out a few important pieces early on so that we don't miss out on them.

Yes. The current implementation is based on the Keras.io blog flow, which skipped CLS and went straight for flattening. There are a few more components we'd want to add, but these were enough to start talking about them, so I didn't add the other ones if we go in the other direction :)

We should ideally keep a list of TODOs somewhere including stuff like this.

sayakpaul:
> For the co-authorship - of course! I wanted to ask how I can properly attribute those to you besides a comment in the code, since I think a lot of the implementation can be re-used when tweaked.

Thanks :) [Here](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-github) you can find more information.

keras_cv/layers/patch_encoder.py
DavidLandup0: Oh wow, I totally missed that for quite some time 🤦‍♂️

keras_cv/layers/patch_merging.py
sayakpaul: `batch_size, num_patches, hidden_dim`.

keras_cv/layers/patch_merging.py
sayakpaul: And for LayerNorm, I guess the default gold standard for Transformers is 1e-6.

keras_cv/layers/patch_encoder.py
sayakpaul: Something we both learned from @fchollet!

keras_cv/layers/patch_merging.py
DavidLandup0: Great call!

DavidLandup0: 👍

DavidLandup0:
Thanks for the comments @LukeWood! There are quite a few things to change on the micro-level, from formatting to optimizing and including docstrings, etc. I'll keep these comments in mind when I start cleaning the code. This is slapped together to form a rough draft only :)

sayakpaul:
> I'll keep these comments in mind when I start cleaning the code. This is slapped together to form a rough draft only :)

I will have to tip my hat for your efforts here. Thank you for jumping in so quickly and working on this. Really, thank you!

DavidLandup0:
Glad to be able to jump in! It seemed like an area that really needed a bit of (multi head) attention ;)

keras_cv/models/swin_transformer.py
```diff
@@ -0,0 +1,411 @@
...
+def to_ntuple(n):
+    def parse(x):
+        if isinstance(x, collections.abc.Iterable):
+            return x
+        return tuple(repeat(x, n))
+
+    return parse
+
+
+class SwinTransformer(keras.Model):
```
innat, line 223: Instead of subclassing, it's better to be built with functional.

cc. @LukeWood @sebastian-sz

sebastian-sz: I think functional is more consistent with the majority of other models implementations in keras-cv? 

I don't have an opinion on which is better here.

Although I think it would be nice to have similar options as other implementations, like: extracting features could be achieved by setting `include_top` instead of using `forward_features`.

DavidLandup0: Yup! That's why the ViT is built with a method using the functional API. This one was copy-pasted from Sayak's repo and is a placeholder for the functional port later. :)

DavidLandup0: Naturally, once the port is finished - it'll conform as much as it can to the other networks in KerasCV, with as many of the shared arguments and commonalities in the API as there can be.
One thing that's worth noting is that we might not want to expose the ViT's features, since it has proven (afaik) itself to encode features that aren't very useful for most downstream tasks, and hence plain ViT's aren't used as backbones. Swins introduced hierarchical abstraction back into the mix, which made them a good backbone for downstream tasks :)

If we expose ViTs features, we could encourage people to use it as a backbone, if they aren't informed about the nuances between models?

LukeWood: > If we expose ViTs features, we could encourage people to use it as a backbone, if they aren't informed about the nuances between models?

Yep!  This is a great point for using functional models.

DavidLandup0:
Any feedback on the overall structure?

- New layers for common operations
- `transformers` directory
- TransformerEncoder (possible duplicate of KerasNLP)

These are the main bottlenecks for further development. If those are fine, I can continue actually porting these :)

innat:
@DavidLandup0 cc. @LukeWood 
> - New layers for common operations

If the layer is more specific to model, then I think it's better to place in the same `{model}.py` file, (like stable-diffusion). 

> - transformers directory

IMO, not necessary. These are layers and can be placed with existed `layer` directory. It doesn't look necessary to create  `cnn/layers` or transformer/layers` for each cases.

> - TransformerEncoder (possible duplicate of KerasNLP)

IMO, the vision specific TE should be placed in `kcv`.   `knlp` already exposes many [transformer blocks](https://keras.io/api/keras_nlp/). If we want to move them to core `keras`, maybe we can for some cases but things may get more sophisticated to maintain in future.

LukeWood:
> 

Yeah I agree with @innat.  Don’t think we need a “transformers” subdirectory for layers.

keras_cv/models/__init__.py
```diff
@@ -55,3 +55,5 @@
 from keras_cv.models.resnet_v2 import ResNet101V2
 from keras_cv.models.resnet_v2 import ResNet152V2
 from keras_cv.models.vgg19 import VGG19
+#from keras_cv.models.swin_transformer import SwinTransformer
```
innat, line 58: @DavidLandup0 It might need to rename `SwinTransformerV1`, [V2](https://arxiv.org/pdf/2111.09883.pdf) is available.

bhack:
Just to quickly recap a bit the HF VIT modularization:
```python
class TFViTEmbeddings(tf.keras.layers.Layer):
class TFViTPatchEmbeddings(tf.keras.layers.Layer):
class TFViTSelfAttention(tf.keras.layers.Layer):
class TFViTSelfOutput(tf.keras.layers.Layer):
class TFViTAttention(tf.keras.layers.Layer):
class TFViTIntermediate(tf.keras.layers.Layer):
class TFViTOutput(tf.keras.layers.Layer):
class TFViTLayer(tf.keras.layers.Layer):
class TFViTEncoder(tf.keras.layers.Layer):
class TFViTMainLayer(tf.keras.layers.Layer):
class TFViTPreTrainedModel(TFPreTrainedModel):
class TFViTModel(TFViTPreTrainedModel):
class TFViTPooler(tf.keras.layers.Layer):
class TFViTForImageClassification(TFViTPreTrainedModel, TFSequenceClassificationLoss):
```

DavidLandup0:
Thank you for the feedback!

Just one small note:

> IMO, not necessary. These are layers and can be placed with existed layer directory. It doesn't look necessary to create cnn/layers or transformer/layers` for each cases.

All of the layers are still in the `keras_cv.layers` directory. Only an `mlp_head.py` is in `transformers` since it's a reusable component. Is that okay or do we totally ditch the directory and keep mini components like that in model files?

DavidLandup0:
Okay, I'll be breaking up the PR into multiple small ones and we can go through each one systematically, starting with adding the basic layers.

keras_cv/layers/patch_encoder.py
```diff
@@ -0,0 +1,21 @@
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+"""
+Linear projection of patches made with `Patching` layer
+"""
+#@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class PatchEncoder(layers.Layer):
```
bhack, line 9: Is this generally expected to be named "Embeddings" or "PositionEmbeddings" etc..?

PR #919, opened 2022-10-13
Allow compute_iou to take boxes of different shape.

tanzhenyu:
Allow _RPNLabelEncoder to take batched inputs.
Move the data labeling from data pipeline to model pipeline. This
 1) _might_ come with a performance cost (but I haven't noticed any regression), but 
 2) in the long run it get us better performance if we want more data augmentation in the data pipeline.
 3) make supporting `model.fit` more native and users don't have to encode labels if they build their own dataset.

### DIFF
--- a/keras_cv/bounding_box/iou.py
+++ b/keras_cv/bounding_box/iou.py
@@ -17,6 +17,47 @@ import tensorflow as tf
 from keras_cv import bounding_box
 
 
+def _compute_area(box):
+    """Computes area for bounding boxes
+
+    Args:
+      box: [N, 4] or [batch_size, N, 4] float Tensor, either batched
+        or unbatched boxes.
+    Returns:
+      a float Tensor of [N] or [batch_size, N]
+    """
+    y_min, x_min, y_max, x_max = tf.split(box, num_or_size_splits=4, axis=-1)
+    return tf.squeeze((y_max - y_min) * (x_max - x_min), axis=-1)
+
+
+def _compute_intersection(boxes1, boxes2):
+    """Computes intersection area between two sets of boxes.
+
+    Args:
+      boxes1: [N, 4] or [batch_size, N, 4] float Tensor boxes.
+      boxes2: [M, 4] or [batch_size, M, 4] float Tensor boxes.
+    Returns:
+      a [N, M] or [batch_size, N, M] float Tensor.
+    """
+    y_min1, x_min1, y_max1, x_max1 = tf.split(boxes1, num_or_size_splits=4, axis=-1)
+    y_min2, x_min2, y_max2, x_max2 = tf.split(boxes2, num_or_size_splits=4, axis=-1)
+    boxes2_rank = len(boxes2.shape)
+    perm = [1, 0] if boxes2_rank == 2 else [0, 2, 1]
+    # [N, M] or [batch_size, N, M]
+    intersect_ymax = tf.minimum(y_max1, tf.transpose(y_max2, perm))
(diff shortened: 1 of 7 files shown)


tanzhenyu:
It looks like many KPL are using dimension 5 instead of 4 and compute_iou should support it for now?

keras_cv/bounding_box/iou.py
```diff
@@ -50,16 +95,19 @@ def compute_iou(
     boxes1_rank = len(boxes1.shape)
     boxes2_rank = len(boxes2.shape)
 
-    if (
-        boxes1_rank != boxes2_rank
-        or boxes1_rank not in [2, 3]
-        or boxes2_rank not in [2, 3]
-    ):
+    if boxes1_rank not in [2, 3]:
```
LukeWood, line 98: can we update the docstring to add more information about the updated input/output shapes?

LukeWood, line 98: perhaps instead of supporting differing ranks we should follow the TensorFlow convention and broadcast only when the first axis is of dim [1].

So instead of:

[boxes, 4], [batch, boxes, 4]

you would pass:

[1, boxes, 4], [batch, boxes, 4]

This is more consistent with the TensorFlow API than automatically broadcasting for users

tanzhenyu: Here the problem is not on [1, boxes, 4] vs [batch, boxes, 4], it's actually [boxes_1, 4] & [batch, boxes_2, 4] should give user [batch, boxes_1, boxes_2, 4], if users pass in [1, boxes_1, 4] and [batch, boxes_2, 4], this will actually error out telling user that the shape doesn't match, which is expected.
If we do ask users to convert to [batch, boxes_1, 4] before calling compute_iou, then users would have to do tf.tile(tf.expand_dims(boxes1, 0), tf.stack([tf.shape(boxes_2)[0], tf.shape(boxes_1)[1:])) in order to call it, which to me is too bloated.

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -329,6 +329,47 @@ def __init__(
...
+            rcnn_box_pred = _decode_deltas_to_boxes(
+                anchors=rois,
+                boxes_delta=rcnn_box_pred,
+                anchor_format="yxyx",
+                box_format=self.bounding_box_format,
+                variance=[0.1, 0.1, 0.2, 0.2],
+            )
+        return rcnn_box_pred, rcnn_cls_pred
+
+    # def train_step(self, data):
```
ianstenbit, line 370: nit: looks like this got left behind

keras_cv/bounding_box/iou.py
tanzhenyu: Done.

tanzhenyu: I have added 4 different scenarios on the documentation part so hopefully this will give user more context

ianstenbit: I'm no expert in the bbox conventions, but personally I think this API looks good.

ianstenbit (approved):

keras_cv/bounding_box/iou.py
LukeWood: Sounds good, that’s fine with me.

LukeWood (approved):

keras_cv/models/object_detection/faster_rcnn.py
tanzhenyu: Oops..removed

PR #923, opened 2022-10-14
Add pretrained EfficientNetV2b1 weights

ianstenbit:
Scores .756 on ImageNet top-1, vs .798 paper-claimed (~95% of claimed result)

### DIFF
--- a/examples/training/classification/imagenet/training_history.json
+++ b/examples/training/classification/imagenet/training_history.json
@@ -64,6 +64,23 @@
             "validation_accuracy": "0.7527"
         }
     },
+    "efficientnetv2b1": {
+        "v0": {
+            "accelerators": 8,
+            "args": {
+                "batch_size": "64",
+                "initial_learning_rate": ".0125"
+            },
+            "contributor": "ianstenbit",
+            "epochs_trained": 288,
+            "script": {
+                "name": "basic_training.py",
+                "version": "e349ca5563b05548996f438fa03b2f34a8231ca3"
+            },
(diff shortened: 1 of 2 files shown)


tanzhenyu:
Is there any particular reason that the train accuracy (~55%) is way lower than val accuracy (~75%)?

ianstenbit:
> Is there any particular reason that the train accuracy (~55%) is way lower than val accuracy (~75%)?

This is due to the use of RandAugment+CutMix, which makes the training dataset much harder than the validation dataset

tanzhenyu:
> > Is there any particular reason that the train accuracy (~55%) is way lower than val accuracy (~75%)?
> 
> This is due to the use of RandAugment+CutMix, which makes the training dataset much harder than the validation dataset

That seems to suggest training loss will keep decreasing if we train more epochs, probably val loss as well?

tanzhenyu (approved):

tanzhenyu:
w/ or w/o training more epochs to verify it, I think this PR is good to go

DavidLandup0:
I've also seen training accuracy being significantly lower than validation accuracy when using RandAugment and CutMix/MixUp (~0.55 training, ~0.95 validation)

I like to think that this keeps the network aware that there's more room for improvement and won't stop updating weights to better fit the data, which is a common issue with high training accuracy. However, with more training, it might conceivably lower the validation accuracy by thinking it's more wrong than it really is? If they start diverging too much, maybe lowering the magnitude of random augmentation might help?

PR #925, opened 2022-10-14
Add model fit for FasterRCNN.

tanzhenyu:
Completing the story of `Model.fit` for FasterRCNN.
This is achieved by overriding `train_step` and `test_step`.
Testing it with Pascal VOC dataset, the performance / quality is similar to CTL.
Currently I'm not adding functionality to upload weights and benchmark test given we don't have a metrics system that produces the same result as pycoco yet.

Some design questions:
1) the compile method is following similar pattern of RetinaNet, i.e., compile(box_loss, cls_loss, ...). I think this is fine, but would like to have a consensus here.
2) the compile method uses `weight_decay` -- not too happy about it, but we cannot ask users to add `kernel_regularizer` for their backbones at this point. I think in TF2.10 we may be able to rely on `optimizers.SGDW` instead, but given weight decay should skip batch norm, and KerasCV should work with tf2.9 (I believe there's a huge optimizer bug fix which didn't get backported to 2.9), this seems to be the only way.

### DIFF
--- a/examples/models/object_detection/faster_rcnn/train.py
+++ b/examples/models/object_detection/faster_rcnn/train.py
@@ -18,16 +18,36 @@ Date created: 2022/09/27
 Last modified: 2022/09/27
 Description: Use KerasCV to train a RetinaNet on Pascal VOC 2007.
 """
-from datetime import datetime
+import sys
 
 import tensorflow as tf
 import tensorflow_datasets as tfds
+from absl import flags
 
 import keras_cv
 
+flags.DEFINE_string(
+    "weights_path", "~", "Directory which will be used to store weight checkpoints."
+)
+flags.DEFINE_string(
+    "tensorboard_path",
+    "~/logs",
+    "Directory which will be used to store tensorboard logs.",
+)
+FLAGS = flags.FLAGS
+FLAGS(sys.argv)
+
 # parameters from FasterRCNN [paper](https://arxiv.org/pdf/1506.01497.pdf)
 
-strategy = tf.distribute.MirroredStrategy()
+# Try to detect an available TPU. If none is present, default to MirroredStrategy
+try:
+    tpu = tf.distribute.cluster_resolver.TPUClusterResolver.connect()
+    strategy = tf.distribute.TPUStrategy(tpu)
+except ValueError:
+    # MirroredStrategy is best for a single machine with one or multiple GPUs
+    strategy = tf.distribute.MirroredStrategy()
+print("Number of accelerators: ", strategy.num_replicas_in_sync)
+
 local_batch = 4
 global_batch = local_batch * strategy.num_replicas_in_sync
 base_lr = 0.01 * global_batch / 16
@@ -241,13 +261,13 @@ eval_ds = eval_ds.prefetch(2)
 with strategy.scope():
     # The keras huber loss will sum all losses and divide by BS * N * 4, instead we want divide by BS
     # using NONE reduction as also summing creates issues
-    rpn_reg_loss_fn = tf.keras.losses.Huber(reduction=tf.keras.losses.Reduction.NONE)
+    rpn_reg_loss_fn = tf.keras.losses.Huber(reduction=tf.keras.losses.Reduction.SUM)
     rpn_cls_loss_fn = tf.keras.losses.BinaryCrossentropy(
-        from_logits=True, reduction=tf.keras.losses.Reduction.NONE
+        from_logits=True, reduction=tf.keras.losses.Reduction.SUM
     )
-    rcnn_reg_loss_fn = tf.keras.losses.Huber(reduction=tf.keras.losses.Reduction.NONE)
+    rcnn_reg_loss_fn = tf.keras.losses.Huber(reduction=tf.keras.losses.Reduction.SUM)
     rcnn_cls_loss_fn = tf.keras.losses.SparseCategoricalCrossentropy(
-        reduction=tf.keras.losses.Reduction.NONE
+        reduction=tf.keras.losses.Reduction.SUM
     )
 
     rpn_reg_metric = tf.keras.metrics.Mean()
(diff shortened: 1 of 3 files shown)


examples/models/object_detection/faster_rcnn/train.py
```diff
@@ -18,16 +18,38 @@
...
-strategy = tf.distribute.MirroredStrategy()
+# Try to detect an available TPU. If none is present, default to MirroredStrategy
+try:
+    tpu = tf.distribute.cluster_resolver.TPUClusterResolver.connect()
+    strategy = tf.distribute.TPUStrategy(tpu)
+except ValueError:
+    # MirroredStrategy is best for a single machine with one or multiple GPUs
+    strategy = tf.distribute.MirroredStrategy()
+print("Number of accelerators: ", strategy.num_replicas_in_sync)
+
```
innat, line 52: It would be better to have a `keras.utility` function to automate the device selection.

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +345,165 @@ def _call_rpn(self, images, training=None):
...
-                anchor_format="yxyx",
-                box_format=self.bounding_box_format,
-                variance=[0.1, 0.1, 0.2, 0.2],
-            )
-            res["rcnn_box_pred"] = rcnn_box_pred
-            res["rcnn_cls_pred"] = rcnn_cls_pred
 
-        return res
+    def train_step(self, data):
+        images = data["images"]
```
ianstenbit, line 478: It seems like this is a bit different from the current RetinaNet box API.

IIRC we wanted to move to dictionary inputs due to the awkwardness of having box tensors include a class and confidence?

I don't have a strong opinion other than that it should be consistent between the two models

examples/models/object_detection/faster_rcnn/train.py
```diff
@@ -241,13 +263,13 @@ def pad_fn(examples):
 with strategy.scope():
     # The keras huber loss will sum all losses and divide by BS * N * 4, instead we want divide by BS
     # using NONE reduction as also summing creates issues
-    rpn_reg_loss_fn = tf.keras.losses.Huber(reduction=tf.keras.losses.Reduction.NONE)
+    rpn_reg_loss_fn = tf.keras.losses.Huber(reduction=tf.keras.losses.Reduction.SUM)
```
ianstenbit, line 266: It seems like these loss_fns are no longer used in this script since we're using the defaults from `compile`?

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +345,165 @@ def _call_rpn(self, images, training=None):
...
+
+    def test_step(self, data):
+        images = data["images"]
+        gt_boxes = data["gt_boxes"]
+        gt_classes = data["gt_classes"]
+        self.compute_loss(images, gt_boxes, gt_classes, training=False)
+        return self.compute_metrics(images, {}, {}, sample_weight={})
+
+
+def _validate_loss(loss, loss_name):
```
ianstenbit, line 501: Should we also validate that, for losses which require a box format, the format matches the format of the rest of the model?

keras_cv/models/object_detection/faster_rcnn.py
LukeWood, line 478: instead of data being a dictionary, lets make data a tuple of (x, y) like a standard Keras model.

Let's make `y` a dictionary with keras {"boxes", "classes"}.  We can migrate RetinaNet to the same format, alongside COCO metrics, and all other components (KPLs, etc)

LukeWood (changes requested):
Minor comments, I see Ian left a review so to avoid spamming I'll keep the current one brief until more changes are pushed.

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +345,165 @@ def _call_rpn(self, images, training=None):
...
+        classification_loss=None,
+        rpn_box_loss=None,
+        rpn_classification_loss=None,
+        weight_decay=0.0001,
+        **kwargs,
+    ):
+        # TODO(tanzhenyu): Add metrics support once COCOMap issue is addressed.
+        # https://github.com/keras-team/keras-cv/issues/915
+        if "metrics" in kwargs.keys():
+            raise ValueError("currently metrics support is not supported intentionally")
```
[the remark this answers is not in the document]
tanzhenyu: metrics defined here are for both train-time and eval

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +345,165 @@ def _call_rpn(self, images, training=None):
...
-        return rcnn_box_pred, rcnn_cls_pred
+        return box_pred, cls_pred
+
+    def compile(
+        self,
+        box_loss=None,
+        classification_loss=None,
+        rpn_box_loss=None,
+        rpn_classification_loss=None,
+        weight_decay=0.0001,
```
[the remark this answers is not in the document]
tanzhenyu: Unfortunately our optimizer doesn't take that as argument. We can use SGDW in tf 2.10, but it's actually quite limited

examples/models/object_detection/faster_rcnn/train.py
tanzhenyu: Good catch! Removed.

keras_cv/models/object_detection/faster_rcnn.py
tanzhenyu: loss is independent to box format, IIUC. So users should be able to change between formats and still use the same loss. However they would have to scale differently

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +345,165 @@ def _call_rpn(self, images, training=None):
...
+        **kwargs,
+    ):
+        # TODO(tanzhenyu): Add metrics support once COCOMap issue is addressed.
+        # https://github.com/keras-team/keras-cv/issues/915
+        if "metrics" in kwargs.keys():
+            raise ValueError("currently metrics support is not supported intentionally")
+        reduction_type = tf.keras.losses.Reduction.SUM
+        if not box_loss:
+            box_loss = tf.keras.losses.Huber(reduction=reduction_type)
```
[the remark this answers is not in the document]
tanzhenyu: Sounds good. Done. We should also do this for RetinaNet then.

examples/models/object_detection/faster_rcnn/train.py
```diff
@@ -18,16 +18,38 @@
...
+try:
+    tpu = tf.distribute.cluster_resolver.TPUClusterResolver.connect()
+    strategy = tf.distribute.TPUStrategy(tpu)
+except ValueError:
+    # MirroredStrategy is best for a single machine with one or multiple GPUs
+    strategy = tf.distribute.MirroredStrategy()
+print("Number of accelerators: ", strategy.num_replicas_in_sync)
+
 local_batch = 4
 global_batch = local_batch * strategy.num_replicas_in_sync
```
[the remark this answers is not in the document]
tanzhenyu: global_batch is still used in data pipeline, and also used to scale our optimizer lr

keras_cv/models/object_detection/faster_rcnn.py
tanzhenyu: Good point. Done.

examples/models/object_detection/faster_rcnn/train.py
tanzhenyu: To me this sounds more like a `tf.distribute.auto_select_strategy` :-)

tanzhenyu: Having a trainer object that takes care of runtime args is definitely not a bad idea

keras_cv/models/object_detection/faster_rcnn.py
ianstenbit: Is IOU loss not relevant for this model? (IIRC that requires a box format?)

examples/models/object_detection/faster_rcnn/train.py
LukeWood:
```
trainer = Trainer(devices=8, accelerator="gpu")
```

isn't this just `MirroredStrategy(devices=8)`

keras_cv/models/object_detection/faster_rcnn.py
tanzhenyu: There seem to be inconsistent losses: 1) iou loss that require a format; 2) smoothl1 loss that doesn't require a format.
In the long run, we should probably by fixing it so that it's format independent. Why? Because there are some weird format (and actually mostly used compared to other formats) such as `tx` `ty` `tw` `th` which we wouldn't be able to scale up by supporting all of them, and clearly we don't want to block users if they want to convert their format to a new one before passing to loss function.

examples/models/object_detection/faster_rcnn/train.py
innat: @LukeWood you're right. Now, the following code is widely used. 

```python
try:
    tpu = tf.distribute.cluster_resolver.TPUClusterResolver.connect()
    strategy = tf.distribute.TPUStrategy(tpu)
except ValueError:
    # MirroredStrategy is best for a single machine with one or multiple GPUs
    strategy = tf.distribute.MirroredStrategy()
print("Number of accelerators: ", strategy.num_replicas_in_sync)
```

That's why I was thinking it might be good to have a simple utility function to automate it, simple and clean.

i.e. `keras.utils.set_strategy(devices=2, accelerator="gpu")`

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +346,169 @@ def _call_rpn(self, images, training=None):
...
+        # TODO(tanzhenyu): Add metrics support once COCOMap issue is addressed.
+        # https://github.com/keras-team/keras-cv/issues/915
+        if "metrics" in kwargs.keys():
+            raise ValueError("currently metrics support is not supported intentionally")
+        box_loss = _validate_and_get_loss(box_loss, "box_loss")
+        classification_loss = _validate_and_get_loss(
+            classification_loss, "classification_loss"
+        )
+        rpn_box_loss = _validate_and_get_loss(rpn_box_loss, "rpn_box_loss")
+        if rpn_classification_loss == "BinaryCrossentropy":
```
LukeWood, line 398: can we move this logic into _validate_and_get_loss?

LukeWood (changes requested):

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +346,169 @@ def _call_rpn(self, images, training=None):
...
                 variance=[0.1, 0.1, 0.2, 0.2],
             )
-        return rcnn_box_pred, rcnn_cls_pred
+        return box_pred, cls_pred
+
+    # TODO(tanzhenyu): Support compile with metrics.
+    def compile(
+        self,
+        box_loss="Huber",
+        classification_loss="SparseCategoricalCrossentropy",
```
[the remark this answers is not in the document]
tanzhenyu: This does not support lower cased losses yet

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +346,169 @@ def _call_rpn(self, images, training=None):
...
                 box_format=self.bounding_box_format,
                 variance=[0.1, 0.1, 0.2, 0.2],
             )
-        return rcnn_box_pred, rcnn_cls_pred
+        return box_pred, cls_pred
+
+    # TODO(tanzhenyu): Support compile with metrics.
+    def compile(
+        self,
+        box_loss="Huber",
```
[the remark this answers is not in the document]
tanzhenyu: This seems reasonable to me. But `loss='default'` is definitely more weird, changing it to require users to specifically pass each loss.

LukeWood (approved):

ianstenbit (approved):

examples/models/object_detection/faster_rcnn/train.py
```diff
@@ -18,16 +18,38 @@
 Last modified: 2022/09/27
 Description: Use KerasCV to train a RetinaNet on Pascal VOC 2007.
```
qlzh727, line 19: File name wise, since this is train wih pascal voc, how about change it to train_pascal_voc.py?

examples/models/object_detection/faster_rcnn/train.py
```diff
@@ -18,16 +18,38 @@
...
 
 import tensorflow as tf
 import tensorflow_datasets as tfds
+from absl import flags
 
 import keras_cv
 
+flags.DEFINE_string(
+    "weights_path",
+    "weights_{epoch:02d}.h5",
```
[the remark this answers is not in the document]
qlzh727: Same here, usually it just need to either keep the last epoch weights, or best epoch weights. Even you have the weights for every epoch, you probably still like to the last or the best epoch weights.

examples/models/object_detection/faster_rcnn/train.py
```diff
@@ -267,167 +269,17 @@ def pad_fn(examples):
 weight_decay = 0.0001
```
qlzh727, line 269: This one is not used anymore here.

examples/models/object_detection/faster_rcnn/train.py
```diff
@@ -267,167 +269,17 @@ def pad_fn(examples):
...
 step = 0
```
qlzh727, line 270: Same for this one.

examples/models/object_detection/faster_rcnn/train.py
```diff
@@ -239,22 +257,6 @@ def pad_fn(examples):
...
-        reduction=tf.keras.losses.Reduction.NONE
-    )
-
-    rpn_reg_metric = tf.keras.metrics.Mean()
-    rpn_cls_metric = tf.keras.metrics.Mean()
-    rcnn_reg_metric = tf.keras.metrics.Mean()
-    rcnn_cls_metric = tf.keras.metrics.Mean()
-
     lr_decay = tf.keras.optimizers.schedules.PiecewiseConstantDecay(
         boundaries=[12000 * 16 / global_batch, 16000 * 16 / global_batch],
```
qlzh727, line 261: i think the boundary here really depends on the size of the dataset (since the ds is not repeat and there isn't a steps per epoch in fit()). We might want to change 12000 and 16000 to something more meaningful (I assume there are 1000 steps per dataset, and 12000 refers to 12 epochs.)

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +346,176 @@ def _call_rpn(self, images, training=None):
...
+    def compile(
+        self,
+        box_loss=None,
+        classification_loss=None,
+        rpn_box_loss=None,
+        rpn_classification_loss=None,
+        weight_decay=0.0001,
+        loss=None,
+        **kwargs,
+    ):
```
qlzh727, line 389: I think we need some docstring for this method.

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +346,176 @@ def _call_rpn(self, images, training=None):
...
+        # TODO(tanzhenyu): Add metrics support once COCOMap issue is addressed.
+        # https://github.com/keras-team/keras-cv/issues/915
+        if "metrics" in kwargs.keys():
+            raise ValueError("currently metrics support is not supported intentionally")
+        if loss is not None:
+            raise ValueError(
+                "`FasterRCNN` does not accept a `loss` to `compile()`. "
+                "Instead, please pass `box_loss` and `classification_loss`. "
```
qlzh727, line 397: how about ` rpn_box_loss` and `rpn_classification_loss`?

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +346,176 @@ def _call_rpn(self, images, training=None):
...
+
+def _validate_and_get_loss(loss, loss_name):
+    if isinstance(loss, str):
+        loss = tf.keras.losses.get(loss)
+    if loss is None or not isinstance(loss, tf.keras.losses.Loss):
+        raise ValueError(
+            f"FasterRCNN only accepts `tf.keras.losses.Loss` for {loss_name}, got {loss}"
+        )
+    if loss.reduction != tf.keras.losses.Reduction.SUM:
+        logging.info(
```
qlzh727, line 517: Should this be a info or a warning?

qlzh727 (dismissed):

keras_cv/models/object_detection/faster_rcnn.py
LukeWood: Can we drop `'gt_'`?

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -346,81 +346,176 @@ def _call_rpn(self, images, training=None):
...
-            )
-            res["rcnn_box_pred"] = rcnn_box_pred
-            res["rcnn_cls_pred"] = rcnn_cls_pred
 
-        return res
+    def train_step(self, data):
+        images, y, sample_weight = tf.keras.utils.unpack_x_y_sample_weight(data)
+        if sample_weight is not None:
+            raise ValueError("`sample_weight` is currently not supported.")
+        gt_boxes = y["gt_boxes"]
```
LukeWood, line 485: lets drop gt_ in the dictionary keys.

Its already implied as they belong to `y`.

tanzhenyu: Oops missed this one before I merge.This sounds good to me. Though if we add `masks` in the future, `gt_masks` seems more

PR #927, opened 2022-10-15
Fixed efficientnet v1 correct pad.

sebastian-sz:
# What does this PR do?
The `correct_pad` function in efficientnet v1 is using the wrong `image_dim`. It assumes `channels_first`, while keras-cv operates on `channels_last`. 

Reference function: [here](https://github.com/keras-team/keras/blob/bb78a0caaea0503728784c0fa5cf547c623633b0/keras/applications/imagenet_utils.py#L441)

The weights should load anyway, but the metric values will be affected.
(I'm not sure if this is big enough for a separate issue first)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood

### DIFF
--- a/keras_cv/models/efficientnet_v1.py
+++ b/keras_cv/models/efficientnet_v1.py
@@ -187,7 +187,7 @@ def correct_pad(inputs, kernel_size):
     Returns:
       A tuple.
     """
-    img_dim = 2
+    img_dim = 1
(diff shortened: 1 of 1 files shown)


sebastian-sz:
Also tagging @ianstenbit for reference.

sebastian-sz:
Yes, happy to see this is merged before EfficientNet's are all trained.

Issue #671, opened 2022-08-06
Support Swin Transformer Model


tanzhenyu:
Related to https://github.com/keras-team/keras-cv/issues/668, this is another transformer based model which we'd like to be supported. For simplicity, focus on only classification task, having in mind that future detection tasks might rely on it.

For contributors, please follow the [contribution guide](https://github.com/keras-team/keras-cv/blob/master/.github/CONTRIBUTING.md).

And specifically, on the [API Design Guide](https://github.com/keras-team/keras-cv/blob/master/.github/API_DESIGN.md), we'd like to see a brief description of how the model would look like before implementation, preferrably conforming to the tf.keras.application style.

ariG23498:
I would like to work on this!

ariG23498:
Hey @tanzhenyu

I got around reading the contriution guideline and the API design document. I was unable to find a guide to contribute a model inside of `keras_cv`.

My main doubts:
- Do we need to port the weights of the model (here SwinT in PyTorch)? If yes (which I hope should be the case), how would I do that?
- Do we also need to validate the weight porting process by validating it wrt the task (classification here) on a benchmark dataset?
- How do we check the correctness of the model?

bhack:
> Hey @tanzhenyu
> 
> I got around reading the contriution guideline and the API design document. I was unable to find a guide to contribute a model inside of `keras_cv`.
> 
> My main doubts:
> 
> * Do we need to port the weights of the model (here SwinT in PyTorch)? If yes (which I hope should be the case), how would I do that?
> * Do we also need to validate the weight porting process by validating it wrt the task (classification here) on a benchmark dataset?
> * How do we check the correctness of the model?

It would be useful to have something like:
https://github.com/keras-team/keras/blob/master/CONTRIBUTING.md#contributing-to-keras-applications

bhack:
Also do we have a duplicate ticket with https://github.com/keras-team/keras-cv/issues/636 ?

tanzhenyu:
> Hey @tanzhenyu
> 
> I got around reading the contriution guideline and the API design document. I was unable to find a guide to contribute a model inside of `keras_cv`.
> 
> My main doubts:
> 
> * Do we need to port the weights of the model (here SwinT in PyTorch)? If yes (which I hope should be the case), how would I do that?
> * Do we also need to validate the weight porting process by validating it wrt the task (classification here) on a benchmark dataset?
> * How do we check the correctness of the model?

[...]

Issue #668, opened 2022-08-05
Support ViT model


tanzhenyu:
Transformer models are becoming more and more relevant. Let's support it by having ViT first.
citing https://arxiv.org/pdf/2010.11929.pdf

sayakpaul:
Happy to work on it.

ariG23498:
I am willing to work on this as well.

tanzhenyu:
> I am willing to work on this as well.

Thanks for providing the help! This is currently assigned, would you be open to contributing other transformer based vision models? for example, swin transformer

ariG23498:
Sure!

tanzhenyu:
> Sure!

https://github.com/keras-team/keras-cv/issues/671

Somehow I can't assign to you through user name...

ariG23498:
> Somehow I can't assign to you through user name...

I think we need to comment on the issue to be assigned?

deepwilson:
If this is not assigned I would love to take this up or help if needed.

sayakpaul:
It's assigned.

![image](https://user-images.githubusercontent.com/22957388/183631170-ec90960f-77bc-46cb-902d-e1ce95f2da61.png)

sayakpaul:
Sorry for the delay. But I still have it planned to contribute to this.

cybergeekgyan:
I want to contribute to this issue, Can anyone guide me through. How to proceed with it..

bhack:
> I want to contribute to this issue, Can anyone guide me through. How to proceed with it..

There is a PR at https://github.com/keras-team/keras-cv/pull/1014

PR #929, opened 2022-10-15
Adding ViT Components/Layers

DavidLandup0:
# What does this PR do?

As mentioned in #916, I'm breaking up the PR into multiple smaller ones, starting with basic components for building ViTs. Added layers:

- `keras_cv.layers.TransformerEncoder`
- `keras_cv.layers.PatchEmbedding`
- `keras_cv.layers.Patching`
- ~~`keras_cv.layers.ClassTokenizing`~~

**Note:** Temporarily removed `ClassTokenizing` to unblock the ViT project, as discussed with @tanzhenyu. We can look into adding it back later for models like DeiT, BeiT, etc.

As per:
image

`Patching` is used to patchify an input batch of images, and return a batch of patches acting like sequences. `PatchEmbedding` adds the class token and encodes the patches to a given projection dims + positionally. `ClassTokenizing` is used to create the class token. `TransformerEncoder` creates a useful encoding for downstream classification.

Here's a Google Colab notebook showcasing the layers: https://colab.research.google.com/drive/1BbCtaKN_mmVEd-iZlWQ8GpgL55FOXrwK?usp=sharing

As proof of concept, here's an approximate ViT that trains with these layers: https://colab.research.google.com/drive/1gzv6jVljae0dNtKhhq0qxRI8IIrejuCU?usp=sharing

With these, we can easily build a ViT with only a few lines, a bit more accurately. That's the next PR :)

A TODO for this PR is to write tests. I might need a bit of help with that. Tagging the people from the previous discussion:

/cc @LukeWood @sayakpaul @innat

## Before submitting
      Pull Request section?
      to it if that's the case. https://github.com/keras-team/keras-cv/pull/916


DavidLandup0:
Individually, they work fine. When building a test model, it builds. But when I try training:

```
 ValueError: tf.function only supports singleton tf.Variables created on the first call. Make sure the tf.Variable is only created once or created outside tf.function.
```

I'm still trying to figure out why...

bhack:
> Individually, they work fine. When building a test model, it builds. But when I try training:
> 
> ```
>  ValueError: tf.function only supports singleton tf.Variables created on the first call. Make sure the tf.Variable is only created once or created outside tf.function.
> ```
> 
> I'm still trying to figure out why...

On which line?

DavidLandup0:
> On which line?

First `LayerNormalization` in the transformer encoder. I think I figured out what was messed up. I was creating trainable layers in the `call()` method instead of initializing them once in the `__init__`, which lead to them being called multiple times?

DavidLandup0:
Okay, I think I got the issue fixed. Trying to train now with some arbitrary dimensionalities (haven't cross-checked the paper for the concrete number of intermediate/hidden dimensionalities).

bhack:
This happen when you are going to intialize tf.variables multiple time inside a tf.function (as the train loop is auto wrapped inside a tf.fuction by default)

See:
https://www.tensorflow.org/guide/function?hl=en#creating_tfvariables
 https://www.tensorflow.org/guide/function?hl=en#using_with_multiple_keras_models

DavidLandup0:
Thanks for the links! The issue was creating layers in the `call()` function in my case, instead of initializing them once in the constructor :)

image

DavidLandup0:
Trains and the loss is going down: https://colab.research.google.com/drive/1gzv6jVljae0dNtKhhq0qxRI8IIrejuCU?usp=sharing

DavidLandup0:
If these look okay - I'll go ahead and build a paper-accurate ViT and its variations for the next PR :)

keras_cv/layers/vit_patching.py
```diff
@@ -0,0 +1,69 @@
...
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class Patching(layers.Layer):
+    """
+    Layer to turn images into a sequence of patches for Vision Transformers from:
```
innat, line 22: It might be usefull to have standard layers for `image-to-patch` and  vice-versa! ex:

```python
class TransformImageToPatch
class TransformPatchToImage
```

cc. @LukeWood

keras_cv/layers/vit_patch_encoding.py
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+from keras_cv.layers.vit_class_tokenizing import ClassTokenizing
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class PatchEncoding(layers.Layer):
```
bhack, line 22: In the HF modularization we have the opportunity to optionally interpolate the position embedding to fine-tune VIT at higher resolution:
https://github.com/huggingface/transformers/blob/main/src/transformers/models/vit/modeling_tf_vit.py#L742-L744

bhack: Is `PatchEncoding` an expected name? Or "PosEmbeddings" etc..

DavidLandup0: The original paper refers to them as patch embeddings. Might be better to rename the layer `PatchEmbedding` or `PathEmbeddings`?

> We refer to the output of this projection as the patch embeddings. Similar to BERT’s [class] token, we prepend a learnable embedding to the sequence [...]. Position embeddings are added to the patch embeddings to retain positional information.

DavidLandup0: Also, good point on the interpolation flag! I'll take a look and try to add it in :)

bhack (commented):
I've some general doubt about the scalability of the `<specialized-network>_<layer_name>.py` approach specially if it is not a specialized version of a more general class/layer that we reuse for other derived network "flavors".
Could we introduce a network subfolder for the specific layers?
E.g. How we are going to naming all the derivative specialized layers?
See how many "variations" are collected in the quite popular repo: 
https://github.com/lucidrains/vit-pytorch/tree/main/vit_pytorch
https://github.com/lucidrains/vit-pytorch

DavidLandup0:
I share your concern, but don't know of a better way to do it, since if we have a general layer, and one for each variation - we'll have 3 instead of 2 files.
I originally wanted to add a `transformers` directory for components like the MLP heads, but it was dropped in the last PR

DavidLandup0:
@innat had a good idea for model-specific layers:

> If the layer is more specific to model, then I think it's better to place in the same {model}.py file, (like stable-diffusion).

I'd say that patch embeddings are general enough not to be part of the file, but for some specific ones, we can avoid scalability issues by merging them.

keras_cv/layers/vit_patching.py
DavidLandup0: Quick question - why would we want to convert patches to images? Or did you mean patch merging like it's used in SwinT?

innat: Here is one cases in [maxim model](https://github.com/google-research/maxim/blob/main/maxim/models/maxim.py#L56-L72).

innat: If keras-team thinks so. However, later I realize that, those image-to-patch and patch-to-image are kind of maxim-design oriented. So, it isn't general. Apart from it, this ticket (https://github.com/tensorflow/tensorflow/issues/6847) should be reopened.

keras_cv/layers/vit_patch_encoding.py
DavidLandup0: Added the `interpolate` flag! The patch sizes would be passed by the configuration of the model in the classfile. The rest (interpolation width/height and whether they want to interpolate) would be left to the user.

Updated the example Google Colab notebook with it too

DavidLandup0: Added support for interpolation! Can you take a look?

keras_cv/layers/vit_patch_embedding.py
```diff
@@ -0,0 +1,141 @@
...
+
+        dimensionality = embedding.shape[-1]
+
+        class_token = tf.expand_dims(embedding[:1, :], 0)
+        patch_positional_embeddings = embedding[1:, :]
+
+        h0 = height // patch_size
+        w0 = width // patch_size
+
+        new_shape = tf.constant(int(math.sqrt(self.num_patches)))
```
DavidLandup0, line 114: One thing that hurts is this because of `tf.consant(int(math...))`.
I tried `tf.math`, but even with several `tf.cast()`s, it still had issues with dtypes :/

DavidLandup0:
Any other feedback to iterate on this if anyone thinks it needs some changes?

DavidLandup0:
Quick question - considering the discussion at #950, are ViTs considered aligned with the current roadmap?
Both #671 and #668 are labeled with 'contributions-welcome', but these don't seem to be aligned with the current roadmap anymore.

While plain ViTs aren't the best backbones, SwinT's, MaxViT, etc. are, and KerasCV is currently missing transformer-oriented blocks and architectures. While that argument could be made - there are major hurdles to be taken care of in regard to object detection, semantic/instance segmentation, and text-to-image/video, so with resource constraints, it might be better to focus on those instead, since CNN-based models work as just-as-good backbones?

Pinging the people from the discussion @LukeWood @bhack @qlzh727 @tanzhenyu @martin-gorner

suvadityamuk:
> Quick question - considering the discussion at #950, are ViTs considered aligned with the current roadmap? Both #671 and #668 are labeled with 'contributions-welcome', but these don't seem to be aligned with the current roadmap anymore.
> 
> While plain ViTs aren't the best backbones, SwinT's, MaxViT, etc. are, and KerasCV is currently missing transformer-oriented blocks and architectures. While that argument could be made - there are major hurdles to be taken care of in regard to object detection, semantic/instance segmentation, and text-to-image/video, so with resource constraints, it might be better to focus on those instead, since CNN-based models work as just-as-good backbones?
> 
> Pinging the people from the discussion @LukeWood @bhack @qlzh727 @tanzhenyu @martin-gorner

Quite agree with most of the discussion here. I believe that transformer-based architectures need to become an integral part of  KerasCV simply because there is a growing demand to try out these architectures, be it in a student/research setting as an experiment or in much more critical workflows in enterprise settings. 
@DavidLandup0 Would be open to working with you on this further if this goes forward.

tanzhenyu:
> Quick question - considering the discussion at #950, are ViTs considered aligned with the current roadmap? Both #671 and #668 are labeled with 'contributions-welcome', but these don't seem to be aligned with the current roadmap anymore.
> 
> While plain ViTs aren't the best backbones, SwinT's, MaxViT, etc. are, and KerasCV is currently missing transformer-oriented blocks and architectures. While that argument could be made - there are major hurdles to be taken care of in regard to object detection, semantic/instance segmentation, and text-to-image/video, so with resource constraints, it might be better to focus on those instead, since CNN-based models work as just-as-good backbones?
> 
> Pinging the people from the discussion @LukeWood @bhack @qlzh727 @tanzhenyu @martin-gorner

Many transformed based architecture can be used as backbones, I believe this is aligned with our roadmap, we're currently looking for contribution on ViT and SwinT.

bhack:
@tanzhenyu's ticket https://github.com/keras-team/keras-cv/issues/668 was already labeled `contribution welcome` also if not  auto-connect this this ticket. 
I think that this label could be regularly aligned by the team on the internal roadmap update/review meetings so we will not going to create doubts on every contribution when we have already tickets that are available in the repository with this label.

DavidLandup0:
@tanzhenyu Thank you for the clarification! I'll await reviews for the code/PR then and update as requested :)
Should I go ahead and do a PR for ViT with these components as well? I wanted to avoid making a ViT PR being built on these if there are any changes required for these, since then we'd have to sync them possibly multiple times (can't work in parallel), but if you feel like it's okay to go ahead on ViTs - I'd gladly submit the model(s)!

DavidLandup0:
Do I go ahead and submit a ViT PR too, or would you prefer to iterate on this first? @tanzhenyu

keras_cv/layers/transformer_encoder.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+        self.dense1 = layers.Dense(self.mlp_units[0], activation=activation)
+        self.dense2 = layers.Dense(self.mlp_units[1])
+
+    def call(self, inputs):
+
+        x1 = self.layer_norm(inputs)
+        attention_output = self.attn(x1, x1)
+        x2 = layers.Add()([attention_output, inputs])
+        x3 = self.layer_norm(x2)
```
tanzhenyu, line 82: does this need to be a separate layer norm, or both inputs and x2 are used to update the same set of variables?

keras_cv/layers/vit_class_tokenizing.py
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class ClassTokenizing(layers.Layer):
+    """
+    Layer to concat learnable class token to input for Vision Transformers from:
```
tanzhenyu, line 22: Not sure which part of this layer is "learnable"?

tanzhenyu (commented):
Thanks for the PR! I have left some comments.
Also on top of those, one main question is how will there layers be used in ViT, can you put an example gist in the description? (The PatchEmbedding and Patching seems coupled, so we'd like to understand more on its usage)
Another question for a follow-up PR would be, how do you plan to provide ViT -- does it come with a training script, or porting weights?

tanzhenyu:
> Do I go ahead and submit a ViT PR too, or would you prefer to iterate on this first? @tanzhenyu

Let's find a common ground then -- iterate on this by describing how it will be used in the ViT PR?
(Sorry for the delayed review, I have been busy with DeepLab and RCNN and just completed those)

keras_cv/layers/transformer_encoder.py
```diff
@@ -0,0 +1,104 @@
...
+from tensorflow.keras import layers
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class TransformerEncoder(layers.Layer):
+    """
+    Transformer encoder block implementation as a Keras Layer.
+    args:
+        - project_dim: the dimensionality of the projection of the encoder
+        - intermediate_dim: the intermediate dimensionality in the MLP head
```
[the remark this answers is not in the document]
DavidLandup0: Good point! The original paper used 768 IIRC. I'll add that as the default value :)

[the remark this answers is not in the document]
DavidLandup0: Done

keras_cv/layers/transformer_encoder.py
```diff
@@ -0,0 +1,104 @@
...
+        - num_heads: the number of heads for the `MultiHeadAttention` layer
+        - dropout: default 0.1, the dropout rate to apply inside the MLP head of the encoder
```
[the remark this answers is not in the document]
DavidLandup0: Done. Actually helped me realize that I accidentally coupled the attention dropout to this parameter too. Now there's `mlp_dropout` applied between the layers of the MLP head and `attention_dropout` applied to the `MultiHeadAttention` layer

keras_cv/layers/transformer_encoder.py
```diff
@@ -0,0 +1,104 @@
...
+        x1 = self.layer_norm(inputs)
+        attention_output = self.attn(x1, x1)
+        x2 = layers.Add()([attention_output, inputs])
+        x3 = self.layer_norm(x2)
+        x3 = self.dense1(x3)
+        x3 = layers.Dropout(self.dropout)(x3)
+        x3 = self.dense2(x3)
+        x3 = layers.Dropout(self.dropout)(x3)
+
+        encoded_patches = layers.Add()([x3, x2])
```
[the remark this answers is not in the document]
DavidLandup0: Makes sense, done!

keras_cv/layers/transformer_encoder.py
DavidLandup0: Great question. I think it should be a separate layer norm. Let me check...

DavidLandup0: Should be separate, fixed. Thanks!

keras_cv/layers/vit_class_tokenizing.py
DavidLandup0: Changed the class token to be initialized as a learnable parameter in the new push. I have to test whether it compiles right now 🤔

DavidLandup0: Yeah, it's throwing an exception because of the unexpected size. Changed to accept the input shape since it's internally being used only with patches that are computed at the time, so we can guarantee input shapes. Let me verify if this works

DavidLandup0:
Updated! Compiles and trains on the Colab notebook

DavidLandup0:
Updated to differentiate between the intermediate dims for the MLP head and the transformer encoder:

image

Our `intermediate_dim` is their "Hidden Size D" and `mlp_dim` is "MLP Size"

tanzhenyu:
> > Thanks for the PR! I have left some comments. Also on top of those, one main question is how will there layers be used in ViT, can you put an example gist in the description? (The PatchEmbedding and Patching seems coupled, so we'd like to understand more on its usage)
> 
> There's a proof-of-concept Google Colab in the PR's description: https://colab.research.google.com/drive/1gzv6jVljae0dNtKhhq0qxRI8IIrejuCU?usp=sharing
> 
> It builds an approximate ViT and shows that it does train on a few epochs using these components with a custom KerasCV-styled implementation.
> 
> > Another question for a follow-up PR would be, how do you plan to provide ViT -- does it come with a training script, or porting weights?
> 
> Since it'd be a custom implementation for KerasCV with these components - I don't think we could port existing weights. I could cobble up a training script or we could use the general basic trainer by @ianstenbit

Ok, looks pretty good. I'd suggest two things:
1) no need for a separate Classtokenizing since it's only being used in one layer
2) let's leave all components except TransformerEncoder under a single file, given it's quite specialized for ViT model at this point.

DavidLandup0:
For reference, Deit has a similar `PatchEmbedding`, with the addition of a distillation token. IMO, we should generalize the layer to fit both ViTs and Deits in that case. Thoughts?

image

Beit is a similar story:

image

tanzhenyu:
> Sure, I'll remove ClassTokenizing and handle the class token in the patch embedding layer instead! Though, isn't `Patching` at least general enough to be a standalone layer? It might be worth generalizing `PatchEmbedding` for other uses too?

If such use cases come up, then we can migrate them to `keras_cv.layers`. Though for now it seems other transformer based models isn't using it yet?

DavidLandup0:
Currently, no, no other layers are using them. Let's ship them together and then separate later if you think it's a good idea :)

tanzhenyu:
> Deit

I haven't followed Deit yet, we might need to do a impact analysis to see if we need to bring them to KerasCV. If yes and indeed this layer can be re-used, then we need to make a patching layer as public offering. For now making it as an internal implementation is the easiest way to unblock the ViT project

DavidLandup0:
Makes sense, thanks for weighing in! I'll put them all under `vit_components.py` temporarily and we'll unroll them later for DeiTs, BeiTs, Swin Transformers, MaxViT, etc. :)

DavidLandup0:
> I haven't followed Deit yet, we might need to do a impact analysis to see if we need to bring them to KerasCV

Once the basic components are done and merged with ViT, I can do a sweep for transformer-based papers and open up a few issues with citation counts and how commonly they're used/referenced in benchmark tables if need be, similar to how other models have been registered in issues so far

tanzhenyu:
> > I haven't followed Deit yet, we might need to do a impact analysis to see if we need to bring them to KerasCV
> 
> Once the basic components are done and merged with ViT, I can do a sweep for transformer-based papers and open up a few issues with citation counts and how commonly they're used/referenced in benchmark tables if need be, similar to how other models have been registered in issues so far

That'd be great. Though at this point I'm more concerned on how the training script should look like, since from the paper it seems to have more tricks that the current training script supports.

DavidLandup0:
True. We might have to create a `transformer_training.py` script for most transformer-based models.
I'll look into creating the basic transformer training script with the first ViT model to bootstrap the development

DavidLandup0:
- Put all of the ViT components into a single file. 
- Turned ClassTokenizing from layer to regular code. When we do other models, we might want do re-introduce it as a layer to keep the rest of the code cleaner (one-line call vs multiple-line call).
- Simplified the PatchEmbedding layer so that it doesn't take `num_patches` or `input_shape`, which are now inferred at build time. 
- Updated docstrings.

DavidLandup0:
- Updated: https://colab.research.google.com/drive/1BbCtaKN_mmVEd-iZlWQ8GpgL55FOXrwK?usp=sharing

DavidLandup0:
For some reason, this no longer works:

```
patches = tf.reshape(
            patches, [batch_size, -1, patch_dims]
)
```

And returns `[batch_size, None, patch_dims]`. Looks like the `-1` is making it lose shape info, for whatever reason. Only seems to work with an ugly workaround:

```
patches = tf.reshape(
            patches, [batch_size, patches.shape[-2] * patches.shape[-2], patch_dims]
)
```

Gotta look into this tomorrow.

DavidLandup0:
Updated: https://colab.research.google.com/drive/1gzv6jVljae0dNtKhhq0qxRI8IIrejuCU?usp=sharing

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,204 @@
...
+    Basic usage:
+
+    ```
+    patches = keras_cv.layers.Patching(patch_size)(batch_img)
+
+    project_dim = 1024
+    num_patches = patches.shape[1] # 196
+
+    encoded_patches = keras_cv.layers.PatchEmbedding(project_dim=project_dim)(patches)
+    print(encoded_patches.shape) # (1, 197, 1024)
```
tanzhenyu, line 99: maybe add an explanation here why it changes from 196 to 197?

DavidLandup0: Sure! I'll expand the docstring to emphasize the addition of the class token too, since it's part of the layer now.

keras_cv/layers/__init__.py
```diff
@@ -73,3 +73,6 @@
 from keras_cv.layers.regularization.squeeze_excite import SqueezeAndExcite2D
 from keras_cv.layers.regularization.stochastic_depth import StochasticDepth
 from keras_cv.layers.spatial_pyramid import SpatialPyramidPooling
+from keras_cv.layers.transformer_encoder import TransformerEncoder
+from keras_cv.layers.vit_layers import PatchEmbedding
+from keras_cv.layers.vit_layers import Patching
```
tanzhenyu, line 78: Also make Patching and PatchEmbeddings private layers

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,204 @@
...
+    print(f"Patch size: {(patch_size, patch_size)}") # Patch size: (16, 16)
+    print(f"Patches: {patches.shape[1]}")            # Patches: 196
+    ```
+    """
+
+    def __init__(self, patch_size, **kwargs):
+        super().__init__(**kwargs)
+        self.patch_size = patch_size
+
+    def call(self, images):
```
tanzhenyu, line 53: add a "args / returns" section that describes what the input / output dtypes and shapes are

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,204 @@
...
+        batch_size = tf.shape(images)[0]
+        patches = tf.image.extract_patches(
```
tanzhenyu, line 55: what if H / W is not divisible by `patch_size` -- can we add more documentation either here or above so readers will have a better understanding, i.e., crop or pad?

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,204 @@
...
+            "patch_size": self.patch_size,
+        }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class PatchEmbedding(layers.Layer):
+    """
+    Layer to concat a class token, positionally embed and create a projection of patches made with the `Patching` layer
```
tanzhenyu, line 79: instead of "concat", how about we say "prepend"?

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+
+    def __init__(self, project_dim, **kwargs):
+        super().__init__(**kwargs)
+        self.project_dim = project_dim
+        self.linear_projection = layers.Dense(self.project_dim)
+
+    def build(self, input_shape):
+        self.class_token = tf.random.normal([1, 1, input_shape[-1]], name="class_token")
+        self.num_patches = input_shape[1]
```
tanzhenyu, line 110: isn't the number of patches dynamic?

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,204 @@
...
+        patch_shape = tf.shape(patch)
+        class_token_broadcast = tf.cast(
+            tf.broadcast_to(
+                self.class_token,
+                [patch_shape[0], 1, patch_shape[-1]],
+            ),
+            dtype=patch.dtype,
+        )
+        patch = tf.concat([class_token_broadcast, patch], 1)
+        # num_patches + class token
```
tanzhenyu, line 133: `# class token + num_patches`

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,204 @@
...
+        self.linear_projection = layers.Dense(self.project_dim)
+
+    def build(self, input_shape):
+        self.class_token = tf.random.normal([1, 1, input_shape[-1]], name="class_token")
+        self.num_patches = input_shape[1]
+        self.position_embedding = layers.Embedding(
+            input_dim=self.num_patches + 1, output_dim=self.project_dim
+        )
+
+    def call(
```
tanzhenyu, line 115: add description for inputs and outputs (similar to other kerascv layers)

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,204 @@
...
+            patch_size,
+        ):
+            raise ValueError(
+                "`None of `interpolate_width`, `interpolate_height` and `patch_size` cannot be None if `interpolate` is True"
+            )
+        else:
+            encoded = self.linear_projection(patch) + self.position_embedding(positions)
+        return encoded
+
+    def interpolate_positional_embeddings(self, embedding, height, width, patch_size):
```
tanzhenyu, line 161: do we expect users to call this? If not then make it private

keras_cv/layers/__init__.py
DavidLandup0: Just removing them from the __init__ file is enough, right?

keras_cv/layers/vit_layers.py
DavidLandup0: There's an "args" already. I'll add in the "returns"

DavidLandup0: Done

keras_cv/layers/vit_layers.py
DavidLandup0: Added note in the main docstring

keras_cv/layers/vit_layers.py
DavidLandup0: Done!

keras_cv/layers/vit_layers.py
DavidLandup0: Whoops - left it in as a reminder during development as an explanation for the +1 in the next line. Removed!

keras_cv/layers/vit_layers.py
DavidLandup0: Done

keras_cv/layers/vit_layers.py
DavidLandup0: The number is inherited from the input, the class token is pre-pended, and if the interpolation flag is set - the number of patches is changed, and the positional embeddings are interpolated for the new shape. I'd say it's dynamic

keras_cv/layers/vit_layers.py
DavidLandup0: Nope. This is called internally when the flag is set to True. 
Prepending `__` is enough, or?

tanzhenyu: yeah `_` is good enough

keras_cv/layers/vit_layers.py
tanzhenyu: yep

keras_cv/layers/vit_layers.py
tanzhenyu: yeah adding comments is good enough -- in general I prefer adding as many comments as possible, because code is our product :-)

keras_cv/layers/vit_layers.py
tanzhenyu: right, as you mentioned, it depends on the input (or the output of the Patching layer). Given the patch size is fixed, this means the number of patches changes as the H / W changes

keras_cv/layers/__init__.py
tanzhenyu: yes

keras_cv/layers/vit_layers.py
DavidLandup0: Done!

keras_cv/layers/vit_layers.py
DavidLandup0: Done

keras_cv/layers/vit_layers.py
DavidLandup0: Hmm, yes - this could pose an issue with datasets where the sizes change between batches. I don't know how common these are, but we'd ideally want to allow for those as well. Good catch!
I'm not sure how we'd make it fully dynamic (None types tend to cause build-type EagerTensor exceptions) but I'll look into how we can make it robust to datasets with batches of images of different shapes

keras_cv/layers/vit_layers.py
DavidLandup0: Agreed! I've found Keras/TF docstrings essential to grasping the inner workings before. I think this double-applies to specialized blocks meant to be re-used as components

keras_cv/layers/__init__.py
DavidLandup0: Done

DavidLandup0:
Thank you for the comments! I think I got them all now. Let me know if you think we should expand the "accepts"/"returns" comments for each `call()` in the `vit_layers.py` file.

keras_cv/layers/vit_layers.py
tanzhenyu: To be more specific, I mean adding this after `call`.
Example:
https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/spatial_pyramid.py#L140-L148

keras_cv/layers/vit_layers.py
tanzhenyu: That depends on the training script, some of our existing models require fixed shape, so they're resized and cropped. You might need to read into the the paper to see if they are doing something similar, or actually passing different shapes across batches -- whatever gets you best results.
But in general, supporting dynamic input shape is desired because users can structure their own dataset to do so

tanzhenyu:
Kicking off the GPU testing as this PR is close to the finish line

tanzhenyu:
you'd also need to run `sh shell/lint.sh` and `sh shell/format.sh` to fix the formatting

bhack:
> you'd also need to run `sh shell/lint.sh` and `sh shell/format.sh` to fix the formatting

Or you can locally automate our linting with `git config core.hooksPath .github/.githooks`

DavidLandup0:
Good point. I'll add those as auto hooks

DavidLandup0:
Linting rn

bhack:
> Good point. I'll add those as auto hooks

It is just a bit more environmentally friendly as we need to re-run the whole CI jobs set also jut for a small format ifx. :smile_cat:

DavidLandup0:
> It is just a bit more environmentally friendly...

*LLMs enter the chat*

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,220 @@
...
+class Patching(layers.Layer):
+    """
+    Layer to turn images into a sequence of patches for Vision Transformers from:
+        - An Image is Worth 16x16 Words: Transformers for Image Recognition at Scale
+        by Alexey Dosovitskiy et al. (https://arxiv.org/abs/2010.11929)
+    Based on Khalid Salama's implementation for:
+        - https://github.com/keras-team/keras-io/blob/master/examples/vision/image_classification_with_vision_transformer.py
+
+    The layer expects a batch of input images and returns batches of patches. If the height and width of the image
+    isn't divisible by the patch size, "VALID" padding is used.
```
LukeWood, line 31: can we expose padding type to be an input argument?

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,220 @@
...
+
+    args:
+        - patch_size: the size (patch_size, patch_size) of each patch created from the image
```
LukeWood, line 34: Can we write this docstring in the Google style?

```
Args:
    patch_size: ....
```

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,220 @@
...
+
+import math
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class Patching(layers.Layer):
+    """
```
LukeWood, line 23: lets start docstrings with a one line description of what the component does

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,220 @@
...
+    print(f"Patches: {patches.shape[1]}")            # Patches: 196
+    ```
+    """
+
+    def __init__(self, patch_size, **kwargs):
+        super().__init__(**kwargs)
+        self.patch_size = patch_size
+
+    """
+    Accepts a batch of images as input.
```
LukeWood, line 57: looks like this docstring is on the wrong line

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,220 @@
...
+        config = {
+            "patch_size": self.patch_size,
+        }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class PatchEmbedding(layers.Layer):
+    """
```
LukeWood, line 86: start with a one line description, then give a full descroption in the follow up paragraph.  This is important for keras.io documentation rendering

keras_cv/layers/vit_layers.py
LukeWood: +1, supporting dynamic input shape when possible is optimal.

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,220 @@
...
+        self.linear_projection = layers.Dense(self.project_dim)
+
+    def build(self, input_shape):
+        self.class_token = tf.random.normal([1, 1, input_shape[-1]], name="class_token")
+        self.num_patches = input_shape[1]
+        self.position_embedding = layers.Embedding(
+            input_dim=self.num_patches + 1, output_dim=self.project_dim
+        )
+
+    """
```
LukeWood, line 126: docstrings should be below call, not above

LukeWood (dismissed):
Some minor comments, thanks @DavidLandup0 !

keras_cv/layers/vit_layers.py
DavidLandup0: Sure! I'll expose it and add a check for the input

keras_cv/layers/vit_layers.py
DavidLandup0: Sure! I think I saw this style somewhere else in KCV, so it might be worth taking a look and standardizing all docs at some point in the future

keras_cv/layers/vit_layers.py
DavidLandup0: Done

keras_cv/layers/vit_layers.py
DavidLandup0: Done

keras_cv/layers/vit_layers.py
DavidLandup0: You mean, before the classnames? They currently start with a one-liner in the class:

> ayer to turn images into a sequence of patches for Vision Transformers from:
        - An Image is Worth 16x16 Words: Transformers for Image Recognition at Scale
        by Alexey Dosovitskiy et al. (https://arxiv.org/abs/2010.11929)

keras_cv/layers/vit_layers.py
DavidLandup0: Could you clarify? Where should it be at? 🤔

keras_cv/layers/vit_layers.py
DavidLandup0: Oh, okay, scratch the previous question. Thanks!

DavidLandup0: Done

keras_cv/layers/vit_layers.py
DavidLandup0: Done

keras_cv/layers/vit_layers.py
DavidLandup0: Got it from another comment, hopefully haha

DavidLandup0:
Btw, I've left the example usage in the docs as `keras_cv.layers.LayerName` even though they're currently only for internal use, as they're likely going to be public layers generalized to other transformer models down the line. Is that okay?

keras_cv/layers/transformer_encoder.py
```diff
@@ -0,0 +1,117 @@
...
+    print(trans_encoded.shape) # (1, 197, 1024)
+    ```
+    """
+
+    def __init__(
+        self,
+        project_dim,
+        num_heads,
+        intermediate_dim=768,
+        mlp_dim=3072,
```
tanzhenyu, line 59: just curious -- is this the default value being used in ViT? 3072 seems a really high dimension to me.

LukeWood: good point; 

we probably don't want a default for either of these given that this is meant to be a generic layer.

DavidLandup0: Yup, a pretty high dimensionality. They set it as the dims for their "base" model for the output of the MLP head in the param tables.

Let me re-read the paper to check if I got it right or not. Even if I did, should we maybe reduce it?

DavidLandup0: MLP dims for ViT-Base:

image

HuggingFace's ViT config also uses 3072 as the default value:

https://huggingface.co/docs/transformers/model_doc/vit#transformers.ViTConfig

DavidLandup0: It _feels_ wrong, but it looks like it's right haha

DavidLandup0: Though, since it's a generic transformer encoder - we might as well have it be 512 or 1024 for other use cases and then supply 3072 for ViTs specifically?

DavidLandup0: If someone wants to build a custom ViTs or ViT derivatives with these, it makes sense to keep the defaults for ViTs though so that they don't have to update the args?
Not sure what's the right choice...

DeiT and BeiT have 3072 too. It might be better to have the defaults stay as 768 and 3072 because they seem to be a common default. I totally get why we'd want to decrease them too though, so whatever you feel like is the right choice works

DavidLandup0: We can either go with having defaults for all of these (12 heads is also a common default) or go with none of them

DavidLandup0: DPT (Dense Prediction Transformer) for CV also have 768 hidden and 3072 mlp dims

DavidLandup0: TL:DR;

ViT, DeiT, BeiT, DPT, ViTMAE, ViTMSE use 768 (transformer hidden dims), 3072 (MLP dims) and 12 heads as the defaults/lower bounds for models.
LeViT and SwinT married convolutions back into the mix and use lower dimensionality for the MLP heads and transformer encoders because of the lower dims coming in (I think).

We can go either with what they've "historically" been (quotes because they're all from the past 2 years) or anticipate more SwinT-type models that use convs and lower dims and remove the defaults fully. Thoughts?

bhack: > Thoughts?

Theoretically it could be ok but I will find hard to communicate to the user base what they are really using when they adopt/retain/fine-tune Keras-cv models.

I mean that generally users expect to have the same arch when we are using "well known names" and we need to handle customization optionally or handling it with a specific care.

DavidLandup0: We can do that when building models. The `MODELS_CONFIG` would have dims set to 768/3072 for ViTs for example, totally regardless of the default transformer args.
The question is whether we want to set the encoder defaults to something or nothing at all?

bhack: Yes I was talking more about the 2nd part:
> or anticipate more SwinT-type models that use convs and lower dims and remove the defaults fully.

DavidLandup0: Maybe we go with no defaults for the generic transformer layer, and enforce the common dimensionality on a per-model basis, since we'll have those set _anyway_

LukeWood: i think having no defaults is a better way to communicate to users what is going on.  I tend to prefer explicitness when any ambiguity is involved.

DavidLandup0: Makes sense! Removed the defaults :)

tanzhenyu (approved):

keras_cv/layers/vit_layers.py
DavidLandup0: Got it! Updated :)

DavidLandup0:
Passed accelerator testing before the latest push (you might not want to re-run the test if it costs any compute/CI minutes)

tanzhenyu:
I see Luke have some minor comments which are already addressed here, so merging it, Luke please create issues if you see additional concern

keras_cv/layers/vit_layers.py
```diff
@@ -0,0 +1,243 @@
...
+    def get_config(self):
+        config = {
+            "patch_size": self.patch_size,
+        }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class PatchEmbedding(layers.Layer):
```
LukeWood, line 98: can we put patching and patch_embedding in their own .py files?

LukeWood:
@DavidLandup0 how do you feel about adding basic unit test coverage?  This should include serialization, call(), graph mode call(), etc.

keras_cv/layers/vit_layers.py
DavidLandup0: They originally were, but were put into a single file after: https://github.com/keras-team/keras-cv/pull/929#issuecomment-1306515688

DavidLandup0:
Sounds good! I'll add the test files :)

keras_cv/layers/vit_layers.py
DavidLandup0: I'd personally opt for having them be standalone layers in different files, since they'll be used/reused for other vision transformers, but whatever is the team ruling is what I'll go with ;)

DavidLandup0:
Any problems with `tf.keras.activations.serialize` before? It's failing to serialize `tf.nn.gelu`:

```
config.update(
            {
                ...
                "activation": keras.activations.serialize(self.activation),
            }
```

It also fails `tanh` and other functions that should work according to the docs: https://www.tensorflow.org/api_docs/python/tf/keras/activations/serialize

I've fixed this by using `'gelu'` instead of `tf.nn.gelu`, but it might be worth looking into why the previous approach doesn't work.

bhack:
> Any problems with `tf.keras.activations.serialize` before? It's failing to serialize `tf.nn.gelu`:
> 
> ```
> config.update(
>             {
>                 ...
>                 "activation": keras.activations.serialize(self.activation),
>             }
> ```
> 
> It also fails `tanh` and other functions that should work according to the docs: https://www.tensorflow.org/api_docs/python/tf/keras/activations/serialize
> 
> I've fixed this by using `'gelu'` instead of `tf.nn.gelu`, but it might be worth looking into why the previous approach doesn't work.

Probably you have misinterpreted this behavior a bit as you was still passed a string  from the test.

So you need something like:

```python
...
        activation="gelu",
        layer_norm_epsilon=1e-06,
        **kwargs
    ):
        super().__init__(**kwargs)

        self.project_dim = project_dim
        self.intermediate_dim = intermediate_dim
        self.mlp_dim = mlp_dim
        self.num_heads = num_heads
        self.mlp_dropout = mlp_dropout
        self.attention_dropout = attention_dropout
        self.activation = tf.keras.activations.get(activation)
...
```

tanzhenyu:
> > Any problems with `tf.keras.activations.serialize` before? It's failing to serialize `tf.nn.gelu`:
> > ```
> > config.update(
> >             {
> >                 ...
> >                 "activation": keras.activations.serialize(self.activation),
> >             }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It also fails `tanh` and other functions that should work according to the docs: https://www.tensorflow.org/api_docs/python/tf/keras/activations/serialize
> > I've fixed this by using `'gelu'` instead of `tf.nn.gelu`, but it might be worth looking into why the previous approach doesn't work.
> 
> Probably you have misinterpreted this behavior a bit as you have still passed a string from the test.
> 
> So you need something like:
> 
> ```python
> ...
>         activation="gelu",
>         layer_norm_epsilon=1e-06,
>         **kwargs
>     ):
>         super().__init__(**kwargs)
> 
>         self.project_dim = project_dim
>         self.intermediate_dim = intermediate_dim
>         self.mlp_dim = mlp_dim
>         self.num_heads = num_heads
>         self.mlp_dropout = mlp_dropout
>         self.attention_dropout = attention_dropout
>         self.activation = tf.keras.activations.get(activation)
> ...
> ```

yep, most likely this

bhack:
As the model training loop is by defaul wrapped/traced in a tf.function it will not go to create overhead.

DavidLandup0:
Noted, I won't waste time on that then. Unit tests incoming :)

keras_cv/layers/vit_layers.py
DavidLandup0: Do we keep them in the same file or do we separate them in the end? @LukeWood

keras_cv/layers/vit_layers.py
```diff
@@ -87,9 +87,7 @@ def call(self, images):
         return patches
```
LukeWood, line 87: can you validate self.padding in the constructor instead of in call()?

DavidLandup0: Added validation in the latest push

DavidLandup0: Done

DavidLandup0:
Added tests for output shapes and dtypes for `TransformerEncoder`, `Patching`, `PatchEmbedding`, as well as for wrong inputs. This lead to a few new checks in the layers themselves, and a scouted bug when using the `interpolate` flag. That's fixed now.
Latest push also includes serialization tests.

DavidLandup0:
The final question is - do we keep Patching and PatchEmbedding as separate files or no? :)

tanzhenyu:
> The final question is - do we keep Patching and PatchEmbedding as separate files or no? :)

Please just put that inside vit_layers.py, if we see fit for other models we will move it into separate files later

DavidLandup0:
Bad wording on my end - they're both in the same file now. Luke mentioned maybe separating them so I was unsure whether or not to keep them in the same file or separate them

keras_cv/layers/vit_layers_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+        output = layer(
+            patch=inputs,
+            interpolate=True,
+            interpolate_width=300,
+            interpolate_height=300,
+            patch_size=12,
+        )
+
+        self.assertTrue(isinstance(output, tf.Tensor))
```
tanzhenyu, line 67: Also add some numerical tests if possible -- especially on the interpolation code path

DavidLandup0: Possibly a stupid question - but can we enforce numerical tests here? There are multiple random elements in the code, so we'd get different results on each run, even with the same input and even when setting a global seed with `tf.keras.utils.set_random_seed(seed)`.

I figured we could probably test the input embeddings against the interpolated and then sampled down embeddings (because they should be approximately the same) but the input to the layer is a sequence of patches, and the result is a linear projection which no longer resembles images but an arbitrary learnable vector, so that numerical test falls down the drain :/

Open to ideas!

bhack:
Can we just very quickly check if we have some interesting tests in:
https://github.com/huggingface/transformers/tree/main/tests/models/vit

DavidLandup0:
I can only see one numerical test for the feature extractor of a pre-trained ViT

DavidLandup0:
The rest of the tests are for dtypes, shapes, and whether the ViT model builds/classifies, etc.

DavidLandup0:
Though, they're testing the model - we're trying to do tests for components of it here 🤔

keras_cv/layers/vit_layers_test.py
tanzhenyu: we should test the numerical outputs of the patching logic, if possible

DavidLandup0: The interpolated output is also summed with a Dense projection before being returned:

```
reshaped_embeddings  = #... resizing, reshaping, etc. with interpolation
linear_projection = self.linear_projection(reshaped_embeddings)
addition = linear_projection + reshaped_embeddings

return tf.concat([class_token, addition], 1)
```

The `Dense` call here that's added to the `reshaped_embeddings` will necessarily add randomness to the `__interpolate_positional_embedings` output. I'll separate the logic for `__interpolate_positional_embeddings` to only interpolate and return the embeddings, and then do the linear projection and summation outside the method. This should make the method testable...

For the other one - sounds good! I'll add that in :)

DavidLandup0: Done! I changed the logic for the interpolation method to just interpolate and the addition with the linear projection is now done outside of it. With this, we can check whether an all 1s tensor is still interpolated as all 1s in the end. The patching layer was tested on this image:

image

And as to whether it correctly returns:

image

...numerically. Thanks for requesting these! Could you take a look? @tanzhenyu

keras_cv/layers/__init__.py
DavidLandup0: Removed in the latest commit (accidentally left them in in the one before for testing on a local Jupyter Notebook)

tanzhenyu:
Thanks for the PR! Please go ahead with two possible path:
1. directly porting the weights from somewhere else, citing the source and making sure we don't break any license
2. have a transformer training script and run for 30% of the total training epochs, check the metrics is improving, and the Keras team can run the script end to end and produce the final weights.

bhack:
>     2. have a transformer training script and run for 30% of the total training epochs, check the metrics is improving, and the Keras team can run the script end to end and produce the final weights.

As a reference point we had some epoch related graph from @sayakpaul at https://github.com/sayakpaul/big_vision_experiments

There are derived from google big vision repo:
https://github.com/google-research/big_vision
https://arxiv.org/abs/2106.10270

DavidLandup0:
Many thanks for the merge, and thank you for the questions and requests in the reviews! 🤗 
I'll take a look at what we can do with the weights to avoid training if possible, or come up with a transformer training script since they differ somewhat from the regular ones.

Thanks for the reference @bhack! Taking a look :)

DavidLandup0:
Before I open up a PR for ViT models - do we want to implement hybrid ViTs or basic ones, or both?

We'd probably want to do this as two sets of classes - `ViTHybrid` and `ViT`. The difference is whether there's a (modified) ResNet stem in the beginning and whether we're feeding raw images or feature maps into the `Patching` layer. The Hybrid version outperforms the one without any conv blocks.

- The official implementation does both with flags: https://github.com/google-research/vision_transformer/blob/main/vit_jax/models_vit.py
- timm does both in separate classes: https://github.com/rwightman/pytorch-image-models/blob/main/timm/models/vision_transformer_hybrid.py, https://github.com/rwightman/pytorch-image-models/blob/main/timm/models/vision_transformer.py
- HuggingFace does hybrid only: https://github.com/huggingface/transformers/blob/v4.24.0/src/transformers/models/vit/modeling_tf_vit.py

If we have to choose one - IMO, we should take the hybrid ViTs. Alternatively, we can do both :)
Thoughts?
Tagging a few people for the discussion: @LukeWood @tanzhenyu @bhack @innat

bhack:
I don't know if I've read table correctly but it seems that we have a reference TB.dev (cifar10, cfar-100, imagenet2012) only for ViT-B_16 right?

DavidLandup0:
TB.dev for ResNet50+ViT-B (ViT hybrid mentioned above), but the rest are not hybrids (ViT-S, ViT-B and ViT-L with various patch sizes)

DavidLandup0:
Curious to see that they got a higher acc on the non-hybrid one. The authors from the paper note otherwise...

bhack:
> TB.dev for ResNet50+ViT-B (ViT hybrid mentioned above), but the rest are not hybrids (ViT-S, ViT-B and ViT-L with various patch sizes)

Yes I meant B_16 for the "Hybrid". I think in general as soon as we have a reference TB.dev is ok.

> Curious to see that they got a higher acc on the non-hybrid one. The authors from the paper note otherwise...

They hybrid seems to be slower also at inference time. 
Have you checked also the table at page 9 in https://arxiv.org/abs/2106.10270

DavidLandup0:
I think it's the opposite - this is measured as throughput, so hybrid models seem to be able to do way more images per second, due to a lower inference time.

> Yes I meant B_16 for the "Hybrid". I think in general as soon as we have a reference TB.dev is ok.

Maybe we do the same setups they did there for starters, and then expand to more hybrid setups when we're able to test them through the lens of KCV?

bhack:
> I think it's the opposite - this is measured as throughput, so hybrid models seem to be able to do way more images per second, due to a lower inference time.

- R26+S/32 -> 327 image/sec
- S/32 -> 2154 image/sec

DavidLandup0:
Yeah, you're right. I misread it.
The 32 patch_size ones seem to be significantly faster at inference time. 16 patch_size is way more computationally expensive (longer sequences when flattened).
Maybe we do the same setups they have and assess whether we want to do more hybrid models?

bhack:
I suggest to start with the smallest one on a small dataset (CIFAR10) to check how far we are from the TB.dev with the first n% epochs.

DavidLandup0:
Deal! Doing ViT-B_16 on CIFAR10 and comparing it. Thanks for weighing in!

DavidLandup0:
So, the replicated results from the papers they noted in https://github.com/google-research/vision_transformer#available-vit-models is actually pre-trained on ImageNet21k and then fine-tuned on CIFAR10, CIFAR100, ImageNet1k, etc. so our runs from scratch on CIFAR10 won't match their TB.dev.
Might be better to benchmark them against a plain ResNet, then? @bhack

bhack:
Yes it was a little bit confusing on the 1st table as in the 2nd one it has a clear upstream:
immagine

So It seems that we don't have any public TB.dev for the pre-training job right?

DavidLandup0:
Yup, the tables don't mention it in the beginning but the paragraph above it does. And the TB logs show a near instant 97% acc on the first epoch 

That being said, I don't know of public ViT TB logs. Gotta take a look but other than results, we might have to guide ourselves with a ResNet baseline?

bhack:
I think we have two solution right now:

- the public curves in the Google big-model fork at https://github.com/sayakpaul/big_vision_experiments. It Isn't like a TB.dev but better then nothing.

- The alternative is to start earlier with per-trainined weight and test the fine-tuning job using the TB.dev references.

@tanzhenyu What do you think?

tanzhenyu:
> I think we have two solutions right now:
> 
> * the public curves in the Google big-model fork at https://github.com/sayakpaul/big_vision_experiments. It Isn't like a TB.dev but ~better then nothing~ the training logs are available.
> * The alternative is to start earlier with per-trainined weights and test out fine-tuning jobs using the TB.dev references.
> 
> @tanzhenyu What do you think?

I think it's ok to start with something simple, and always make sure that we can reproduce the result (a core value of this repo). @DavidLandup0 FYI.
(So ViT tiny sounds good to me)

bhack:
Yes the problem was that we have full log only for S16.

For the others and smallers we have "fine-tuning" logs starting from pretained weights.

tanzhenyu:
> https://github.com/google-research/vision_transformer#available-vit-models

@lucasb-eyer is it possible to share the pre-training logs for tiny / small models so TF2 based library can reproduce the results?

lucasb-eyer:
It's a bit too much work to dig those from "how to train your vit" out and publish them now.

However, we already did publish the logs for [our recommended S/16 baseline](https://arxiv.org/abs/2205.01580) , with [a colab to read/plot them](https://colab.research.google.com/drive/1Iau-_yVPLsAur4PP6NTMdoBKl0lct8nS), hope that helps!

bhack:
> It's a bit too much work to dig those from "how to train your vit" out and publish them now.
> 
> However, we already did publish the logs for [our recommended S/16 baseline](https://arxiv.org/abs/2205.01580) , with [a colab to read/plot them](https://colab.research.google.com/drive/1Iau-_yVPLsAur4PP6NTMdoBKl0lct8nS), hope that helps!

I don't know if @DavidLandup0 has enough hw resources to target `90 epochs of training surpass 76% top-1` that it seems it is the first available datapoints on that model. The target required >6h `on TPUv3-8`

DavidLandup0:
I'm running a single GTX 1660 Super, so probably not *but* I we could use Kaggle's TPU for 6h 🤔

tanzhenyu:
> I'm running a single GTX 1660 Super, so probably not _but_ I we could use Kaggle's TPU for 6h 🤔

If you can mimic the training logic there and train for 6 hours, the keras team should be able to carry this on and train end to end

PR #933, opened 2022-10-17
Add bugfix to MaP IoU computation

LukeWood:


keras_cv/metrics/coco/mean_average_precision.py
```diff
@@ -229,7 +229,7 @@ def update_state(self, y_true, y_pred, sample_weight=None):
                 )
 
                 ious = iou_lib.compute_iou(
-                    ground_truths_by_category, detections_by_category, "yxyx"
+                    ground_truths_by_category, detections_by_category, "xyxy"
```
ianstenbit, line 232: Is there a test case we could write which would have caught this?

ianstenbit (approved):

keras_cv/metrics/coco/mean_average_precision.py
bhack: As bboxes argument/objects don't transport any metadata about the format it is hard.

https://github.com/keras-team/keras-cv/blob/3a9b8f020ba499af590ccfffee6603bc1a4c0764/keras_cv/bounding_box/iou.py#L65-L97

You could check if x <= height  and y <= width but just in the case we don't have out of "frame" coordinates.

tanzhenyu: +1

LukeWood:
Removed in favor of the final OD COCO metric PyMetric approach

PR #935, opened 2022-10-18
Add RandomCropAndResize to ImageNet classification training script

ianstenbit:
@LukeWood @tanzhenyu 

Training results with this change are pending (although I've verified that the training script still runs correctly). I'll train an EfficientNet with this change and compare results to the previous version.

### DIFF
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -131,7 +131,6 @@ train_ds = imagenet.load(
     split="train",
     tfrecord_path=FLAGS.imagenet_path,
     batch_size=BATCH_SIZE,
-    img_size=IMAGE_SIZE,
 )
 test_ds = imagenet.load(
     split="validation",
@@ -148,6 +147,11 @@ We define a set of augmentation layers and then apply them to our input dataset.
 
 
 AUGMENT_LAYERS = [
+    keras_cv.layers.RandomCropAndResize(
+        target_size=IMAGE_SIZE,
+        crop_area_factor=(0.8, 1),
+        aspect_ratio_factor=(3 / 4, 4 / 3),
+    ),
     keras_cv.layers.RandomFlip(mode="horizontal"),
     keras_cv.layers.RandAugment(value_range=(0, 255), magnitude=0.3),
     keras_cv.layers.CutMix(),
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -130,6 +130,7 @@ information about preparing this dataset at keras_cv/datasets/imagenet/README.md
 train_ds = imagenet.load(
     split="train",
     tfrecord_path=FLAGS.imagenet_path,
+    shuffle_buffer=BATCH_SIZE*2,
 )
 test_ds = imagenet.load(
     split="validation",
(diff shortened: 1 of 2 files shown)


tanzhenyu:
To clarify -- we only use this if it is not worse than existing ResNet50?

ianstenbit:
> To clarify -- we only use this if it is not worse than existing ResNet50?

Yeah I figured I'd do a training run against a model for which we already have weights and verify that this performs better (I figured I'd do an EfficientNet since it's quick to train)

tanzhenyu:
> > To clarify -- we only use this if it is not worse than existing ResNet50?
> 
> Yeah I figured I'd do a training run against a model for which we already have weights and verify that this performs better (I figured I'd do an EfficientNet since it's quick to train)

Given this has a broader impact, let's make sure it also runs well on ResNet?

ianstenbit:
> Given this has a broader impact, let's make sure it also runs well on ResNet?

SGTM! I'll kick off a ResNet50V2 training run

ianstenbit:
> > Given this has a broader impact, let's make sure it also runs well on ResNet?
> 
> SGTM! I'll kick off a ResNet50V2 training run

The training run isn't done yet, but it's currently outperforming our previous ResNet50V2 runs (72% top-1 accuracy after 78 epochs and it still hasn't Reduced Lr on Plateau yet)

So I think it's fair to say this is working well. I think we can merge the change, and once the ResNet run finishes I will upload our new best weights

tanzhenyu:
> > > Given this has a broader impact, let's make sure it also runs well on ResNet?
> > 
> > 
> > SGTM! I'll kick off a ResNet50V2 training run
> 
> The training run isn't done yet, but it's currently outperforming our previous ResNet50V2 runs (72% top-1 accuracy after 78 epochs and it still hasn't Reduced Lr on Plateau yet)
> 
> So I think it's fair to say this is working well. I think we can merge the change, and once the ResNet run finishes I will upload our new best weights

Isn't the ResNet50V2 at 76% after 90 epochs?

ianstenbit:
> Isn't the ResNet50V2 at 76% after 90 epochs?

Our best run was 75.5% after 168 epochs. It didn't pass 71% until after its first LR reduction

LukeWood (approved):

tanzhenyu:
> > Isn't the ResNet50V2 at 76% after 90 epochs?
> 
> Our best run was 75.5% after 168 epochs. It didn't pass 71% until after its first LR reduction

Oh...that doesn't seem right at all, I'd assume it should pass 71% at 60 epochs

ianstenbit:
> Oh...that doesn't seem right at all, I'd assume it should pass 71% at 60 epochs

In our previous run it got quite close to that: https://tensorboard.dev/experiment/TQ5r1EhXS4SDDagBD84rgA/#scalars
This is likely due to using a high LR with DecreaseLROnPlateau instead of a decaying LR.

Once I am able to test #908 we will include the warmed-up Cosine decay which may improve the speed of convergence (if not the actual final scores)

ianstenbit:
The new ResNet50V2 training run with this change has now surpassed our current weights' validation accuracy by 0.5% (and the training is still running).

Any objections to my merging this @tanzhenyu ?

tanzhenyu (approved):
Thanks!

PR #938, opened 2022-10-18
removing the usage of len to allow saving models in the saved_model format

gustheman:
this len forbids the framework from saving these models in the saved_model format replacing with the shape is what it's recommended

# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood


LukeWood:
Ideally we assert on this a bit, but merging as this layer is an internal API only.  The true core keras version will handle input validation.

tanzhenyu:
This is merged from another PR. Thanks for bringing this up!

PR #941, opened 2022-10-18
Introduce testing decorators

LukeWood:
These decorators allow you to skip tests if they require lots of resources, time, or custom C++ ops.

They're pretty trivial, and just make tests opt-in.

### DIFF
--- a/keras_cv/testing/utils.py
+++ b/keras_cv/testing/utils.py
@@ -1,0 +1,30 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


keras_cv/testing/utils.py
```diff
@@ -0,0 +1,35 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+
+
+def integration(test_fn):
```
ianstenbit, line 17: Do we have a use for the integration annotation? If not, maybe we should make this PR just add requires_custom_ops for now?

Also, FWIW, it seems like we could make a meta function here which creates a decorator with a given tag (e.g. `requires_custom_ops = build_opt_in_test_decorator("TEST_CUSTOM_OPS")`)

ianstenbit:
@LukeWood do you still want to move forward with this concept?

jbischof:
KerasNLP already has this functionality set up. We may want to follow that pattern: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/conftest.py

LukeWood:
> KerasNLP already has this functionality set up. We may want to follow that pattern: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/conftest.py

yeah this looks like a better setup than what we have with decorators.

LukeWood:
closed as obselete

Issue #807, opened 2022-09-16
Add shortlist of strings for support in RetinaNet compile() method.


LukeWood:
basically, `box_loss="l1"` `classification_loss="focal"` etc.

martin-gorner:
and also
loss = "default"
OR
box_loss="default" classification_loss="default" 
as decided in [sync meeting](https://groups.google.com/a/google.com/g/keras-cv/c/uO8zqo14xVI)

PR #942, opened 2022-10-18
Update the RetinaNet example

LukeWood:
Here are the final COCO metric results of this pipeline as evaluated by both KerasCV & MSCOCO:

Screen Shot 2022-10-18 at 12 06 23 PM

Here are sample detections:
download-1

closes: https://github.com/keras-team/keras-cv/issues/807


bhack:
It is hard to understand the diff as you moved the file. What kind of changes we had?

LukeWood:
> It is hard to understand the diff as you moved the file. What kind of changes we had?

migrated input pipeline away from the KPLs to @tanzhenyu 's from RCNN

LukeWood:
@tanzhenyu

tanzhenyu:
> It is hard to understand the diff as you moved the file. What kind of changes we had?

Maybe we could use `git mv files` so it's easier to check what happened exactly?

LukeWood:
> > It is hard to understand the diff as you moved the file. What kind of changes we had?
> 
> Maybe we could use `git mv files` so it's easier to check what happened exactly?

hmm - I keep trying to do this but I can't get it to work.

tanzhenyu:
> > > It is hard to understand the diff as you moved the file. What kind of changes we had?
> > 
> > 
> > Maybe we could use `git mv files` so it's easier to check what happened exactly?
> 
> hmm - I keep trying to do this but I can't get it to work.

You might need to git reset in order to do so. If this is too complicated, it's fine not to do it

LukeWood:
@tanzhenyu I also updated the loss strings FYI

LukeWood:
> > > > It is hard to understand the diff as you moved the file. What kind of changes we had?
> > > 
> > > 
> > > Maybe we could use `git mv files` so it's easier to check what happened exactly?
> > 
> > 
> > hmm - I keep trying to do this but I can't get it to work.
> 
> You might need to git reset in order to do so. If this is too complicated, it's fine not to do it

yeah, sorry about that - its basically an entirely new script at this point though.

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,567 @@
...
+)
+eval_ds = eval_ds.apply(
+    tf.data.experimental.dense_to_ragged_batch(8, drop_remainder=True)
+)
+
+
+def visualize_dataset(dataset, bounding_box_format, path=None):
+    example = next(iter(dataset))
+    images, boxes = example["images"], example["bounding_boxes"]
+    visualization.plot_bounding_box_gallery(
```
tanzhenyu, line 253: Is this part of keras_cv?

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,567 @@
...
+
+We'll use the KerasCV API to construct a RetinaNet model.  In this tutorial we use
+a pretrained ResNet50 backbone using weights.  In order to perform fine-tuning, we
+freeze the backbone before training.  When `include_rescaling=True` is set, inputs to
+the model are expected to be in the range `[0, 255]`.
+"""
+
+model = keras_cv.models.RetinaNet(
+    # number of classes to be used in box classification
+    classes=20,
```
tanzhenyu, line 301: len(class_ids)?

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,567 @@
...
+callbacks = [
+    keras.callbacks.TensorBoard(log_dir="logs"),
+    keras.callbacks.ReduceLROnPlateau(patience=5),
+    keras.callbacks.ModelCheckpoint(CHECKPOINT_PATH, save_weights_only=True),
+]
+
+history = model.fit(
+    train_ds,
+    validation_data=eval_ds.take(10),
+    epochs=100,
```
tanzhenyu, line 353: 100 epochs seem too much?

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,567 @@
...
+    font_scale=1,
+    class_mapping=class_mapping,
+    path="inference.png",
+)
+
+"""
+## Evaluation
+"""
+
+coco_suite = [
```
tanzhenyu, line 418: why not just model.fit(train_ds, validation_data=eval_ds)?

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -116,7 +116,6 @@ def __init__(
         classification_head=None,
         box_head=None,
         evaluate_train_time_metrics=False,
-        name="RetinaNet",
```
tanzhenyu, line 119: why removing the name?

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -506,6 +503,36 @@ def _parse_backbone(backbone, include_rescaling, backbone_weights):
...
+def _parse_box_loss(loss):
+    if not isinstance(loss, str):
+        # support arbitrary callables
+        return loss
+
+    # case insensitive comparison
+    if loss.lower() == "smoothl1":
+        return keras_cv.losses.SmoothL1Loss(l1_cutoff=1.0, reduction="none")
+
+    raise ValueError(
```
tanzhenyu, line 515: any particular reason that this cannot support Huber loss?

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -506,6 +503,36 @@ def _parse_backbone(backbone, include_rescaling, backbone_weights):
...
+
+
+def _parse_classification_loss(loss):
+    if not isinstance(loss, str):
+        # support arbitrary callables
+        return loss
+
+    # case insensitive comparison
+    if loss.lower() == "focal":
+        return keras_cv.losses.FocalLoss(from_logits=True, reduction="none")
```
tanzhenyu, line 528: Do we check `from_logits` if the user pass a callable?

tanzhenyu (commented):
Thanks for the PR!

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,567 @@
+import tensorflow as tf
+import tensorflow_datasets as tfds
+from luketils import visualization
```
innat, line 3: Some may not recognize `luketils`.

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,567 @@
...
+        return {"images": image, "bounding_boxes": bounding_boxes}
+
+    return apply
+
+
+train_ds = train_ds.map(
+    proc_train_fn("xywh", image_size), num_parallel_calls=tf.data.AUTOTUNE
+)
+train_ds = train_ds.apply(
+    tf.data.experimental.dense_to_ragged_batch(8, drop_remainder=True)
```
innat, line 239: `BATCH_SIZE`

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,567 @@
...
+    # and also causes a massive performance hit.  It can, however be useful to produce
+    # train time metrics when debugging your model training pipeline.
+    evaluate_train_time_metrics=False,
+)
+# Fine-tuning a RetinaNet is as simple as setting backbone.trainable = False
+model.backbone.trainable = False
+
+metrics = [
+    keras_cv.metrics.COCOMeanAveragePrecision(
+        class_ids=range(21),
```
innat, line 325: `range(len(class_ids)+1)`

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,567 @@
...
+    return apply
+
+
+eval_ds = tfds.load("voc/2007", split="test", with_info=False, shuffle_files=True)
+eval_ds = eval_ds.map(
+    proc_eval_fn("xywh", [640, 640, 3]), num_parallel_calls=tf.data.AUTOTUNE
+)
+eval_ds = eval_ds.apply(
+    tf.data.experimental.dense_to_ragged_batch(8, drop_remainder=True)
+)
```
innat, line 563: `BATCH_SIZE`

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,567 @@
...
+eval_ds = eval_ds.prefetch(2)
+
```
innat, line 565: Any reason to pick 2?
`...eval_ds.prefetch(tf.data.AUTOTUNE)`

examples/models/object_detection/retina_net/pascal_voc/train.py
LukeWood: @tanzhenyu should we just add the visualization functions to keras_cv?  We can just copy paste them, but it requires OpenCV.   I originally had them here just to avoid copy-pasting them all over.

examples/models/object_detection/retina_net/pascal_voc/train.py
LukeWood: Should we add it?  My thought was to upstream it at some point - but I wasn't quite ready for a full API review.

examples/models/object_detection/retina_net/pascal_voc/train.py
LukeWood: Added early stopping callback

examples/models/object_detection/retina_net/pascal_voc/train.py
LukeWood: if you evaluate the coco metrics during training after each epoch it takes way too long to run.

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: Added back - wasn't sure if we wanted/needed default names.  No strog opinion

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: Sure, why not!  added.

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: No we don't, should we just accept loss objects?

examples/models/object_detection/retina_net/pascal_voc/train.py
tanzhenyu: Yeah, I think it's definitely fine to achieve the minimal here: only train_ds without a COCO evaluation or visualizaiton. In the end we should have those, but given it's not fully flushed-out yet, better to exclude it for this PR.

keras_cv/models/object_detection/retina_net/retina_net.py
tanzhenyu: I think loss objects was added to Keras *after* loss callables, with intention to be the default, so yeah only accepting objects is fine with me

examples/models/object_detection/retina_net/pascal_voc/train.py
tanzhenyu: Maybe we should have some canonical training workflow like:
```
model.compile(..., metrics=[cocomAP_suite])
model.fit(train_ds, validation_data=eval_ds, epochs=?)
model.evaluate(eval_ds)
```
Though with cocomAP accuracy still under investigation, it's definitely ok not to include them for now. No strong opinion here.

examples/models/object_detection/retina_net/pascal_voc/train.py
tanzhenyu: Sounds good. No strong opinion whether you want to follow the paper strictly by epochs, or early stopping. (It's not easy to set the `patience` part for those callbacks though)

LukeWood:
@tanzhenyu does this look good to you?

examples/models/object_detection/retina_net/pascal_voc/train.py
```diff
@@ -0,0 +1,464 @@
...
+    classification_loss=keras_cv.losses.FocalLoss(from_logits=True, reduction="none"),
+    box_loss=keras_cv.losses.SmoothL1Loss(l1_cutoff=1.0, reduction="none"),
+    optimizer=optimizer,
+    metrics=coco_suite,
+)
+
+model.load_weights(CHECKPOINT_PATH)
+
+
+def proc_eval_fn(bounding_box_format, target_size):
```
tanzhenyu, line 407: Consider refactoring this with `proc_train_fn`? I didn't really use this other than in the eval.py, so I copied a lot of code for convenience

tanzhenyu (approved):

examples/models/object_detection/retina_net/pascal_voc/train.py
LukeWood: I think most of this will go away with KPLs anyways - should we just focus our efforts there?

PR #943, opened 2022-10-18
Add pascal voc 2012 dataset with segmentation data.

qlzh727:
# What does this PR do?
Add a new data loader function for Pascal voc that contains the segmentation mask data (both class segmentation and instance segmentation.)

## Before submitting
      Pull Request section?
      to it if that's the case.

Unit test will be added soon.

### DIFF
--- a/keras_cv/datasets/pascal_voc/segmentation.py
+++ b/keras_cv/datasets/pascal_voc/segmentation.py
@@ -99,7 +99,7 @@ def _download_pascal_voc_2012(data_url, local_dir_path=None, override_extract=Fa
     if override_extract or not os.path.exists(
         os.path.join(data_directory, "VOCdevkit")
     ):
-        logging.info(f"Extract data into %s", data_directory)
+        logging.info("Extract data into %s", data_directory)
         with tarfile.open(data_file_path) as f:
             f.extractall(data_directory)
     return os.path.join(data_directory, "VOCdevkit", "VOC2012")
--- a/keras_cv/datasets/pascal_voc/segmentation.py
+++ b/keras_cv/datasets/pascal_voc/segmentation.py
@@ -40,7 +40,6 @@ import xml
 
 import numpy as np
 import tensorflow as tf
-from PIL import Image
 
 DATA_URL = "http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar"
 
@@ -71,6 +70,41 @@ CLASSES = [
 ]
 # This is used to map between string class to index.
 CLASS_TO_INDEX = {name: index for index, name in enumerate(CLASSES)}
+
+# For the mask data in the PNG file, the encoded raw pixel value need be to converted
+# to the proper class index. In the following map, [0, 0, 0] will be convert to 0, and
+# [128, 0, 0] will be conveted to 1, so on so forth. Also note that the mask class is 1
+# base since class 0 is reserved for the background. The [128, 0, 0] (class 1) is mapped
+# to `aeroplane`.
+VOC_PNG_COLOR_VALUE = [
+    [0, 0, 0],
+    [128, 0, 0],
+    [0, 128, 0],
+    [128, 128, 0],
+    [0, 0, 128],
+    [128, 0, 128],
+    [0, 128, 128],
+    [128, 128, 128],
+    [64, 0, 0],
+    [192, 0, 0],
+    [64, 128, 0],
+    [192, 128, 0],
+    [64, 0, 128],
+    [192, 0, 128],
+    [64, 128, 128],
+    [192, 128, 128],
+    [0, 64, 0],
+    [128, 64, 0],
+    [0, 192, 0],
+    [128, 192, 0],
+    [0, 64, 128],
+]
+VOC_PNG_COLOR_MAPPING = [0] * (256**3)
+for i, colormap in enumerate(VOC_PNG_COLOR_VALUE):
+    VOC_PNG_COLOR_MAPPING[(colormap[0] * 256 + colormap[1]) * 256 + colormap[2]] = i
+# There is a special mapping with [224, 224, 192] -> 255
+VOC_PNG_COLOR_MAPPING[224 * 256 * 256 + 224 * 256 + 192] = 255
+VOC_PNG_COLOR_MAPPING = tf.constant(VOC_PNG_COLOR_MAPPING)
 
 
 def _download_pascal_voc_2012(data_url, local_dir_path=None, override_extract=False):
@@ -155,38 +189,38 @@ def _get_image_ids(data_dir, split):
         return image_ids
 
 
-def _convert_pascal_voc_segmentation_mask_encoding(
-    segmentation_file_path, use_cache=True
-):
-    """Convert the original segmentation PNG file to proper encoding for tf.io API.
-
-    The original PNG was in a 2D (without channel dimension), which will get a wrong value
-    when directly read by tf.io API. In this function, the original image will be first
-    read with PIL.image API, expand the last dimention and save back by TF API. This will
-    ensure the converted image can be properly consumed by TF API.
-
-    If a file with `xxx_converted.png` already exists, and `use_cache` is True, the
-    conversion will be skipped.
-
-    Args:
-        segmentation_file_path: string, the original PNG file path from the dataset.
-        override: boolean, whether to use the existing cached file result.
-    Returns:
-        the file path for converted PNG file, it will be suffixed with '_converted'.
-    """
-    dir_and_file_name, _ = os.path.splitext(segmentation_file_path)
-    updated_file_path = dir_and_file_name + "_updated.png"
-    if os.path.exists(updated_file_path) and use_cache:
-        return updated_file_path
-
-    with tf.io.gfile.GFile(segmentation_file_path, "rb") as f:
-        mask = Image.open(f)
-
-    original_mask = np.expand_dims(np.asarray(mask), axis=-1)
-    # # Write the mask to PNG via the TF io API, so that it can be read properly.
-    encoded_png = tf.io.encode_png(original_mask)
-    tf.io.write_file(updated_file_path, encoded_png)
-    return updated_file_path
+# def _convert_pascal_voc_segmentation_mask_encoding(
+#     segmentation_file_path, use_cache=True
+# ):
+#     """Convert the original segmentation PNG file to proper encoding for tf.io API.
+#
+#     The original PNG was in a 2D (without channel dimension), which will get a wrong value
+#     when directly read by tf.io API. In this function, the original image will be first
+#     read with PIL.image API, expand the last dimention and save back by TF API. This will
+#     ensure the converted image can be properly consumed by TF API.
+#
+#     If a file with `xxx_converted.png` already exists, and `use_cache` is True, the
+#     conversion will be skipped.
+#
+#     Args:
+#         segmentation_file_path: string, the original PNG file path from the dataset.
+#         override: boolean, whether to use the existing cached file result.
+#     Returns:
+#         the file path for converted PNG file, it will be suffixed with '_converted'.
+#     """
+#     dir_and_file_name, _ = os.path.splitext(segmentation_file_path)
+#     updated_file_path = dir_and_file_name + "_updated.png"
+#     if os.path.exists(updated_file_path) and use_cache:
+#         return updated_file_path
+#
+#     with tf.io.gfile.GFile(segmentation_file_path, "rb") as f:
+#         mask = Image.open(f)
+#
+#     original_mask = np.expand_dims(np.asarray(mask), axis=-1)
+#     # # Write the mask to PNG via the TF io API, so that it can be read properly.
+#     encoded_png = tf.io.encode_png(original_mask)
+#     tf.io.write_file(updated_file_path, encoded_png)
+#     return updated_file_path
 
 
 def _parse_single_image(image_file_path):
@@ -196,14 +230,8 @@ def _parse_single_image(image_file_path):
     class_segmentation_file_path = os.path.join(
         data_dir, "SegmentationClass", image_id + ".png"
     )
-    class_segmentation_file_path = _convert_pascal_voc_segmentation_mask_encoding(
-        class_segmentation_file_path
-    )
     object_segmentation_file_path = os.path.join(
         data_dir, "SegmentationObject", image_id + ".png"
-    )
-    object_segmentation_file_path = _convert_pascal_voc_segmentation_mask_encoding(
-        object_segmentation_file_path
     )
     annotation_file_path = os.path.join(data_dir, "Annotations", image_id + ".xml")
     image_annotations = _parse_annotation_data(annotation_file_path)
@@ -255,6 +283,16 @@ def _build_metadata(data_dir, image_ids):
     return result
 
 
+def _decode_png_mask(mask):
+    """Decode the raw PNG image and convert it to 2D tensor with probably class."""
+    # Cast the mask to int32 since the original uint8 will overflow when multiple with 256
+    mask = tf.cast(mask, tf.int32)
+    mask = mask[:, :, 0] * 256 * 256 + mask[:, :, 1] * 256 + mask[:, :, 2]
+    mask = tf.expand_dims(tf.gather(VOC_PNG_COLOR_MAPPING, mask), -1)
+    mask = tf.cast(mask, tf.uint8)
+    return mask
+
+
 def _load_images(example):
     image_file_path = example.pop("image/file_path")
     segmentation_class_file_path = example.pop("segmentation/class/file_path")
@@ -264,9 +302,11 @@ def _load_images(example):
 
     segmentation_class_mask = tf.io.read_file(segmentation_class_file_path)
     segmentation_class_mask = tf.image.decode_png(segmentation_class_mask)
+    segmentation_class_mask = _decode_png_mask(segmentation_class_mask)
 
     segmentation_object_mask = tf.io.read_file(segmentation_object_file_path)
     segmentation_object_mask = tf.image.decode_png(segmentation_object_mask)
+    segmentation_object_mask = _decode_png_mask(segmentation_object_mask)
 
     example.update(
         {
--- a/keras_cv/datasets/pascal_voc/segmentation.py
+++ b/keras_cv/datasets/pascal_voc/segmentation.py
@@ -18,7 +18,9 @@ The image classification and object detection (bounding box) data is covered by 
 TF datasets in https://www.tensorflow.org/datasets/catalog/voc. The segmentation data (
 both class segmentation and instance segmentation) are included in the VOC 2012, but not
 offered by TF-DS yet. This module is trying to fill this gap while TFDS team can
-address this feature (b/252870855 and https://github.com/tensorflow/datasets/issues/27).
+address this feature (b/252870855, https://github.com/tensorflow/datasets/issues/27 and
+https://github.com/tensorflow/datasets/pull/1198).
+
 The schema design is similar to the existing design of TFDS, but trimmed to fit the need
 of Keras CV models.
 
@@ -42,6 +44,9 @@ from PIL import Image
 
 DATA_URL = "http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar"
 
+# Note that this list doesn't contain the background class. In the classification use
+# case, the label is 0 based (aeroplane -> 0), whereas in segmentation use case, the 0 is
+# reserved for background, so aeroplane maps to 1.
 CLASSES = [
     "aeroplane",
     "bicycle",
@@ -78,8 +83,13 @@ def _download_pascal_voc_2012(data_url, local_dir_path=None, override_extract=Fa
         the path to the folder of extracted Pascal VOC data.
     """
     if not local_dir_path:
-        local_dir_path = "pascal_voc_2012/data.tar"
-    data_file_path = tf.keras.utils.get_file(fname=local_dir_path, origin=data_url)
+        fname = "pascal_voc_2012/data.tar"
+    else:
+        # Make sure the directory exists
+        if not os.path.exists(local_dir_path):
+            os.makedirs(local_dir_path, exist_ok=True)
+        fname = os.path.join(local_dir_path, "data.tar")
+    data_file_path = tf.keras.utils.get_file(fname=fname, origin=data_url)
     logging.info("Received data file from %s", data_file_path)
     # Extra the data into the same directory as the tar file.
     data_directory = os.path.dirname(data_file_path)
@@ -176,9 +186,10 @@ def _convert_pascal_voc_segmentation_mask_encoding(
     # # Write the mask to PNG via the TF io API, so that it can be read properly.
     encoded_png = tf.io.encode_png(original_mask)
     tf.io.write_file(updated_file_path, encoded_png)
+    return updated_file_path
 
 
-def parse_single_image(image_file_path):
+def _parse_single_image(image_file_path):
     data_dir, image_file_name = os.path.split(image_file_path)
     data_dir = os.path.normpath(os.path.join(data_dir, os.path.pardir))
     image_id, _ = os.path.splitext(image_file_name)
@@ -215,8 +226,9 @@ def _build_metadata(data_dir, image_ids):
     image_file_paths = [
         os.path.join(data_dir, "JPEGImages", i + ".jpg") for i in image_ids
     ]
-    p = multiprocessing.Pool(10)
-    metadata = p.map(parse_single_image, image_file_paths)
+    pool_size = 10 if len(image_ids) > 10 else len(image_ids)
+    with multiprocessing.Pool(pool_size) as p:
+        metadata = p.map(_parse_single_image, image_file_paths)
 
     # Transpose the metadata which convert from list of dict to dict of list.
     keys = [
--- a/keras_cv/datasets/pascal_voc/segmentation.py
+++ b/keras_cv/datasets/pascal_voc/segmentation.py
@@ -37,14 +37,10 @@ import tarfile
 import xml
 
 import numpy as np
+import tensorflow as tf
 from PIL import Image
 
-import tensorflow as tf
-
-logging.info = print
-
-
-DATA_URL = 'http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar'
+DATA_URL = "http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar"
 
 CLASSES = [
     "aeroplane",
@@ -68,6 +64,9 @@ CLASSES = [
     "train",
     "tvmonitor",
 ]
+# This is used to map between string class to index.
+CLASS_TO_INDEX = {name: index for index, name in enumerate(CLASSES)}
+
 
 def _download_pascal_voc_2012(data_url, local_dir_path=None, override_extract=False):
     """Fetch the original Pascal VOC 2012 from remote URL.
@@ -79,20 +78,21 @@ def _download_pascal_voc_2012(data_url, local_dir_path=None, override_extract=Fa
         the path to the folder of extracted Pascal VOC data.
     """
     if not local_dir_path:
-        local_dir_path = 'pascal_voc_2012/data.tar'
-    data_file_path = tf.keras.utils.get_file(
-        fname=local_dir_path, origin=data_url)
-    logging.info('Received data file from %s', data_file_path)
+        local_dir_path = "pascal_voc_2012/data.tar"
+    data_file_path = tf.keras.utils.get_file(fname=local_dir_path, origin=data_url)
+    logging.info("Received data file from %s", data_file_path)
     # Extra the data into the same directory as the tar file.
     data_directory = os.path.dirname(data_file_path)
     # Note that the extracted data will be located in a folder `VOCdevkit` (from tar).
     # If the folder is already there and `override_extract` is False, then we will skip
     # extracting the folder again.
-    if override_extract or not os.path.exists(os.path.join(data_directory, 'VOCdevkit')):
-        logging.info(f'Extract data into %s', data_directory)
+    if override_extract or not os.path.exists(
+        os.path.join(data_directory, "VOCdevkit")
+    ):
+        logging.info(f"Extract data into %s", data_directory)
         with tarfile.open(data_file_path) as f:
             f.extractall(data_directory)
-    return os.path.join(data_directory, 'VOCdevkit', 'VOC2012')
+    return os.path.join(data_directory, "VOCdevkit", "VOC2012")
 
 
 def _parse_annotation_data(annotation_file_path):
@@ -111,42 +111,43 @@ def _parse_annotation_data(annotation_file_path):
         objects = []
         for obj in root.findall("object"):
             # Get object's label name.
-            label = obj.find("name").text.lower()
+            label = CLASS_TO_INDEX[obj.find("name").text.lower()]
             # Get objects' pose name.
             pose = obj.find("pose").text.lower()
-            is_truncated = (obj.find("truncated").text == "1")
-            is_difficult = (obj.find("difficult").text == "1")
+            is_truncated = obj.find("truncated").text == "1"
+            is_difficult = obj.find("difficult").text == "1"
             bndbox = obj.find("bndbox")
             xmax = int(bndbox.find("xmax").text)
             xmin = int(bndbox.find("xmin").text)
             ymax = int(bndbox.find("ymax").text)
             ymin = int(bndbox.find("ymin").text)
-            objects.append({
-                "label": label,
-                "pose": pose,
-                "bbox": [ymin, xmin, ymax, xmax],
-                "is_truncated": is_truncated,
-                "is_difficult": is_difficult,
-            })
+            objects.append(
+                {
+                    "label": label,
+                    "pose": pose,
+                    "bbox": [ymin, xmin, ymax, xmax],
+                    "is_truncated": is_truncated,
+                    "is_difficult": is_difficult,
+                }
+            )
 
-        return {
-            "width": width,
-            "height": height,
-            "objects": objects
-        }
+        return {"width": width, "height": height, "objects": objects}
 
 
 def _get_image_ids(data_dir, split):
-    data_file_mapping = {
-        'train': 'train.txt', 'eval': 'val.txt', None: 'trainval.txt'
-    }
-    with tf.io.gfile.GFile(os.path.join(
-            data_dir, 'ImageSets', 'Segmentation', data_file_mapping[split]), 'r') as f:
+    data_file_mapping = {"train": "train.txt", "eval": "val.txt", None: "trainval.txt"}
+    with tf.io.gfile.GFile(
+        os.path.join(data_dir, "ImageSets", "Segmentation", data_file_mapping[split]),
+        "r",
+    ) as f:
         image_ids = f.read().splitlines()
-        logging.info(f'Received {len(image_ids)} images for {split} dataset.')
+        logging.info(f"Received {len(image_ids)} images for {split} dataset.")
+        return image_ids
 
 
-def _convert_pascal_voc_segmentation_mask_encoding(segmentation_file_path, use_cache=True):
+def _convert_pascal_voc_segmentation_mask_encoding(
+    segmentation_file_path, use_cache=True
+):
     """Convert the original segmentation PNG file to proper encoding for tf.io API.
 
     The original PNG was in a 2D (without channel dimension), which will get a wrong value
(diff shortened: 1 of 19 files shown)


keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,320 @@
...
+    conversion will be skipped.
+
+    Args:
+        segmentation_file_path: string, the original PNG file path from the dataset.
+        override: boolean, whether to use the existing cached file result.
+    Returns:
+        the file path for converted PNG file, it will be suffixed with '_converted'.
+    """
+    dir_and_file_name, _ = os.path.splitext(segmentation_file_path)
+    updated_file_path = dir_and_file_name + "_updated.png"
```
tanzhenyu, line 168: Some high-level question -- it seems this is currently being used to directly load from raw data. In this case, do we still need to write back to a "_updated.png" file, or can we simply pass the expanded image to the data pipeline?
On the other hand, once we support "produce tfrecords", this converted data will be directly written to record file, I believe?

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,320 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Data loader for Pascal VOC 2012 segmentation dataset.
+
+The image classification and object detection (bounding box) data is covered by existing
+TF datasets in https://www.tensorflow.org/datasets/catalog/voc. The segmentation data (
+both class segmentation and instance segmentation) are included in the VOC 2012, but not
+offered by TF-DS yet. This module is trying to fill this gap while TFDS team can
+address this feature (b/252870855 and https://github.com/tensorflow/datasets/issues/27).
```
bhack, line 21: And also the contribution we have lost with https://github.com/tensorflow/datasets/pull/1198

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,320 @@
...
+):
+    """Load the Pacal VOC 2012 dataset.
+
+    This function will download the data tar file from remote if needed, and untar to
+    the local `data_dir`, and build dataset from it.
+
+    Args:
+        split: string, can be 'train', 'eval', or None. When None, both train and eval data
+            will be loaded. Default to `train`
+        data_dir: string, local directory path for the loaded data. This will be used to
```
tanzhenyu, line 301: a generic comment unrelated to this PR: it'd be nice if Keras support environment variables so the data can be populated to user defined directory

tanzhenyu (commented):
Thanks for the PR!

keras_cv/datasets/pascal_voc/segmentation.py
qlzh727: Sure.

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,320 @@
...
+import tarfile
+import xml
+
+import numpy as np
+import tensorflow as tf
+from PIL import Image
+
+DATA_URL = "http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar"
+
+CLASSES = [
```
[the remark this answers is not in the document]
qlzh727: Done.

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,320 @@
...
+    # Note that the extracted data will be located in a folder `VOCdevkit` (from tar).
+    # If the folder is already there and `override_extract` is False, then we will skip
+    # extracting the folder again.
+    if override_extract or not os.path.exists(
+        os.path.join(data_directory, "VOCdevkit")
+    ):
+        logging.info(f"Extract data into %s", data_directory)
+        with tarfile.open(data_file_path) as f:
+            f.extractall(data_directory)
+    return os.path.join(data_directory, "VOCdevkit", "VOC2012")
```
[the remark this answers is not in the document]
qlzh727: I didn't download the tar for the voc 2007, but I would expect some of the annotation/metadata is changed.

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,320 @@
...
+            bndbox = obj.find("bndbox")
+            xmax = int(bndbox.find("xmax").text)
+            xmin = int(bndbox.find("xmin").text)
+            ymax = int(bndbox.find("ymax").text)
+            ymin = int(bndbox.find("ymin").text)
+            objects.append(
+                {
+                    "label": label,
+                    "pose": pose,
+                    "bbox": [ymin, xmin, ymax, xmax],
```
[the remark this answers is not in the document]
qlzh727: we could. The current PR is just trying to reuse the existing feature mapping from tfds. My worry is that when user shift from tfds to our model, the change of key will force them to update the data input pipeline config.

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+def _convert_pascal_voc_segmentation_mask_encoding(
+    segmentation_file_path, use_cache=True
+):
+    """Convert the original segmentation PNG file to proper encoding for tf.io API.
+
+    The original PNG was in a 2D (without channel dimension), which will get a wrong value
+    when directly read by tf.io API. In this function, the original image will be first
+    read with PIL.image API, expand the last dimention and save back by TF API. This will
```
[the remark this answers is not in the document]
qlzh727: Done. Thanks for catching this.

keras_cv/datasets/pascal_voc/segmentation.py
qlzh727: If we don't write back to "x_updated.png", then we need to include some quite unreadable logic to convert the data format. This also becomes a cost per invocation of load(), whereas the x_updated.png is a one time cost.

When we change to tfrecord, it will be generated from x_updated.png as well, and the x_updated.png can be removed afterwards.

keras_cv/datasets/pascal_voc/segmentation.py
qlzh727: the data_dir is the designed for this. If you want to have a path that contains any reference to env vars, it should be able to resolve correctly.

keras_cv/datasets/pascal_voc/segmentation.py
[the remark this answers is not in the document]
tanzhenyu: That is a good point. Hmm...it's a tradeoff then,
1) if we're more consistent with tfds, then all our training script will need to have a parsing function that maps from ["objects"]["bbox"] to ["boxes"]
2) if we deviate from it, as you said, users will need to change the code.

keras_cv/datasets/pascal_voc/segmentation.py
bhack: The dataset url is temp unreachable. Have you tried to open it explicit channels with:
 ```python 
 tf.io.decode_png(
    contents,
    channels=1,
    dtype=tf.dtypes.uint8
)
```

File: keras_cv/datasets/pascal_voc/segmentation.py [hunk 8]
```diff
@@ -0,0 +1,320 @@
...
+    result["labels"] = sorted(labels)
+    return result
+
+
+def _build_metadata(data_dir, image_ids):
+    # Parallel process all the images.
+    image_file_paths = [
+        os.path.join(data_dir, "JPEGImages", i + ".jpg") for i in image_ids
+    ]
+    p = multiprocessing.Pool(10)
```
ianstenbit, line 218: Is there any way to use a dynamically sized threadpool here? Using a static size seems a bit machine-specific, but I suppose it's probably not really a problem?

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,320 @@
...
+    split="train",
+    data_dir=None,
+):
+    """Load the Pacal VOC 2012 dataset.
+
+    This function will download the data tar file from remote if needed, and untar to
+    the local `data_dir`, and build dataset from it.
+
+    Args:
+        split: string, can be 'train', 'eval', or None. When None, both train and eval data
```
ianstenbit, line 299: TFDS defaults to `None` for split, and IMO that experience is a bit more intuitive. It feels weird to have to pass split=None to get both splits

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,320 @@
...
+    dir_and_file_name, _ = os.path.splitext(segmentation_file_path)
+    updated_file_path = dir_and_file_name + "_updated.png"
+    if os.path.exists(updated_file_path) and use_cache:
+        return updated_file_path
+
+    with tf.io.gfile.GFile(segmentation_file_path, "rb") as f:
+        mask = Image.open(f)
+
+    original_mask = np.expand_dims(np.asarray(mask), axis=-1)
+    # # Write the mask to PNG via the TF io API, so that it can be read properly.
```
ianstenbit, line 176: nit: double commented

ianstenbit (approved):

keras_cv/datasets/pascal_voc/segmentation.py
qlzh727: Added in the latest commit.

keras_cv/datasets/pascal_voc/segmentation.py
qlzh727: To tanzhenyu, I think your confusion point is why can't we just convert the PNG and make it part of the dataset without saving it. The tricky part here is the dataset building logic. When creating the dataset from metadata (a python dict), it only contains the file path (not the raw png data, otherwise it will quickly OOM). The map function for dataset will use tf.io ops to read the image, and convert them to float tensor. The loading logic could only rely on TF ops, not any PIL or pure python logic, otherwise dataset will complain and require us to use tf.pyfunction, and it will have a performance impact.

The conversion function in pure tf ops format is quite unreadable, see https://github.com/tensorflow/models/blob/master/research/deeplab/utils/get_dataset_colormap.py#L317-L331. It can be just replaced by a PIL load and convert, but the PIL can not happen in the dataset.map function.

To bhack: I had several tries before going to this solution. See following example. The 150 value is not correct, since the dataset only contains 20 classes. The 150 value should be mapped to 19 (train).

```
>>> content = tf.io.read_file("2011_000277.png")
>>> png = tf.io.decode_png(content, channels=1, dtype=tf.uint8)
>>> print(png)
tf.Tensor(
[[[150]
  [150]
  [150]
  ...
  [  0]
  [  0]
  [  0]]
......

```

keras_cv/datasets/pascal_voc/segmentation.py
qlzh727: Not a big problem so far, I can change it to read the local cpu count, but might be a bit overkill. 

Updated a bit to not use too many thread when the data entry is less than 10.

keras_cv/datasets/pascal_voc/segmentation.py
qlzh727: For the keras fit/eval workflow, I think default to train actually make more sense, since the fit() method doesn't need eval data, or the eval data are actually specified by a separate param (validation_data).

keras_cv/datasets/pascal_voc/segmentation.py
bhack: Can you share this image? The web site is still down.

Thanks.

bhack: Thanks I have found it also in a mirror.

Yes the function it was not reading the color mapping but it is quite simple. 

You need to read it 3 channel not 1 then this is the mapping on the 20 Voc classes:

```python

VOC_COLORMAP = [[0, 0, 0], [128, 0, 0], [0, 128, 0], [128, 128, 0],
                [0, 0, 128], [128, 0, 128], [0, 128, 128], [128, 128, 128],
                [64, 0, 0], [192, 0, 0], [64, 128, 0], [192, 128, 0],
                [64, 0, 128], [192, 0, 128], [64, 128, 128], [192, 128, 128],
                [0, 64, 0], [128, 64, 0], [0, 192, 0], [128, 192, 0],
                [0, 64, 128]]
```

```python
VOC_CLASSES = ['background', 'aeroplane', 'bicycle', 'bird', 'boat',
               'bottle', 'bus', 'car', 'cat', 'chair', 'cow',
               'diningtable', 'dog', 'horse', 'motorbike', 'person',
               'potted plant', 'sheep', 'sofa', 'train', 'tv/monitor']
```

So it is not 150 but with 3 channels:
```python
Shape (375, 500, 3)

tf.Tensor(
[[[128 192   0]
  [128 192   0]
  [128 192   0]
  ...
```

That is 19 (train).

bhack: (As we totally ignore colormapped imges libpng FLAG in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/lib/png/png_io.cc)

But as exposed above it is quite easy to remmap it (Pytorch example):
https://d2l.ai/chapter_computer-vision/semantic-segmentation-and-dataset.html

qlzh727: Note that to convert the raw tensor value from 
```
[[[128 192   0]
  [128 192   0]
  [128 192   0]
```
 to
```
[[19], [19], [19]]
```
it will need a lookup from the VOC_COLORMAP, which will probably need a https://www.tensorflow.org/api_docs/python/tf/lookup/StaticHashTable. It is computation heavy, and this will happen in the map function which is not a one time cost.

bhack: I don't know if it is the TF faster solution have you seen the link? 
Do you think that it is faster to pass twice in the filesystem?

qlzh727: The temp file write will only happen once and get cached. It also happens before creating the dataset and training, whereas the lookup/map will happen within the data input pipeline.

qlzh727: Thanks for the link https://d2l.ai/chapter_computer-vision/semantic-segmentation-and-dataset.html. Let me take a closer look.

bhack: As you like:

A) We could always use the TF lookup/map in the same section as Pillow and not use an addition dependency 
B) We could always patch to use the libpng colormap flag as in pillow https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/lib/png/png_io.cc (best effect on the ecosystem)
C) We could add pillow as a dependency and use its Image

qlzh727: Thanks. I agree that we can use the TF ops to do the convert rather than rely on the extra deps of Pillow.

The only concern I have is whether convert cost should happen in dataset.map or as one time pre-process step.

So far the speed I got from local temp file with Pillow is quite fast (~2 sec for the whole dataset with 10 threads). I don't have the number of the dataset.map() yet.

I think the final question is whether we prefer:
1. Temp file with pre-converted PNGs (either with PIL or TF ops)
2. dataset.map() with TF ops only (PIL doesn't work within map())

bhack: The best solution is to patch https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/lib/png/png_io.cc as at the end both PIL and tf.io rely on libpng. So just 1 single filesystem read and decode.

But I am realistic and I know that this level of ecosystem cooperation is quite "utopian" :smile_cat: 

Then IMHO for your 2) I have an extra question are we covered by XLA with the OPset/s that we want to use?

bhack:
@qlzh727 I've protected our `devcontainer` with a specific Github Action as we was always drifting when we introduce new dependencies or we introduce version change. So you need to add the dependency to the [Dockerfile](https://github.com/keras-team/keras-cv/blob/master/.devcontainer/Dockerfile#L9).

keras_cv/datasets/pascal_voc/segmentation.py
qlzh727: Good question for XLA. Let me write a TF function and have a try.

tanzhenyu: IIUC, 
1) `create_pascal_label_colormap` is from label index to colormap, this is used for visualization;
2) `voc_label_indices` is from colormap to label index, this is for converting input mask to label.
So all we need is have a tf version of `voc_label_indices` here to make it work?

keras_cv/datasets/pascal_voc/segmentation.py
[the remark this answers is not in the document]
tanzhenyu: Ok, maybe add a TODO here to add year info as argument?

keras_cv/datasets/pascal_voc/segmentation.py
qlzh727: Tested with another image, which somehow miss from the mapping. The pixel value I got from the raw png is 
[224 224 192], which doesn't map to any of the items above.

bhack: Is `[224 224 192]` -> `255` ?

As I see 255 with `Image.open(f)`

bhack: I think is 255. See https://stackoverflow.com/a/55246638

bhack: See also a TF implementation variant at https://stackoverflow.com/a/71413414 I don't know if you have a faster impl idea but this one is XLA covered.

qlzh727: Thanks. I think the all the ops will work with the XLA.

```
import os

import tensorflow as tf

LOCAL_DATA_PATH = '~/pascal_voc_2012/VOCdevkit/VOC2012/'

VOC_COLORMAP = [[0, 0, 0], [128, 0, 0], [0, 128, 0], [128, 128, 0],
                [0, 0, 128], [128, 0, 128], [0, 128, 128], [128, 128, 128],
                [64, 0, 0], [192, 0, 0], [64, 128, 0], [192, 128, 0],
                [64, 0, 128], [192, 0, 128], [64, 128, 128], [192, 128, 128],
                [0, 64, 0], [128, 64, 0], [0, 192, 0], [128, 192, 0],
                [0, 64, 128]]

def convert_pascal_voc_segmentation_mask_encoding(mask_file_path):
  color_map_to_label = [0] * (256 ** 3)
  for i, colormap in enumerate(VOC_COLORMAP):
    color_map_to_label[
      (colormap[0] * 256 + colormap[1]) * 256 + colormap[2]] = i
  # There is a special mapping with [224, 224, 192] -> 255
  color_map_to_label[224 * 256 * 256 + 224 * 256 + 192] = 255
  color_map_to_label = tf.constant(color_map_to_label)

  mask = tf.io.decode_png(tf.io.read_file(mask_file_path))
  mask = tf.cast(mask, tf.int32)
  mask = mask[:, :, 0] * 256 * 256 + mask[:, :, 1] * 256 + mask[:, :, 2]
  mask = tf.expand_dims(tf.gather(color_map_to_label, mask), -1)
  return mask

mask_file_path = os.path.expanduser(LOCAL_DATA_PATH) + 'SegmentationClass/2011_003255.png'
mask = convert_pascal_voc_segmentation_mask_encoding(mask_file_path)
print(mask)
```

qlzh727: The creation of color_map_to_label can be lifted out, and we don't have to create a constant for every step.

bhack: Do you think it is faster then https://stackoverflow.com/a/71413414 ?

bhack: NIT: if you don't want to manually fill the colormap constant [check this](https://github.com/tensorflow/models/blob/238922e98dd0e8254b5c0921b241a1f5a151782f/research/deeplab/utils/get_dataset_colormap.py#L317-L331)

bhack: Note that don't have a registered XLA kernel for `read_file` and `decode_png` so they need to be out of the XLA section. These are the optimized HLO graph of the two implementations (please note the related fusions blocks):
https://colab.research.google.com/gist/bhack/88ddd268735208d27664ca401c51f449/voc_hlo.ipynb

keras_cv/datasets/pascal_voc/segmentation.py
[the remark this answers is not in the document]
LukeWood: A personal thought would be to be consistent with TFDS in this instance in particular and include a parsing function in our training script.  Another benefit of having parsing functions is it increases visibility into the data forma for users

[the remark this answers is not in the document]
tanzhenyu: Scott up to you. My personal thought here is we should be more friendly to KerasCV models so each training script don't need to come with duplicated parsing

keras_cv/datasets/pascal_voc/segmentation.py
bhack: We can still use XLA in tf.data as you see will still gain some fusions.

> I think this decoding logic is fine,
We have two decoding logic if you check the cloab. We just evaluating the performance between these two alternatives.

qlzh727: Updated to use the tf.gather with constant tensor value lookup.

qlzh727:
> 

Ack, since we change the approach, the extra deps of PILLOW is not needed.

keras_cv/datasets/pascal_voc/segmentation.py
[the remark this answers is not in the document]
qlzh727: I will keep it consistent with the tfds fields, so that the preprocessing doesn't have to be changed when we switch between dataset.

keras_cv/datasets/pascal_voc/segmentation_test.py
```diff
@@ -0,0 +1,285 @@
...
+from absl import flags
+
+from keras_cv.datasets.pascal_voc import segmentation
+
+
+class PascalVocSegmentationDataTest(tf.test.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.tempdir = self.get_tempdir()
+        # Note that this will not work with bazel, can need to be rewrite into rely on
```
tanzhenyu, line 28: maybe reformat this sentence a little bit

qlzh727: Done.

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,337 @@
...
+    [0, 192, 0],
+    [128, 192, 0],
+    [0, 64, 128],
+]
+VOC_PNG_COLOR_MAPPING = [0] * (256**3)
+for i, colormap in enumerate(VOC_PNG_COLOR_VALUE):
+    VOC_PNG_COLOR_MAPPING[(colormap[0] * 256 + colormap[1]) * 256 + colormap[2]] = i
+# There is a special mapping with [224, 224, 192] -> 255
+VOC_PNG_COLOR_MAPPING[224 * 256 * 256 + 224 * 256 + 192] = 255
+VOC_PNG_COLOR_MAPPING = tf.constant(VOC_PNG_COLOR_MAPPING)
```
lucasdavid, line 106: This variable occupies 64 MB, regardless if the dataset is being used or not.
I think this should be a sparse tensor, if possible, or at least lazy loaded:
```py
VOC_PNG_COLOR_MAPPING = None

def voc_png_color_mapping():
  global VOC_PNG_COLOR_MAPPING
  
  if VOC_PNG_COLOR_MAPPING is None:
    # generate...

  return VOC_PNG_COLOR_MAPPING
...

... tf.gather(voc_png_color_mapping(), ...)
```

qlzh727: Good point. Change to lazy load to avoid unnecessary memory consumption when loading the python module.

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -0,0 +1,349 @@
...
+    for key in ["label", "pose", "bbox", "is_truncated", "is_difficult"]:
+        values = []
+        objects = [value["objects"] for value in metadata]
+        for object in objects:
+            values.append([o[key] for o in object])
+        result["objects/" + key] = values
+    return result
+
+
+def _decode_png_mask(mask):
```
bhack, line 262: Do we want to jit_compile this?

bhack: Too late..

qlzh727: Opps, submitted too quickly. I don't have the number here. Since usually the dataset load/map happens on CPU, the gain of jit_compile=True might not be very big.

Will run some performance benchmark as a follow up task.

qlzh727: Run with following snippet, and the result is like below:
```
============================================
py function result:
warmup_time: 0.016756296157836914 secs
total_time: 3.9043142795562744 secs
============================================
tf function result:
warmup_time: 0.1970975399017334 secs
total_time: 0.6691527366638184 secs
============================================
tf function with xla result:
warmup_time: 0.5804996490478516 secs
total_time: 0.43773317337036133 secs
```


```
import os
import time

import tensorflow as tf

LOCAL_DATA_PATH = '~/pascal_voc_2012/VOCdevkit/VOC2012/'

VOC_COLORMAP = [[0, 0, 0], [128, 0, 0], [0, 128, 0], [128, 128, 0],
                [0, 0, 128], [128, 0, 128], [0, 128, 128], [128, 128, 128],
                [64, 0, 0], [192, 0, 0], [64, 128, 0], [192, 128, 0],
                [64, 0, 128], [192, 0, 128], [64, 128, 128], [192, 128, 128],
                [0, 64, 0], [128, 64, 0], [0, 192, 0], [128, 192, 0],
                [0, 64, 128]]
color_map_to_label = [0] * (256 ** 3)
for i, colormap in enumerate(VOC_COLORMAP):
  color_map_to_label[
    (colormap[0] * 256 + colormap[1]) * 256 + colormap[2]] = i
# There is a special mapping with [224, 224, 192] -> 255
color_map_to_label[224 * 256 * 256 + 224 * 256 + 192] = 255
color_map_to_label = tf.constant(color_map_to_label)


def decode_with_python_function(mask_file_path):
  mask = tf.io.decode_png(tf.io.read_file(mask_file_path))
  mask = tf.cast(mask, tf.int32)
  mask = mask[:, :, 0] * 256 * 256 + mask[:, :, 1] * 256 + mask[:, :, 2]
  mask = tf.expand_dims(tf.gather(color_map_to_label, mask), -1)
  return mask

decode_with_jit_compile = tf.function(func=decode_with_python_function, jit_compile=True)
decode_with_tf_function = tf.function(func=decode_with_python_function)

mask_file_path = os.path.expanduser(LOCAL_DATA_PATH) + 'SegmentationClass/2011_003255.png'


def benchmark_test(fn, mask_file_path, iteration):
  start = time.time()
  # warmup
  fn(mask_file_path)
  warmup_time = time.time() - start
  
  start = time.time()
  for _ in range(iteration):
    fn(mask_file_path)
    
  end = time.time()
  total_time = end - start
  print(f"warmup_time: {warmup_time} secs")
  print(f"total_time: {total_time} secs")


print("============================================")
print("py function result:")
benchmark_test(decode_with_python_function, mask_file_path, 1000)

print("============================================")
print("tf function result:")
benchmark_test(decode_with_tf_function, mask_file_path, 1000)

print("============================================")
print("tf function with xla result:")
benchmark_test(decode_with_jit_compile, mask_file_path, 1000)

```

qlzh727: We can ignore the pure python function result, since dataset.map will compile the function into tf.function.

There is about 0.4 sec overhead for XLA compilation, and 1000 step result is 0.2 sec faster. Since 0.4 is just a one time cost, the more image we have, the more gain we will have for using xla.

bhack: Nice, I suppose that your function, as you in see we have others colormapping type for other segmentation datasets, converted as keras preprocessing layer, will probably let use to preprocess at the batch level on GPU also faster.

qlzh727: Also, did a comparison with https://stackoverflow.com/questions/50244982/read-image-in-tensorflow-with-indexed-color-value-for-semantic-segmentation-task/71413414#71413414, that solution is much more slower (90 sec for 1000 records), probably due to match of 1d tensor (eg [128, 0, 0]), instead of encoded value.

qlzh727: https://github.com/keras-team/keras-cv/pull/953 for adding the tf.function annotation.

lucasdavid: Encoding the three pixels with `r*256**2+g*256+b` really is a good idea, but i think your benchmark "cheats" by always using the same sample. That doesn't trigger retracing, while optimizing caching by always using the same data, hence favoring access operations.

I improved the benchmark and ran it multiple times, using the pixel encoding in both strategies, single and multiple samples, with and without `experimental_follow_type_hints`.

I believe my implementation is faster in most cases, while not requiring the allocation of a tensor of shape [256**3].

Benchmark code: https://gist.github.com/lucasdavid/c48d39399b0bbe167cc2f5c056beadf0
<details>
<summary>Benchmark logs</summary>

```shell
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt a --sampling first --data-dir workspace/datasets/voc/VOCdevkit/VOC2012
[INFO MESSAGES]
Benchmarking opt a sampling:first iterations:1000 follow-type-hints:False
py_fn | warmup:  0.0226 s | total:  4.1935 s
py_fn | warmup:  0.1803 s | total:  0.6522 s
py_fn | warmup:  0.5773 s | total:  0.4188 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt b --sampling first --data-dir workspace/datasets/voc/VOCdevkit/VOC2012
[INFO MESSAGES]
Benchmarking opt b sampling:first iterations:1000 follow-type-hints:False
py_fn | warmup:  0.0238 s | total:  4.8490 s
py_fn | warmup:  0.2218 s | total:  0.2861 s
py_fn | warmup:  0.0594 s | total:  0.2358 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt b --sampling first --data-dir workspace/datasets/voc/VOCdevkit/VOC2012
[INFO MESSAGES]
Benchmarking opt b sampling:first iterations:1000 follow-type-hints:False
py_fn | warmup:  0.0229 s | total:  4.8222 s
py_fn | warmup:  0.2196 s | total:  0.2447 s
py_fn | warmup:  0.0623 s | total:  0.2472 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt a --sampling first --data-dir workspace/datasets/voc/VOCdevkit/VOC2012
[INFO MESSAGES]
Benchmarking opt a sampling:first iterations:1000 follow-type-hints:False
py_fn | warmup:  0.0209 s | total:  4.1461 s
py_fn | warmup:  0.1806 s | total:  0.6624 s
py_fn | warmup:  0.5778 s | total:  0.4153 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt a --sampling first --data-dir workspace/datasets/voc/VOCdevkit/VOC2012 --experimental-follow-type-hints
[INFO MESSAGES]
Benchmarking opt a sampling:first iterations:1000 follow-type-hints:True
py_fn | warmup:  0.0223 s | total:  4.1025 s
py_fn | warmup:  0.1827 s | total:  4.1046 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt b --sampling first --data-dir workspace/datasets/voc/VOCdevkit/VOC2012 --experimental-follow-type-hints
[INFO MESSAGES]
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
Benchmarking opt b sampling:first iterations:1000 follow-type-hints:True
py_fn | warmup:  0.0229 s | total:  4.8926 s
py_fn | warmup:  0.2413 s | total:  3.6635 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt a --sampling first --data-dir workspace/datasets/voc/VOCdevkit/VOC2012 --experimental-follow-type-hints
[INFO MESSAGES]
Benchmarking opt a sampling:first iterations:1000 follow-type-hints:True
py_fn | warmup:  0.0215 s | total:  4.1227 s
py_fn | warmup:  0.1848 s | total:  4.1191 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt b --sampling first --data-dir workspace/datasets/voc/VOCdevkit/VOC2012 --experimental-follow-type-hints
[INFO MESSAGES]
Benchmarking opt b sampling:first iterations:1000 follow-type-hints:True
py_fn | warmup:  0.0224 s | total:  4.9606 s
py_fn | warmup:  0.2549 s | total:  3.7402 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt a --sampling many --data-dir workspace/datasets/voc/VOCdevkit/VOC2012 --experimental-follow-type-hints
[INFO MESSAGES]
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
Benchmarking opt a sampling:many iterations:1000 follow-type-hints:True
py_fn | warmup:  0.0231 s | total:  4.1516 s
py_fn | warmup:  0.1907 s | total:  4.0629 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt b --sampling many --data-dir workspace/datasets/voc/VOCdevkit/VOC2012 --experimental-follow-type-hints
[INFO MESSAGES]
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
Benchmarking opt b sampling:many iterations:1000 follow-type-hints:True
py_fn | warmup:  0.0233 s | total:  4.8790 s
py_fn | warmup:  0.2338 s | total:  3.8315 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt b --sampling many --data-dir workspace/datasets/voc/VOCdevkit/VOC2012 --experimental-follow-type-hints
[INFO MESSAGES]
Benchmarking opt b sampling:many iterations:1000 follow-type-hints:True
py_fn | warmup:  0.0224 s | total:  4.8583 s
py_fn | warmup:  0.2371 s | total:  3.8537 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt a --sampling many --data-dir workspace/datasets/voc/VOCdevkit/VOC2012 --experimental-follow-type-hints
[INFO MESSAGES]
Benchmarking opt a sampling:many iterations:1000 follow-type-hints:True
py_fn | warmup:  0.0208 s | total:  4.2101 s
py_fn | warmup:  0.1872 s | total:  3.9629 s
```
</details>

bhack: Thanks, note that `experimental_follow_type_hint` is formally deprecated.

It is also important that the function could work directly in batches. 
See:
https://github.com/keras-team/keras-cv/pull/953#issuecomment-1287003209

lucasdavid: Well, the current implementation in main branch isn't batched, is it? I guess it doesn't matter anymore, but for honesty purposes, I [re-implemented the benchmark](https://gist.github.com/lucasdavid/c48d39399b0bbe167cc2f5c056beadf0) using batches and without experimental_follow_type_hint.

qlzh727's implementation is indeed faster when batching (~3x). Still, it has the drawback of huge memory footprint (64MB) for storing all color codes in a (256**3)-sized array.

```shell
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt a --data-dir workspace/datasets/voc/VOCdevkit/VOC2012
Benchmarking opt a iterations:1000
mask_decode_py  | warmup:  0.0704 s | total:  0.0642 s
mask_decode_tf  | warmup:  0.0851 s | total:  0.0469 s
mask_decode_jit | warmup:  0.0274 s | total:  0.0153 s
(t) ldavid@shepard:~$ CUDA_VISIBLE_DEVICES="" python test_voc_colormap.py --opt b --data-dir workspace/datasets/voc/VOCdevkit/VOC2012
Benchmarking opt b iterations:1000
mask_decode_py  | warmup:  0.1805 s | total:  0.1733 s
mask_decode_tf  | warmup:  0.2060 s | total:  0.1662 s
mask_decode_jit | warmup:  0.3092 s | total:  0.3028 s
```

PR #953, opened 2022-10-20
Add tf.function(jit_compile=True) for the PNG decode function.

qlzh727:
# What does this PR do?
Change the decode function to be backed with XLA for better performance.

See https://github.com/keras-team/keras-cv/pull/943#discussion_r1001092882 for the actual performance result. 

In short, XLA compile introduce 0.4 sec overhead, but save 0.2 sec per 1000 examples.

## Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/keras_cv/datasets/pascal_voc/segmentation.py
+++ b/keras_cv/datasets/pascal_voc/segmentation.py
@@ -259,6 +259,9 @@ def _build_metadata(data_dir, image_ids):
     return result
 
 
+# With jit_compile=True, there will be 0.4 sec compilation overhead, but save about 0.2
+# sec per 1000 images. See https://github.com/keras-team/keras-cv/pull/943#discussion_r1001092882
+# for more details.
 @tf.function(jit_compile=True)
 def _decode_png_mask(mask):
     """Decode the raw PNG image and convert it to 2D tensor with probably class."""
--- a/keras_cv/datasets/pascal_voc/segmentation.py
+++ b/keras_cv/datasets/pascal_voc/segmentation.py
@@ -259,6 +259,7 @@ def _build_metadata(data_dir, image_ids):
     return result
 
 
+@tf.function(jit_compile=True)
 def _decode_png_mask(mask):
     """Decode the raw PNG image and convert it to 2D tensor with probably class."""
     # Cast the mask to int32 since the original uint8 will overflow when multiple with 256


qlzh727:
@bhack

bhack (approved):
Just a side note. I don't know if you want to comment inline...
Do you think that this function could work in batch in the case it will be transformed in a KPL?
This just eventually in the case we have a CPU bottleneck and we starve GPU cycles for input images.

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -259,6 +259,7 @@ def _build_metadata(data_dir, image_ids):
     return result
 
 
+@tf.function(jit_compile=True)
```
tanzhenyu, line 262: Maybe add a comment here for some data: w/ or w/o the change, what happens, i.e., compile time change, process time change, etc (though it's highly hardware dependent)

tanzhenyu (approved):

keras_cv/datasets/pascal_voc/segmentation.py
bhack: Yes in theory we have git blame to got to this PR and then to the benchmark comment but not many users has this habit... :smiley_cat:

qlzh727: Done. Added comment.

qlzh727:
> Just a side note. I don't know if you want to comment inline... Do you think that this function could work in batch in the case it will be transformed in a KPL? This just eventually in the case we have a CPU bottleneck and we starve GPU cycles for input images.

Are you referring to tf.vectorized_map in the KPL? This function is not used in the KPL, and I think the batched dataset doesn't use tf.vectorized_map as implementation.

bhack:
> Are you referring to tf.vectorized_map in the KPL? This function is not used in the KPL,

No I meant these colormapping inversions could be exactly KPLs, like a crop or any other preprocessing layer we have.
Than we could rely on KPL tf.vectorized_map, in this case I think it will work without any fallback, but probably this function it is quite "batch ready" to work with batches directly.

bhack:
Other then this KPL hypothesis (for the GPU) also considering to be still in `tf.dataset` I think that if it is easy to have a batched version of that function we could then batch before map `_decode_png_mask` instead of only parallel mapping the whole load:

// dataset = dataset.map(_load_images, num_parallel_calls=tf.data.AUTOTUNE) 

See https://www.tensorflow.org/guide/data_performance#vectorizing_mapping

bhack:
Just to make a small gist example if you want still maintain this in the dataset processing (in the case you don't want to transform this in a KPL to run it on GPU):

https://colab.research.google.com/gist/bhack/2afa97dcb7455ee3edd23ffe29ec93e9/vectorized_colormappign_inversion.ipynb

Of course you can still maintain at the same time `dataset.map(_load_images, num_parallel_calls=tf.data.AUTOTUNE)` that I've just removed to simplify the example.

qlzh727:
Thanks for the example. 

I think the current function can be executed correctly with batched pngs. At this stage, we will keep the dataset as is so that user can add batch/shuffle/cache by themselves down the stream. If we choose to extend the `load()` API to include more params for batch/shuffle or more, we can certainly move this function after batch and let it process the batched pngs.

Eventually, I think we will provide a function to convert the current dataset into tfrecords format, and user will just load the converted data from it, without further decoding, etc.

bhack:
> I think the current function can be executed correctly with batched pngs

Yes the function it is already running in batches "slightly modified" (as you don't have the diff) in the colab

PR #957, opened 2022-10-25
Step 1/3. add dummy semantic segmentation training script.

tanzhenyu:
Step 2: add SBD data
Step 3: add random resize data augmentation that improves training


examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
```diff
@@ -0,0 +1,90 @@
...
+        y_true = y_true.to_tensor()
+    return image, y_true
+
+
+train_ds = train_ds.map(pad_fn, num_parallel_calls=tf.data.AUTOTUNE)
+eval_ds = eval_ds.map(pad_fn, num_parallel_calls=tf.data.AUTOTUNE)
+
+with strategy.scope():
+    backbone = keras_cv.models.ResNet50V2(
+        include_rescaling=True, weights="imagenet/classification-v2", include_top=False
```
ianstenbit, line 75: (optional): use just `"imagenet"` to use the best ImageNet weights (in its current form it will pin to v2)

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
```diff
@@ -0,0 +1,90 @@
...
+    ).as_backbone()
+    model = DeepLabV3(classes=21, backbone=backbone, include_rescaling=True)
```
ianstenbit, line 77: Question about rescaling:

if the backbone is rescaling and the DeepLab model is also rescaling, are we double-rescaling?

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
```diff
@@ -0,0 +1,90 @@
...
+    optimizer = tf.keras.optimizers.SGD(
+        learning_rate=0.0001, momentum=0.9, clipnorm=2.0
```
ianstenbit, line 79: Should this be flag-controlled? (Maybe also batch size?)
(Fine with me if this will come in part 2 or 3)

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
```diff
@@ -0,0 +1,90 @@
...
+    loss_fn = tf.keras.losses.SparseCategoricalCrossentropy(
+        from_logits=True, reduction=tf.keras.losses.Reduction.NONE
+    )
+
+    metrics = [
+        tf.keras.metrics.SparseCategoricalCrossentropy(),
+        tf.keras.metrics.MeanIoU(num_classes=21, sparse_y_pred=False),
+    ]
+model.compile(optimizer, loss="SparseCategoricalCrossentropy", metrics=metrics)
+model.fit(train_ds, epochs=50, validation_data=eval_ds)
```
ianstenbit, line 90: Should we add callbacks for BackupAndRestore and ModelCheckpoint?

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2022 The KerasCV Authors
```
ianstenbit, line 1: Can we move this to `examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py`?

(Or something generic in that folder like `basic_training.py` if this can support other models)

ianstenbit (changes requested):
Thanks!

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
tanzhenyu: Sounds good. task/dataset/model is better than task/model/dataset

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
tanzhenyu: IMHO we should not expose `include_rescaling` at meta arch level, if it accepts `backbone` as another argument. However the trick here is when `backbone` also accept a string we don't know whether we should include that or not. Not particularly related to this PR, but we should definitely discuss this and have a consensus.

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
tanzhenyu: Done.

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
tanzhenyu: Added ModelCheckpoint.

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
tanzhenyu: Done. I assume if the weights is not found, error will be thrown?

qlzh727 (approved):
The metric calculation might need some update, to ignore the class 255 (boundary).

tanzhenyu:
> The metric calculation might need some update, to ignore the class 255 (boundary).

It is currently ignored in the class_segmentation_map during data pipeline as background mask. But yeah good point! We should move this to metrics part instead of data pipeline.

examples/models/semantic_segmentation/deeplab_v3/pascal_voc/train.py
ianstenbit: I agree that this should be backbone-configured (maybe that means for the meta-arch we have an optional `include_rescaling` which should be supplied iff using a string backbone)

ianstenbit (approved):
LGTM

PR #959, opened 2022-10-25
Fix build target for custom ops

ianstenbit:


ianstenbit:
Also renaming `BUILD` to `BUILD.bazel` to make `setup.py` happy on case-insensitive file systems

tanzhenyu (approved):

PR #961, opened 2022-10-26
Step 2/3: add Sbd data

tanzhenyu:
Step 2: add SBD data -- without 3, this gives meanIOU ~ 37%, still heavily overfitted
Step 3: add random resize data augmentation that improves training

Using the generator pattern with scipy, I haven't noticed any obvious drop in training performance compared to loading voc data using tf ops, actually the generator is even faster (140ms/step compared to 220ms/step), so I leave this as a TODO later to generate TFRecords


bhack:
>  Step 3: add random resize data augmentation that improves training

Can you check step 3 overhead on training an epoch? Thanks..

tanzhenyu:
> > Step 3: add random resize data augmentation that improves training
> 
> Can you check step 3 overhead on training an epoch? Thanks..

Which overhead, training performance or model quality?

bhack:
Just the training performance overhead you have introducing 3.

tanzhenyu:
> Just the training performance overhead you have introducing 3.

I see. I can test it when we're getting to Step 3. But if you just care about step 3, it's easy to comment this out in Object Detection training scripts and compare the diff -- they use pretty much the same random resizing logic

bhack:
> I see. I can test it when we're getting to Step 3. 

Yes it is just that you are already preparing to launch these jobs. So if you can annotate some numbers it could be useful (or if you could log the metric in the Tensorboard logs)

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -117,38 +129,38 @@ def _maybe_populate_voc_color_mapping():
...
-def _download_pascal_voc_2012(data_url, local_dir_path=None, override_extract=False):
-    """Fetch the original Pascal VOC 2012 from remote URL.
+def _download_data_file(
+    data_url, extracted_dir, local_dir_path=None, override_extract=False
+):
+    """Fetch the original VOC or Semantic Boundaries Dataset from remote URL.
 
     Args:
-        data_url: string, the URL for the Pascal VOC data, should be in a tar package.
+        data_url: string, the URL for the SBD data, should be in a zipped tar package.
```
ianstenbit, line 138: I think this should be "the URL for the SBD or VOC data", right?

(Same for the Returns: docs)

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -336,7 +444,7 @@ def load(
             download the data file, and unzip. It will be used as a cach directory.
             Default to None, and `~/.keras/pascal_voc_2012` will be used.
     """
-    supported_split_value = ["train", "eval", None]
+    supported_split_value = ["train", "eval", "trainval", "sbd_train", "sbd_eval"]
```
ianstenbit, line 447: One other API approach to consider:
load(`split="train", semantic_boundaries=True`)

(As a user, I feel like whether or not I'm using semantic boundaries is a little bit orthogonal to which split I'm using) What do you think?

ianstenbit: (Regardless of exact API decision, I think some docs on the `load` method about the meaning of `sbd` would be good)

keras_cv/datasets/pascal_voc/segmentation.py
tanzhenyu: Done.

keras_cv/datasets/pascal_voc/segmentation.py
tanzhenyu: The original tfds proposal is to use "sbd_train" to differentiate with voc data, see [here](https://github.com/AdamHillier/datasets/blob/38a1b5d73c4e6e3700d77fbe31c9cc469f3f5918/tensorflow_datasets/object_detection/voc.py#L80)

Either way seems fine to me -- we could probably even just get rid of voc data for segmentation tasks.

keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -300,6 +362,25 @@ def _load_images(example):
     return example
 
 
+def _load_sbd_images(image_file_path, seg_cls_file_path, seg_obj_file_path):
+    image = tf.io.read_file(image_file_path)
+    image = tf.image.decode_jpeg(image)
+
+    segmentation_class_mask = scipy.io.loadmat(seg_cls_file_path)
```
ianstenbit, line 369: Can we do this simply without a scipy dependency?

ianstenbit (approved):

keras_cv/datasets/pascal_voc/segmentation.py
bhack: `tfds.core.lazy_imports.scipy.io.loadmat`

tanzhenyu: Done. I was under the impression that tfds will still try to import scipy.io? What makes one succeed and the other fail in cloudbuild?

Issue #855, opened 2022-09-27
Track % of boxes matched in LabelEncoder of RetinaNet in a Keras metric


LukeWood:
This would be a super valuable debugging tool

PR #964, opened 2022-10-27
Track encoded boxes in the RetinaNet in a Keras metric

LukeWood:
fixes https://github.com/keras-team/keras-cv/issues/855

### DIFF
--- a/keras_cv/layers/object_detection/retina_net_label_encoder.py
+++ b/keras_cv/layers/object_detection/retina_net_label_encoder.py
@@ -36,6 +36,11 @@ class RetinaNetLabelEncoder(layers.Layer):
         background_class: (Optional) The class ID used for the background class.
             Defaults to -1.
         ignore_class: (Optional) The class ID used for the ignore class. Defaults to -2.
+        track_matched_boxes: (Optional) Whether or not to track the number of boxes that
+            are matched with an anchor box in a Keras metric.  To use this metric, you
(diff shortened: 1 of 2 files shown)


tanzhenyu (commented):
I wonder whether we should do this through a callback, instead of impacting the API?

LukeWood:
> I wonder whether we should do this through a callback, instead of impacting the API?

Good point.  I feel this actually relates to the PyMetric in a way; in particular we want to add a metric to each epoch outside of the train/test step.

keras_cv/layers/object_detection/retina_net_label_encoder.py
```diff
@@ -36,6 +36,11 @@ class RetinaNetLabelEncoder(layers.Layer):
         background_class: (Optional) The class ID used for the background class.
             Defaults to -1.
         ignore_class: (Optional) The class ID used for the ignore class. Defaults to -2.
+        track_matched_boxes: (Optional) Whether or not to track the number of boxes that
```
tanzhenyu, line 39: discussed offline -- let's just turn on this without impacting API

LukeWood: (This is in the encoder - RetinaNet still doesn't have a way to disable)

LukeWood: Updated

tanzhenyu (approved):

PR #967, opened 2022-10-28
Added mixed_precision support for preprocessing layers.

### LINKED ISSUE #734: Is mixed precision, mixed_float16, not supported for certain augmentations
I found that certain augmentations fail when the mixed precision policy is set to mixed_float16 but work with float32.  So far I've found this to be true for RandomColorDegeneration and Solarization - both run individually and within the RandAugment layer.

Is there a work around? I did try specifying the dtype for the RandAugment layer but that didn't change the outcome.

sebastian-sz:
# What does this PR do?
It was reported that some preprocessing layers do not run in mixed float16. This PR is an attempt to fix this.

Fixes #734 

This PR introduces two things:
1. Adds test for running in mixed precision for Keras-CV preprocessing layers.
2. Fix those layers that can be fixed from Python side.

### Affected layers:
Running (added in this PR) `keras_cv/layers/preprocessing/with_mixed_precision_test.py`, without changing the existing code, shows which layers are affected:
* GridMask
* CutMix
* Mosaic (throws some `OP_REQUIRES failed at functional_ops.cc:373` warning)
* RandomColorDegeneration 
* RandomCutout 
* RandomGaussianBlur
* RandomSharpness
* Solarization
* RandomColorJitter  
* RandomSaturation 

I managed to fix all, but two layers: RandomColorJitter and RandomSaturation. Both of these use `tf.image.adjust_saturation` which doesn't have CPU Float16 kernel registered. I believe this would have to be fixed from Tensorflow side.

The above layers should still work on the GPU though.

# Limitations
The majority of the problems were caused by explicitly casting to float32, which worked if the images were also in float32. I mostly added casting to `images.dtype` or fetching some Factor component with different dtype.

I am not sure, however, about running some computations in float16, as previously they were running in float32.
For example in: Mosaic and RandomGaussianBlur. I added a TODO so you could provide a comment whether you think this is ok or not.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood @tanzhenyu @ianstenbit

### DIFF
--- a/keras_cv/layers/preprocessing/random_color_degeneration.py
+++ b/keras_cv/layers/preprocessing/random_color_degeneration.py
@@ -55,7 +55,7 @@ class RandomColorDegeneration(BaseImageAugmentationLayer):
         self.seed = seed
 
     def get_random_transformation(self, **kwargs):
-        return self.factor()
+        return self.factor(dtype=kwargs["image"].dtype)
 
     def augment_image(self, image, transformation=None, **kwargs):
         degenerate = tf.image.grayscale_to_rgb(tf.image.rgb_to_grayscale(image))
--- a/keras_cv/layers/preprocessing/random_color_degeneration.py
+++ b/keras_cv/layers/preprocessing/random_color_degeneration.py
@@ -55,7 +55,7 @@ class RandomColorDegeneration(BaseImageAugmentationLayer):
         self.seed = seed
 
     def get_random_transformation(self, **kwargs):
-        return self.factor(dtype=kwargs["image"].dtype)
+        return self.factor(dtype=self.compute_dtype)
 
     def augment_image(self, image, transformation=None, **kwargs):
         degenerate = tf.image.grayscale_to_rgb(tf.image.rgb_to_grayscale(image))
--- a/keras_cv/layers/preprocessing/random_gaussian_blur.py
+++ b/keras_cv/layers/preprocessing/random_gaussian_blur.py
@@ -59,7 +59,7 @@ class RandomGaussianBlur(BaseImageAugmentationLayer):
                 )
 
     def get_random_transformation(self, **kwargs):
-        factor = self.factor(dtype=kwargs["image"].dtype)
+        factor = self.factor()
         blur_v = RandomGaussianBlur.get_kernel(factor, self.y)
         blur_h = RandomGaussianBlur.get_kernel(factor, self.x)
         blur_v = tf.reshape(blur_v, [self.y, 1, 1, 1])
@@ -72,8 +72,12 @@ class RandomGaussianBlur(BaseImageAugmentationLayer):
 
         num_channels = tf.shape(image)[-1]
         blur_v, blur_h = transformation
-        blur_h = tf.tile(blur_h, [1, 1, num_channels, 1])
-        blur_v = tf.tile(blur_v, [1, 1, num_channels, 1])
+        blur_h = tf.cast(
+            tf.tile(blur_h, [1, 1, num_channels, 1]), dtype=self.compute_dtype
+        )
+        blur_v = tf.cast(
+            tf.tile(blur_v, [1, 1, num_channels, 1]), dtype=self.compute_dtype
+        )
         blurred = tf.nn.depthwise_conv2d(
             image, blur_h, strides=[1, 1, 1, 1], padding="SAME"
         )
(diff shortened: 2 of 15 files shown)


keras_cv/layers/preprocessing/random_color_degeneration.py
```diff
@@ -55,7 +55,7 @@ def __init__(
         self.seed = seed
 
     def get_random_transformation(self, **kwargs):
-        return self.factor()
+        return self.factor(dtype=kwargs["image"].dtype)
```
ianstenbit, line 58: Everywhere where we're making this change -- instead of using `kwargs["image"]` let's just add `image=None` to the method signature

keras_cv/layers/preprocessing/with_mixed_precision_test.py
```diff
@@ -0,0 +1,136 @@
...
+NO_CPU_FP16_KERNEL_LAYERS = [
+    preprocessing.RandomSaturation,
+    preprocessing.RandomColorJitter,
+]
+
+
+class WithMixedPrecisionTest(tf.test.TestCase, parameterized.TestCase):
+    @parameterized.named_parameters(
+        *TEST_CONFIGURATIONS,
+        ("CutMix", preprocessing.CutMix, {}),
```
ianstenbit, line 106: I think these can just be included in TEST_CONFIGURATIONS

keras_cv/layers/preprocessing/random_gaussian_blur.py
```diff
@@ -94,12 +94,13 @@ def augment_segmentation_mask(self, segmentation_mask, transformation, **kwargs)
...
     def get_kernel(factor, filter_size):
+
         x = tf.cast(
-            tf.range(-filter_size // 2 + 1, filter_size // 2 + 1), dtype=tf.float32
-        )
-        blur_filter = tf.exp(
-            -tf.pow(x, 2.0) / (2.0 * tf.pow(tf.cast(factor, dtype=tf.float32), 2.0))
+            tf.range(-filter_size // 2 + 1, filter_size // 2 + 1), dtype=factor.dtype
         )
+        # TODO: are we ok with running this in FP16?
```
ianstenbit, line 101: I would expect this to be okay in fp16, and unless there's some good evidence that this is bad, I don't think we should special case this.

ianstenbit (changes requested):
This is great! Thanks for the PR 💯

keras_cv/layers/preprocessing/random_color_degeneration.py
tanzhenyu: +1

keras_cv/layers/preprocessing/random_gaussian_blur.py
tanzhenyu: exponential computation might be a problem in fp16, how do we verify if it's ok to do this?

bhack: In PyTorch vision they have test coverage for fp16:

https://github.com/pytorch/vision/blob/main/test/test_functional_tensor.py#L1097-L1146

sebastian-sz: I inspected the output pixel values of `RandomGaussianBlur`. If we calculate the kernel in FP16 the results are quite off, compared to FP32 version (e.g. given value range of 0-255, FP32 will output 240 and FP16 will output 160). 

I will keep kernel calculation in FP32, as this fixes the issue. The blurring is still done in FP16 (after casting the kernel to adequate dtype).

keras_cv/layers/preprocessing/with_mixed_precision_test.py
sebastian-sz: Added.

keras_cv/layers/preprocessing/random_gaussian_blur.py
sebastian-sz: Moved (only) kernel calculation fo FP32.

sebastian-sz:
@ianstenbit @tanzhenyu 
I updated the PR.  
* Removed `kwargs['image']` - use `self.compute_dtype` instead. I also added `self.compute_dtype` to most casts as I think it's more elegant (?).
* Moved `RandomGaussianBlur` kernel calculation to float32 (blurring, aka `tf.nn.depthwise_conv2d` still happens in float16).
* Added `dtype=self.compute_dtype` to `transform_value_range` functions, as it was force casting to float32 by default.

keras_cv/layers/preprocessing/random_gaussian_blur.py
```diff
@@ -94,13 +98,12 @@ def augment_segmentation_mask(self, segmentation_mask, transformation, **kwargs)
 
     @staticmethod
     def get_kernel(factor, filter_size):
-
         x = tf.cast(
-            tf.range(-filter_size // 2 + 1, filter_size // 2 + 1), dtype=factor.dtype
+            tf.range(-filter_size // 2 + 1, filter_size // 2 + 1), dtype=tf.float32
```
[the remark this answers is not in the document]
sebastian-sz: Added comment in `get_kernel` method.

PR #968, opened 2022-10-28
Add Dice Loss

DavidLandup0:
# What does this PR do?

As introduced in #296 discussed in #371, the modified PR includes:

- Fixed typos/extended documentation
- Test cases for output shapes and values of CategoricalDice, SparseDice and BinaryDice
- Check for `axis` which adjusts the channel format (whether the input is channels_first or channels_last)
- Check for `axis` values 

Tagging the people from the first PR and @innat whose original PR was modified. 
/cc @LukeWood @bhack @qlzh727 

## Before submitting
      Pull Request section?
      to it if that's the case. #296 #371

## Who can review?


tanzhenyu (commented):
Can you help us report the mean iou after 30 epochs? (w/ or w/o mixed precision is fine)

DavidLandup0:
Sure! Just to check - there are 9963 images in the set, right?

tanzhenyu:
> Sure! Just to check - there are 9963 images in the set, right?

Yes it is around that. I have yet to submit a "Step 4" PR to augment it to around ~10000 images, but it's on the right order

DavidLandup0:
And the `sbd_train` is `train+val` splits, or? I can't seem to find the exact info 🤔

tanzhenyu:
> And the `sbd_train` is `train+val` splits, or? I can't seem to find the exact info 🤔

you can use whatever the training script is using today

DavidLandup0:
Running the script. It might take a while, so I'll run it locally on my workstation to avoid interruptions in training. I'll report the val mean iou after 30 epochs when they run out :)

tanzhenyu:
> Running the script. It might take a while, so I'll run it locally on my workstation to avoid interruptions in training. I'll report the val mean iou after 30 epochs when they run out :)

on a 2-gpu setup, it takes ~4 hours I think? Hopefully this can be done simply by changing `model.compile` in the training script.

DavidLandup0:
Yup, the only difference is the change from `SparseCategoricalCrossentropy` to `SparseDice`!

I'm running it on a single GPU system, since Kaggle's 2-GPU environment throws unsupported Conv2D operations (a recent example by @bhack worked, but I haven't been able to get it running for KCV models yet). Due to the limited VRAM, I can only fit the training with a batch_size of `1`, so hopefully the training doesn't go awry because of the super small batch size.

What's the reference `val_mean_io_u` to aim for?

DavidLandup0:
The 30 epoch run is done, best epoch was:

```
Epoch 22/30
932/932 [==============================] - 1137s 1s/step - loss: 0.9042 - sparse_categorical_crossentropy: 0.5024 - mean_io_u: 0.5802 - sparse_categorical_accuracy: 0.8944 - val_loss: 0.9084 - val_sparse_categorical_crossentropy: 0.5301 - val_mean_io_u: 0.5937 - val_sparse_categorical_accuracy: 0.8920
```

And it tapered off in the end, with somewhat unstable metrics. The LR might've been a bit too high? 
Does this look okay? @tanzhenyu

tanzhenyu:
> The 30 epoch run is done, best epoch was:
> 
> ```
> Epoch 22/30
> 932/932 [==============================] - 1137s 1s/step - loss: 0.9042 - sparse_categorical_crossentropy: 0.5024 - mean_io_u: 0.5802 - sparse_categorical_accuracy: 0.8944 - val_loss: 0.9084 - val_sparse_categorical_crossentropy: 0.5301 - val_mean_io_u: 0.5937 - val_sparse_categorical_accuracy: 0.8920
> ```
> 
> And it tapered off in the end, with somewhat unstable metrics. The LR might've been a bit too high? Does this look okay? @tanzhenyu

shouldn't it report something like "sparse_dice" instead of "sparse_categorical_crossentropy"?

DavidLandup0:
> shouldn't it report something like "sparse_dice" instead of "sparse_categorical_crossentropy"?

It's using a Dice Loss, so the `loss` and `val_loss` are the main metric. I've included the sparse categorical crossentropy as a metric (`keras.metrics.SparseCategoricalCrossentropy`) from the original script, because I thought it would help compare the dice loss and CE loss runs with a common metric :)

We don't currently have a dice score as a _metric_. I can add it in if you'd like?

tanzhenyu:
> > shouldn't it report something like "sparse_dice" instead of "sparse_categorical_crossentropy"?
> 
> It's using a Dice Loss, so the `loss` and `val_loss` are the main metric tied to the dice scores. I've included the sparse categorical crossentropy as a metric (`keras.metrics.SparseCategoricalCrossentropy`) from the original script, because I thought it would help compare the dice loss and CE loss runs with a common metric :)
> 
> ```
> loss_fn = keras_cv.losses.SparseDice()
> metrics = [
>         tf.keras.metrics.SparseCategoricalCrossentropy(ignore_class=255),
>         tf.keras.metrics.MeanIoU(num_classes=21, sparse_y_pred=False),
>         tf.keras.metrics.SparseCategoricalAccuracy(),
> ]
> ```
> 
> We don't currently have a dice score as a _metric_.

Oh I see. yeah the metrics and loss looks solid to me

DavidLandup0:
Awesome! Wanna review the code and tests?
I can add dice variants as metrics if you think they'd be useful to have 🤔

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,592 @@
...
+
+    if class_ids is not None:
+        if isinstance(class_ids, float) or any(isinstance(x, float) for x in class_ids):
+            raise ValueError(
+                f"The indices should be int or a list of integer. Got {class_ids}"
+            )
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class CategoricalDice(tf.keras.losses.Loss):
```
tanzhenyu, line 85: A high-level recommendation -- can you split this up into separate PRs? For example, if you only use SparseDice, maybe start with that first. This greatly help us to expedite the review process (given we want to review for correctness as well)

tanzhenyu (changes requested):

keras_cv/losses/dice.py
bhack: I hope that we could enforce this policy also with internal sourcing PRs.

Recently I saw very large PRs.

DavidLandup0: Sure! They were in one originally, so I kept them that way. Separating into three PRs.
Do I also do Dice as a metric too?

DavidLandup0: Also, one file for all three PRs or three files? It might be a bit verbose for the FS if there are three files given the shared method(s) between all three losses

DavidLandup0: Separated into three PRs, one for each loss type. Since they share two methods, I've put them all in the same file. We could have a separate file for util methods, but no other losses have this format.
Is this okay? @tanzhenyu

DavidLandup0: Chaining onto @bhack's comment - what's a large PR?
I'll make mine smaller going forward, but some wrapper would help in trying to make them more manageable for reviews :)

DavidLandup0: Can we revisit this since it should be near the finish line? @tanzhenyu

keras_cv/losses/dice.py
```diff
@@ -0,0 +1,261 @@
...
+    def __init__(
+        self,
+        beta=1,
+        from_logits=False,
+        class_ids=None,
+        axis=[1, 2],
+        loss_type=None,
+        label_smoothing=0.0,
+        epsilon=1e-07,
+        per_image=False,
```
innat, line 124: > Note: I haven't been able to test the per_image argument. @innat mentioned that it provides a performance boost in some cases, but I'm not sure which tasks these are. Could you share some that we can benchmark the argument?

https://github.com/keras-team/keras-cv/issues/296#issuecomment-1326053384

innat:
@jbischof 
Any thoughts on this PRs. https://github.com/keras-team/keras-cv/pull/1050 https://github.com/keras-team/keras-cv/pull/1049
cc. @DavidLandup0

jbischof:
Seems interesting @innat if @DavidLandup0 or someone else wants to continue the work!

innat:
@DavidLandup0 
Would you like to move forward this PR?

PR #974, opened 2022-11-01
image generation with same seed and same prompt is fixed.

faizankarim:
Fixes # (issue)
same seed value will now return same images if the prompt is also same, 
previously if model is loaded and user give same seed value to model with same prompt multiple times different images were generated.


LukeWood:
I think  @ianstenbit did some work to fix this issue; so perhaps it is already fixed.  Adding Ian to review; feel free to close if already fixed.

ianstenbit:
It looks like this is making a very similar change to #962 (and has a merge conflict due to that one).

@faizankarim after pulling past #962 are you still seeing lack of reproducability?

faizankarim:
yes #962 fixes it, closing it now.

PR #975, opened 2022-11-01
Step 3/3: Update DeeplabV3 model

tanzhenyu:
(There might be several broken tests for the 1st commit, but requesting review to go through several design questions first)
Step 2: add SBD data -- without 3, this gives meanIOU ~ 37%, still heavily overfitted
Step 3 (original plan): add random resize data augmentation that improves training
Step 3 (new plan): updates DeepLabV3 model that improves training, this gives meanIOU ~ 63%
Step 4 (newly added): adds a "trainaug" dataset that helps the model to achieve ~70%, and adds visualization utils to convert back from label to colormap.

This PR adds the following:
1) gets rid of the decoder module, adds ASPP. I realized that we should only adds decoder for DeepLabV3+, not DeepLabV3.
2) modifies ResNetV2 module, adds dilation rate argument and removes zero padding, coincidentally this is also how GluonCV [works](https://github.com/dmlc/gluon-cv/blob/master/gluoncv/model_zoo/resnetv1b.py#L181-L194)
3) no need for an internal segmentation head anymore (to be removed depending on the review)
4) ablation study between adding random data augmentation and only resizing, finds resizing actually works better, which is different than object detection and classification.
5) gets rid of dict input for ASPP, the main reason is that dilation rate should be different per level, so we should probably have a dict of {level: ASPP_Module} instead of ASPP(level).
6) finds out a nan issue with ASPP, a keras issue to be filed. 
7) more ablation study and leaves several TODOs, the current mIOU is ~63 mIOU. The next PR will introduce more data to get 70 mIOU.

TO be discussed:
1) whether taking both backbone and weights are necessary at model API. It is actually not easy for users to pass backbone as a separate model because they have to deal with dilation rates.


keras_cv/datasets/pascal_voc/segmentation.py
```diff
@@ -406,10 +409,13 @@ def _build_dataset_from_metadata(metadata):
 def _build_sbd_dataset_from_metadata(metadata):
     img_filepath = metadata["image/file_path"]
     cls_filepath = metadata["segmentation/class/file_path"]
-    obj_file_path = metadata["segmentation/object/file_path"]
+    obj_filepath = metadata["segmentation/object/file_path"]
 
     def md_gen():
-        for img_fp, cls_fp, obj_fp in zip(img_filepath, cls_filepath, obj_file_path):
+        c = list(zip(img_filepath, cls_filepath, obj_filepath))
+        random.shuffle(c)
```
qlzh727, line 416: Is it possible to have the shuffle after the dataset is built? or you would like the original dataset to be in random order?

keras_cv/layers/spatial_pyramid.py
```diff
@@ -31,22 +31,15 @@ class SpatialPyramidPooling(tf.keras.layers.Layer):
     inp = tf.keras.layers.Input((384, 384, 3))
     backbone = tf.keras.applications.EfficientNetB0(input_tensor=inp, include_top=False)
     layer_names = ['block2b_add', 'block3b_add', 'block5c_add', 'top_activation']
```
qlzh727, line 33: The layer_names is no longer used here anymore.

keras_cv/layers/spatial_pyramid.py
```diff
@@ -31,22 +31,15 @@ class SpatialPyramidPooling(tf.keras.layers.Layer):
...
-
-    backbone_outputs = {}
-    for i, layer_name in enumerate(layer_names):
-        backbone_outputs[i+2] = backbone.get_layer(layer_name).output
-
-    # output_dict is a dict with 4 as keys, since it only process the level 4 backbone
-    # inputs
-    output_dict = keras_cv.layers.SpatialPyramidPooling(
-        level=4, dilation_rates=[6, 12, 18])(backbone_outputs)
+    backbone_outputs = backbone(inp)
```
qlzh727, line 34: I didn't see any change on the resnet50 to let it produce backbone level output, eg calling model.as_backbone(). Do u want to update this example, or I missed anything.

keras_cv/layers/spatial_pyramid.py
```diff
@@ -157,47 +143,23 @@ def build(self, input_shape):
...
-        The input the of the layer will be a dict of {`level`, `tf.Tensor`}, and layer
-        will pick the actual input based on the `level` from its init args.
-        The output of the layer will be a dict of {`level`, `tf.Tensor`} with only one level.
-
         Args:
-          inputs: A `dict` of `tf.Tensor` where
-            - key: A `int` of the level of the multilevel feature maps.
-            - values: A `tf.Tensor` of shape [batch, height_l, width_l,
-              filter_size].
+          inputs: A `tf.Tensor` of shape [batch, height, width, channels]
```
qlzh727, line 147: so you would like the backbone network to directly produce one tensor at the specific feature level. This somehow impose a more strong coupling between the backbone network and the ASPP layer, and user might not easily swap the components.

keras_cv/models/segmentation/deeplab.py
```diff
@@ -64,89 +76,89 @@ def __init__(
...
 
     def train_step(self, data):
-        images, y_true, _ = tf.keras.utils.unpack_x_y_sample_weight(data)
+        images, y_true, sample_weight = tf.keras.utils.unpack_x_y_sample_weight(data)
         with tf.GradientTape() as tape:
             y_pred = self(images, training=True)
-            total_loss = self.compute_loss(images, y_true, y_pred)
+            total_loss = self.compute_loss(images, y_true, y_pred, sample_weight)
             reg_losses = []
             if self.weight_decay:
```
qlzh727, line 154: Where is this field get set?

keras_cv/models/segmentation/deeplab.py
```diff
@@ -64,89 +76,89 @@ def __init__(
...
-                output_scale_factor = pow(2, self.decoder.min_level)
-            else:
-                output_scale_factor = None
-
-            segmentation_head = (
-                keras_cv.models.segmentation.__internal__.SegmentationHead(
-                    classes=classes, output_scale_factor=output_scale_factor
-                )
+            segmentation_head = tf.keras.Sequential(
+                [
```
qlzh727, line 115: I vaguely remember there are some conv2d + BN needed before the final classification head. Did u find that in the original papar/code?

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
```diff
@@ -56,49 +55,80 @@
...
+# TODO(tanzhenyu): add a diff dataset.
+# all_ds = load(split="sbd_train", data_dir=None)
+# all_ds = all_ds.concatenate(load(split="sbd_eval", data_dir=None))
+# train_ds = all_ds.take(10000)
+# eval_ds = all_ds.skip(10000).concatenate(load(split="diff", data_dir=None))
 train_ds = load(split="sbd_train", data_dir=None)
 eval_ds = load(split="sbd_eval", data_dir=None)
 
-resize_layer = tf.keras.layers.Resizing(512, 512)
+resize_layer = tf.keras.layers.Resizing(512, 512, interpolation="nearest")
```
ianstenbit, line 66: Should we use standard bilinear interpolation for the image, even though we need to use "nearest" for the segmask? (I guess maybe it doesn't really matter?)

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
```diff
@@ -56,49 +55,80 @@
...
+
+image_size = [512, 512, 3]
+
+
+def flip_fn(image, cls_seg):
```
ianstenbit, line 71: our RandomFlip KPL supports seg masks -- can we use that here instead?

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
```diff
@@ -56,49 +55,80 @@
...
-    ).as_backbone()
-    model = DeepLabV3(classes=21, backbone=backbone, include_rescaling=True)
+    model = DeepLabV3(
+        classes=21, backbone="resnet50_v2", include_rescaling=True, weights="imagenet"
+    )
     optimizer = tf.keras.optimizers.SGD(
         learning_rate=lr_decay, momentum=0.9, clipnorm=10.0
     )
-    loss_fn = tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True)
+    loss_fn = tf.keras.losses.SparseCategoricalCrossentropy(ignore_class=255)
```
ianstenbit, line 127: Since 255 is a magic number in this dataset, should we export it as a constant from the dataloader with a descriptive name?

keras_cv/models/resnet_v2.py
```diff
@@ -464,6 +484,7 @@ def ResNet18V2(
         stackwise_filters=MODEL_CONFIGS["ResNet18V2"]["stackwise_filters"],
         stackwise_blocks=MODEL_CONFIGS["ResNet18V2"]["stackwise_blocks"],
         stackwise_strides=MODEL_CONFIGS["ResNet18V2"]["stackwise_strides"],
+        stackwise_dilations=MODEL_CONFIGS["ResNet50V2"]["stackwise_dilations"],
```
ianstenbit, line 487: It looks like these are not defined in this file (and in this case it's using the config from the wrong ResNet)

keras_cv/models/segmentation/deeplab.py
```diff
@@ -64,89 +76,89 @@ def __init__(
...
-                    include_rescaling=include_rescaling, include_top=False
+                backbone = ResNetV2(
+                    stackwise_filters=BACKBONE_CONFIG["ResNet50V2"][
+                        "stackwise_filters"
+                    ],
+                    stackwise_blocks=BACKBONE_CONFIG["ResNet50V2"]["stackwise_blocks"],
+                    stackwise_strides=BACKBONE_CONFIG["ResNet50V2"][
+                        "stackwise_strides"
+                    ],
+                    stackwise_dilations=BACKBONE_CONFIG["ResNet50V2"][
```
ianstenbit, line 87: Should stackwise_dilations be defined for all ResNets, or only for this particular application?

It seems like this should break our weights -- is that correct?

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
```diff
@@ -56,49 +55,80 @@
...
     metrics = [
-        tf.keras.metrics.SparseCategoricalCrossentropy(from_logits=True),
+        tf.keras.metrics.SparseCategoricalCrossentropy(ignore_class=255),
```
innat, line 129: @tanzhenyu @qlzh727 
`ignore_class` and [`class_ids`](https://github.com/keras-team/keras-cv/blob/c8064e64043a35360e42023be2f411c9f5520b29/keras_cv/metrics/coco/mean_average_precision.py#L32) are similar in nature. It looks confusing to have both in the API.

keras_cv/datasets/pascal_voc/segmentation.py
tanzhenyu: I tested only shuffling before the generator vs shuffling at each time. Apparently the latter gives ~10+ MeanIOU

keras_cv/layers/spatial_pyramid.py
tanzhenyu: `backbone_outputs` here is a single Tensor. Renaming it to `output` intead.

keras_cv/layers/spatial_pyramid.py
tanzhenyu: Yeah, because users have to carefully choose the dilation rate for each level.
I think in the future it might be better if we allow creating multiple ASPP layers, and each take different level and dilation rates, but all accept dictionary of inputs -- but I'd probably figure out what's the best API once I start adding DeepLabV3+. For now, it might be better to do the minimal?

keras_cv/models/segmentation/deeplab.py
tanzhenyu: in Line 149?

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
tanzhenyu: It probably doesn't matter, but if we don't choose the same method, then merging this to KPL will cause an issue?

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
tanzhenyu: added a TODO, will test the BEFORE and AFTER difference.

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
tanzhenyu: We paid for something like this from other keras datasets (such as mnist) when we attach class_ids to it. But the issue is that this will get lost when adding another transformation on the returned dataset.
So adding a comment here for users who need to copy the template.

keras_cv/models/resnet_v2.py
tanzhenyu: good catch. removed.

keras_cv/models/segmentation/deeplab.py
tanzhenyu: It shouldn't break our weights, i.e., the current classification models will not use this dilation, but deeplab models will use it but still load from classification model pretrained weights

keras_cv/layers/spatial_pyramid.py
tanzhenyu: (The more I think, the more I feel like even in DeepLabV3+, we need something like this:
```
level_to_dilation = {2: [4, 6, 8], 4: [2, 4, 6]}
level_to_aspp = {key: SpatialPyramidPooling(key, value) for key, value in level_to_dilation.items()}
backbone_outputs = backbone(inputs).get(level_to_dilation.keys())
...
```
So passing a dict of inputs (2, 3, 4, 5) and getting a dict of outputs (2) seems weird and confusing to users, instead we accept single tensor and return single tensor seems more straightforward.

keras_cv/models/segmentation/deeplab.py
tanzhenyu: From the paper:
"The resulting features from all the branches are then concatenated and pass through another 1x1 conv (also with 256 filters and batch norm) before the final 1x1 conv which generates the final logits"

From [Gluon](https://cv.gluon.ai/_modules/gluoncv/model_zoo/deeplabv3.html), it's a single 3x3 conv

From previous segmentation head, it's 2 3x3 conv layers.

They all seem very different. And TBH I don't see how this adds any value given it's mapping from 256 channels to 256 channels. But I chose to stick with what the paper says regardless.

keras_cv/layers/spatial_pyramid.py
tanzhenyu: Done.

keras_cv/models/segmentation/deeplab.py
tanzhenyu: Some update -- adding another conv_bn_relu makes the meanIOU drop by 1.0. But sticking to the paper sounds fine with me.

keras_cv/models/resnet_v2.py
```diff
@@ -265,6 +276,7 @@ def ResNetV2(
     stackwise_strides,
     include_rescaling,
     include_top,
+    stackwise_dilations=None,
```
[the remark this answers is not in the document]
tanzhenyu: but other `stackwise_.*` are necessary params, this is optional param?

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
tanzhenyu: Done.

keras_cv/models/resnet_v2.py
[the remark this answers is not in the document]
ianstenbit: you're right -- I totally missed that  🤦

keras_cv/layers/spatial_pyramid.py
```diff
@@ -133,9 +112,13 @@ def build(self, input_shape):
...
                 tf.keras.layers.Reshape((1, 1, channels)),
                 tf.keras.layers.Conv2D(
-                    filters=self.num_channels, kernel_size=(1, 1), use_bias=False
+                    filters=self.num_channels,
+                    kernel_size=(1, 1),
+                    use_bias=False,
                 ),
-                tf.keras.layers.BatchNormalization(),
+                tf.keras.layers.experimental.SyncBatchNormalization(),
+                # TODO(tanzhenyu): File a bug to Keras for NAN issue.
```
ianstenbit, line 120: maybe link to https://github.com/keras-team/keras/issues/17204 or https://github.com/tensorflow/tensorflow/issues/48845?

ianstenbit (approved):
Excited that this is almost ready! Awesome 🎊

keras_cv/layers/spatial_pyramid.py
bhack: @tanzhenyu Can you quickly verify if you can reproduce it with `drop_remainder=True`?

tanzhenyu: Looks like the problem goes away. (I'm running the script to see if quality wise it makes any difference)

Issue #915, opened 2022-10-13
Inconsistent behavior between pycoco and keras_cv.metrics.COCOMeanAveragePrecision


tanzhenyu:
During the development of FasterRCNN, I have noticed reasonable loss and reasonable prediction outcome, but always get low mAP metrics (~0.18). The evaluation loop takes ~13 minutes.
So I instead hack a COCO wrapper (thanks to Model Garden) and directly call pycoco, this time it gets me 0.44 mAP which is more reasonable. The evaluation loop takes ~8 minutes.
I wonder how we can investigate those inconsistencies.
I have put my reproducible code in a separate branch, this is a standalone script that can run end-to-end:
https://github.com/keras-team/keras-cv/blob/map/examples/models/object_detection/faster_rcnn/eval.py

I have also tried to configure iou_thresholds, recall_thresholds to try to match the result from pycoco, but with no luck.

You can also copy the script to colab if you want to run the `visualize_image` section along with mAP computation.

LukeWood:
> During the development of FasterRCNN, I have noticed reasonable loss and reasonable prediction outcome, but always get low mAP metrics (~0.18). The evaluation loop takes ~13 minutes. So I instead hack a COCO wrapper (thanks to Model Garden) and directly call pycoco, this time it gets me 0.44 mAP which is more reasonable. The evaluation loop takes ~8 minutes. I wonder how we can investigate those inconsistencies. I have put my reproducible code in a separate branch, this is a standalone script that can run end-to-end: https://github.com/keras-team/keras-cv/blob/map/examples/models/object_detection/faster_rcnn/eval.py
> 
> I have also tried to configure iou_thresholds, recall_thresholds to try to match the result from pycoco, but with no luck.
> 
> You can also copy the script to colab if you want to run the `visualize_image` section along with mAP computation.

Thanks for the error report!  I'll take this on and try to figure out whats going on here.

bhack:
> The evaluation loop takes ~8 minutes.

@tanzhenyu Other then precision that could be fixed/covered by the right test we still have some perf. inefficiency as we cannot XLA compile these things. See https://github.com/keras-team/keras-cv/issues/895#issuecomment-1273430678

tanzhenyu:
> n loop takes ~8 minutes.

Those metrics that require lots of dynamic processing are usually something we cannot make it happen in XLA anyway, so right now this only works on CPU, we don't have plans to support it in GPU/TPU

tanzhenyu:
[...]

PR #949, opened 2022-10-19
[Draft] Improve MeanAveragePrecision metric accuracy for good models

LukeWood:
Good object detection models make a ton of predictions with confidence scores >0.9.  As such, most boxes end up in the high end of the buckets.  

download-3

Fixes https://github.com/keras-team/keras-cv/issues/915

Histogram of confidence scores of a well optimized model

download-2

---

Action items:

- try out just a simple exponential function prioritizing values over 0.8
- plot histogram of scores of a good RetinaNet
- make this configurable to users

Another point:

After NMS boxes below score threshold X are pruned, so any buckets UNDER that score threshold are ENTIRELY unused.  I should also make this configurable.

PR #979, opened 2022-11-02
[Draft] Implement a dynamic bucketization algorithm for MeanAveragePrecision

LukeWood:
Fixes https://github.com/keras-team/keras-cv/issues/915

Histogram of confidence scores of a well optimized model:
196818643-d4fc5db6-1879-482e-8d7d-ad62705da054

Sequel to: https://github.com/keras-team/keras-cv/pull/949

First, I tried to map boxes to buckets nonlinearly using an exponential function.  This increases the number of buckets are the end of the confidence thresholds (i.e..in the range [0.9, 1.0] confidence range). This did not work well enough; many good models map almost all boxes to 0.999 confidence.

This branch implements a dynamic bucketization algorithm.  In this algorithm all boxes are first assigned round-robin to new buckets, then after all buckets are filled the buckets are sorted and re-organized to be in ascending order.  Boxes are then assigned based on the first matching confidence score.

---

This does not currently pass tests.  Unsure why, opening as a draft to keep track of this, but I am shelving this for a little bit.


LukeWood:
Removed in favor of the final PyCOCOTools OD metrics

PR #980, opened 2022-11-02
[Backbone Training] added 50/50 between cutmix and mixup, changed default hold steps

DavidLandup0:
# What does this PR do?

Tested on Imagenette - shows a ~2% increase in accuracy over the original script. Unclear how much this translates to ImageNet (if it does).

- Hold steps updated to default to 45% (decay starts around the first reduction on plateau in the `ReduceLROnPlateau` alternative)
- Each batch is augmented either with Mixup or Cutmix with a 50/50 probability between them
- The script that produced the plot uses RandAugment with `magnitude` of `0.7` instead of `0.3` for both runs

image

P.S. I made the labels a bit weird - both use the same CutMix/MixUp policy, not just the other one. The only difference between these is the LR schedule (reduce on plateau vs warmup + decay).

@ianstenbit 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -186,6 +186,7 @@ AUGMENT_LAYERS = [
     keras_cv.layers.RandAugment(value_range=(0, 255), magnitude=0.3),
 ]
 
+
 @tf.function
 def augment(img, label):
     inputs = {"images": img, "labels": label}
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -192,7 +192,7 @@ def augment(img, label):
     inputs = {"images": img, "labels": label}
     for layer in AUGMENT_LAYERS:
         inputs = layer(inputs)
-    if tf.random().uniform(()) > 0.5:
+    if tf.random().uniform() > 0.5:
         inputs = keras_cv.layers.CutMix()(inputs)
     else:
         inputs = keras_cv.layers.MixUp()(inputs)
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -192,7 +192,7 @@ def augment(img, label):
     inputs = {"images": img, "labels": label}
     for layer in AUGMENT_LAYERS:
         inputs = layer(inputs)
-    if tf.random().uniform() > 0.5:
+    if tf.random.uniform() > 0.5:
         inputs = keras_cv.layers.CutMix()(inputs)
     else:
         inputs = keras_cv.layers.MixUp()(inputs)
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -99,7 +99,7 @@ flags.DEFINE_float(
 
 flags.DEFINE_float(
     "warmup_hold_steps_percentage",
-    0.1,
+    0.45,
     "For how many steps expressed in percentage (0..1 float) of total steps should the schedule hold the initial learning rate after warmup is finished, and before applying cosine decay.",
 )
 
@@ -184,15 +184,18 @@ def crop_and_resize(img, label):
 AUGMENT_LAYERS = [
     keras_cv.layers.RandomFlip(mode="horizontal"),
     keras_cv.layers.RandAugment(value_range=(0, 255), magnitude=0.3),
-    keras_cv.layers.CutMix(),
 ]
-
 
 @tf.function
 def augment(img, label):
     inputs = {"images": img, "labels": label}
     for layer in AUGMENT_LAYERS:
         inputs = layer(inputs)
+    if tf.random().uniform(()) > 0.5:
+        inputs = keras_cv.layers.CutMix()(inputs)
+    else:
+        inputs = keras_cv.layers.MixUp()(inputs)
+
     return inputs["images"], inputs["labels"]
 
 


DavidLandup0:
P.S. We might want to look into weight regularization (as per the "bag of tricks for cnns") in general for the training script. I'll do some test runs and see if we can generalize it to all models.

ianstenbit:
> P.S. We might want to look into weight regularization (as per the "bag of tricks for cnns") in general for the training script. I'll do some test runs and see if we can generalize it to all models.

Agreed. I am currently exploring weight decay as an option, doing some test runs on ImageNet

DavidLandup0:
Excited to see the results of regularization! I'll try to do some runs as well. Currently, trying out the advertised benefits of progressive learning with adaptive regularization for augmentation. Got the script to train ~30% faster, but the accuracy has dropped too.

Posting updates on progressive learning if there's something promising down the line :)

DavidLandup0:
The tweaks come from wanting to provide good pretrained weights for downstream tasks, which currently aren't at proposed paper levels. If we can find a sweet spot with a _general_ trainer script that can be reused for most architectures - that'd speed up the development and distribution of new models.

The original training procedures might not be ideal for each architecture - if ResNets were proposed in 2022, they'd probably have used CutMix as well, so drifting from the original training procedures can, IMO, lead to better weights than originally proposed in these papers. For example, in the EffNetV2 paper, the authors train a plain ResNet with progressive learning and achieve better results.

On the production end - the main considerations are FLOPS and accuracy. If the FLOPS/accuracy hierarchy is changed through a better training pipeline, especially if a low-FLOP model starts performing better, IMO, that's only for the better.

bhack:
> On the production end - the main considerations are FLOPS and accuracy. If the FLOPS/accuracy hierarchy is changed through a better training pipeline, especially if a low-FLOP model starts performing better, IMO, that's only for the better.

Yes, what I meant is that this in the end could really go to impact the criteria of accepting new models especially if we are going to overpopulate the same flops/accuracy area. In the case we narrow the margins what is the advantages of working on all these architectures or just introducing a new one?

I understand more the logic of having reference models in research as they are useful as recurrent baselines for writing new papers/experiment when introducing new models or forking original experiments, but what is the logic in production?

I would really like to understand when we come up with a table what kind of FLOPS / accuracy ratio we have on these archs with a training protocol that is in de-facto a custom one.

DavidLandup0:
I think you bring up a good question. We should probably sit down and figure this out when we hit paper-claimed levels on most archs and can see improvements beyond that point in the near horizon. :) @bhack 

Have you had the chance to try the new setup out? @ianstenbit

bhack:
> We should probably sit down and figure this out when we hit paper-claimed levels on most archs and can see improvements beyond that point in the near horizon

Or it could be the case we are already going to flat many models on the same flops/accuracy area with our custom training protocol. Who knows?  
The only solution is to start to populate FLOPS/accurancy markdown tables from who finally run the training jobs and it is already committing the weights in the repository.

DavidLandup0:
I can make the tables given the current training history runs. Should it be a new MD file?

DavidLandup0:
IMO - it should have:

- Model name
- Param count
- Estimated FLOPS
- Top-1 and Top-5 validation accuracy
- Size (in megabytes, relevant for edge device deployment)
- Inference time (CPU and GPU)


In other words - pretty much what the current Keras Applications page has :)

bhack:
Yes something like that. But the internal team need to join/agree to this effort as currently the models and the related weights are mainly trained and committed with their computing resources/manual work (also if we have margins to automate this https://github.com/keras-team/keras-cv/discussions/954).

This activity could help also expose to the community the best practices for https://github.com/keras-team/keras/issues/16699 or any other nice tool collect to automate the production of these values for a given checkpoint.

ianstenbit:
> IMO - it should have:
> 
> * Model name
> * Param count
> * Estimated FLOPS
> * Top-1 and Top-5 validation accuracy
> * Size (in megabytes, relevant for edge device deployment)
> * Inference time (CPU and GPU)
> 
> In other words - pretty much what the current Keras Applications page has :)

I think this would be great -- ideally most of this information is either
- Included in the training history JSON files
- Computable based on the models theselves

I would really like it if we could do this automatically with a script to parse our training records and update such a table.

One missing piece is that we're not currently exporting top-5 accuracy in our history JSON file. This would be an easy fix as we'd just have to add top-5 accuracy to the training script and it should show up in tensorboard logs and then can be parsed into the JSON

ianstenbit:
> I think you bring up a good question. We should probably sit down and figure this out when we hit paper-claimed levels on most archs and can see improvements beyond that point in the near horizon. :) @bhack
> 
> Have you had the chance to try the new setup out? @ianstenbit

I have not been able to run an experiment with this yet. I am hoping to do so in the next week or so, but can't make any promises. Thank you for your great contributions here -- I haven't lost track of this and I'll make sure it gets properly tested and merged if it looks good.

DavidLandup0:
> I would really like it if we could do this automatically with a script to parse our training records and update such a table.
One missing piece is that we're not currently exporting top-5 accuracy in our history JSON file. This would be an easy fix as we'd just have to add top-5 accuracy to the training script and it should show up in tensorboard logs and then can be parsed into the JSON

We couldn't get it for previous runs though, right? Or is it tracked but not displayed? Anyhow, yes - automating it with a script would be great :)
It's worth standardizing FLOPS estimation, and on which hardware inference/training is tested. Since we'd expect people to use it on consumer hardware, using an "average" setup might be a good idea

bhack:
> It's worth standardizing FLOPS estimation, and on which hardware inference/training is tested. Since we'd expect people to use it on consumer hardware, using an "average" setup might be a good idea

We had already an interesting invite from @gfursin in the Keras ticket https://github.com/keras-team/keras/issues/16389#issuecomment-1251152900

examples/training/classification/imagenet/basic_training.py
```diff
@@ -193,6 +192,11 @@ def augment(img, label):
     inputs = {"images": img, "labels": label}
     for layer in AUGMENT_LAYERS:
         inputs = layer(inputs)
+    if tf.random().uniform(()) > 0.5:
+        inputs = keras_cv.layers.CutMix()(inputs)
+    else:
+        inputs = keras_cv.layers.MixUp()(inputs)
```
innat, line 198: Don't we have any functionality to achieve this? https://github.com/keras-team/keras-cv/issues/408
cc @LukeWood

innat: Maybe! [RandomAugmentationPipeline](https://keras.io/api/keras_cv/layers/preprocessing/random_augmentation_pipeline/)

DavidLandup0: Good catch! Trying a run with:

```
AUGMENT_LAYERS = [
    keras_cv.layers.RandomFlip(mode="horizontal"),
    keras_cv.layers.RandAugment(value_range=(0, 255), magnitude=0.3),
    keras_cv.layers.RandomAugmentationPipeline(layers=[keras_cv.layers.CutMix(), keras_cv.layers.MixUp()], augmentations_per_image=1, rate=1)
]
```

DavidLandup0:
```
AUGMENT_LAYERS = [
    keras_cv.layers.RandomFlip(mode="horizontal"),
    keras_cv.layers.RandAugment(value_range=(0, 255), magnitude=0.3),
    keras_cv.layers.RandomAugmentationPipeline(layers=[keras_cv.layers.CutMix(), keras_cv.layers.MixUp()], augmentations_per_image=1, rate=1)
]

@tf.function
def augment(img, label):
    inputs = {"images": img, "labels": label}
    for layer in AUGMENT_LAYERS:
        inputs = layer(inputs)
        
    return inputs["images"], inputs["labels"]


train_set = (
    train_set.map(crop_and_resize, num_parallel_calls=tf.data.AUTOTUNE)
    .batch(BATCH_SIZE)
    .map(augment, num_parallel_calls=tf.data.AUTOTUNE)
    .prefetch(tf.data.AUTOTUNE)
)
```

🤔 

```
ValueError: CutMix received a single image to `call`.  The layer relies on combining multiple examples, and as such will not behave as expected.  Please call the layer with 2 or more samples.
```

DavidLandup0: But works perfectly fine for the code in the PR, which should be functionally equivalent

innat: Some augmentation require batch of inputs, such as cut-mix, mix-up, fourier-mix, mosaic-mix.

```python
keras_cv.layers.RandomAugmentationPipeline(
   layers=[
        keras_cv.layers.CutMix(), 
        keras_cv.layers.MixUp()], 
       augmentations_per_image=1, # < shuld be > 1?
       rate=1 # is it ok? 
)
```

DavidLandup0: Yes - CutMix and MixUp need batches of images and labels to mix up, but I think that we are providing batches? 
The initial code deals with the same batched input and they correctly work. I'll look into the potential difference in input

DavidLandup0: The augmentations per image are set to 1 with a rate of 1 to ensure that on each augmentation call, we apply exactly 1 aug out of the two with a 50/50 probability

examples/training/classification/imagenet/basic_training.py
```diff
@@ -193,6 +192,11 @@ def augment(img, label):
     inputs = {"images": img, "labels": label}
     for layer in AUGMENT_LAYERS:
         inputs = layer(inputs)
+    if tf.random().uniform(()) > 0.5:
```
ianstenbit, line 195: This should just be `tf.random.uniform()`

DavidLandup0: Whoops, good catch!

DavidLandup0:
Awesome, thanks for the update! Excited to hear how it goes :D

ianstenbit (changes requested):
After some testing, this seems to marginally improve performance, so I think let's go ahead and merge it.

Just need to fix the `tf.random().uniform()` -> `tf.random.uniform()` issue.

Thank you!

DavidLandup0:
Awesome, thanks for the update! Any TB.dev logs for reference?
What counts as marginal improvement? 😅 

Updated the random() call

ianstenbit:
> Awesome, thanks for the update! Any TB.dev logs for reference? What counts as marginal improvement? 😅

0.3% top-1 improvement for a DarkNet training run. Don't have a side-by-side comparison but the results after the change are [here](https://tensorboard.dev/experiment/665BYHNUQlmSpxkyyvrKng/#scalars)

DavidLandup0:
Awesome, thanks!
Did the prolonged hold steps help at all? 🤔

bhack:
> After some testing, this seems to marginally improve performance, so I think let's go ahead and merge it.
> 
> Just need to fix the `tf.random().uniform()` -> `tf.random.uniform()` issue.
> 
> Thank you!

Not for this probably but for the next one can we evaluate the flops/accuracy ratio before introducing new tricks?

bhack:
E.g. users cloud not be interested in a 0.3% improvement if we add a large overhead in epoch/steps.

DavidLandup0:
> Not for this probably but for the next one can we evaluate the flops/accuracy ratio before introducing new tricks?

This one should be pretty equal since it's either one or the other, and the only extra call is the `tf.uniform.random()` on each batch, but I agree that down the line, a super small improvement that makes the script run for way longer (or makes it more expensive to run) is a bad idea.

I like the idea of testing out "Super Convergence" - a largely unnoticed paper (only <600 citations): https://arxiv.org/pdf/1708.07120.pdf

I've known about it for a while but never tried testing it. If we can speed up the training by any factor,  it'd be welcome 🤔 
They specifically train ResNet, WideResNet, Inception and DenseNet (all of which we have here and can test) on ImageNet, CIFAR10 and CIFAR100

DavidLandup0:
I'll do some runs on Super Convergence today and if it bears any fruit, send the results

bhack:
> This one should be pretty equal since it's either one or the other, and the only extra call is the tf.uniform.random() on each batcu

But we need also to consider also if we have a perf gap between mixup and cutmix.

ianstenbit:
re: speed of cutmix vs mixup

I agree there is the possibility of a performance delta here. Based on the TensorBoard logs, I don't think there was a meaningful performance difference, but I don't have great benchmark data.

I did do a quick 10-epoch test of training EfficientNetV2B0 with/without this PR and there was no noticeable difference in step time.

ianstenbit (approved):

PR #983, opened 2022-11-03
Shuffle dataset every epoch for basic_training

ianstenbit:
I've verified that this doesn't meaningfully change epoch time, and IIUC this should be enabled in all cases for imagenet training.


tanzhenyu (approved):
I wonder if we should also this for pascal voc loading

ianstenbit:
> I wonder if we should also this for pascal voc loading

I think most likely we should (although I'm not certain of the specifics for OD models, afaik this is best practice in almost all cases)

Issue #821, opened 2022-09-22
Disable regularization loss for RetinaNet when it is zero


LukeWood:
per feedback on https://github.com/keras-team/keras-io/pull/1065#discussion_r978022228

Issue #824, opened 2022-09-23
RetinaNet with tf.set_soft_device_placement(False) fails


LukeWood:
Setting this then running the RetinaNet guide should work

sachinprasadhs:
Can we close this issue since the above linked PR is merged? Thanks!

PR #984, opened 2022-11-03
[Draft] Polish the RetinaNet 

LukeWood:
Fixes https://github.com/keras-team/keras-cv/issues/821, https://github.com/keras-team/keras-cv/issues/824


keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -419,10 +442,16 @@ def _backward(self, y_true, y_pred):
 
         self.classification_loss_metric.update_state(classification_loss)
```
tanzhenyu, line 443: unrelated to this PR -- consider refactoring RetinaNet, for example, `compute_loss` called in `_backward` is weird, as it's part of forward path.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -419,10 +442,16 @@ def _backward(self, y_true, y_pred):
...
         self.box_loss_metric.update_state(box_loss)
-        self.regularization_loss_metric.update_state(regularization_loss)
+
+        if self._track_regularization:
+            self.regularization_loss_metric.update_state(regularization_loss)
         self.loss_metric.update_state(loss)
         return loss
 
+    @property
+    def _track_regularization(self):
```
tanzhenyu, line 452: under what circumstances would `len(self.losses)` = 0?

keras_cv/models/object_detection/retina_net/retina_net_inference_test.py
```diff
@@ -31,7 +31,10 @@ class RetinaNetTest(tf.test.TestCase):
     @pytest.fixture(autouse=True)
     def cleanup_global_session(self):
```
tanzhenyu, line 32: Add a comment as to why do we need a `cleanup_global_session`

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -278,6 +283,24 @@ def decode_training_predictions(self, x, train_predictions):
...
+        distinctions.  Primarily, all metrics must support bounding boxes, and all
+        two losses must be provided: `box_loss` and `classification_loss`.
+
+        Args:
+            box_loss: a keras loss to use for box offset regression.  Preconfigured
+                losses are provided when the string "huber" or "smoothl1" are passed.
+            classification_loss: a keras loss to use for box classification.
+                A preconfigured `FocalLoss` is provided when the string "focal" is
+                passed.
+            metrics: a list of Keras Metrics that accept bounding boxes as inputs, i.e.
```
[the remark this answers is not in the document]
LukeWood: Hmm, I'd like to keep using the KerasCV COCO metrics in my debugging and testing.  Perhaps the constructor of THOSE metrics should come with a disclaimer?

Recall is fully accurate, MaP is the one that would need a disclaimer.

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: no regularization losses.  losses does not hold `compiled_loss`, only extra losses added via add_loss() & regularization.

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: Good point, that is a really bad name.  I'll have to update it.  Thanks!

LukeWood:
ok @tanzhenyu all comments addressed

keras_cv/models/object_detection/retina_net/retina_net.py
tanzhenyu: I see. makes sense to use as-is for now, but we should definitely fix it as Keras level.

tanzhenyu (approved):

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -278,6 +283,24 @@ def decode_training_predictions(self, x, train_predictions):
     def compile(
         self, box_loss=None, classification_loss=None, loss=None, metrics=None, **kwargs
     ):
+        """compiles the RetinaNet.
+
+        compile() mirrors the standard Keras compile() method, but has a few key
+        distinctions.  Primarily, all metrics must support bounding boxes, and all
+        two losses must be provided: `box_loss` and `classification_loss`.
```
ianstenbit, line 290: s/all two/both?

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -278,6 +283,24 @@ def decode_training_predictions(self, x, train_predictions):
...
+
+        Args:
+            box_loss: a keras loss to use for box offset regression.  Preconfigured
```
ianstenbit, line 293: here and in classification_loss: s/keras/Keras

ianstenbit (approved):
(Probably for another PR, but following up on the discussion from #1000)

Right now RetinaNet has a slightly different API expectation for training data than the FasterRCNN. RetinaNet wants `boxes` to be a tensor with boxes and classes, while FasterRCNN wants `boxes` to be a dict with `gt_boxes` and `gt_classes`.

Can we make the two consistent?

Issue #581, opened 2022-07-13
BUG: `RandomContrast` slows down training 6-fold


DavidLandup0:
Augmentation will obviously slow down training, but it shouldn't be a 6-fold slowdown. This happens with the `RandomContrast` layer, which makes the training per epoch grow from ~100s to ~600s.

I'd share a Colab notebook but there seems to be an issue with KerasCV on Colab on importing, so here are the steps to reproduce:

```
import keras_cv
print(f'KerasCV version {keras_cv.__version__}')
import tensorflow as tf
print(f'TF version {tf.__version__}')
from tensorflow import keras
print(f'Keras version {keras.__version__}')
```

```
(x_train, y_train), (x_test, y_test) = tf.keras.datasets.cifar10.load_data()
```

```
aug = keras.Sequential([
    keras_cv.layers.RandomContrast(factor=0.2)
])
```

```
model = keras.Sequential([
    keras.layers.InputLayer(input_shape=(32, 32, 3)),
    aug,
    keras.applications.EfficientNetV2B0(weights=None, include_top=False),
    keras.layers.GlobalAveragePooling2D(),
    keras.layers.Dense(10, activation='softmax')
    
])
model.compile(optimizer='adam', loss='sparse_categorical_crossentropy')

model.fit(x_train, y_train)
```

```
model = keras.Sequential([
    keras.layers.InputLayer(input_shape=(32, 32, 3)),
    keras.applications.EfficientNetV2B0(weights=None, include_top=False),
    keras.layers.GlobalAveragePooling2D(),
    keras.layers.Dense(10, activation='softmax')
    
])
model.compile(optimizer='adam', loss='sparse_categorical_crossentropy')

model.fit(x_train, y_train)
```

bhack:
> I will pin this issue. This is a significant slowdown, maybe we need to consider manually vectorizing - which would be very unfortunate.

I think that it will more useful to brainstorm a solution for the old and more general problem at https://github.com/keras-team/keras-cv/issues/291 then pinning every single issue as this is already the 2nd one (https://github.com/tensorflow/tensorflow/issues/56242).

bhack:
As the superset of this issue is https://github.com/keras-team/keras-cv/issues/291 and the root cause is our choice of `vectorized_map` and within the batch randomization as we discussed in many early tickets on our augmentation design we could:

- Understand what to do with the missing `vectorized_map` converters and what kind of issue they could solve
[...]

Issue #372, opened 2022-04-27
Repeated augmentation layer


sayakpaul:
Repeated augmentation layer [1] has also become an important recipe to train SoTA image classification models. The abstract of [1] pretty much sums up what it is:

> Large-batch SGD is important for scaling training of deep neural networks. However, without fine-tuning hyperparameter schedules, the generalization of the model may be hampered. We propose to use batch augmentation: replicating instances of samples within the same batch with different data augmentations. Batch augmentation acts as a regularizer and an accelerator, increasing both generalization and performance scaling for a fixed budget of optimization steps. We analyze the effect of batch augmentation on gradient variance and show that it empirically improves convergence for a wide variety of networks and datasets. Our results show that batch augmentation reduces the number of necessary SGD updates to achieve the same accuracy as the state-of-the-art. Overall, this simple yet effective method enables faster training and better generalization by allowing more computational resources to be used concurrently.

**References**

[1] [Augment Your Batch: Improving Generalization Through Instance Repetition](https://openaccess.thecvf.com/content_CVPR_2020/papers/Hoffer_Augment_Your_Batch_Improving_Generalization_Through_Instance_Repetition_CVPR_2020_paper.pdf)

bhack:
We have talked about this some month ago at  https://github.com/keras-team/keras-cv/pull/146#issuecomment-1048315498.

At that time we have not received any feedback.

bhack:
P.s. We have still many performance  fallback related to the batch randomizzation strategy:
 https://github.com/keras-team/keras-cv/issues/291

sayakpaul:
Makes sense to have a dedicated issue for the layer rather than keeping it within comments. That way it becomes more visible and all the related concerns (like the ones you mentioned) can be discussed specifically.

bhack:
Probably, but as you can see we was in a design phase on Feb and comments emergerged in that context.
Now, at least, the history is complete also here for who is landing on this ticket directly.

PR #986, opened 2022-11-04
Add base augmentation layer for 3D preception.

lengzq:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/keras_cv/layers/preprocessing3d/base_augmentation_layer_3d_test.py
+++ b/keras_cv/layers/preprocessing3d/base_augmentation_layer_3d_test.py
@@ -18,8 +18,8 @@ from keras_cv.layers.preprocessing3d.base_augmentation_layer_3d import (
     BaseAugmentationLayer3D,
 )
 
-POINT_CLOUDS = BaseAugmentationLayer3D.POINT_CLOUDS
-BOUNDING_BOXES = BaseAugmentationLayer3D.BOUNDING_BOXES
+POINT_CLOUDS = "point_clouds"
+BOUNDING_BOXES = "bounding_boxes"
 
 
 class RandomAddLayer(BaseAugmentationLayer3D):
--- a/keras_cv/layers/preprocessing3d/base_augmentation_layer_3d_test.py
+++ b/keras_cv/layers/preprocessing3d/base_augmentation_layer_3d_test.py
@@ -21,25 +21,36 @@ from keras_cv.layers.preprocessing3d.base_augmentation_layer_3d import (
 POINT_CLOUDS = "point_clouds"
 BOUNDING_BOXES = "bounding_boxes"
 
+
 class RandomAddLayer(BaseAugmentationLayer3D):
     def __init__(self, translate_noise=(0.0, 0.0, 0.0), **kwargs):
         super().__init__(**kwargs)
         self._translate_noise = translate_noise
 
     def get_random_transformation(self, **kwargs):
-        random_x = self._random_generator.random_normal((), mean=0.0, stddev=self._translate_noise[0])
-        random_y = self._random_generator.random_normal((), mean=0.0, stddev=self._translate_noise[1])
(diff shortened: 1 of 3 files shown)


lengzq:
@tanzhenyu

bhack:
With all the performance gap we had with https://github.com/keras-team/keras-cv/issues/581#issuecomment-1271579684 and the other related KPL tickets are we sure that we want to go ahead with a within the batch augmentation policy?

tanzhenyu:
> #581 (comment)

So quickly looking over the issue, do you mean `vectorized_map` is the main reason for performance gap?

bhack:
It is not strictly or exclusively the vectorized map but I think our within the batch randomizattion policy when the underline TF ops  that we want to use doen't have a batched argument so also a working vectorized map can "miracles" (this was quite frequent with `tf.image` ops in KLP)

Also the main problem is that after many months and users tickets we haven't done any step ahead, also with the `vectorized_map` API owners involved in TF, in this analysis. On the other side it seems that we don't have any specific paper support point on the superior performance or faster convergence of the within the batch randomization policy.

As I know the only thing that we have for the batch augmentation policy with a support paper is:
https://github.com/keras-team/keras-cv/issues/372#issuecomment-1110827256

tanzhenyu:
> It is not strictly or exclusively the vectorized map but I think our within the batch randomizattion policy when the underline TF ops that we want to use don't have batched argument (this was quite frequent with `tf.image` ops in KLPs) . So also a working vectorized map converter cannot do "miracles".
> 
> Also the main problem is that after many months and users tickets we haven't done any step ahead, also with the `vectorized_map` API owners involved in TF, in this analysis. On the other side it seems that we don't have any specific paper/support point related to the superior performance or faster convergence of the within the batch randomization policy.
> 
> As I know the only thing that we have for the batch augmentation policy with a support paper is: [#372 (comment)](https://github.com/keras-team/keras-cv/issues/372#issuecomment-1110827256)

then let's just keep using `tf.map_fn` and other batched implementations

bhack:
> then let's just keep using tf.map_fn and other batched implementations

I think you missed the point. It will be probably slow as the vectorized_map fallback.

The problem is that if we want randomize over every single batch sample when the TF ops use in the preprocessing function don't accept vectorized arguments we will go to heavily impact  the training process performance.  
The point is: for what? You haven't expose any empirical experiment or paper that sustain this big performance gap compensating it with an higher convergence rate.

At least https://github.com/keras-team/keras-cv/issues/372#issuecomment-1110827256 could limit the performance gap a we could use at least a TF ops in "sub-batches".

keras_cv/layers/preprocessing3d/base_augmentation_layer_3d_test.py
```diff
@@ -0,0 +1,112 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import numpy as np
+import tensorflow as tf
+
+from keras_cv.layers.preprocessing3d.base_augmentation_layer_3d import (
+    BaseAugmentationLayer3D,
+)
+
+POINT_CLOUDS = "point_clouds"
```
tanzhenyu, line 21: just use BaseAugmentationLayer3D.POINT_CLOUDS?

tanzhenyu (approved):
LGTM, some minor comments.

bhack:
If you prefer I could delete my comments so the PR is more readable. Let me know.

tanzhenyu:
> As it is approved, If you prefer I could delete my comments so the PR is more readable. Let me know.

It's ok, I think both the contributor and I are aware of the current status

bhack:
> It's ok, I think both the contributor and I are aware of the current status

A more explicit rollout is at https://github.com/keras-team/keras-cv/issues/581#issuecomment-1304554115 if you want to subscribe.

PR #989, opened 2022-11-07
Lazy-load components of StableDiffusion and make them visible to users

ianstenbit:
Open question:
Should we accept components as constructor parameters for StableDiffusion?
I am leaning toward yes. That way, a user could do something like this:

```
from keras_cv.models.generative.stable_diffusion import TextEncoder

text_encoder = TextEncoder()
# Do something to the text_encoder (e.g. add tokens via textual inversion)

sd = StableDiffusion(512, 512, text_encoder=text_encoder)
```

Thoughts?


LukeWood (approved):
This is much better than what we had before, thank you!

Issue #995, opened 2022-11-09
BUG: StableDiffusion() broken after recent update


BFauber:
Unfortunately, a recent update appears to have broken the `StableDiffusion()` class function found in `keras_cv.models.generative`.  Loading the  `StableDiffusion()` function now throws the following error and requires ~20 min to run, whereas it previously ran in <5 seconds (NVIDIA V100 GPU, CUDA v11.8) with the `keras-cv` version available on GitHub on Friday 11/4:

`WARNING:tensorflow:5 out of the last 54 calls to <function Model.make_predict_function.<locals>.predict_function at 0x7fc4486947a0> triggered tf.function retracing. Tracing is expensive and the excessive number of tracings could be due to (1) creating @tf.function repeatedly in a loop, (2) passing tensors with different shapes, (3) passing Python objects instead of tensors. For (1), please define your @tf.function outside of the loop. For (2), @tf.function has reduce_retracing=True option that can avoid unnecessary retracing. For (3), please refer to https://www.tensorflow.org/guide/function#controlling_retracing and https://www.tensorflow.org/api_docs/python/tf/function for  more details.`

Here is the code that throws the error:

```python
from tensorflow import keras
from keras_cv.models import StableDiffusion

# Enable mixed precision
keras.mixed_precision.set_global_policy("mixed_float16")

model = StableDiffusion(
    img_height=512, 
    img_width=512, 
    jit_compile=True
)

prompt = "photograph of coffee beans"
seed = 44079

img0 = model.text_to_image(
    prompt=prompt,
    batch_size=1,
    num_steps=25,
    unconditional_guidance_scale=7.5,
    seed=seed,
)
```

Updates #973 and #989 appear to have occurred within the past few days, but it is unknown if these updates are related to the error.

bhack:
Yes it was introduced with the `lazy_load` refactoring in https://github.com/keras-team/keras-cv/pull/989

ianstenbit:
Thank you @BFauber  for posting the issue, and thank you @bhack for the fix (I'll merge once comments are resolved)
I agree that #989 is the cause of this issue.

LukeWood:
Surprised that would be causing that.  Interested to hear the conclusion..

bhack:
> Surprised that would be causing that. Interested to hear the conclusion..

It is that every time we access to the property we execute `.compile` and it is going to trigger the autograph and so on...

LukeWood:
Ahhhh I see.

PR #997, opened 2022-11-09
Avoid retracing with python lazy_loading

bhack:
# What does this PR do?
Lazy_loading refactoring at https://github.com/keras-team/keras-cv/pull/989 introduces retraces.

Fixes https://github.com/keras-team/keras-cv/issues/995 

## Before submitting
- [ x] Did you read the [contributor guideline](https://github.com/keras-team/keras-cv/blob/master/.github/CONTRIBUTING.md),
      Pull Request section?
      to it if that's the case.

## Who can review?

@ianstenbit

### DIFF
--- a/keras_cv/models/generative/stable_diffusion/stable_diffusion.py
+++ b/keras_cv/models/generative/stable_diffusion/stable_diffusion.py
@@ -98,10 +98,7 @@ class StableDiffusion:
         self._tokenizer = None
 
         self.jit_compile = jit_compile
-        self._decoder_compiled = False
-        self._diffusion_model_compiled = False
-        self._text_encoder_compiled = False
-        self._image_encoder_compiled = False
+
         print(
             "By using this model checkpoint, you acknowledge that its usage is "
             "subject to the terms of the CreativeML Open RAIL-M license at "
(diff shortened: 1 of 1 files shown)


keras_cv/models/generative/stable_diffusion/stable_diffusion.py
```diff
@@ -423,8 +426,9 @@ def image_encoder(self):
         """
         if self._image_encoder is None:
             self._image_encoder = ImageEncoder(self.img_height, self.img_width)
-        if self.jit_compile:
+        if self.jit_compile and not self._image_encoder_comiled:
```
ianstenbit, line 429: Instead of adding a tracking variable for whether these are compiled, can we just nest this if statement inside the block that only runs if the encoder was none?

e.g. 
```
if self._image_encoder is None:
    self._image_encoder = ImageEncoder(self.img_height, self.img_width)
    if self.jit_compile:
        self._image_encoder.compile(jit_compile=True)
```

ianstenbit (commented):
Thank you for this fix! This was a silly oversight in my PR and I appreciate the fix.

keras_cv/models/generative/stable_diffusion/stable_diffusion.py
bhack: Mhh.. do we have use cases where we need to explicitly control the compile phase?

ianstenbit: Not yet -- if we do, we can add guarding logic for that when it's necessary.

For now, compiling only at initialization time for the component models should bring us back to the functionality from pre-#989

bhack: Ok

bhack: Done

ianstenbit:
Looks like there's a linter error. Other than that, looks good!

PR #1000, opened 2022-11-10
[Object Detection] Introduce a callback-based coco metric evaluator

ianstenbit:
Example Colab using this:
https://colab.research.google.com/drive/16Lb6rMcFvtoic7DLd89Ap5-VxLa1dX_4?resourcekey=0-sDnh3nGFCx-6c7l_jnxShQ&usp=sharing

Example TensorBoard (from 3 epochs of training a RetinaNet on TPU: https://tensorboard.dev/experiment/Dmvt2HYDTxamopnh5QgJkQ/#scalars&_smoothingWeight=0&runSelectionState=eyJ0cmFpbiI6dHJ1ZX0%3D

Basic API is:

```
from keras_cv.callbacks import PyCOCOCallback
callbacks = [
    PyCOCOCallback(eval_ds, bounding_box_format="xywh"),
    keras.callbacks.TensorBoard(log_dir="logs"),
]

model.fit(
    train_ds,
    validation_data=eval_ds,
    callbacks=callbacks,
)
```

~The evaluation dataset passed to `model.fit` is used for the COCO metrics at epoch end.~ I've updated the API to request its own evaluation dataset, as our call into the validation dataset from the model wasn't stable.
The COCO metrics are created in the logs and should automatically show up in TensorBoard (note that this is currently dependent on the `COCOPyMetric` callback before the `TensorBoard` callback in the callbacks list -- this can be fixed with proper PyMetric support in core Keras)


keras_cv/metrics/coco/pycoco_utils.py
```diff
@@ -0,0 +1,218 @@
...
+
+            ann["area"] = ann["bbox"][2] * ann["bbox"][3]
+            ann["segmentation"] = [[x1, y1, x1, y2, x2, y2, x2, y1]]
+
+        res.dataset["annotations"] = copy.deepcopy(predictions)
+        res.createIndex()
+        return res
+
+
+def _yxyx_to_xywh(boxes):
```
ianstenbit, line 85: Should just adapt this to use bounding_box.convert_format

bhack: Yes I think it will be useful

bhack:
Do we need add a test for this also with a `dummy model` with `compile=True`?

ianstenbit:
> Do we need add a test for this also with a `dummy model` with `compile=True`?

Yes absolutely -- this will need tests. I want to get some feedback from @LukeWood and @tanzhenyu on the API surface first

LukeWood:
Some feedback:

- lets name the callback COCOEvaluator() or something
- can we support evaluation for the train set from the callback too?

keras_cv/metrics/coco/pycoco_utils.py
```diff
@@ -0,0 +1,231 @@
...
+
+    gt_dataset = _convert_groundtruths_to_coco_dataset(groundtruths)
+    coco_gt = COCOWrapper(gt_dataset=gt_dataset)
+    coco_predictions = _convert_predictions_to_coco_annotations(predictions)
+    coco_dt = coco_gt.loadRes(predictions=coco_predictions)
+    image_ids = [ann["image_id"] for ann in coco_predictions]
+
+    coco_eval = cocoeval.COCOeval(coco_gt, coco_dt, iouType="bbox")
+    coco_eval.params.imgIds = image_ids
+    coco_eval.evaluate()
```
LukeWood, line 220: does this print a ton of logs?

tanzhenyu:
One high-level question is : we should probably first provide a VOCEvaluator given our existing training are all on Pascal VOC dataset. That would require some python level coding, but no need to rely on pycoco

innat:
@ianstenbit 
Would it be better to introduce `callback` directory?

```python
from keras_cv.metrics import COCOPyMetric # current
from keras_cv.callback import COCOPyMetric # what if
```

Also note, sooner or later, it may be needed introduce **display** type callback for the segmentation or generative model training.

keras_cv/metrics/__init__.py
```diff
@@ -12,5 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from keras_cv.metrics.coco.coco_pymetric import COCOPyMetric
```
tanzhenyu, line 15: make a callback folder

keras_cv/metrics/coco/coco_pymetric.py
```diff
@@ -0,0 +1,106 @@
...
+                    gt[k].append(v)
+
+        metrics = compute_pycoco_metrics(gt, preds)
+        # Mark these as validation metrics by prepending a val_ prefix
+        metrics = {"val_" + name: val for name, val in metrics.items()}
+
+        logs.update(metrics)
+
+    def _eval_batch(self, batch, index):
+        images, gt_boxes = batch
```
tanzhenyu, line 60: This should be structured as 
images, gt_boxes, gt_classes = batch,
instead of gt_boxes [..., 5]
as the latter should be removed before 0.4

keras_cv/metrics/coco/coco_pymetric.py
```diff
@@ -0,0 +1,106 @@
...
+        width = images.shape[2]
+
+        gt_boxes = bounding_box.convert_format(
+            gt_boxes, source=self.bounding_box_format, target="yxyx"
+        )
+
+        gt_classes = gt_boxes[:, :, 4]
+        gt_boxes = gt_boxes[:, :, :4]
+
+        source_ids = tf.strings.join(
```
tanzhenyu, line 72: I'd recommend taking the source id from the input dataset, this ideally should just be from "00032.jpg" -> 32

keras_cv/metrics/coco/pycoco_utils.py
```diff
@@ -0,0 +1,231 @@
...
+    coco_eval.accumulate()
+    coco_eval.summarize()
+    coco_metrics = coco_eval.stats
+
+    metrics = coco_metrics
+
+    metrics_dict = {}
+    for i, name in enumerate(METRIC_NAMES):
```
tanzhenyu, line 228: do we need this given coco_eval.summarize() already gives the expected log?

keras_cv/metrics/coco/pycoco_utils.py
```diff
@@ -0,0 +1,231 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import copy
+
+import numpy as np
+import tensorflow as tf
+from pycocotools import coco
```
tanzhenyu, line 18: should do a try import and raise error and ask user to pip install

keras_cv/metrics/coco/pycoco_utils.py
```diff
@@ -0,0 +1,231 @@
...
+    "ARmax1",
+    "ARmax10",
+    "ARmax100",
+    "ARs",
+    "ARm",
+    "ARl",
+]
+
+
+class COCOWrapper(coco.COCO):
```
tanzhenyu, line 37: structure wise, let's make it:
1. keras_cv/metrics/pycoco_wrapper.py
2. keras_cv/callbacks/pycoco_callback.py
3. in the future if this becomes a PyMetrics, then keras_cv/metrics/pycoco_metrics.py

And we can make sure 1 can be re-used in 2 and 3 when implementing 3.

tanzhenyu (commented):
Thanks for the PR!

keras_cv/metrics/coco/pycoco_utils.py
ianstenbit: Yes it does -- I don't think we can't really do anything about it without forking the Pycocotools code (or with a hack like this: https://stackoverflow.com/questions/8391411/how-to-block-calls-to-print)

ianstenbit:
> * can we support evaluation for the train set from the callback too?

We can do this, yes. But I think the right approach for that is to support it in the Keras PyMetrics construct, which is forthcoming. For now, I think this should be eval-focused as this is a stopgap solution until we can build a proper PyMetric for this

keras_cv/metrics/coco/coco_pymetric.py
ianstenbit: I agree that this should change, but I think it needs to change at the same time that we restructure this package-wide. Right now, for it to work as a callback it needs to accept data in the same format that the models do, which (at least for RetinaNet, which is what I'm testing with) is `images, boxes`, where boxes include the class.

I think this also may end up as something like
```
images, annotated_boxes = batch
boxes, classes = annotated_boxes
```
if we end up going with the nested concept of box classes (idk where we are on that conversation)

keras_cv/metrics/coco/coco_pymetric.py
ianstenbit: IMO this is a bad user experience, because it requires that the dataset has the source ID as a field. This isn't really exposed to the user in any meaningful way, it's just used internally by pycocoutils as a key for a row in the dataset.

If we require "source_id" as a field in the dataset, then users will have to manufacture a source ID themselves for datasets that don't already have one

keras_cv/metrics/coco/pycoco_utils.py
ianstenbit: Just verified this, and `coco_metrics` here is a list of numbers (without any associated metric names) -- so this is required to associated the metric names with the metrics for the output logs

keras_cv/metrics/coco/pycoco_utils.py
ianstenbit: sgtm -- done

keras_cv/metrics/coco/pycoco_utils.py
ianstenbit: Good call -- done

keras_cv/metrics/coco/coco_pymetric.py
LukeWood: +1 to waiting to restructure on a package wide basis.

ianstenbit:
> One high-level question is : we should probably first provide a VOCEvaluator given our existing training are all on Pascal VOC dataset. That would require some python level coding, but no need to rely on pycoco

I'm not sure I follow. This callback should work independently of which KerasCV model we're using for OD, as well as which dataset we're training it on.

keras_cv/metrics/coco/__init__.py
```diff
@@ -13,5 +13,12 @@
 # limitations under the License.
 
 from keras_cv.metrics.coco.mean_average_precision import COCOMeanAveragePrecision
-from keras_cv.metrics.coco.pycoco_wrapper import PyCOCOWrapper
-from keras_cv.metrics.coco.pycoco_wrapper import compute_pycoco_metrics
+
+try:
+    from keras_cv.metrics.coco.pycoco_wrapper import PyCOCOWrapper
```
LukeWood, line 18: This seems risky.  Can we instead do this on object instantiation?
```
def get_coco_eval():
    try:
        import cocoeval
        return cocoeval
     except e:
          raise ValueError(...)
```


then in the class:
```
cocoeval = get_coco_eval()
```

keras_cv/metrics/coco/pycoco_utils.py
LukeWood: unfortunate, but if theres no easy solution it should be fine.

LukeWood:
> One high-level question is : we should probably first provide a VOCEvaluator given our existing training are all on Pascal VOC dataset. That would require some python level coding, but no need to rely on pycoco

what would this look like?

tanzhenyu:
> > One high-level question is : we should probably first provide a VOCEvaluator given our existing training are all on Pascal VOC dataset. That would require some python level coding, but no need to rely on pycoco
> 
> I'm not sure I follow. This callback should work independently of which KerasCV model we're using for OD, as well as which dataset we're training it on.

COCO metrics is only for coco dataset. VOC metrics is defined slightly different and only need very few AP with fixed IOU threshold

keras_cv/metrics/coco/__init__.py
ianstenbit: I don't think this will work, because we are defining a class which is a subclass of `COCO`.
Maybe there's a way to do this which I'm not aware of, but as far as I can tell, we can't define that class if the import fails.

keras_cv/metrics/coco/pycoco_utils.py
tanzhenyu: last time I tried it, it gives something like AP50, AP75, APSmall, APLarge, etc, is that not the case anymore?

keras_cv/metrics/coco/coco_pymetric.py
tanzhenyu: It is part of tfds, correct? https://www.tensorflow.org/datasets/catalog/coco
So in the future, we will probably provide some tools for users to find out bad predictions based on the metric output, it'd be nice to associate image id / filename with bad predictions

keras_cv/metrics/coco/coco_pymetric.py
tanzhenyu: Hmmm, both the RetinaNet and FasterRCNN pipeline does not use any KPLs, so it's actually returning boxes and classes, and our model prediction returns boxes and classes separately, so I'm not sure what's blocking this?

keras_cv/callbacks/pycoco_callback.py
```diff
@@ -0,0 +1,106 @@
...
+                    gt[k].append(v)
+
+        metrics = compute_pycoco_metrics(gt, preds)
+        # Mark these as validation metrics by prepending a val_ prefix
+        metrics = {"val_" + name: val for name, val in metrics.items()}
+
+        logs.update(metrics)
+
+    def _eval_batch(self, batch, index):
+        images, gt_boxes = batch
```
tanzhenyu, line 60: I think we should only make sure this works with retinanet.fit and fasterrcnn.fit at this moment, so accepting boxes and classes separately is fine.

tanzhenyu (changes requested):

keras_cv/metrics/coco/pycoco_utils.py
ianstenbit: coco_eval.eval gives a dictionary with e.g. `recall` and `precision`, but on a per-image basis. stats just gives an array. Implementation [here](https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/cocoeval.py).

keras_cv/metrics/coco/coco_pymetric.py
ianstenbit: Yes, this is part of TFDS for COCO. However, using it here:
- requires users to parse that field and add it to the train/val datasets
- adds a lot of friction for using these metrics with any other dataset

Is there any value add to using it? From what Ic an see, we get the exact same behavior with a simpler API by leaving it out.

keras_cv/callbacks/pycoco_callback.py
ianstenbit: retinanet.fit assumes that boxes and classes are together in one tensor. This currently works with retinanet.fit but not fasterrcnn.fit. If I changed it, that would flip flop. More discussion in the thread above.

tanzhenyu: Then it seems we should make sure this works with fasterrcnn.fit not retinanet.fit -- given we have discussed this several times and it seems that we have a consensus that we should separate them out, I'd prefer not introduce more tech debt.

keras_cv/metrics/coco/coco_pymetric.py
tanzhenyu: Ok, that makes sense

keras_cv/metrics/coco/coco_pymetric.py
tanzhenyu: I believe we agreed that we should use : 
```
for examples in train_ds:
    image, gt_boxes, gt_classes = examples
```

This is the most native format for any detection workflow.

keras_cv/callbacks/pycoco_callback.py
ianstenbit: Fine by me -- I will update this to work with fasterrcnn, not retinanet.

Once that's done, and I've added test, I'll re-request review.

tanzhenyu (changes requested):
See comments inline

keras_cv/callbacks/pycoco_callback.py
LukeWood: I believe we settled on a dictionary format, maybe it’s time migrate all components to the dictionary format for bounding boxes.

keras_cv/metrics/coco/coco_pymetric.py
LukeWood: Are we sure we want to diverge?  right now we have all bounding boxes in one format, and MaskRCNN in another format.

ianstenbit:
I've updated the callback to use the ```images, {"gt_boxes": boxes, "gt_classes": classes}``` approach for structuring input data, and I've added unit testing (using RetinaNet.fit for now, because FasterRCNN doesn't have a prediction decoder)

@tanzhenyu and @LukeWood PTAL -- and thank you both for your feedback so far!

ianstenbit:
BTW, test failures seem unrelated to this change and I am looking into it.

ianstenbit:
> BTW, test failures seem unrelated to this change and I am looking into it.

Turns out failures were due to this PR :) -- should be fixed now

keras_cv/callbacks/pycoco_callback.py
```diff
@@ -0,0 +1,102 @@
...
+import six
+import tensorflow as tf
+from keras.callbacks import Callback
+
+from keras_cv import bounding_box
+from keras_cv.metrics.coco import compute_pycoco_metrics
+
+
+class PyCOCOCallback(Callback):
+    def __init__(self, validation_data, bounding_box_format):
```
LukeWood, line 23: add **kwargs and propogate to __init__

ianstenbit: Done

tanzhenyu (approved):
LGTM, only need to make sure:
1) it's using the input format as (images, {"gt_boxes", "gt_classes"})
2) having a description on the PR on the eval results (for example, 3 epochs)

ianstenbit:
> LGTM, only need to make sure:
> 
> 1. it's using the input format as (images, {"gt_boxes", "gt_classes"})
> 2. having a description on the PR on the eval results (for example, 3 epochs)

SG -- I am working on getting the eval results. I figured I'd get it working on TPU at the same time, so it's taking a bit. I'll wait to merge until that is sorted out.

ianstenbit:
I've updated the description with a TensorBoard link to an example run training a RetinaNet with this callback.

I think there are some improvements to be made on the performance which are likely going to be tied to more efficiently getting predictions from the model during evaluation time (calling the model for each batch instead of for the full dataset is super inefficient). I will work on those improvements as a follow-up PR.

tanzhenyu:
> https://tensorboard.dev/experiment/Dmvt2HYDTxamopnh5QgJkQ/#scalars&_smoothingWeight=0&runSelectionState=eyJ0cmFpbiI6dHJ1ZX0%3D

Sounds good to me. Feel free to go ahead and merge, with follow-up PRs

PR #1005, opened 2022-11-11
Implement bounding boxes for center crop

AmedeoBiolatti:
# What does this PR do?

Implement # (https://github.com/keras-team/keras-cv/issues/731)

centercrop

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood and @tanzhenyu

### DIFF
--- a/keras_cv/layers/preprocessing/center_crop.py
+++ b/keras_cv/layers/preprocessing/center_crop.py
@@ -1,0 +1,171 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv import bounding_box
+from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
+    BaseImageAugmentationLayer,
+)
+from tensorflow import keras
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class CenterCrop(BaseImageAugmentationLayer):
+    """A preprocessing layer which crops images.
+
+    The input images should have values in the `[0-255]` or `[0-1]` range.
+
+    Input shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Output shape:
+        3D (unbatched) or 4D (batched) tensor with shape:
+        `(..., height, width, channels)`, in `"channels_last"` format
+
+    Args:
+        height: Integer, the height of the output shape.
+        width: Integer, the width of the output shape.
+
+    Call arguments:
+        inputs: Tensor representing images of shape
+            `(batch_size, width, height, channels)`, with dtype tf.float32 / tf.uint8,
+            ` or (width, height, channels)`, with dtype tf.float32 / tf.uint8
+        training: A boolean argument that determines whether the call should be run
+            in inference mode or training mode. Default: True.
+
+    Usage:
+    ```python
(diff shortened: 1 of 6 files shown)


keras_cv/layers/preprocessing/center_crop.py
```diff
@@ -0,0 +1,171 @@
...
+from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
+    BaseImageAugmentationLayer,
+)
+from tensorflow import keras
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class CenterCrop(BaseImageAugmentationLayer):
+    """A preprocessing layer which crops images.
+
```
tanzhenyu, line 27: This layers crops the central portion of the images to a target size. If an image is smaller than the target size, it will be resized and cropped so as to return the largest possible window in the image that matches the target aspect ratio.

keras_cv/layers/preprocessing/center_crop.py
```diff
@@ -0,0 +1,171 @@
...
+            image=None,
+            label=None,
+            bounding_boxes=None,
+            keypoints=None,
+            segmentation_mask=None,
+    ):
+        image_shape = tf.shape(image)
+        return image_shape[-3], image_shape[-2]
+
+    def _transform_bounding_boxes(self, bounding_boxes, transformation):
```
tanzhenyu, line 79: do you need to filter boxes that has no overlap with the center crop region?

tanzhenyu (commented):
Thanks for the PR!

keras_cv/layers/preprocessing/center_crop.py
AmedeoBiolatti: This is handled by `bounding_box.clip_to_image` inside `augment_bounding_boxes,` as seen in the unit test it seems to work

keras_cv/layers/preprocessing/center_crop.py
bhack: >  it will be resized..

Do we need to introduce some safety warning log about the ratio? As It could happen to have extreme values processing bulk datasets...

bhack: ...Or at least a explicit flag to enable these cases..

keras_cv/layers/preprocessing/center_crop.py
LukeWood: Yep!  clip_to_image() does in fact filter out of bounds bounding boxes!

LukeWood:
make sure to add entries to:

- serialization_test
- with_labels_test

keras_cv/layers/preprocessing/center_crop.py
AmedeoBiolatti: Is it possible to have more details? Should be raise a warning when the aspect ratio is too much different or when the size is too different?

keras_cv/layers/preprocessing/center_crop.py
LukeWood: you might also need to add a call to `bounding_box.filter_sentinels()` to go from dense -1 padded tensor -> ragged

LukeWood:
Here's a thought: CenterCrop is not really ideal for bounding boxes because you can literally crop entire labels out.
Perhaps we should intentionally not support it, and just raise an informative error message telling users to use Resizing() instead?


@tanzhenyu thoughts?

divyashreepathihalli:
> 

@LukeWood - I think that would be the case for random crop  and random resize crop as well. I think there is value is supporting augment bounding boxes for this with a warning. WDYT?

LukeWood:
> > 
> 
> @LukeWood - I think that would be the case for random crop and random resize crop as well. I think there is value is supporting augment bounding boxes for this with a warning. WDYT?

@AmedeoBiolatti lets add a warning for center crop with bounding boxes indicating that this may be causing many boxes to be dropped.

AmedeoBiolatti:
> @AmedeoBiolatti lets add a warning for center crop with bounding boxes indicating that this may be causing many boxes to be dropped.

@LukeWood  Done! Now I will try to resolve the conflicts

AmedeoBiolatti:
@LukeWood the conflicts are solved, let me know if anything else is needed

LukeWood (changes requested):
Lets remove the box portion of this PR and just merge CenterCrop as a KerasCV layer for now.

divyashreepathihalli:
closing because this is stale.

PR #1006, opened 2022-11-11
[YOLOX] Step 1/? : add GIoU loss and update IoU loss

quantumalaviya:
# What does this PR do?
This PR is one of many upcoming PRs that will attempt to break down #982 so that it's easier to review. This one deals with the losses:

- adds GIoU loss 
- edits IoU to add an axis argument in the API.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood, @tanzhenyu @ianstenbit


keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,151 @@
...
+    smallest box enclosing both the boxes being considered for the iou. The length of
+    the last dimension should be atleast 4 to represent the bounding boxes. While
+    this dimension can have more than 4 values, these values will be ignored for the
+    calculation of this loss.
+    Args:
+        bounding_box_format: a case-insensitive string (for example, "xyxy").
+            Each bounding box is defined by at least these 4 values. The inputs
+            may contain additional information such as classes and confidence after
+            these 4 values but these values will be ignored while calculating
+            this loss. For detailed information on the supported formats, see the
```
tanzhenyu, line 35: I think just referring to the bbox documentation is good enough.
But here it should clarify what does it represent. Is this bbox for y_pred, or for y_true, or both, and also will there be any cases where the pred and label format need to be different

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,151 @@
...
+        ):
+            raise ValueError(
+                "`GIoULoss` expects both boxes to be batched, or both "
+                f"boxes to be unbatched.  Received `len(boxes1.shape)`={boxes1_rank}, "
+                f"`len(boxes2.shape)`={boxes2_rank}.  Expected either `len(boxes1.shape)`=2 AND "
+                "`len(boxes2.shape)`=2, or `len(boxes1.shape)`=3 AND `len(boxes2.shape)`=3."
+            )
+
+        if self.bounding_box_format.startswith("rel"):
+            target = "rel_yxyx"
```
tanzhenyu, line 78: any reason we cannot always use a single target format?

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,151 @@
...
+            boxes2_height = tf.maximum(zero, boxes2_ymax - boxes2_ymin)
+            boxes1_area = boxes1_width * boxes1_height
+            boxes2_area = boxes2_width * boxes2_height
+            intersect_ymin = tf.maximum(boxes1_ymin, boxes2_ymin)
+            intersect_xmin = tf.maximum(boxes1_xmin, boxes2_xmin)
+            intersect_ymax = tf.minimum(boxes1_ymax, boxes2_ymax)
+            intersect_xmax = tf.minimum(boxes1_xmax, boxes2_xmax)
+            intersect_width = tf.maximum(zero, intersect_xmax - intersect_xmin)
+            intersect_height = tf.maximum(zero, intersect_ymax - intersect_ymin)
+            intersect_area = intersect_width * intersect_height
```
tanzhenyu, line 111: please re-use bbox.compute_iou

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,151 @@
...
+            return iou - tf.math.divide_no_nan(
+                (enclose_area - union_area), enclose_area
+            )
+
+        if boxes1_rank == 2:
+            return compute_giou_for_batch((boxes1, boxes2))
+        else:
+            return tf.vectorized_map(compute_giou_for_batch, elems=(boxes1, boxes2))
+
+    def call(self, y_true, y_pred):
```
tanzhenyu, line 133: support `sample_weight`?

keras_cv/losses/giou_loss_test.py
```diff
@@ -0,0 +1,59 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_cv.losses.giou_loss import GIoULoss
+
```
tanzhenyu, line 18: please add numerical tests

keras_cv/losses/iou_loss.py
```diff
@@ -55,14 +60,15 @@ class IoULoss(tf.keras.losses.Loss):
     ```
     """
 
-    def __init__(self, bounding_box_format, mode="log", **kwargs):
+    def __init__(self, bounding_box_format, mode="log", axis=None, **kwargs):
```
tanzhenyu, line 63: can axis come with a default value?

keras_cv/losses/iou_loss_test.py
```diff
@@ -54,26 +54,6 @@ def test_output_shape_relative(self):
             [0.2, 0.1, 0.3, 0.3, 7, 0.48],
         ]
 
-        iou_loss = IoULoss(bounding_box_format="rel_xyxy")
+        iou_loss = IoULoss(bounding_box_format="rel_xyxy", axis=-1)
 
         self.assertAllEqual(iou_loss(y_true, y_pred).shape, ())
-
-    def test_output_value(self):
```
tanzhenyu, line 61: why removing numerical test?

keras_cv/models/object_detection/yolox/__internal__/binary_crossentropy.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2022 The KerasCV Authors
```
tanzhenyu, line 1: Suggestion: remove all yolox/__internal__ at this PR, so it just focus on giou and iou, for faster review and merge

tanzhenyu: Meanwhile what is the BCE loss trying to fix, compared to keras.BCE loss?

tanzhenyu (commented):
Thanks for the PR!

Also you can rename the PR title to be "[YOLOX] Step 1/? : add GIoU loss and update IoU loss"

keras_cv/losses/giou_loss.py
quantumalaviya: This is mostly to remove dependency on images for this loss.

For example if the passed bbox format is non-rel and we use only rel for this function, the images will be required for the dimension calculation (discussed in #604 and fixed for `compute_iou` in #612)

keras_cv/losses/giou_loss.py
quantumalaviya: This will be impossible without making the API noisy and changing `compute_iou`. In my opinion, this is not ideal for a user-facing API.

Alternatively, using `compute_iou` for what it already has will reduce the lines of code by only ~8 lines and make the code considerably more confusing.

keras_cv/losses/iou_loss.py
quantumalaviya: Hmm, I left `None` as default because it would mean no mean at all as we would use it in YoloX

Alternatively, I can add `axis=-1` as the default value maybe? The output of this loss, in that case, would be (batch_size,) if reduction="none"

keras_cv/losses/iou_loss_test.py
quantumalaviya: The IoU loss is not pairwise anymore. I can write a new numerical test for this.

keras_cv/models/object_detection/yolox/__internal__/binary_crossentropy.py
quantumalaviya: For this one `axis=None` means no mean over any axis.

quantumalaviya: Alright, I will remove `binary_crossentropy` from this one.

keras_cv/losses/giou_loss.py
LukeWood: > I think just referring to the bbox documentation is good enough.
But here it should clarify what does it represent. Is this bbox for y_pred, or for y_true, or both, and also will there be any cases where the pred and label format need to be different

In general we only support one and I think that makes for the best UX.  We can just rely on users transforming formats before calling the loss.

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,151 @@
...
+    this dimension can have more than 4 values, these values will be ignored for the
+    calculation of this loss.
+    Args:
+        bounding_box_format: a case-insensitive string (for example, "xyxy").
+            Each bounding box is defined by at least these 4 values. The inputs
+            may contain additional information such as classes and confidence after
+            these 4 values but these values will be ignored while calculating
+            this loss. For detailed information on the supported formats, see the
+            [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+        axis: the axis along which to mean the ious. Defaults to `None` which implies
```
LukeWood, line 37: I wonder if axis=None should actually mean over all axes, and we should instead ship axis="some-special-string"

keras_cv/losses/giou_loss.py
LukeWood: I think that this is automatic in convert_format() now, right?

keras_cv/models/object_detection/yolox/__internal__/binary_crossentropy.py
LukeWood: I think we should have a special string to ignore axis.  axis=None usually means mean over all axes in TF

bhack: We had a thread at https://github.com/keras-team/keras/issues/17207

keras_cv/losses/giou_loss.py
quantumalaviya: I was also wondering this. Using up None will confuse existing Keras users.

Should I use axis = "none" or axis = "yolox_output" here? Since I can't think of any good string for this, I can also add axis as a kwargs argument. When it's not passed, we can just perform the loss with no mean and when it is passed, we can use it as used in Keras BCE 🤔

LukeWood: axis="no_reduction" perhaps?

keras_cv/losses/giou_loss.py
quantumalaviya: I don't think it is. `convert_format` just has support for rel->rel conversions without an image.
Earlier, even rel->rel conversion required images because the intermediate conversion was xyxy.

This part ensures that rel source has rel targets and the (same for non-rel ones).

keras_cv/losses/giou_loss_test.py
quantumalaviya: added numerical tests

keras_cv/losses/giou_loss.py
quantumalaviya: Added this. However, there's a possibility users will get confused with `reduction` for this.

keras_cv/losses/giou_loss.py
LukeWood: Ah I see; ok that makes sense.  I bet we can write a utility for this.

Something like:
```
preserve_rel('xywh', bounding_box_format=bounding_box_format)
```

Returns 'rel_xywh' if bounding_box_format starts with rel, otherwise 'xywh'.

quantumalaviya: Although I think the current way is better, I have opened a PR for this (#1016).

Lmk if this is what you were looking for and I will rebase this one and add it to `_compute_giou` as well! @LukeWood

LukeWood (approved):
Looks good to me, I think if you address sample weight we should be good to go.

keras_cv/losses/iou_loss.py
tanzhenyu: Hmmm...then why do we need to support a `axis` argument here, if YOLOX uses `None`? (as other keras losses don't have such argument)

keras_cv/losses/giou_loss.py
quantumalaviya: On it

keras_cv/losses/iou_loss.py
quantumalaviya: In case other models use the loss in the future.

keras_cv/losses/giou_loss.py
quantumalaviya: Sorry for the delay on this. 

@tanzhenyu Is there a specific reason we need to add this? It's inbuilt in Keras and Addons implementations uses the `LossFunctionWrapper`. Is this worth the complex code we will need to add to support it?

quantumalaviya:
@LukeWood

keras_cv/losses/giou_loss.py
tanzhenyu: But this loss is not inhering from `LossFunctionWrapper`, correct?

keras_cv/losses/giou_loss.py
tanzhenyu: besides the benefit of reducing code complexity, if we find bugs for computing iou, it'd be easier to track where this needs to be fixed as well

tanzhenyu (changes requested):

keras_cv/losses/giou_loss.py
quantumalaviya: Yeah, its not

LukeWood (approved):
Looks good to me, an explicit error on `sample_weight` would be nice, but other than that LGTM.  Thanks!

quantumalaviya:
@LukeWood This is ready (added the error as well)

keras_cv/losses/numerical_tests/iou_giou_loss_numerical_test.py
```diff
@@ -0,0 +1,148 @@
+# Copyright 2022 The KerasCV Authors
```
tanzhenyu, line 1: no need for this file.

keras_cv/losses/iou_loss_test.py
```diff
@@ -57,23 +57,3 @@ def test_output_shape_relative(self):
         iou_loss = IoULoss(bounding_box_format="rel_xyxy")
 
         self.assertAllEqual(iou_loss(y_true, y_pred).shape, ())
```
tanzhenyu, line 59: please add numerical tests here. It should have some easy to understand numerics and comments like this:
https://github.com/keras-team/keras-cv/blob/master/keras_cv/losses/penalty_reduced_focal_loss_test.py#L53-L91

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,156 @@
...
+            enclose_height = tf.maximum(zero, enclose_ymax - enclose_ymin)
+            enclose_area = enclose_width * enclose_height
+            return iou - tf.math.divide_no_nan(
+                (enclose_area - union_area), enclose_area
+            )
+
+        if boxes1_rank == 2:
+            return compute_giou_for_batch((boxes1, boxes2))
+        else:
+            return tf.vectorized_map(compute_giou_for_batch, elems=(boxes1, boxes2))
```
tanzhenyu, line 131: Does vectorized_map work in TPU?

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,156 @@
...
+
+        boxes2 = bounding_box.convert_format(
+            boxes2, source=self.bounding_box_format, target=target
+        )
+
+        def compute_giou_for_batch(boxes):
+            boxes1, boxes2 = boxes
+            zero = tf.convert_to_tensor(0.0, boxes1.dtype)
+            boxes1_ymin, boxes1_xmin, boxes1_ymax, boxes1_xmax = tf.unstack(
+                boxes1[..., :4], 4, axis=-1
```
tanzhenyu, line 94: just tf.unstack(boxes1, 4, axis=-1)

keras_cv/losses/giou_loss.py
tanzhenyu: I'm not sure I understand why we need to change `compute_iou` to make it work here, can you elaborate?

tanzhenyu (changes requested):

keras_cv/losses/giou_loss.py
quantumalaviya: Not sure about this. I can't seem to find any particular note in the documentation. 
@LukeWood Can you help here? I will update the code to `map_fn` if this is an issue.

keras_cv/losses/giou_loss.py
quantumalaviya: > I'm not sure I understand why we need to change compute_iou to make it work here, can you elaborate?

The area of the enclosing box is also required for GIoU. So the code starting line 116 is added to `compute_iou` to form `compute_giou`.

An alternative if we had to reuse `compute_iou` was to update the API to take in a string indicating whether IoU or GIoU is being computed (as in the addons implementation). But since this is a user-facing API, I decided to let it be after consulting Luke.

tanzhenyu: I mean from line 101 to line 112 can be replaced by `compute_iou`, while the giou part can still be manually computed here, correct?

keras_cv/losses/giou_loss.py
tanzhenyu: I think we should always prefer batched implementation and not rely on vectorized_map or map_fn in loss computation. However sometimes that comes with more work, If your YOLOX implementation does not require > rank 2 inputs, I suggest you can error it out for now

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,154 @@
...
+    smallest box enclosing both the boxes being considered for the iou. The length of
+    the last dimension should be 4 to represent the bounding boxes.
+
+    Args:
+        bounding_box_format: a case-insensitive string (for example, "xyxy").
+            Each bounding box is defined by these 4 values.For detailed information
+            on the supported formats, see the
+            [KerasCV bounding box documentation](https://keras.io/api/keras_cv/bounding_box/formats/).
+        axis: the axis along which to mean the ious. Passing the string "no_reduction" implies
+            mean across no axes. Defaults to -1.
```
tanzhenyu, line 35: I prefer None = do not reduce on any axis here. Otherwise we're not consistent with how reduction works with TF/Keras losses

tanzhenyu (changes requested):

keras_cv/losses/giou_loss.py
quantumalaviya: `None` implies reduction across all axes in Keras losses. That's why I added a new string specifically for the use case of YoloX.

keras_cv/losses/giou_loss.py
quantumalaviya: This might be somewhat fixed now that I am using utils from `compute_iou`

quantumalaviya:
@tanzhenyu Refactored most of the `compute_giou` code and it solves (somewhat) most of your review comments. Please take a look.

keras_cv/losses/giou_loss.py
tanzhenyu: axis=None actually means axis=-1:
https://github.com/tensorflow/tensorflow/blob/d5b57ca93e506df258271ea00fc29cf98383a374/tensorflow/python/ops/nn_ops.py#L4106-L4107

The point here is that taking a string is quite odd, can we remove `axis` from this PR, and add a TODO to unify axis naming convention together with `keras.losses.Huber` later? (I believe that loss also needs to take some reduction axis as well)

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,172 @@
...
+            )
+        if boxes2_rank not in [2, 3]:
+            raise ValueError(
+                "compute_iou() expects boxes2 to be batched, or "
+                f"to be unbatched. Received len(boxes1.shape)={boxes1_rank}, "
+                f"len(boxes2.shape)={boxes2_rank}. Expected either len(boxes2.shape)=2 AND "
+                "or len(boxes2.shape)=3."
+            )
+
+        target = bounding_box.preserve_rel(
```
tanzhenyu, line 100: If it's converted to rel_yxyx, or rel_yxhw, don't we need to provide `image_shape` during convert_format so it can support ALL input box format?
In any case not converting to relative format sounds more clear to me

tanzhenyu (changes requested):

keras_cv/losses/giou_loss.py
tanzhenyu: NVM, this can be resolved

keras_cv/losses/giou_loss.py
quantumalaviya: Sure, I'll remove this for now and add an issue.

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,169 @@
...
+        - [TFAddons Implementation](https://www.tensorflow.org/addons/api_docs/python/tfa/losses/GIoULoss)
+
+    Sample Usage:
+    ```python
+    y_true = tf.random.uniform((5, 10, 5), minval=0, maxval=10, dtype=tf.dtypes.int32)
+    y_pred = tf.random.uniform((5, 10, 4), minval=0, maxval=10, dtype=tf.dtypes.int32)
+    loss = GIoULoss(bounding_box_format = "xyWH")
+    loss(y_true, y_pred).numpy()
+    ```
+    Usage with the `compile()` API:
```
LukeWood, line 47: newlines between sections for documentation rendering

keras_cv/losses/giou_loss.py
```diff
@@ -0,0 +1,169 @@
...
+        - [GIoU paper](https://arxiv.org/pdf/1902.09630)
+        - [TFAddons Implementation](https://www.tensorflow.org/addons/api_docs/python/tfa/losses/GIoULoss)
+
+    Sample Usage:
+    ```python
+    y_true = tf.random.uniform((5, 10, 5), minval=0, maxval=10, dtype=tf.dtypes.int32)
+    y_pred = tf.random.uniform((5, 10, 4), minval=0, maxval=10, dtype=tf.dtypes.int32)
+    loss = GIoULoss(bounding_box_format = "xyWH")
+    loss(y_true, y_pred).numpy()
+    ```
```
LukeWood, line 46: same here needs newline

keras_cv/losses/iou_loss.py
```diff
@@ -55,14 +56,15 @@ class IoULoss(tf.keras.losses.Loss):
...
 
-    def __init__(self, bounding_box_format, mode="log", **kwargs):
+    def __init__(self, bounding_box_format, mode="log", axis=-1, **kwargs):
         super().__init__(**kwargs)
         self.bounding_box_format = bounding_box_format
         self.mode = mode
+        self.axis = axis
 
-        if self.mode not in ["linear", "square", "log"]:
+        if self.mode not in ["linear", "squared", "log"]:
```
LukeWood, line 65: rename squared to quadratic.  `linear` is akin to `quadratic` in math

LukeWood (changes requested):
I left 3~ minor changes but overall this is

LukeWood (approved):
LGTM

tanzhenyu (approved):
Thanks for the patience, looks good now!!

PR #1013, opened 2022-11-15
fixed transformer encoder dims per head

DavidLandup0:
# What does this PR do?

Fixes overlooked dimensionality issue with the use of the `MultiHeadAttention` layer in `keras_cv.layers.TransformerEncoder`. It should be:

```
        self.attn = layers.MultiHeadAttention(
            num_heads=self.num_heads,
            key_dim=self.project_dim // self.num_heads,
            dropout=self.attention_dropout,
        )
```

And not:

```
        self.attn = layers.MultiHeadAttention(
            num_heads=self.num_heads,
            key_dim=self.project_dim,
            dropout=self.attention_dropout,
        )
```

The latter makes the params skyrocket. Updated tests as well, and removed unused arg from the call.
My bad!

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@tanzhenyu

### DIFF
--- a/keras_cv/layers/transformer_encoder.py
+++ b/keras_cv/layers/transformer_encoder.py
@@ -59,9 +59,7 @@ class TransformerEncoder(layers.Layer):
         **kwargs,
     ):
         super().__init__(**kwargs)
-        # hidden_dim by authors
         self.project_dim = project_dim
-        # intermediate_dim by authors
         self.mlp_dim = mlp_dim
         self.num_heads = num_heads
         self.mlp_dropout = mlp_dropout
--- a/keras_cv/layers/transformer_encoder.py
+++ b/keras_cv/layers/transformer_encoder.py
@@ -23,7 +23,6 @@ class TransformerEncoder(layers.Layer):
 
     Args:
         project_dim: the dimensionality of the projection of the encoder
-        intermediate_dim: the intermediate dimensionality of the transformer encoder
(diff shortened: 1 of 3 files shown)


keras_cv/layers/transformer_encoder.py
```diff
@@ -23,7 +23,6 @@ class TransformerEncoder(layers.Layer):
 
     Args:
         project_dim: the dimensionality of the projection of the encoder
```
tanzhenyu, line 25: add statement that this is the output of multihead

tanzhenyu (changes requested):

keras_cv/layers/transformer_encoder.py
DavidLandup0: Sure!

DavidLandup0: Done

DavidLandup0:
The output shapes stay the same - it's just a difference in the internal dimensionality that wasn't tangible/testable before I tried stacking 12 encoders on top of each other and realizing that they have way too many params each...

DavidLandup0:
Updated serialization test

DavidLandup0:
Verified that these build to the right param counts when compared to the papers:

image

DavidLandup0:
Should be ready now - can you take a look again? @tanzhenyu 
Thanks!

tanzhenyu (approved):

DavidLandup0:
Thanks for the approval!

Issue #907, opened 2022-10-11
Express cloudbuild as github action


bhack:
Can we express our CloudBuild with the Gooogle  Github Action?

This is an official example from the official google cloud Action repo:
https://github.com/google-github-actions/setup-gcloud/tree/main/example-workflows/cloud-build

https://github.com/google-github-actions/setup-gcloud/blob/main/example-workflows/cloud-build/.github/workflows/cloud-build.yml

bhack:
1. We have all the orchestration on a common well know DSL (Githu Action)
2. I don't know specific limit, it is maintained by Google.

bhack:
Also it is quite annoying that the Cloudbuild is always in failing by default for every new PR (when it is still not manually triggered).

It makes the PRs status and overview/lists very confusing.

LukeWood:
> Also it is quite annoying that the Cloudbuild is always in failing by default for every new PR (when it is still not manually triggered).

This is a good point, but I'm not sure we want to invest the resources to fix this at this time.  Unfortunately, we don't have the GPU resources to run the tests for every single PR commit.

bhack:
> Unfortunately, we don't have the GPU resources to run the tests for every single PR commit.

I never asked this but as we manually request it to run it doesn't need to fail by default as it is not mandatory to have a default fail status.

LukeWood:
> I never asked this but as we manually request it to run it doesn't need to fail by default as it is not mandatory to have a default fail status.

Ok, how do we configure that?  I'm happy to accept configuration for that.  That would be really nice to have it default to yellow status.

bhack:
> Ok, how do we configure that? I'm happy to accept configuration for that. That would be really nice to have it default to yellow status.

I am not a manager of this repo so you need to set yourself:
https://cloud.google.com/build/docs/automating-builds/github/build-repos-from-github#different_types_of_github-based_triggers

Then Github Action help community PRs, but it is a distinct topic (it was the other point of this ticket).

bhack:
Just to make a blackbox reverse engineering, what I suppose is that you have configured it to trigger also on PRs but it is not required as you have probably configure also a custom trigger for the `/gcburn` pattern.

sachinprasadhs:
Thanks for reporting the issue! We have consolidated the development of KerasCV into the new KerasHub package, which supports image, text, and multi-modal models. Please read https://github.com/keras-team/keras-hub/issues/1831. KerasHub will support all the core functionality of KerasCV.

[...]

PR #1014, opened 2022-11-15
Adding ViT Models

### LINKED ISSUE #668: Support ViT model

DavidLandup0:
# What does this PR do?

Closes #668

As discussed in #916 and #929, and fixed in #1013 - adding support for Vision Transformers, using the new ViT components:

- Tiny-16/32 (5.5M/5.9M)
- S-16/32 (21.6M/22.5M)
- B-16/32 (85.8M/87.5M)
- L-16/32 (303M/305.5M)
- H-16/32 (630M/633.7M)
Ported weights on ImageNetV2 validation set (ms/sep, top-1 and top-5 ACC):

Tiny16: 87ms/step - loss: 1.6895 - accuracy: 0.6067 - sparse_top_k_categorical_accuracy: 0.8285
S16:      178ms/step - loss: 1.2600 - accuracy: 0.6884 - sparse_top_k_categorical_accuracy: 0.8916
B16:      464ms/step - loss: 1.0631 - accuracy: 0.7293 - sparse_top_k_categorical_accuracy: 0.9199
L16:      1s/step - loss: 0.9622 - accuracy: 0.7545 - sparse_top_k_categorical_accuracy: 0.9321
H16:      NA
Tiny32: NA
S32:      54ms/step - loss: 1.6650 - accuracy: 0.6090 - sparse_top_k_categorical_accuracy: 0.8335
B32:      122ms/step - loss: 1.3684 - accuracy: 0.6650 - sparse_top_k_categorical_accuracy: 0.8781
L32:       NA
H32:      NA

### Still TODO:

- [x] Add documentation in the `vit.py` file
- [x] Attempt to convert weights from existing runs, such as https://github.com/google-research/vision_transformer#available-vit-models
- [x] Validate by training locally
- [x] Write test files
- [x] Update model tests 

While possible weight conversions are taking place (hopefully successful) the architecture itself is ready for review!
@LukeWood @tanzhenyu @innat @bhack 

### Notes

Because the patching logic depends on `tf.reshape` and `tf.resize` - we can't use `(None, None, 3)` as the default `input_shape` during instantiation like with other CNN-based models. I debated using a default of `(224, 224, 3)` but after the discussion in #529, it's left as `(None, None, 3)`.

The `models_tests.py` was updated to accommodate for non-CNN output shapes, just as with MLPMixer. ViTs have an extra class token so the shapes weren't reused fully. There's a lengthy chain of "or X not in app.__name__", so we'll probably want to create a list of model names like these instead for cleaner conditionals.

## Before submitting
      Pull Request section?
      to it if that's the case. #668 


tanzhenyu (commented):
Thanks for the PR! On a high level, this is looking great!
On the other hand, we should have a weight porting or training script that comes along with the model def.

DavidLandup0:
Sure! I'll cobble up the trainer first, since it'll take less time and we'll need it anyway :)

DavidLandup0:
Quick question! Is the training script to validate that the model _can_ train to the ~paper-claimed accuracy, or is it to be runnable to actually produce them, or both?
I've tried pre-training the Tiny model on Imagenette for a small number of epochs (50) and then "fine tuned" it on CIFAR10 for another 50.

Considering that it's the Tiny model (fairly small), ViTs require lots of data to train well (which is why most benchmarks include them pretrained on ImageNet21k or JFT), typically use tricks (I haven't included any yet, this was a super basic training script) and that it was trained on just 50 epochs each - a 70% accuracy on CIFAR10 sounds about what I've seen being mentioned in other places.

I'm struggling to find ViTs trained on CIFAR10 from scratch to compare the performances, which might indicate that they don't do so well without massive pre-training. Here are the only two references I found:

- https://github.com/kentaroy47/vision-transformers-cifar10
- https://github.com/omihub777/ViT-CIFAR

Here was my short run:

image

bhack:
Yes one of the few probably reliable/paper experiment without a large scale pretraining is:

https://arxiv.org/abs/2203.02358

https://github.com/Freder-chen/Swelling-ViT

Instead I don't know if we have enough resources to proxy this other target:
https://github.com/keras-team/keras-cv/pull/929#issuecomment-1319369917

DavidLandup0:
I'll try hitting [#](https://github.com/keras-team/keras-cv/pull/929#issuecomment-1319369917) and continue trying to port the weights. Assuming it does hit the mark at 90 epochs, we'll will want to at least try to port the weights if possible

bhack:
I think this is a quite plastic evidence on how the community it is often at risk to hit issues similar to what was described some months ago at: https://github.com/keras-team/keras-cv/issues/488#issuecomment-1154998812

Probably it is not exactly the VIT case but if we are starting from pretraining and "fine-tuning" how much this case is really different?

Of course stable diffusion, where its "large scale is quite clear" when  it was integrated/contributed in this library has no training script.

So the point again here what is this threshold?

tanzhenyu:
There is no specific threshold to transformer models, we have to make judgemental calls based on what we think KerasCV's value is. For conv models, reproducible training and leadership is important, but that doesn't really apply well to transformer. For transformers, what's more important is that it is modular enough so it shares the same "plug-and-play" with conv models. So my suggestion here is roughly we can reproduce tiny transformers (T + S), but port weights for large transformers (B, L, H).

DavidLandup0:
> So the point again here what is this threshold?

The favorite answer of any data scientist - it depends 😆 
Jokes aside - yeah, I think it depends on where we draw the line, based on the architecture, how commonly it's used, etc. In this case, I think that we can re-train ViT-Tiny since I think public weights don't exist for it, but port for the other ones *given* that Tiny trained well (and expecting the other ones to do so as well) to save compute. 

I agree with @tanzhenyu - IMO, the most important part of this PR is that it's using the reusable components from the previous one, which made the implementation super simple. It made porting weights harder, because the original ViT does patching with Conv2Ds and patch embedding somewhat differently. The transformer encoders are the exact same, so I might be able to port 99% of the weights, without proper porting for patch embedding...

Speaking of which, we might want to update the MixerMLP model to use `Patching` instead of the `Conv2D` ;)
Design questions + pure reformatting (no logic changes) for MixerMLP are in: https://github.com/keras-team/keras-cv/pull/1024

bhack:
> So my suggestion here is roughly we can reproduce tiny transformers (T + S), but port weights for large transformers (B, L, H).

I think the point here is that if we need to handle VIT as fine-tuning models or not.

And I think this also apply to any other models where we don't have a small enough proxy target to reach with cheap enough computing resources to let  then train with Keras-cv internal resources. 

Then probably foundation tier models are in another category and they need to have a separate strategy/policy (e.g. Stable diffusion).

In this specific cases, as discussed, we have tiny transformers but with fine-tuning only public data points/logs.

So if you want to train a tiny VIT model from scratch, without starting from pre-computed weights, what is the expected target?

tanzhenyu:
> > So my suggestion here is roughly we can reproduce tiny transformers (T + S), but port weights for large transformers (B, L, H).
> 
> I think the point here is that if we need to handle VIT as fine-tuning models or not.
> 
> And I think this also apply to any other models where we don't have a small enough proxy target to reach with cheap enough computing resources to let then train it with Keras-cv internal resources.
> 
> Then probably foundation tier models are in another category and they need to have a separate strategy/policy (e.g. Stable diffusion).
> 
> In this specific cases, as discussed, we have tiny transformers but with fine-tuning only public data points/logs.
> 
> So if you want to train a tiny VIT model from scratch, without starting from pre-computed weights, what is the expected target?

So that we can treat this as a backbone, and train with segmentation meta architectures which sometimes trains everything from scratch anyone.
Generally the larger the model is, the more valuable on it being used as a backbone for fine-tuning.
To summarize, having the weights here is a goal more important than reproducibility, that's why I will leave this up to the contributor

bhack:
> that's why I will leave this up to the contributor

I will add to contributor "computing resources".

I still think that we have a problem cause it could be hard to verify a proposed training from scratch protocol/script on tiny models when we don't have a reference point/anchor.
This could create a false confidence about the training protocol and its related scripts to the end users in the case we are not confident about a defined performance gap.

What I think is that we could still train small tiny model from scratch blacbox and then fine-tuning with the public log but at least we need to have some number from the team about the accuracy starting point of these pretrained weights.

@DavidLandup0 If we cannot have this data I suggest to train a tiny model unitill reaching a reasonable delta from  the first fine-tuning data point in on the small dataset in the public TB.dev. 
So that you can start to use the public reference fine-tuning log without a too large pretraining GAP.

bhack:
@DavidLandup0 There is a specific from scratch target accuracy for the tiny one at:

https://arxiv.org/abs/2205.14443v1

But without a specified number of required epochs.

Probably you could try with that target and then fine tune on one of the small datasets.

DavidLandup0:
I managed to get a full Imagenet run going on Kaggle for ViT_S_16, with _just_ horizontal flips for augmentations (to isolate the network itself) and _no_ special tricks (the script isn't the exact same one as `basic_training`, but isn't far from it since both are kinda basic).

2 epochs in, 20% Val Top-5 Accuracy, and 9% Val Top-1 Accuracy, with a LR warmup 🤔 
I want to see if I can get it to at least kind of replicate this: https://github.com/sayakpaul/big_vision_experiments

The linked run uses several tricks from the training paper published after ViTs, which bumped the accuracy from 70% to 76%

DavidLandup0:
> What I think is that we could still train small tiny model from scratch blackbox and then fine-tuning with the public log but at least we need to have some number from the team about the accuracy starting point of these pretrained weights. @DavidLandup0 If we cannot have this data I suggest to train a tiny model unitill reaching a reasonable delta from the first fine-tuning data point in on the small dataset in the public TB.dev.
So that you can start to use the public reference fine-tuning log without a too large pretraining GAP.

Let me just check if I got you right - you think the team should try making a single checkpoint for Tiny (since we currently have none public), so I can try tuning it on CIFAR10 to reference the public logs? If not, I should try training a Tiny model on Imagenet until reaching a stable point, from which I can try fine-tuning on CIFAR10 (and other smaller datasets) to replicate the public logs?

DavidLandup0:
> @DavidLandup0 There is a specific "from scratch" target accuracy for the tiny one at:
https://arxiv.org/abs/2205.14443v1
But without a specified number of required epochs.
Probably you could try with that target and then fine tune on one of the small datasets.

So, the only other runs I've been able to find from scratch get up to ~76% in ~90 epochs, on ViT-S, so I'd assume that you'd need ~100 epochs for the 72% for ViT Tiny. Should be doable on the Kaggle TPU too 🤔 

I'll keep you posted on the S run since we have a full training curve public for it. If it does well, let's try doing a ~100 epoch Tiny run and build on top of it for  fine-tuning

bhack:
Just as a reminder we had many works on let VIT works on small dataset. Just to mention two:

https://arxiv.org/abs/2210.14319

https://keras.io/examples/vision/vit_small_ds/

So we could also evaluate the path to contribute a VIT on small datasets solution and instead pretrained backbones where we only test that we could fine-tuning on the downstream task.

 It really depend on the library strategy and to what we think is our target.

DavidLandup0:
Here's a small-ds run: https://colab.research.google.com/drive/1CCdyxLhHO7tpvJfuQbNGjMxIXpkDiQnu?usp=sharing
52% on Imagenette and 56% on CIFAR10 with transferred weights from the Imagenette run in 17 epochs.
IIRC, it did 52% on a cold start on CIFAR10 in ~50 epochs.

DavidLandup0:
Okay, so it looks like I almost managed to port the weights - there's *one* layer that I think is blocking it from working.
Here's the script so far: https://colab.research.google.com/drive/1cnZfZ8jx7sk0zUbfBiF-nLMyZVbQuWc1?usp=sharing

The *only* layer that differs is our `patch_embedding` and their layer, because they use a `Con2D` layer to couple together both patching and embedding, while we use `Patching` and `PathEmbedding` which has a *dense* projection, making that layer differ in the shape:

```
Original: (16, 16, 3, 768)
Ours: (768, 768)
```

We could reshape it with `(-1, 768)`, and it works to load the weights, but I think that it'll totally mess up the way they work and make the network delusional down the line because the embedded patch weights are encoding something totally different. 

Unless there's a more standard way to do this that anyone can give advice on - I'm thinking about porting all of the weights like this, and then freezing all layers except the `PatchEmbedding` layers and letting them train on ImageNet. That way, we'll only re-train the super cheap, simple layer, and won't train 99% of the weights. Thoughts?

@tanzhenyu @bhack @LukeWood

bhack:
> while we use Patching and PathEmbedding which has a dense projection

What was the advantage of this design?

DavidLandup0:
I think it makes sense to separate the steps in a ViT:

image

Namely:

- Patching
- Embedding patches (linear projection)
- Encoder
- MLP Head

Other transformer models like DEiTs also have these steps, and if we separate them into individual operations, we can reuse them for others (which, in my understanding, is the purpose of KCV layers - allowing people to rapidly build and prototype using common components).

The use of "linear" in "Linear Projection of Flattened Patches" and "we split the image into fixed-size patches, linearly embed each of them, add position encodings..." also implied a `layers.Dense()` projection, added with positional embeddings, which is what we have verbatim:

```
encoded = self.linear_projection(patch) + self.position_embedding(positions)
```

Collapsing the entire step-by-step process into a `Conv2D` would blur the line between the steps, and prevent people from using these (IMO) clearly *separate steps* in any different way. Many changes in architectures come from slightly differently using some small step in a process, so allowing this is IMO very important.

This is also in line with what we did at:
- https://keras.io/examples/vision/image_classification_with_vision_transformer/
- https://keras.io/examples/vision/vit_small_ds/

(Patch using `tf.extract_patches` and then linearly project instead of patch and project using `Conv2D`).

If this isn't a wise choice - I can replace the original design...

bhack:
> Patching
    Embedding patches (linear projection)
    Encoder
    MLP Head

> Collapsing the entire step-by-step process into a Conv2D would blur the line between the steps, 

Only between the first two steps (Patch + Embedding patches) right?

It seems that pytorch has the same convolutional approach:
https://pytorch.org/vision/main/_modules/torchvision/models/vision_transformer.html

Personally I am ok with this modular approach and it is more "linear"  also by "an education" point of view.

But we also need to consider, other that failing the full pre-trained weights portability of the backbone, the piratical impact of `tf.extract_patches` that it will let the XLA compilation to fail (fail fast) as it is not supported by the default TF bridge.

```python
import tensorflow as tf
def test_patches():
  n = 10
  # images is a 1 x 10 x 10 x 1 array that contains the numbers 1 through 100
  images = [[[[x * n + y + 1] for y in range(n)] for x in range(n)]]

  # We generate two outputs as follows:
  # 1. 3x3 patches with stride length 5
  # 2. Same as above, but the rate is increased to 2
  tf.image.extract_patches(images=images,
                           sizes=[1, 3, 3, 1],
                           strides=[1, 5, 5, 1],
                           rates=[1, 1, 1, 1],
                           padding='VALID')
test_patches()
```

```python
ExtractImagePatches (No registered 'ExtractImagePatches...
```
It is supported by  the "new" "MLIR" bridge (`tf.config.experimental.enable_mlir_bridge()`) that it is not enable by default in TF and personally it is hard for me to grasp how much it is mature/safe right now with the current TF stable release, also I've not tested this Op on TPU with the new bridge only its support on CPU/GPU.

DavidLandup0:
> Only between the first two steps (Patch + Embedding patches) right?

Yes.

> It seems that pytorch has the same convolutional approach: https://pytorch.org/vision/main/_modules/torchvision/models/vision_transformer.html

Hmm, you're right. I've just dug out the HuggingFace embedding layer and they've borrowed it from `timm`, which borrowed it from the original paper, which is a `Conv2D` call. I'll combine the `Patching` and `PatchEmbedding` layers then into a `PatchingAndEmbedding`, if that's okay?

> Personally I am ok with this modular approach and it is more "linear" also by "an education" point of view.
> 
> But we also need to consider, other that failing the full pre-trained weights portability of the backbone, the piratical impact of `tf.extract_patches` that it will let the XLA compilation to fail (fail fast) as it is not supported by the default TF bridge.
> 
> ```python
> import tensorflow as tf
> def test_patches():
>   n = 10
>   # images is a 1 x 10 x 10 x 1 array that contains the numbers 1 through 100
>   images = [[[[x * n + y + 1] for y in range(n)] for x in range(n)]]
> 
>   # We generate two outputs as follows:
>   # 1. 3x3 patches with stride length 5
>   # 2. Same as above, but the rate is increased to 2
>   tf.image.extract_patches(images=images,
>                            sizes=[1, 3, 3, 1],
>                            strides=[1, 5, 5, 1],
>                            rates=[1, 1, 1, 1],
>                            padding='VALID')
> test_patches()
> ```
> 
> ```python
> ExtractImagePatches (No registered 'ExtractImagePatches...
> ```
> 
> It is supported by the "new" "MLIR" bridge (`tf.config.experimental.enable_mlir_bridge()`) that it is not enable by default in TF and personally it is hard for me to grasp how much it is mature/safe right now with the current TF stable release. Also I've not tested this Op on TPU with the new bridge only its support on CPU/GPU.

This is also a really important point, IMO. If we can't do XLA compatibility with `tf.extract_patches`, it's a strong case against it. I automatically assumed that a `tf.image` operation was compatible with XLA, my bad.

I'll fuse the two layers then, and see if that enables the port to work correctly :)
I have a hunch that it will, which means that we'll have the pre-trained weights for all but the Tiny model then with the PR.

bhack:
>  if that's okay?

Personally it could be ok but it is just my personal opinion for the full weights portability + current XLA support.

DavidLandup0:
Updated in this PR to reflect the original, `timm` and HF implementation, and to allow us to port weights. Running the porting again.

DavidLandup0:
So - the weights are ported with the new design of the layer, and they're a 1-to-1 match now, but the probs on a cat image are all low (max confidence is like 2%), so it looks like the weights are random, even though they aren't...

bhack:
> So - the weights are ported with the new design of the layer, and they're a 1-to-1 match now, but the probs on a cat image are all low (max confidence is like 2%), so it looks like the weights are random, even though they aren't...

IMHO another interesting point to evaluate, when we have "JAX backbones", is if we could rely or not to the official `jax2tf` tool in JAX.

E.g.:
https://github.com/sayakpaul/ViT-jax2tf

DavidLandup0:
Tried it, it correctly makes a `SavedModel` and loads it, but actually assigning the trained variables bore no fruit in my case. Maybe I did it wrong...
The main problem, IMO, is that it's a `SavedModel` that's non-trainable and non-convertible to a trainable model, which makes the inspection of the structure and working with it super hard.

bhack:
It seems that loaded as  feature extractor backbone we need to keep that "feature extractor"  frozen. Also in the the official example we have `trainable=False` caused by `b/123499169` but I don't have access to this internal ticket description and `git blame` on this line tells that it a 2 years old comment:

https://github.com/google/jax/blob/main/jax/experimental/jax2tf/examples/keras_reuse_main.py#L50-L51

bhack:
> Edit: I've reached out to Sayak and he shared a new repository with a conversion notebook for the new MAXIM model that's cleaner and shorter than the previous ones, so I might be able to replicate it for ViTs as well. In the meantime, I'm doing an ImageNet1K run with the updated components :)

If `b/123499169` is not solved probably it is a better solution to fully unlock the "feature extractor" backbone on finetuning.

bhack:
Ok here it is explained better by @gnecula's comment:

https://github.com/gnecula/jax/blob/main/jax/experimental/jax2tf/examples/saved_model_lib.py#L111-L114

bhack:
@DavidLandup0 I've asked for a clarification at https://github.com/google/jax/issues/13329 as it seems to me not valid anymore.

bhack:
I think it will work:
https://colab.research.google.com/github/sayakpaul/ViT-jax2tf/blob/main/conversion.ipynb#scrollTo=J2e1h-F2SmDB

You need to enable `with_gradient` in the cell (also for the feature extractor use case):
```python
predict_fn = lambda params, inputs: model.apply(
    dict(params=params), inputs, train=False
)

with_gradient = False if num_classes else True
tf_fn = jax2tf.convert(
    predict_fn,
    with_gradient=with_gradient,
    polymorphic_shapes=[None, "b, 224, 224, 3"],
)
```

Then you could use `hub.KerasLayer(feature_model_dir, trainable=True)`

DavidLandup0:
Thank you for probing @bhack!
I'll test out the trainable model as well - seriously, thank you for helping out with this, it means a lot :)

In the meantime, it looks like the Imagenet run is going good so far with just a horizontal flip:
image

If it continues going well, I'll share the parameters (optimizer, LR, etc.) if we can run a full training job here!

DavidLandup0:
Okay, so so far, the output looks pretty much like Sayak's run on: https://github.com/sayakpaul/big_vision_experiments
Around 10 epochs in, at ~22% validation top-1 accuracy. I expect it start levelling off below Sayak's run because his uses other augmentations too, whilst this run only does a horizontal flip.

Does this qualify for the GCS run here? @tanzhenyu

bhack:
We had:
```
{"step": 13000, "global_schedule": 0.9978464841842651, "training_loss": 4.419351577758789, "l2_grads": 1.041631817817688, "l2_params": 295.06610107421875, "l2_updates": 1.0085276365280151, "examples_seen": 13312000.0, "epoch": 10.49548430841523, "img/sec/core": 592.0690997183742, "core_hours_TPU v3": 6.395378464592827, "core_hours": 6.395378464592827, "minival/prec@1": 0.3978301592257259, "minival/loss": 2.7889692227318625, "val/prec@1": 0.37626, "val/loss": 2.899968134765625, "real/prec@1": 0.4283152208723872, "real/loss": 3.790358957244673, "v2/prec@1": 0.2965, "v2/loss": 3.454605224609375}
```

DavidLandup0:
I managed to misread the chart 🤦‍♂️

bhack:
> I managed to misread the chart man_facepalming

I the repo's assets you can download the log directly.

DavidLandup0:
So, Kaggle environments don't allow for the latest TF version, which KCV depends on, so I can't use MixUp that should be used in the training run. So far, the run's training loss is lower than Sayak's but his validation loss is lower than mine. I don't know whether this is fully attributable to the MixUp (I'd expect it to not be that dramatic in the first 10 epochs), but even so, it doesn't look too far away.

Maybe I'll just bite the bullet and let it run for a while on my consumer GPU since the latest versions of TF and KCV are installed there 🤷‍♂️

bhack:
> So, Kaggle environments don't allow for the latest TF version, which KCV depends on, so I can't use MixUp that should be used in the training run. 

Isn't this enough?

```python
!pip install tensorflow==2.10.0
import tensorflow as tf
from cloud_tpu_client import Client

c = Client()
c.configure_tpu_version('tensorflow==2.10.0', restart_type='ifNeeded')
print("Tensorflow version " + tf.__version__)

try:
  tpu = tf.distribute.cluster_resolver.TPUClusterResolver()  # TPU detection
  print('Running on TPU ', tpu.cluster_spec().as_dict()['worker'])
except ValueError:
  raise BaseException('ERROR: Not connected to a TPU runtime; please see the previous cell in this notebook for instructions!')

tf.config.experimental_connect_to_cluster(tpu)
tf.tpu.experimental.initialize_tpu_system(tpu)
tpu_strategy = tf.distribute.experimental.TPUStrategy(tpu)

**Comment by DavidLandup0** [comment]
Last time I tried - it didn't work due to compatability issues. I know that CUDA definitely doesn't compile with the new TF versions on Kaggle due to limitations in their environments, even when they should. I'll try this again with the TPU, thanks!

**Comment by bhack** [comment]
> Last time I tried - it didn't work due to compatability issues. I know that CUDA definitely doesn't compile with the new TF versions on Kaggle due to limitations in their environments, even when they should. I'll try this again with the TPU, thanks!

For GPU is different you need to use something like:
`!apt install --allow-change-held-packages libcudnn8=8.1.0.77-1+cuda11.2` using the right version for TF version that you want to use.

**Comment by bhack** [comment]
P.s. you could retrieve the right cuda/cudnn version to install with:
```python
sys_details = tf.sysconfig.get_build_info()
cuda_version = sys_details["cuda_version"]
print(cuda_version)  
cudnn_version = sys_details["cudnn_version"]  
print(cudnn_version)
```

DavidLandup0:
> For GPU is different you need to use something like:
!apt install --allow-change-held-packages libcudnn8=8.1.0.77-1+cuda11.2 using the right version for TF version that you want to use.

This should work and it does on Colab, but doesn't on Kaggle. TF imports but throws unsupported kernel operations while training. A similar issue with versioning that highlights their issues with keeping environments updated is also discussed in: https://www.kaggle.com/product-feedback/296467 and https://github.com/Kaggle/docker-python/issues/1114 and https://www.kaggle.com/getting-started/210020

I can try it again, but remember not being able to find any solutions to upgrading TF-GPU on Kaggle about a month ago, without throwing unsupported kernel operations while training. :/

When using:

> !pip install tensorflow==2.10.0
import tensorflow as tf
from cloud_tpu_client import Client
...

The environment won't import Tensorflow, and throws a:

```
...
/opt/conda/lib/python3.7/site-packages/tensorflow/python/client/pywrap_tf_session.py in <module>
     17 # pylint: disable=invalid-import-order,g-bad-import-order, wildcard-import, unused-import
     18 from tensorflow.python import pywrap_tensorflow
---> 19 from tensorflow.python.client._pywrap_tf_session import *
     20 from tensorflow.python.client._pywrap_tf_session import _TF_SetTarget
     21 from tensorflow.python.client._pywrap_tf_session import _TF_SetConfig

ImportError: initialization failed
```

bhack:
I can run a basic mnist training without problem with

```notebook
!python -m pip install --upgrade pip
!pip install tensorflow==2.10.0
```
```python
import tensorflow as tf
sys_details = tf.sysconfig.get_build_info()
cuda_version = sys_details["cuda_version"]
print(cuda_version)  
cudnn_version = sys_details["cudnn_version"]  
print(cudnn_version)
print(tf.__version__)
mnist = tf.keras.datasets.mnist

(x_train, y_train), (x_test, y_test) = mnist.load_data()
x_train, x_test = x_train / 255.0, x_test / 255.0
model = tf.keras.models.Sequential([
  tf.keras.layers.Flatten(input_shape=(28, 28)),
  tf.keras.layers.Dense(128, activation='relu'),
  tf.keras.layers.Dropout(0.2),
  tf.keras.layers.Dense(10)
])
predictions = model(x_train[:1]).numpy()
predictions
loss_fn = tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True)
loss_fn(y_train[:1], predictions).numpy()
model.compile(optimizer='adam',
              loss=loss_fn,
              metrics=['accuracy'])
model.fit(x_train, y_train, epochs=5)
```

bhack:
For the trainable jax converted backbone  it is confirmed that we could fine-tune also the converted feature extractor:

https://github.com/google/jax/commit/b390426af0c8807de42ef9cd2e553459d6b03861

DavidLandup0:
After a bunch of testing, just increasing the initial learning rate got the script to behave more like Sayak's run at: https://github.com/sayakpaul/big_vision_experiments

image

It's not 100% the same, but with further training script tweaks, I think it will since it's close to it right now. Can we do a code review while I try out other runs or we do a GC run here? :)
@bhack @tanzhenyu @LukeWood

tanzhenyu:
> After a bunch of testing, just increasing the initial learning rate got the script to behave more like Sayak's run at: https://github.com/sayakpaul/big_vision_experiments
> 
> image
> 
> It's not 100% the same, but with further training script tweaks, I think it will since it's close to it right now. Can we do a code review while I try out other runs or we do a GC run here? :) @bhack @tanzhenyu @LukeWood

Seems to be long thread here, what's the TLDR?

DavidLandup0:
> Seems to be long thread here, what's the TLDR?

We were debating porting vs training. I tried porting the weights, and was successful in loading them in for each individual component, but the final model didn't have any predictive power (almost like the weights were random). The components were updated to more closely mimick the paper-method and to support XLA compilation (replaced `tf.image.extract_patches` with `Conv2D`). Bhack noted that the training results I sent initially weren't as good as the reference run from Sayak, so this is the new run which is closer.

I think I got it all?

tanzhenyu:
> > Seems to be long thread here, what's the TLDR?
> 
> We were debating porting vs training. I tried porting the weights, and was successful in loading them in for each individual component, but the final model didn't have any predictive power (almost like the weights were random). The components were updated to more closely mimick the paper-method and to support XLA compilation (replaced `tf.image.extract_patches` with `Conv2D`). Bhack noted that the training results I sent initially weren't as good as the reference run from Sayak, so this is the new run which is closer.
> 
> I think I got it all?

I see. by porting weights, is it feasible to compare outputs layer by layer? Seems like a debugging process to me if the ported model does not have any predictive power

DavidLandup0:
> I see. by porting weights, is it feasible to compare outputs layer by layer? Seems like a debugging process to me if the ported model does not have any predictive power

I don't _think_ it can be done? I haven't used JAX before, but as far as I've been able to see, we couldn't "chop off" the layers and observe the outputs to find which one(s) produce different outputs and isolate them that way. Same goes for the reconstructed `SavedModel` on TF-Hub (it's not a Keras instance and we don't get to peel the layers back)

bhack:
> Seems to be long thread here, what's the TLDR?

My point is that as now the jax2tf converted "features" could be trained in keras it has still it own value to test the reproducibility of the "downstream" training job.

Of course it would be nice to have the VIT "features" trained from scratch but as it is a complex job that requires also enough computing resources given a certain time budget I suppose that many users just want to train the pre-trained features on the downstream dataset.

@DavidLandup0 have you checked?
https://github.com/rwightman/pytorch-image-models/blob/main/timm/models/vision_transformer.py

As I see a lot of tricks concerting big-vision VIT weights in pytorch.

DavidLandup0:
I agree that we'll want to reproduce fine tuning too. Though, to get there, we'll need to either convert or pre-train a base feature extractor and then test fine tuning. I can do fine tuning on my own device, thankfully, so extensive testing, hyperparam searches, etc. isn't an issue.

I've referenced the timm implementation and lucidrains ViTs. Both have weights ported and seem to have the same structure as ours when I inspected them manually...
I'll cross-reference their JAX port again.

bhack:
> Though, to get there, we'll need to either convert or pre-train a base feature extractor and then test fine tuning. 

My open points are:
- can we optionally replicate the fine-tuning/downstream training as we have seen that we can now train the feature with the jax2tf approach?
- I suppose that porting weights without using jax2tf conversion is useful only if we want strictly have an  "unified code path".
- I think that it is useful to train from scratch but in the case you cannot replicate the performance or the gap is too large users are still strongly interested to train on their downstream datasets without starting from "a too large pre-trained penalty".

So generally I think that having a reproducible training from scratch it could be useful if you want to experiment changes in the training pipeline or in some model component.

But for the downstream task what is really the difference between optionally training on jax2tf converted features vs manually porting all the weights? 
I think that with a constrained time budget personally I am more then happy to replicate the downstream training from jax2tf features and at the same time having the best we could have from scratch without conflicts.

Then on the long run we could see what we can achieve with training from scratch so If the gap is too large we could always not going to impact the downstream training jobs.

DavidLandup0:
Oh, I think I get what you were saying now - I think we might've been misunderstanding each other for a bit there.
You think we should use the JAX2TF feature base and *serve it* as the pre-trained model for fine-tuning, and offer this PR for building the model for training from scratch?

Or did I misunderstand?

bhack:
Yes something like that as also the downstream training scripts/jobs need to be tested for reproducibility and it is another task to be contributed.

If we could make it optional in the API IMHO it could be fine.

Then if we will have a from scratch exceptional performance at some point in time we could not care about this option (tf2xa) anymore.

DavidLandup0:
I'm testing out minor differences like whether `broadcast_to` or `tile` is used, QKV bias True or False (timm sets it to False, but the bias does exist in JAX weights) etc.

DavidLandup0:
I am totally lost as to why this doesn't work: https://colab.research.google.com/drive/1cnZfZ8jx7sk0zUbfBiF-nLMyZVbQuWc1#scrollTo=lEz9usTbwRJo
Is there a difference in how MHA runs between TF and PyTorch, even if they have the same parameter shapes maybe?

sebastian-sz:
@DavidLandup0 I think you might have missed the GeLU activation function in ViT's. By default, activation is [set to None](https://github.com/keras-team/keras-cv/pull/1014/files#diff-ccc83ecec316037011be9cd8122b3cb75b7cf9def427cde534f123c8d8939124R188).

This also shows when we print the `TransformerEncoder`:
```
#!pip install --upgrade tensorflow
#!pip install "git+https://github.com/DavidLandup0/keras-cv@vit"
import keras_cv
model = keras_cv.models.ViT_B_16(include_rescaling=True, include_top=True, classes=1000, weights=None, classifier_activation=None, input_shape=(224, 224, 3))
for layer in model.layers:
    if isinstance(layer, keras_cv.layers.transformer_encoder.TransformerEncoder):
        print(layer.activation)  # None
```
In the contrast, when we inspect timm model in the Colab you linked - the GeLU activation function is used.

sebastian-sz:
@DavidLandup0 One last thing regarding Gelu: In jax they use tanh approximation ON by default:
https://jax.readthedocs.io/en/latest/_autosummary/jax.nn.gelu.html

This approximation is OFF by default in Keras and Pytorch:
https://www.tensorflow.org/api_docs/python/tf/keras/activations/gelu 
https://pytorch.org/docs/stable/generated/torch.nn.GELU.html

which might slightly throw your values off.

DavidLandup0:
Oh my god, I totally missed the missing activation, because I was sure it was set to 'gelu' by default instead of None, and didn't even check it. You're a lifesaver @sebastian-sz! Dziekuje bardzo :)

sebastian-sz:
@DavidLandup0 Proszę! Always happy to help! :)

DavidLandup0:
So it wasn't set to `None`:

```
    def __init__(self, activation='gelu', ...,
    ):
        super().__init__(**kwargs)
        ...
        self.activation = activation
        ...
        self.dense1 = layers.Dense(self.mlp_units[0], activation=activation)
```

Instead of:

```
self.dense1 = layers.Dense(self.mlp_units[0], activation=self.activation)
```

I *think* this might've caused it to fall back to `None`, for loaded models? Testing it out on the conversion script again

DavidLandup0:
Great news! The port seems to work :)

The ported ViTB16 does 72% top-1 and 92% top-5 on ImageNetV2.

- Porting and test set eval: https://colab.research.google.com/drive/1cnZfZ8jx7sk0zUbfBiF-nLMyZVbQuWc1#scrollTo=ixhjqYEvrIDt

Could we run a model against the same set as we're evaluating KCV models on to ensure that this isn't an apples to oranges comparison? @ianstenbit @tanzhenyu

bhack:
What we want to do in general to maintain the converting scripts in the repo for reproducibility?

As I think this is relevant (as it is in the jax2tf case) especially if we cannot sistematically train all the flavours from scratch without generating a too impacting pretraining gap on downstream training jobs.

bhack:
> What we want to do in general to maintain the conversion scripts in the repo for reproducibility?
> 
> As I think this is relevant (as it is in the jax2tf case) especially if we cannot sistematically train all the flavours from scratch without generating a too impacting pretraining gap on downstream training jobs.

Probably we have a WIP example about how to contribute weights conversion scripts at https://github.com/keras-team/keras-cv/pull/1056/files#diff-1ab9e852f691788ec4a67d2521318069cf1ded78299ac8b964c2356575f3d934

DavidLandup0:
Since the weight port works (#1063), should we start the architecture review?
It 
@tanzhenyu @LukeWood 

A TL:DR; for the longer thread is:

- Changed Patching and PatchEmbedding into PatchingAndEmbedding (to provide XLA support and weight porting)
- Added architecture
- Added/updated tests, and updated model tests

Sorry for on-the-fly changes to some of the components - before this PR, the main focus was modularity, but weight porting and XLA required a change since patching and embedding can't be separated as I originally wanted it to be.

tanzhenyu (commented):
This is great progress! Would you mind adding the port-weight script to this PR as well?

DavidLandup0:
Sure! Adding it as a script. Where should I put it? The examples directory?

tanzhenyu:
> Sure! Adding it as a script. Where should I put it? The examples directory?

you can create another directory , like this:
https://github.com/keras-team/keras-nlp/tree/master/tools/checkpoint_conversion

DavidLandup0:
Sure. Adding it and updating the PR in a moment.
Since it was a notebook that I manually updated for each model, I'll update it to be generic

DavidLandup0:
Yikes - the ipynb file is 4.4k lines long. Should I switch to a .py script instead?

tanzhenyu:
> Yikes - the ipynb file is 4.4k lines long. Should I switch to a .py script instead?

It's fine to keep it as ipynb, as long as we don't save the intermediate results.
This gives other users a chance to interact with it

tanzhenyu:
I think this PR is in a good state to be merged. But before we do that, can you create a demo .py file to verify the inference (and top-5 accuracy) for validation dataset?

DavidLandup0:
Awesome, thanks!
Sure thing - adding the script in the examples directory. 
There's a small part of the conversion notebook on Google Colab that does the inference (which is where I got the numbers from), but it'd be nice to have a `validate_on_imagenet.py` that can be run for any KCV model and just spits out the result, inference time, etc as a small report.

bhack:
> but it'd be nice to have a validate_on_imagenet.py that can be run for any KCV model and just spits out the result, inference time, etc as a small report.

I think this really need to be part of the CI with some custom trigger.

DavidLandup0:
Perhaps after gcbrun, we could have /benchmark?
We'd need a flag for including rescaling as well

bhack:
> Perhaps after gcbrun, we could have /benchmark? We'd need a flag for including rescaling as well

It could help to contribute if we could use the Gitbub action syntax for GPU resource https://github.com/keras-team/keras-cv/issues/907

tanzhenyu:
> Awesome, thanks! Sure thing - adding the script in the examples directory. There's a small part of the conversion notebook on Google Colab that does the inference (which is where I got the numbers from), but it'd be nice to have a `validate_on_imagenet.py` that can be run for any KCV model and just spits out the result, inference time, etc as a small report.

Do you have the colab link that I could help verify?

DavidLandup0:
Finished writing the script, commited locally, just need to push when I get wifi.

Here's the colab notebook with the weight porting and the ImageNetV2 inference in the end:
https://colab.research.google.com/drive/1cnZfZ8jx7sk0zUbfBiF-nLMyZVbQuWc1#scrollTo=ixhjqYEvrIDt

DavidLandup0:
That's ViT Tiny16. It has a significantly lower acc. Also, the paper reported acc is on ImageNet, while this is ImageNetV2 which tends to be 10% less in top-1 :)
This can also be seen in Sayak's runs

tanzhenyu:
> https://github.com/sayakpaul/big_vision_experiments

I see. In that case, can you reproduce S/16 in the colab?

DavidLandup0:
I did V2 because it's available in TFDS. Neither Val nor Real are available there for a quick download/eval. I can try to get my hands on the Val set, but it might be difficult to get it up on Collab 🤔

ayulockin:
Just here to admire the work going into it. Thanks all. <3

DavidLandup0:
Thanks Ayush! <3

DavidLandup0:
Re-run the colab with S16. 68% top-1 on ImageNetV2 as previously reported in the description. Can you take a look? @tanzhenyu

DavidLandup0:
Added the benchmark script! You run it as:

```
python examples/benchmarking/imagenet_v2.py --model_name=NAME
```

With flags for rescaling, batch size,  weight path (defaults to 'imagenet'), model name and model kwargs, and it spits out the results for V2. We might want to add imagenet_real and imagenet_val (the same set @ianstenbit uses for validation while training, which is on the team's GCS).

tanzhenyu (approved):

tanzhenyu:
> Added the benchmark script! You run it as:
> 
> ```
> python examples/benchmarking/imagenet_v2.py --model_name=NAME
> ```
> 
> With flags for rescaling, batch size, weight path (defaults to 'imagenet'), model name and model kwargs, and it spits out the results for V2. We might want to add imagenet_real and imagenet_val (the same set @ianstenbit uses for validation while training, which is on the team's GCS).

Excellent! Thank you for the hard work and contribution. Merging

tanzhenyu:
manual merging due to gcp testing broken

keras_cv/models/__init__.py
```diff
@@ -60,3 +60,13 @@
 from keras_cv.models.resnet_v2 import ResNet152V2
 from keras_cv.models.stable_diffusion import StableDiffusion
 from keras_cv.models.vgg19 import VGG19
+from keras_cv.models.vit import ViT_B_16
```
LukeWood, line 63: can we rename these to remove the `_`s?   ViT_B_16 deviates from the rest of the Keras naming convention.  I guess We can do ViTB16 instead?

DavidLandup0: Deal! Updating in a moment :)

DavidLandup0:
Awesome, thank you for the merge, patience and reviews @tanzhenyu! :)
Really excited to get SwinTs, DeiTs, etc. on their way!

tanzhenyu:
> Awesome, thank you for the merge, patience and reviews @tanzhenyu! :) Really excited to get SwinTs, DeiTs, etc. on their way!

As the next step, can you come up with a transformer training script that can run Tiny / Small for ~10 epochs?

DavidLandup0:
I already have one, and the first 10 epochs are clean, but after a few more, a divergence happens and the model starts overfitting.
Since it's on TPU on Kaggle, I don't have access to new TF versions, so I did my own implementations for MixUp, but it's not enough to regularize the model.

If the GC enviro here can run KerasCV, can I share the setup for a run here? Turns out that ViTs trained with a basic script (much like the ones for CNNs) actually perform better than the original protocols with tricks. We can use our basic script with dialled down augmentations and it should do the trick just fine :)

tanzhenyu:
> I already have one, and the first 10 epochs are clean, but after a few more, a divergence happens and the model starts overfitting. Since it's on TPU on Kaggle, I don't have access to new TF versions, so I did my own implementations for MixUp, but it's not enough to regularize the model.
> 
> If the GC enviro here can run KerasCV, can I share the setup for a run here? Turns out that ViTs trained with a basic script (much like the ones for CNNs) actually perform better than the original protocols with tricks. We can use our basic script with dialled down augmentations and it should do the trick just fine :)

Meanwhile, would you be interested in MaxViT? That has a higher priority than SwinT / DeiT and easier to implement since you already have some re-usable block

DavidLandup0:
Yes! Could you assign me on the issue?
I'll start reading the relevant paper(s) and get to it soon :)

tanzhenyu:
> Yes! Could you assign me on the issue? I'll start reading the relevant paper(s) and get to it soon :)

https://github.com/keras-team/keras-cv/issues/1144
Feel free to start with it, but I will keep this open just in case anyone else also wants to contribute

PR #1017, opened 2022-11-15
[Object Detection] Update retinanet predict loop to enable TPU support

ianstenbit:
I've tested the full training + evaluation loop on TPU on [Colab](https://colab.sandbox.google.com/drive/16Lb6rMcFvtoic7DLd89Ap5-VxLa1dX_4?resourcekey=0-sDnh3nGFCx-6c7l_jnxShQ) with this change, and it's all working:


tanzhenyu:
> Colab

Need to set soft_device_placement to False in order to verify it

PR #1020, opened 2022-11-15
Customized op for WithinBox3d.

tanzhenyu:
pip uninstall keras_cv
bazel build keras_cv/custom_ops:all
mv bazel-bin/keras_cv/custom_ops/*.so keras_cv/custom_ops/
bazel build build_pip_pkg
./bazel-bin/build_pip_pkg wheels
pip install wheels/keras_cv-0.0.0-cp39-cp39-linux_x86_64.whl


keras_cv/ops/point_cloud.py
```diff
@@ -13,6 +13,16 @@
 # limitations under the License.
 
 import tensorflow as tf
+from tensorflow.python.framework import load_library
+from tensorflow.python.platform import resource_loader
+
+custom_ops = load_library.load_op_library(
```
ianstenbit, line 19: In order to support wheels without custom binaries, I think we'll need to try-catch this so that we don't break at import time if the .so files are missing.

For IoU3D I did this by making IoU3D an object, but IIRC we decided that's probably not the right solution, so maybe we just want to try/except this and raise an Exception when the op is called if we weren't able to load the custom ops at load time.

wdyt?

bhack: Check https://github.com/tensorflow/addons/pull/869

tanzhenyu: Done.

keras_cv/ops/within_box_3d_test.py
```diff
@@ -0,0 +1,188 @@
...
+import tensorflow as tf
+
+import keras_cv
+
+num_points = 200000
+num_boxes = 1000
+box_dimension = 20.0
+
+
+def get_points_boxes():
```
lengzq, line 30: Do you ensure no overlap between two bounding boxes? The v2 op assumes no overlaps, right?

lengzq (approved):

keras_cv/custom_ops/kernels/withinbox_op.cc
```diff
@@ -0,0 +1,139 @@
...
+        points_y_indices[idx] = p_set;
+      }
+    };
+    device.parallelFor(num_boxes, cost, set_fn_y);
+
+    // for the intersection of x indices set and y indices set, check if
+    // those points are within the box
+    auto within_fn = [&points_x_indices, &points_y_indices, &boxes_vec, &points_vec, &boxes_indices_t](int64_t begin, int64_t end) {
+      for (int64_t idx = begin; idx < end; ++idx) {
+        std::unordered_set<int>& set_a = points_x_indices[idx];
```
Guowang, line 113: A side note is https://github.com/abseil/abseil-cpp/blob/master/absl/container/flat_hash_map.h available here?

keras_cv/custom_ops/kernels/withinbox_op.cc
```diff
@@ -0,0 +1,139 @@
...
+        std::unordered_set<int>& set_b = points_y_indices[idx];
+        std::unordered_set<int> p_set;
+        for (auto val : set_a) {
```
Guowang, line 116: Nit: std::set_intersection?

Guowang (approved):

keras_cv/ops/within_box_3d_test.py
```diff
@@ -0,0 +1,188 @@
...
+        "TEST_CUSTOM_OPS" not in os.environ or os.environ["TEST_CUSTOM_OPS"] != "true",
+        reason="Requires binaries compiled from source",
+    )
+    def test_many_points(self):
+        points, boxes = get_points_boxes()
+
+        for _ in range(5):
+            print(datetime.now())
+            res = keras_cv.ops.within_box3d_index(points, boxes)
+            self.assertAllClose(res.shape, points.shape[:1])
```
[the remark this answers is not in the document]
tanzhenyu: will compare this once we have "is_within_any_box3d"

keras_cv/ops/within_box_3d_test.py
tanzhenyu: Good point. Added docstring

keras_cv/custom_ops/kernels/withinbox_op.cc
tanzhenyu: that only works for sorted range of set_a and set_b

keras_cv/custom_ops/kernels/withinbox_op.cc
tanzhenyu: can you elaborate? (I'm not sure if I understand the comment)

Guowang: flat_hash_set is more efficient than std:: unordered_set, but we can check the profiling first to see whether this is the bottleneck.

PR #1021, opened 2022-11-16
[Object Detection] RetinaNet: custom predict_function

ianstenbit:
This allows us to perform the model inference using our distributed strategy and perform decoding on the local CPU.

Decoding on distributed devices doesn't work for TPU, and this change fixes that.
Eventually, TPUs should support the necessary op for decoding in a distributed fashion.

Missing op message:
```
No registered 'CombinedNonMaxSuppression' OpKernel for XLA_TPU_JIT
```

Unfortunately, to make this work I had to copy-paste the `keras.Model` implementation of `make_predict_function`, with the only changes being:

- Accessing `jit_compile` via `self._jit_compile` because `self.jit_compile` is a property defined on a superclass
- Replacing this code:
```
data = next(iterator)
outputs = model.distribute_strategy.run(run_step, args=(data,))
outputs = reduce_per_replica(
    outputs, self.distribute_strategy, reduction="concat"
)
return outputs
```
with
```
data = next(iterator)
outputs = model.distribute_strategy.run(run_step, args=(data,))
outputs = reduce_per_replica(
    outputs, self.distribute_strategy, reduction="concat"
)
if not isinstance(data, tf.Tensor):
    data = tf.concat(data.values, axis=0)
return self.decode_training_predictions(data, outputs)
```

I think this is a bit of a hack, but let's to discuss

### DIFF
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -484,8 +484,10 @@ class RetinaNet(ObjectDetectionBaseModel):
             outputs = reduce_per_replica(
                 outputs, self.distribute_strategy, reduction="concat"
             )
+            if not isinstance(data, tf.Tensor):
+                data = tf.concat(data.values, axis=0)
             return self.decode_training_predictions(
-                tf.concat(data.values, axis=0), outputs
+                data, outputs
             )
 
         # Special case if steps_per_execution is one.
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -14,6 +14,9 @@
 
 import numpy as np
 import tensorflow as tf
+from keras.engine.training import _minimum_control_deps
+from keras.engine.training import reduce_per_replica
+from keras.utils import tf_utils
 from tensorflow import keras
 
 import keras_cv
@@ -457,9 +460,74 @@ class RetinaNet(ObjectDetectionBaseModel):
         self._update_metrics(y_for_metrics, predictions)
         return {m.name: m.result() for m in self.metrics}
 
-    def predict(self, x, **kwargs):
-        predictions = super().predict(x, **kwargs)
-        return self.decode_training_predictions(x, predictions)
+    def make_predict_function(self, force=False):
+        if self.predict_function is not None and not force:
+            return self.predict_function
+
+        def step_function(model, iterator):
+            """Runs a single evaluation step."""
+
+            def run_step(data):
+                outputs = model.predict_step(data)
+                # Ensure counter is updated only if `test_step` succeeds.
+                with tf.control_dependencies(_minimum_control_deps(outputs)):
+                    model._predict_counter.assign_add(1)
+                return outputs
+
+            if self._jit_compile:
+                run_step = tf.function(
+                    run_step, jit_compile=True, reduce_retracing=True
+                )
+
+            data = next(iterator)
+            outputs = model.distribute_strategy.run(run_step, args=(data,))
+            outputs = reduce_per_replica(
+                outputs, self.distribute_strategy, reduction="concat"
+            )
+            return self.decode_training_predictions(
+                tf.concat(data.values, axis=0), outputs
+            )
+
+        # Special case if steps_per_execution is one.
+        if (
+            self._steps_per_execution is None
+            or self._steps_per_execution.numpy().item() == 1
+        ):
+
+            def predict_function(iterator):
+                """Runs an evaluation execution with a single step."""
+                return step_function(self, iterator)
+
+        else:
+
+            def predict_function(iterator):
+                """Runs an evaluation execution with multiple steps."""
+                outputs = step_function(self, iterator)
+                for _ in tf.range(self._steps_per_execution - 1):
+                    tf.autograph.experimental.set_loop_options(
+                        shape_invariants=[
+                            (
+                                outputs,
+                                tf.nest.map_structure(
+                                    lambda t: tf_utils.get_tensor_spec(
+                                        t, dynamic_batch=True
+                                    ).shape,
+                                    outputs,
+                                ),
+                            )
+                        ]
+                    )
+                    step_outputs = step_function(self, iterator)
+                    outputs = tf.nest.map_structure(
+                        lambda t1, t2: tf.concat([t1, t2]), outputs, step_outputs
+                    )
+                return outputs
+
+        if not self.run_eagerly:
+            predict_function = tf.function(predict_function, reduce_retracing=True)
+        self.predict_function = predict_function
+
+        return self.predict_function
 
     def _update_metrics(self, y_true, y_pred):
         y_true = bounding_box.convert_format(
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -486,9 +486,7 @@ class RetinaNet(ObjectDetectionBaseModel):
             )
             if not isinstance(data, tf.Tensor):
                 data = tf.concat(data.values, axis=0)
-            return self.decode_training_predictions(
-                data, outputs
-            )
+            return self.decode_training_predictions(data, outputs)
 
         # Special case if steps_per_execution is one.
         if (
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -14,9 +14,6 @@
 
 import numpy as np
 import tensorflow as tf
-from keras.engine.training import _minimum_control_deps
-from keras.engine.training import reduce_per_replica
-from keras.utils import tf_utils
 from tensorflow import keras
 
 import keras_cv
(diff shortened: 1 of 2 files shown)


bhack:
I think cause:
https://github.com/keras-team/keras-cv/blob/fc13f60a29c1ea294bc72029e510ca81e3e65c56/keras_cv/layers/object_detection/non_max_suppression.py#L123

Is not covered by the TF Bridge/XLA

ianstenbit:
> It is supported already by the new TF MLIR bridge:
> 
> ```python
> tf.config.experimental.enable_mlir_bridge()
> ```

Thanks for this pointer @bhack!
After enabling the MLIR bridge, I'm still seeing XLA errors when trying to run on TPU, this time related to the loss function it seems.

```
(0) INVALID_ARGUMENT: {{function_node __inference_train_function_70007}} TF to XLA legalization failed: <unknown>:0: error: loc(callsite(fused["SigmoidGrad:", "gradient_tape/FocalLoss/Sigmoid/SigmoidGrad@__inference_run_step_69903"] at fused["StatefulPartitionedCall:", "StatefulPartitionedCall"])): 'mhlo.constant' op result #0 must be statically shaped tensor of floating-point or pred (AKA boolean or 1-bit integer) or 8/16/32/64-bit signless integer or 8/16/32/64-bit unsigned integer or complex type with 32-bit float or 64-bit float elements values, but got 'tensor<?x?x20xf32>'
<unknown>:0: note: loc(fused["StatefulPartitionedCall:", "StatefulPartitionedCall"]): called from
<unknown>:0: note: loc(callsite(fused["SigmoidGrad:", "gradient_tape/FocalLoss/Sigmoid/SigmoidGrad@__inference_run_step_69903"] at fused["StatefulPartitionedCall:", "StatefulPartitionedCall"])): see current operation: %5928 = "mhlo.constant"() {value = dense<1.000000e+00> : tensor<?x?x20xf32>} : () -> tensor<?x?x20xf32>
```.

I am still digging to try to work out root cause of this one.

**Comment by bhack** [comment]
> After enabling the MLIR bridge, I'm still seeing XLA errors when trying to run on TPU, this time related to the loss function it seems.

Yes and It could be not the lat one as it is currently the fail fast design of XLA (that I would contribute to change):

https://github.com/tensorflow/tensorflow/issues/57923

**Comment by bhack** [comment]
I think that it is failing for @ukoxyz check at https://github.com/tensorflow/tensorflow/commit/76079c00ac45b2bd48b3749a490e464149f4b244#diff-20844a653ebc1cb9b4b2adb26672259647f08dc1e78a06458d2e753a1449336bR2070 

In our https://github.com/keras-team/keras-cv/blob/de196f5295063abe754bf255dea4316fe7023a7d/keras_cv/losses/focal.py#L81-L82

**Comment by bhack** [comment]
@ianstenbit Let me know if you could share a TPU Colab to reproduce the issue.

**Comment by ianstenbit** [comment]
> tensorflow/tensorflow@76079c0#diff-20844a653ebc1cb9b4b2adb26672259647f08dc1e78a06458d2e753a1449336bR2070

@bhack here is a Colab which uses this change and demonstrates RetinaNet working on TPU:
https://colab.research.google.com/gist/ianstenbit/d1408a030b9ee1a97d417fcbb78e5913/retinanet-on-tpu.ipynb

If you update the pip install at the top to install from keras-team/keras-cv.git@master, youll see the TPU/XLA issues when you call `model.predict`

**Comment by smit-hinsu** [comment]
CombinedNonMaxSuppression is not supported by any (old or new) bridges.

**Comment by ianstenbit** [comment]
I've now confirmed that with this change, RetinaNet training and evaluation is working on [TPU](https://colab.research.google.com/drive/16Lb6rMcFvtoic7DLd89Ap5-VxLa1dX_4?resourcekey=0-sDnh3nGFCx-6c7l_jnxShQ#scrollTo=s-CIYXnq52vy) and [GPU](https://colab.research.google.com/drive/1uTgwEVNuwoAK-DmXKdvAKK1BAJ7CxGSk?resourcekey=0-0kjR4ofdWirv9IUNcS4mHQ#scrollTo=s-CIYXnq52vy)

**Comment by bhack** [comment]
> CombinedNonMaxSuppression is not supported by any (old or new) bridges.

Do you mean with new bridge also on CPU/GPU?

**Comment by bhack** [comment]
@smit-hinsu Is this correct https://colab.research.google.com/gist/bhack/fbd3ead4379d687e115175529d0490d3/untitled309.ipynb ?

**Comment by smit-hinsu** [comment]
That's a pleasant suprise. TPU might be using outside compilation for the op but GPU doesn't have that so I am not sure.  In that case, the Outside compilation should have worked on old bridge as well.

File: keras_cv/models/object_detection/retina_net/retina_net.py [hunk 1]
```diff
@@ -14,6 +14,9 @@
 
 import numpy as np
 import tensorflow as tf
+from keras.engine.training import _minimum_control_deps
```
LukeWood, line 17: @fchollet opinion on importing these from `keras` directly?

LukeWood (changes requested):
This approach looks good to me.  

Upon further inspection I DO think we should move this to ObjectDetectionBaseModel; accepting a `prediction_decoder` attribute.  This allows us to keep the RetinaNet super clean and concise, while the OD base model handles the unfortunate TensorFlow code

bhack:
@smit-hinsu Do you know instead what is the shape issue we had in Sigmoid grad at https://github.com/keras-team/keras-cv/pull/1021#issuecomment-1317494118?

It seems that the shape it is not fully defined.

ianstenbit:
> This approach looks good to me.
> 
> Upon further inspection I DO think we should move this to ObjectDetectionBaseModel; accepting a `prediction_decoder` attribute. This allows us to keep the RetinaNet super clean and concise, while the OD base model handles the unfortunate TensorFlow code

Okay I've moved this to the base OD model, which now has a default implementation of `decode_predictions` (which I renamed from `decode_training_predictions`)

smit-hinsu:
> @smit-hinsu Do you know instead what is the shape issue we had in Sigmoid grad at [#1021 (comment)](https://github.com/keras-team/keras-cv/pull/1021#issuecomment-1317494118)?
> 
> It seems that the shape it is not fully defined.

could you provide the full error log? That might help see what is going on.

bhack:
@ianstenbit Can you share a more extended chunk  of the sigmoidgrad error log?

ianstenbit:
> @ianstenbit Can you share a more extended chunk of the sigmoidgrad error log?

Repro colab [here](https://colab.research.google.com/drive/1ycEDn-aW4ygrEOWOnmc130Vk_RPZwn92?resourcekey=0--zpX8Ftx0Z5nrBan84Nuxw&usp=sharing)
[Specific cell with Sigmoidgrad error](https://colab.research.google.com/drive/1ycEDn-aW4ygrEOWOnmc130Vk_RPZwn92?resourcekey=0--zpX8Ftx0Z5nrBan84Nuxw#scrollTo=4sPJM_SMyChC)

bhack:
> > @ianstenbit Can you share a more extended chunk of the sigmoidgrad error log?
> 
> Repro colab [here](https://colab.research.google.com/drive/1ycEDn-aW4ygrEOWOnmc130Vk_RPZwn92?resourcekey=0--zpX8Ftx0Z5nrBan84Nuxw&usp=sharing)
> [Specific cell with Sigmoidgrad error](https://colab.research.google.com/drive/1ycEDn-aW4ygrEOWOnmc130Vk_RPZwn92?resourcekey=0--zpX8Ftx0Z5nrBan84Nuxw#scrollTo=4sPJM_SMyChC)

It is private

ianstenbit:
> It is private

https://colab.research.google.com/gist/ianstenbit/6cbcf6e034fbf63e22208f3d9a479297/sigmoid-loss-issue-retinanet-on-tpu.ipynb

bhack:
@smit-hinsu Can you check the legalization failed in that colab run?

smit-hinsu:
It doesn't seem to have the full context so difficult to say. We will need to get a reproducer with the full graph using TF_DUMP_GRAPH_PREFIX to some filesystem path. It should emit a .mlir file there. I don't know how colab interacts with filesystem but you may know.

bhack:
@ianstenbit I didn't find an available free TPU on colab in the last hours. Can you try to dump and share the graph setting `TF_DUMP_GRAPH_PREFIX` path  with `os.environ()`?

LukeWood (approved):

LukeWood:
LGTM, I like the approach of the base model having a decode predictions function.

ianstenbit:
> @ianstenbit I didn't find an available free TPU on colab in the last hours. Can you try to dump and share the graph setting `TF_DUMP_GRAPH_PREFIX` path with `os.environ()`?

@bhack I wasn't able to get this to dump logs. I tried setting the logs path to a GCS bucket, since TPU Colabs seem to prefer file I/O via GCS, but no dice.

I can take another look next week but for now hopefully access to the Colab is helpful at least?

bhack:
If this is still valid today it will be hard to dump or not?

https://stackoverflow.com/questions/63278991/dump-hlo-ir-for-tpu-while-using-tpuclusterresolver

bhack:
> @bhack I wasn't able to get this to dump logs. I tried setting the logs path to a GCS bucket, since TPU Colabs seem to prefer file I/O via GCS, but no dice.

I have now a free TPU colab instance. Can you share a pre-populated `gs://` of the dataset as the extraction is super slow.
Thanks.

PR #1022, opened 2022-11-16
Revert "[Object Detection] Update retinanet predict loop to enable TPU support"

ianstenbit:
Reverts keras-team/keras-cv#1017

As per our discussion, this seems to have broken `predict` for tf.data.Dataset, so we'll need to find another path forward. For the time being, I think we should roll this back.


LukeWood (approved):
Looks good, thanks and looking forward to fixing tomorrow!

PR #1024, opened 2022-11-16
[Codebase Consistency] MLPMixer Refactoring

DavidLandup0:
# What does this PR do?

The MLPMixer architecture is relevant in the general vision transformer project, and there's a difference in conventions and design choices between MLP Mixer and general KCV conventions, which are proposed to be ironed out in this PR. Highlights:

- Removed necessary `patch_size` argument from the main model call to conform to all other KerasCV models
- The `patch_size` shouldn't be modifiable or exposed for a given architecture, such as MLPMixerB16 because the *16* in the name stands for the patch size (similarly, we don't offer a `depth` argument for `ResNet50` because the `50` is the depth)
- The `patch_size` was changed from a tuple (ex: `(16, 16`) to an int (ex: `16`). They have to be of the same size and a check in the model raises a `ValueError` if they're not uniform in size. Instead of enforcing this through a check, and having a more verbose version, we can collapse it to a single `int` like with `ViT` and most Keras arguments such as `strides=3` which gets converted into `strides=(3, 3)` under the hood
- Updated tests to reflect these changes
- The original file was missing `MODEL_CONFIGS` for ease of defining various setups. 
- The original file was missing a base docstring that's updated dynamically.

After the update - all tests pass, and the model loads and builds just fine - these were refactoring changes, no logic (other than unnecessary conditionals) was changed.

### Additional Considerations

- The paper notes S16, S32, B16, B32, L16 and L32 setups, but we currently only have B16, B32 and L16. Should we include the other ones too?
- We now have a patching and embedding layer built specifically to be used as reusable components for models like these. We might want to replace the `Conv2D` that does patching with the tested components instead. This would be a design change, so I held it off before consulting. Thoughts?

## Before submitting
      Pull Request section?

## Who can review?

@LukeWood @tanzhenyu 
Also tagging @sayakpaul for thoughts because he contributed the model.


DavidLandup0:
Pinging for codebase/API consistency since no functional difference is made 
@tanzhenyu @LukeWood

tanzhenyu (approved):

tanzhenyu:
Looks like there are merge conflicts

DavidLandup0:
Whoops - looks like some new merge collided with a couple of lines in a file. Resolved. Let's just check whether all tests run fine

PR #1026, opened 2022-11-17
[Object Detection] Speed up the PyCOCO callback by only calling model.predict once

ianstenbit:
I did some benchmarking on Colab ([example](https://colab.research.google.com/drive/1ucPTx6GWI_YG6GMe6L8rkF1BdRxFtDqw?resourcekey=0-WcM6cnxkRvJmKdbDDgisRA#scrollTo=4sPJM_SMyChC) notebook using this), and found this is about 30x speedup over calling `model.predict` in a loop and manually batching. It makes the code a lot easier to understand, too IMO.

~This is a draft because it's currently contingent upon the evaluation dataset not being shuffled. (Shuffling causes inconsistent iteration order and this approach requires two iterations through the evaluation dataset -- one in the call to `model.predict`, and one to convert gt data to the COCO format)~

### DIFF
--- a/keras_cv/callbacks/pycoco_callback.py
+++ b/keras_cv/callbacks/pycoco_callback.py
@@ -36,7 +36,7 @@ class PyCOCOCallback(Callback):
             return boxes
 
         images_only_ds = self.val_data.map(images_only)
-        y_pred = model.predict(images_only_ds)
+        y_pred = self.model.predict(images_only_ds)
 
         gt = [boxes for boxes in self.val_data.map(boxes_only)]
         gt_boxes = tf.concat(
--- a/keras_cv/callbacks/pycoco_callback.py
+++ b/keras_cv/callbacks/pycoco_callback.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import six
 import tensorflow as tf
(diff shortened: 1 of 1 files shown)


keras_cv/callbacks/pycoco_callback.py
```diff
@@ -20,83 +19,88 @@
...
+                Note that if the validation dataset has shuffling of any kind
+                (e.g from `shuffle_files=True` in a call to TFDS.load or a call
+                to tf.data.Dataset.shuffle() with `reshuffle_each_iteration=True`),
+                you **must** cache the dataset to preserve iteration order. This
+                will store your entire dataset in main memory, so for large datasets
+                consider avoiding shuffle operations and passing `cache=False`.
+        """
         self.model = None
         self.val_data = validation_data
+        if cache:
```
tanzhenyu, line 41: why this cannot be cached when building the pipeline?
```
eval_ds = ...
eval_ds = eval_ds.cache()
```

tanzhenyu: If cache is for performance related, instead of determinism, maybe this is fine? Not related to coco, but for other evaluation dataset, it's generally better to be cached

ianstenbit: Caching by default is to enforce determinism. The purpose of structuring the API this way is to give users the power to control caching and determinism themselves if they want.

tanzhenyu (approved):

Issue #904, opened 2022-10-11
Xception architecture


DavidLandup0:
**Short Description**

Do we want to add Xception to KerasCV? Seems like a natural fit, and I've had Xception perform quite similar to EfficientNets (1-3) in most cases. It has nearly 10k citations, isn't hard to implement, and can be ported with the same pre-trained weights from `keras.applications` if need be.

**Other Information**

If so, sign me up to port it!

tanzhenyu:
This is Francois's domain, I will let him comment.
I think we should :-)

jbischof:
Closing as stale for now. Xception is not a SoTA arch now.

PR #1030, opened 2022-11-17
Implement xception

AmedeoBiolatti:
Implement Xception (https://github.com/keras-team/keras-cv/issues/904)

Using a downloaded version of the _keras.applications_ weights it works, should be sufficient to set them up in the _weights.py_ file to make it work completely

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood and @tanzhenyu


keras_cv/models/xception.py
```diff
@@ -0,0 +1,342 @@
...
+    **kwargs,
+):
+    return GeneralXception(
+        MODEL_CONFIGS["Xception"]["entry_flow_filters"],
+        MODEL_CONFIGS["Xception"]["middle_flow_filters"],
+        MODEL_CONFIGS["Xception"]["exit_flow_filters"],
+        include_rescaling,
+        include_top,
+        name=name,
+        weights=parse_weights(weights, include_top, "xception"),
```
tanzhenyu, line 332: wouldn't the parse_weights error out given this PR doesn't provide any weights?

keras_cv/models/xception.py
```diff
@@ -0,0 +1,342 @@
...
+    "Xception": {
+        "entry_flow_filters": [32, 64, 128, 256, 728],
+        "middle_flow_filters": [728 for _ in range(8)],
+        "exit_flow_filters": [(728, 1024), 1536, 2048],
+    }
+}
+
+BN_AXIS = 3
+
+BASE_DOCSTRING = """Instantiates the {name} architecture.
```
tanzhenyu, line 36: can you try this model with our existing training script?

tanzhenyu (commented):
Thanks for the PR!

keras_cv/models/xception.py
AmedeoBiolatti: At the moment it can handle two cases 

- `weights=None` -> default random initialization
- `weights="some_local_file.h5"` -> load the weights from a local file

It gives an error if you try `weights="imagenet"`, since the corresponding file is not defined/available

keras_cv/models/xception.py
AmedeoBiolatti: I will run it on Colab and share the link asap

AmedeoBiolatti: I am having issues downloading imagenet since I am not affiliated with any research institution. Looking forward to any possible solution

AmedeoBiolatti: I finally managed to get it run, but something is not working right...

xception

I did use `use_mixed_precision=True`, is it possible that the mixed precision leads to this kind of divergence early on? I used the defaults for all the other parameters.

Otherwise it could be something in the preprocessing of the data, since currently there is no detailed information on how to do it I had to use my gut feeling. Let's say that I would not expect that kind of problem given some minor difference in the preprocessing.

DavidLandup0: Quick question - why not download and re-use the existing weights from `keras.applications`?
I don't think we need to re-train Xception, since it's quite expensive to train (Separable convolutions don't utilize hardware optimizations well, even though they're technically more efficient than regular Conv2Ds).

This doesn't look like an MP issue, but rather a setup issue. Did you try to replicate the setup from: https://arxiv.org/pdf/1610.02357.pdf?

I can help out with the setup and porting if need be.

AmedeoBiolatti: The `keras.applications` weights work fine, but I think that the objective here was more about testing the model than getting a new set of trained weights

DavidLandup0: If it's ported from keras.applications, the model should work fine - the only difference being the beginning rescaling layer (probably won't work with original weights, but should train just as well from scratch).
Have you tried the basic_script.py from KCV with Xception-noted parameters or a custom script?

The original training protocol from the paper uses a LR decay every two epochs, which currently isn't present in the training script here, but the schedule shouldn't change too much. 

Besides that, just using the same parameters as in the paper should yield a similar result, as the training protocol outlined there wasn't tricky.

AmedeoBiolatti: Yeah, to make the original weights work you need to exclude the rescaling, but should be quite easy to make a different version that works with rescaling.

Regarding the differences in the original paper I notice

- image size (224 vs 299)
- learning rate schedule (reduce on plateau vs decay every 2 epochs)
- weight decay (no decay vs l2=1e-5)
- dropout (no dropout vs rate 0.5 before softmax layer)
- augmentation (no details)

I would wait someone from the KCV team to decide how to proceed

tanzhenyu: The new model should work with the `basic_training` script we have and ideally return metrics similar to the original paper.
@AmedeoBiolatti would you mind running that if you have access to Kaggle kernels or colab GPUs? Ideally we should have a mechanism to allow external contributors to use some GCP resources, however we haven't had the chance to set that up yet

tanzhenyu: Please kindly let us know if the xception model can be trained with our training script

AmedeoBiolatti: @tanzhenyu Before running the training script again I would like to understand what did go wrong in the previous attempt.  Sadly the dataset is too big for Kaggle kernels and free access colab GPU are too slow and sessions too short, so I did use a RunPod instance.

Could it be the usage of mixed precision? IMHO the difference in performance it too big to be so.

Otherwise it could be something in the preprocessing. Since there are still no details on the tfrecords generation I did try to get it right by intuition, maybe some error on my side... I did also try to reduce the image size before saving the tfrecods to reduce the file sizes, maybe I was too aggressive on that side too

jbischof:
Closing as stale. Feel free to file a new issue if you want to propose adding in our new API @AmedeoBiolatti!

Issue #1034, opened 2022-11-21
Exception encountered when calling layer 'group_normalization_60


charbull:
Referencing this issue: https://github.com/keras-team/keras-cv/issues/1033#issuecomment-1321775815 and opening a pull request accordingly as suggested.

```
[/usr/local/lib/python3.7/dist-packages/keras_cv/models/generative/stable_diffusion/__internal__/layers/group_normalization.py](https://localhost:8080/#) in _create_broadcast_shape(self, input_shape)
     85 
     86     def _create_broadcast_shape(self, input_shape):
---> 87      broadcast_shape = [1] * len(input_shape) 
     88         broadcast_shape[self.axis] = input_shape[self.axis] // self.groups

TypeError: Exception encountered when calling layer 'group_normalization_60' (type GroupNormalization).

len is not well defined for a symbolic Tensor (Shape:0). Please call `x.shape` rather than `len(x)` for shape information.

Call arguments received by layer 'group_normalization_60' (type GroupNormalization):
  • args=('tf.Tensor(shape=(None, 64, 64, 320), dtype=float32)',)
  • kwargs=<class 'inspect._empty'>
  ```

bhack:
We could close this.

charbull:
sgtm ! Thank you !

Issue #1033, opened 2022-11-21
StableDiffusion Tensorflow to TF Lite


charbull:
Hi @LukeWood,

For fun, I tried converting stable Diffusion model from Tensorflow to TF lite, so that I can run it on coral/edge tpu.

I tried two approaches:
I- Saved model approach:
II- Go through h5

will try to document them as much as possible. (sorry in advance for the long traces)

for both:
```
!pip install git+https://github.com/divamgupta/stable-diffusion-tensorflow --upgrade 
!pip install tensorflow tensorflow_addons ftfy --upgrade
```
Using `!pip install --upgrade keras-cv` I was not able to save the model for both.

I- Saved model approach:
1. Saved the model in a directory
```
from stable_diffusion_tf.stable_diffusion import StableDiffusion
model = StableDiffusion(
    img_height=512,
    img_width=512,
)
model.diffusion_model.save('/saved_model')
```
2. lets try to load it:
```
import tensorflow as tf
model2 = tf.keras.models.load_model('/saved_model')
converter = tf.lite.TFLiteConverter.from_keras_model(model2)
tflite_model = converter.convert()
open("converted_model.tflite", "wb").write(tflite_model)
```

3. The following error is thrown:
```
WARNING:tensorflow:No training configuration found in save file, so the model was *not* compiled. Compile it manually.
WARNING:absl:Found untraced functions such as dense_328_layer_call_fn, dense_328_layer_call_and_return_conditional_losses, dense_329_layer_call_fn, dense_329_layer_call_and_return_conditional_losses, group_normalization_173_layer_call_fn while saving (showing 5 of 1200). These functions will not be directly callable after loading.
INFO:tensorflow:Assets written to: /tmp/tmpicdc9dmk/assets
INFO:tensorflow:Assets written to: /tmp/tmpicdc9dmk/assets
---------------------------------------------------------------------------
ConverterError                            Traceback (most recent call last)
<ipython-input-15-52d23a3e5390> in <module>
      2 model2 = tf.keras.models.load_model('mydata/ivo/pythalpha/saved_model')
      3 converter = tf.lite.TFLiteConverter.from_keras_model(model2)
----> 4 tflite_model = converter.convert()
      5 open("converted_model.tflite", "wb").write(tflite_model)

/opt/conda/envs/rapids/lib/python3.8/site-packages/tensorflow/lite/python/lite.py in wrapper(self, *args, **kwargs)
    931   def wrapper(self, *args, **kwargs):
[...]

PR #1035, opened 2022-11-21
Update group_normalization.py

charbull:
opening a pull request to fix the len while saving the model as suggested by @freedomtan

reference context: https://github.com/keras-team/keras-cv/issues/1033
specific issue: https://github.com/keras-team/keras-cv/issues/1034

[/usr/local/lib/python3.7/dist-packages/keras_cv/models/generative/stable_diffusion/__internal__/layers/group_normalization.py](https://localhost:8080/#) in _create_broadcast_shape(self, input_shape)
     85 
     86     def _create_broadcast_shape(self, input_shape):
---> 87      broadcast_shape = [1] * len(input_shape) 
     88         broadcast_shape[self.axis] = input_shape[self.axis] // self.groups

TypeError: Exception encountered when calling layer 'group_normalization_60' (type GroupNormalization).

len is not well defined for a symbolic Tensor (Shape:0). Please call `x.shape` rather than `len(x)` for shape information.

Call arguments received by layer 'group_normalization_60' (type GroupNormalization):
  • args=('tf.Tensor(shape=(None, 64, 64, 320), dtype=float32)',)
  • kwargs=<class 'inspect._empty'>

# What does this PR do?
changes from len(input) to input.shape.rank

Fixes # (issue)
https://github.com/keras-team/keras-cv/issues/1034

## Before submitting
      Pull Request section?
      to it if that's the case.
    The issue discussed with @LukeWood. This is just a minor patch.

### DIFF
--- a/keras_cv/models/stable_diffusion/__internal__/layers/group_normalization.py
+++ b/keras_cv/models/stable_diffusion/__internal__/layers/group_normalization.py
@@ -84,7 +84,7 @@ class GroupNormalization(tf.keras.layers.Layer):
         return gamma, beta
 
     def _create_broadcast_shape(self, input_shape):
-        broadcast_shape = [1] * len(input_shape)
+        broadcast_shape = [1] * input_shape.shape.rank
         broadcast_shape[self.axis] = input_shape[self.axis] // self.groups
         broadcast_shape.insert(self.axis, self.groups)
         return broadcast_shape


bhack:
@LukeWood @ianstenbit  Why this was not already catched by our `jit_compile=True` XLA?

ianstenbit:
Thanks @charbull!
This looks good to me -- it's consistent with how I implemented [GroupNorm in core Keras](https://github.com/keras-team/keras/blob/v2.11.0/keras/layers/normalization/group_normalization.py#L215) (which is now available in 2.11).

I'd like to eventually update our SD implementation to use Keras GroupNorm instead of our own, but that'll add a 2.11 dependency which I don't think we want to do just yet.

I am going to manually test this change against some SD workflows (too large to run on GitHub tests). If they look good, I will merge this.

@bhack - I suspect we didn't encounter this because this seems to be TFLite specific?

ianstenbit (approved):
LGTM -- thank you!

My test workflows all look good with this change, and it seems like a low-risk improvement

PR #1039, opened 2022-11-22
Add testdata file of one Waymo Open Dataset Frame

Guowang:
Testdata for following data transformer, extracted from https://storage.googleapis.com/waymo_open_dataset_v_1_0_0_individual_files/training/segment-10017090168044687777_6380_000_6400_000.tfrecord.


tanzhenyu:
update: reviewers will wait until the transformer is provided

PR #1040, opened 2022-11-22
Add group_points_by_bounding_boxes data augmentation.

lengzq:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


lengzq:
@tanzhenyu

keras_cv/layers/preprocessing3d/crop_objects.py
```diff
@@ -0,0 +1,198 @@
...
+        The first 5 features are [x, y, z, class, range].
+      bounding_boxes: 3D (multi frames) float32 Tensor with shape
+        [num of frames, num of boxes, num of box features].
+        The first 8 features are [x, y, z, dx, dy, dz, phi, box class].
+
+    Output shape:
+      A dictionary of Tensors with the same shape as input Tensors and two additional items for object_point_clouds and OBJECT_BOUNDING_BOXES.
+
+    Arguments:
+      label_index: A int scalar sets the target object index.
```
tanzhenyu, line 45: this still seems to generic, can you elaborate on what "label_index" mean? for example, crop all objects that does not equal to `label_index`

keras_cv/layers/preprocessing3d/crop_objects.py
```diff
@@ -0,0 +1,198 @@
...
+        points_in_bounding_boxes = is_within_box3d(
+            point_clouds, object_bounding_boxes[:, :, :7]
+        )
+        # Filter bounding boxes using the current frame.
+        min_points_filter = (
+            tf.reduce_sum(tf.cast(points_in_bounding_boxes[0], dtype=tf.int32), axis=0)
+            >= self._min_points_per_bounding_boxes
+        )
+
+        object_bounding_boxes = tf.boolean_mask(
```
tanzhenyu, line 99: to me, this is still bounding_boxes, but it's a filtered version of bounding_boxes, no? (I was trying to understand why we have two different formats, one is BOUNDING_BOXES, the other one is OBJECT_BOUNDING_BOXES

keras_cv/layers/preprocessing3d/crop_objects.py
```diff
@@ -0,0 +1,198 @@
...
+BOUNDING_BOXES = base_augmentation_layer_3d.BOUNDING_BOXES
+OBJECT_POINT_CLOUDS = base_augmentation_layer_3d.OBJECT_POINT_CLOUDS
+OBJECT_BOUNDING_BOXES = base_augmentation_layer_3d.OBJECT_BOUNDING_BOXES
+
+
+class CropObjects(base_augmentation_layer_3d.BaseAugmentationLayer3D):
+    """A preprocessing layer which crops objects from point clouds during training.
+
+    This layer will crop points based on object bounding boxes and generates OBJECT_POINT_CLOUDS and OBJECT_BOUNDING_BOXES tensors.
+    During inference time, the output will be identical to input. Call the layer with `training=True` to crop the input points.
```
tanzhenyu, line 31: given that during training time, this changes from point_cloud to object_point_cloud, I assume the layers that follow this layer will need to accept object_point_cloud. In that case, does returning point_cloud during inference time make sense?

tanzhenyu (commented):
Thanks for the PR!

keras_cv/layers/preprocessing3d/crop_objects.py
lengzq: OBJECT_BOUNDING_BOXES is a subset of BOUNDING_BOXES

keras_cv/layers/preprocessing3d/crop_objects.py
lengzq: The following layers will accept both object_point_cloud and the original point_cloud. I think we need to return point_cloud as well

keras_cv/layers/preprocessing3d/crop_objects.py
lengzq: Yes, it is a filtered version of BOUNDING_BOXES

tanzhenyu: so, can we re-use BOUNDING_BOXES?

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
```diff
@@ -0,0 +1,198 @@
...
+    Input shape:
+      point_clouds: 3D (multi frames) float32 Tensor with shape
+        [num of frames, num of points, num of point features].
+        The first 5 features are [x, y, z, class, range].
+      bounding_boxes: 3D (multi frames) float32 Tensor with shape
+        [num of frames, num of boxes, num of box features].
+        The first 8 features are [x, y, z, dx, dy, dz, phi, box class].
+
+    Output shape:
+      A dictionary of Tensors with the same shape as input Tensors and two additional items for OBJECT_POINT_CLOUDS and OBJECT_BOUNDING_BOXES.
```
ianstenbit, line 42: Maybe let's include the shape of OBJECT_POINT_CLOUDS and OBJECT_BOUNDING_BOXES here?

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Arguments:
+      label_index: A int scalar sets the target object index. Bounding boxes and corresponding point clouds with box class == label_index will be saved as OBJECT_BOUNDING_BOXES and OBJECT_POINT_CLOUDS.
```
ianstenbit, line 45: Based on the implementation, it seems like 0 is a special case for this parameter where we include all boxes with non-zero classes, right?

If so, let's document that here.

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes_test.py
```diff
@@ -0,0 +1,116 @@
...
+)
+
+POINT_CLOUDS = base_augmentation_layer_3d.POINT_CLOUDS
+BOUNDING_BOXES = base_augmentation_layer_3d.BOUNDING_BOXES
+OBJECT_POINT_CLOUDS = base_augmentation_layer_3d.OBJECT_POINT_CLOUDS
+OBJECT_BOUNDING_BOXES = base_augmentation_layer_3d.OBJECT_BOUNDING_BOXES
+
+
+class GroupPointsByBoundingBoxesTest(tf.test.TestCase):
+    def test_augment_point_clouds_and_bounding_boxes(self):
```
ianstenbit, line 29: It's probably worth including a test case which demonstrates that:

- calling the layer with training=False is a noop
- the layer preserves the input point clouds and bounding boxes (and potentially any other keys thrown at it)

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
ianstenbit: Also, if we're special casing this, I think it's probably more intuitive to a user to have the default be `None`, and have that reflect all classes (instead of 0, which seems a little bit like it's a real class ID and not a sentinel value)

ianstenbit (changes requested):

ianstenbit:
thanks for the PR!

keras_cv/layers/preprocessing3d/crop_objects.py
lengzq: No augmentation at all during inference.

keras_cv/layers/preprocessing3d/crop_objects.py
lengzq: I think it is better to use different names to distinguish OBJECT_BOUNDING_BOXES from BOUNDING_BOXES. Same for POINT_CLOUDS and OBJECT_POINT_CLOUDS.

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
lengzq: Done.

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
lengzq: Done. I set default label_index = None.

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes_test.py
lengzq: 1. Done.
2. Do you mean the following?
```
        self.assertAllClose(inputs[POINT_CLOUDS], outputs[POINT_CLOUDS])
        self.assertAllClose(inputs[BOUNDING_BOXES], outputs[BOUNDING_BOXES])
```

ianstenbit: Yes I think that would be reasonable (and maybe also passing some other random key as an input to the layer to make sure it gets preserved in the output)

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+        object_point_clouds = tf.gather(
+            object_point_clouds, sort_valid_index, axis=2, batch_dims=2
+        )[:, :, : self._max_points_per_bounding_boxes, :]
+
+        object_point_clouds = tf.where(
+            sort_valid_mask[:, :, :, tf.newaxis] > 0, object_point_clouds, 0.0
+        )
+
+        return (
```
tanzhenyu, line 129: following up on the bb vs object_bb -- the output of this layer, or the input to all following layers, should it be a dict, and should the keys be "point_cloud" and "bounging_boxes"?

lengzq: Yes, I think the output of this layer should be a dictionary of {'point_clouds': tf.Tensor, 'bounding_boxes': tf.Tensor, 'object_point_clouds': tf.Tensor, 'object_bounding_boxes': tf.Tensor}

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
```diff
@@ -0,0 +1,199 @@
...
+        points_in_bounding_boxes = tf.boolean_mask(
+            points_in_bounding_boxes, min_points_filter, axis=2
+        )
+        # [num of frames, num of boxes, num of points].
+        points_in_bounding_boxes = tf.transpose(points_in_bounding_boxes, [0, 2, 1])
+        points_in_bounding_boxes = tf.cast(points_in_bounding_boxes, tf.int32)
+        sort_valid_index = tf.argsort(
+            points_in_bounding_boxes, axis=-1, direction="DESCENDING"
+        )
+        sort_valid_mask = tf.gather(
```
tanzhenyu, line 113: for padding -- shouldn't it pad with all zeros, instead of padding with points that are not within the box?

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
```diff
@@ -0,0 +1,199 @@
...
+        result = {
+            OBJECT_POINT_CLOUDS: object_point_clouds,
+            OBJECT_BOUNDING_BOXES: object_bounding_boxes,
+        }
+
+        # preserve any additional inputs unmodified by this layer.
+        for key in inputs.keys() - result.keys():
+            result[key] = inputs[key]
+        return result
+
```
tanzhenyu, line 161: nit: maybe you can start with results = inputs, and then results.update({OBJECT_POINT_CLOUD, OBJECT_BOUNDING_BOXES})

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
```diff
@@ -0,0 +1,199 @@
...
+    #             )
+    #             return inputs
+    #         else:
+    #             raise ValueError(
+    #                 "Point clouds augmentation layers are expecting inputs point clouds and bounding boxes to "
+    #                 "be rank 3D (Frame, Point, Feature) or 4D (Batch, Frame, Point, Feature) tensors. Got shape: {} and {}".format(
+    #                     point_clouds.shape, bounding_boxes.shape
+    #                 )
+    #             )
+    #     else:
```
tanzhenyu, line 198: does the commented out branch need to be updated for the new output format?

tanzhenyu (approved):

tanzhenyu:
also the formatting test is failing

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
lengzq: Yes, padded object_point_clouds is replaced with zero before return, i.e.,

```
        object_point_clouds = tf.where(
            sort_valid_mask[:, :, :, tf.newaxis] > 0, object_point_clouds, 0.0
        )
```

tanzhenyu (approved):

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
```diff
@@ -0,0 +1,200 @@
...
+        The first 5 features are [x, y, z, class, range].
+      bounding_boxes: 3D (multi frames) float32 Tensor with shape
+        [num of frames, num of boxes, num of box features].
+        The first 8 features are [x, y, z, dx, dy, dz, phi, box class].
+
+    Output shape:
+      A dictionary of Tensors with the same shape as input Tensors and two additional items for OBJECT_POINT_CLOUDS (shape [num of frames, num of valid boxes, max num of points, num of point features]) and OBJECT_BOUNDING_BOXES (shape [num of frames, num of valid boxes, num of box features]).
+
+    Arguments:
+      label_index: An optional int scalar sets the target object index. Bounding boxes and corresponding point clouds with box class == label_index will be saved as OBJECT_BOUNDING_BOXES and OBJECT_POINT_CLOUDS. If label index is None, all valid boundinhg boxes (box class !=0) are used.
```
ianstenbit, line 45: nit: s/boundinhg/bounding

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes_test.py
```diff
@@ -0,0 +1,154 @@
...
+            [[[[0, 1, 2, 3, 4], [0, -1, 2, 3, 4]], [[10, 1, 2, 3, 4], [0, 0, 0, 0, 0]]]]
+            * 2
+        ).astype("float32")
+        object_bounding_boxes = np.array(
+            [[[0, 0, 0, 4, 4, 4, 0, 1], [10, 1, 2, 2, 2, 2, 0, 1]]] * 2
+        ).astype("float32")
+        self.assertAllClose(inputs[POINT_CLOUDS], outputs[POINT_CLOUDS])
+        self.assertAllClose(inputs[BOUNDING_BOXES], outputs[BOUNDING_BOXES])
+        self.assertAllClose(inputs["dummy_item"], outputs["dummy_item"])
+        # Sort the point clouds due to the orders of points are different when using Tensorflow and Metal+Tensorflow (MAC).
```
ianstenbit, line 65: just curious -- is this a known behavior? Should this be an issue we open with TF?

ianstenbit (approved):
Great -- thank you!

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes_test.py
lengzq: Not to my knowledge.
Here is a simple example:
```
 tf.argsort([1, 1, 0], direction="DESCENDING")

TF on mac
 <tf.Tensor: shape=(3,), dtype=int32, numpy=array([1, 0, 2], dtype=int32)>
TF on linux 
<tf.Tensor: shape=(3,), dtype=int32, numpy=array([0, 1, 2], dtype=int32)>
```

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
lengzq: Done.

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
lengzq: Uncommented

keras_cv/layers/preprocessing3d/group_points_by_bounding_boxes.py
lengzq: Done.

PR #1041, opened 2022-11-22
[Object Detection] Remove train_time_metrics from RetinaNet

ianstenbit:
@LukeWood

### DIFF
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -453,6 +453,12 @@ class RetinaNet(ObjectDetectionBaseModel):
         y_pred = self(x, training=False)
         _ = self._backward(y_training_target, y_pred)
 
+        # Early exit for no train time metrics
+        if not self.evaluate_train_time_metrics:
+            # To minimize GPU transfers, we update metrics AFTER we take grads and apply
+            # them.
+            return {m.name: m.result() for m in self.train_metrics}
+
         predictions = self.decode_predictions(y_pred, x)
         self._update_metrics(y_for_metrics, predictions)
         return {m.name: m.result() for m in self.metrics}
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -429,7 +429,15 @@ class RetinaNet(ObjectDetectionBaseModel):
         gradients = tape.gradient(loss, trainable_vars)
         self.optimizer.apply_gradients(zip(gradients, trainable_vars))
 
-        return {m.name: m.result() for m in self.train_metrics}
+        # Early exit for no train time metrics
+        if len(super().metrics) == 0:
+            # To minimize GPU transfers, we update metrics AFTER we take grads and apply
+            # them.
+            return {m.name: m.result() for m in self.train_metrics}
+
+        predictions = self.decode_predictions(y_pred, x)
+        self._update_metrics(y_for_metrics, predictions)
+        return {m.name: m.result() for m in self.metrics}
 
     def test_step(self, data):
         x, y = data
@@ -437,7 +445,15 @@ class RetinaNet(ObjectDetectionBaseModel):
         y_pred = self(x, training=False)
         _ = self._backward(y_training_target, y_pred)
 
-        return {m.name: m.result() for m in self.train_metrics}
+        # Early exit for no train time metrics
+        if len(super().metrics) == 0:
+            # To minimize GPU transfers, we update metrics AFTER we take grads and apply
+            # them.
+            return {m.name: m.result() for m in self.train_metrics}
+
+        predictions = self.decode_predictions(y_pred, x)
+        self._update_metrics(y_for_metrics, predictions)
+        return {m.name: m.result() for m in self.metrics}
 
     def _update_metrics(self, y_true, y_pred):
         y_true = bounding_box.convert_format(
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -94,12 +94,6 @@ class RetinaNet(ObjectDetectionBaseModel):
         box_head: (Optional) A `keras.Layer` that performs regression of
             the bounding boxes.  If not provided, a simple ConvNet with 1 layer will be
             used.
-        evaluate_train_time_metrics: (Optional) whether or not to evaluate metrics
-            passed in `compile()` inside of the `train_step()`.  This is NOT
-            recommended, as it dramatically reduces performance due to the synchronous
-            label decoding and COCO metric evaluation.  For example, on a single GPU on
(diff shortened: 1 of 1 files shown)


tanzhenyu:
Isn't the conclusion to remove this completely?

ianstenbit:
> Isn't the conclusion to remove this completely?

My understanding was that we were going to keep it for now, defaulted to False, because otherwise the train-time keras-cv native COCO metrics aren't available at all. But I may have misunderstood. Happy to remove it if that's what we want to do. @LukeWood @martin-gorner

tanzhenyu:
> > Isn't the conclusion to remove this completely?
> 
> My understanding was that we were going to keep it for now, defaulted to False, because otherwise the train-time keras-cv native COCO metrics aren't available at all. But I may have misunderstood. Happy to remove it if that's what we want to do. @LukeWood @martin-gorner

in the spirit of stable API, I'd rather prefer not having this argument for 0.4, and add this in later releases once we're very confident that this brings train-time eval, otherwise the PyMetrics solution might be the long term solution.

ianstenbit:
> > > Isn't the conclusion to remove this completely?
> > 
> > 
> > My understanding was that we were going to keep it for now, defaulted to False, because otherwise the train-time keras-cv native COCO metrics aren't available at all. But I may have misunderstood. Happy to remove it if that's what we want to do. @LukeWood @martin-gorner
> 
> in the spirit of stable API, I'd rather prefer not having this argument for 0.4, and add this in later releases once we're very confident that this brings train-time eval, otherwise the PyMetrics solution might be the long term solution.

That sgtm. Luke and Martin -- any objections to dropping the train_time_metrics parameter for now?

ianstenbit:
I've updated the PR to drop train_time_metrics from RetinaNet entirely for now

tanzhenyu (approved):
Thanks for the quick update!

LukeWood:
Yeah we can drop this as long as we keep the eval metrics

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -437,25 +429,15 @@ def train_step(self, data):
...
     def test_step(self, data):
         x, y = data
         y_for_metrics, y_training_target = y
         y_pred = self(x, training=False)
         _ = self._backward(y_training_target, y_pred)
 
-        predictions = self.decode_predictions(y_pred, x)
-        self._update_metrics(y_for_metrics, predictions)
-        return {m.name: m.result() for m in self.metrics}
+        return {m.name: m.result() for m in self.train_metrics}
```
LukeWood, line 440: can we keep the eval metrics?  I actually use these a lot.

ianstenbit: This breaks test_step on TPU currently due to the call to decode_predictions.

But maybe I should hold off on this PR anyway, since @tanzhenyu has an idea about how to fix decoding compatibility when training on TPU which may affect this?

LukeWood: > This breaks test_step on TPU currently due to the call to decode_predictions.
> 
> But maybe I should hold off on this PR anyway, since @tanzhenyu has an idea about how to fix decoding compatibility when training on TPU which may affect this?

we could also check if metrics is empty and only decode when metrics are not empty

ianstenbit: > we could also check if metrics is empty and only decode when metrics are not empty

This approach SGTM. I'll update the code.

LukeWood: > Okay I've updated this. I'm just using `len(super().metrics)` as a gating function, but we can also store a flag based on whether `metrics` were passed to `compile` (although this requires that the model be compiled)

I like your approach of using len().  I do the same with regularization losses.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -438,7 +430,7 @@ def train_step(self, data):
         self.optimizer.apply_gradients(zip(gradients, trainable_vars))
 
         # Early exit for no train time metrics
-        if not self.evaluate_train_time_metrics:
+        if len(super().metrics) == 0:
```
martin-gorner, line 433: With this change, RetinaNet fails again on TPU (error: XLA cannot compile NMS). It worked on the previous commit with the evaluate_train_time_metrics=False. I do not have any metrics in .compile().

LukeWood: I think the issue is that super().metrics always has loss in it, so we might have to store a flag in `compile()` indicating if a metric was passed or not.

bhack: > With this change, RetinaNet fails again on TPU (error: XLA cannot compile NMS). It worked on the previous commit with the evaluate_train_time_metrics=False. I do not have any metrics in .compile().

It is always a surprise what things compile and what not in the library. 
I hope that soon or later we could regularly add compiler decorated versions to our tests (eventually with expected failures).

As our NMS layer it is using `tf.image.combined_non_max_suppression` it is supported only enabling the new MLIR XLA bridge:
https://github.com/keras-team/keras-cv/pull/1021#issuecomment-1317925675

As XLA [it is still fail fast](https://github.com/tensorflow/tensorflow/issues/57923) the next issue is:

```python
INVALID_ARGUMENT: TF to XLA legalization ... 'tf.TensorListSetItem' op found tf.TensorListSetItemOp on unknown TensorList.
```

That i suppose it is caused by our `TensorArray` manipulation:
https://github.com/keras-team/keras-cv/blob/51260f0689df27525ee307dd15caca87800f0765/keras_cv/layers/object_detection/non_max_suppression.py#L144-L171


Do you have a more atomic test/input for this?
```python
def _decode_nms_boxes_to_tensor(self, nmsed_boxes):
```

So that we could open a ticket in TF with a minimized but standalone gist.

ianstenbit:
> With this change, RetinaNet fails again on TPU

@martin-gorner my latest commit should fix this. It turns out the `RetinaNetLabelEncoder` adds a metric which was tracked by super.metrics -- this adds some custom logic to check for actual custom train metrics.

LukeWood:
Looks good to me - a much better approach than a constructor argument to control this.  Thanks!

LukeWood (approved):

LukeWood:
didn't realize https://github.com/keras-team/keras-cv/issues/1068 was an issue until updating the RetinaNet tutorial, will get a fix ready

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -437,8 +434,8 @@ def train_step(self, data):
         gradients = tape.gradient(loss, trainable_vars)
         self.optimizer.apply_gradients(zip(gradients, trainable_vars))
 
-        # Early exit for no train time metrics
-        if not self.evaluate_train_time_metrics:
+        # Early exit for no custom metrics
```
LukeWood, line 437: we should still gate this behind a constructor or compile() argument, in the GPU case you only want the COCO metrics on a tiny subset of your eval metrics

PR #1047, opened 2022-11-23
(Ragged OD Pipeline Step 1/3) Support RaggedImages and Bounding Boxes in BaseImageAugmentationLayer

LukeWood:
# What does this PR do?

This layer adds support for ragged images & cleans up ragged bounding box support.

layers.RandomFlip() will now support ragged images.

One API delta is that we always output ragged bounding boxes, but this is our native format so it is reasonable to do so.

---

another tiny delta that got included - adds a batchwise argument to MaybeApply so we can use it with CutMix & MixUp

### DIFF
--- a/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
@@ -224,7 +224,8 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
             return True
         if BOUNDING_BOXES in inputs:
             return True
-        # TODO(lukewood): exhaustively check cases
+        if KEYPOINTS in inputs:
+            return True
         return False
 
     def _map_fn(self, func, inputs):
--- a/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
@@ -466,20 +466,6 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
             inputs = {IMAGES: inputs}
             return inputs, metadata
 
-        valid_keys = [
-            IMAGES,
-            LABELS,
-            TARGETS,
-            BOUNDING_BOXES,
-            KEYPOINTS,
-            SEGMENTATION_MASKS,
-        ]
-        if not all([x in valid_keys for x in inputs.keys()]):
-            raise ValueError(
-                "Expected all keys in `inputs` to be in the list of "
-                f"valid keys, {valid_keys}"
-            )
-
         if not isinstance(inputs, dict):
             raise ValueError(
                 f"Expect the inputs to be image tensor or dict. Got inputs={inputs}"
--- a/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
@@ -12,11 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import functools
-
 import tensorflow as tf
 
-from keras_cv import bounding_box
 from keras_cv.utils import preprocessing
 
 # In order to support both unbatched and batched inputs, the horizontal
--- a/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import functools
+
 import tensorflow as tf
 
 from keras_cv import bounding_box
@@ -27,7 +29,6 @@ LABELS = "labels"
 TARGETS = "targets"
 BOUNDING_BOXES = "bounding_boxes"
 KEYPOINTS = "keypoints"
-RAGGED_BOUNDING_BOXES = "ragged_bounding_boxes"
 SEGMENTATION_MASKS = "segmentation_masks"
 IS_DICT = "is_dict"
 USE_TARGETS = "use_targets"
@@ -108,6 +109,24 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
         super().__init__(seed=seed, **kwargs)
 
     @property
+    def force_output_ragged_images(self):
+        """Control whether to force outputting of ragged images."""
+        return getattr(self, "_force_output_ragged_images", False)
+
+    @force_output_ragged_images.setter
+    def force_output_ragged_images(self, force_output_ragged_images):
+        self._force_output_ragged_images = force_output_ragged_images
+
+    @property
+    def force_output_dense_images(self):
+        """Control whether to force outputting of dense images."""
+        return getattr(self, "_force_output_dense_images", False)
+
+    @force_output_dense_images.setter
+    def force_output_dense_images(self, force_output_dense_images):
+        self._force_output_dense_images = force_output_dense_images
+
+    @property
     def auto_vectorize(self):
         """Control whether automatic vectorization occurs.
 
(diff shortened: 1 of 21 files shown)


bhack:
Can we have only random flip here?

LukeWood:
> Can we have only random flip here?

not really, the updates to base layer are not compatible with old layer versions and its easier to batch update.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -107,6 +105,24 @@ def augment_image(self, image, transformation):
     def __init__(self, seed=None, **kwargs):
         super().__init__(seed=seed, **kwargs)
 
+    @property
+    def force_output_ragged_images(self):
+        """Control whether to force outputting of ragged images."""
+        return getattr(self, "_force_output_ragged_images", False)
```
ianstenbit, line 111: What's the tradeoff between using these settable properties vs a constructor argument?

I feel like `force_ragged=True` or similar could be a nice construction-time control of this

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -129,12 +145,101 @@ def __init__(self):
...
+            ragged_spec = tf.RaggedTensorSpec(
+                shape=images.shape[1:],
+                ragged_rank=1,
+                dtype=self.compute_dtype,
+            )
+            return ragged_spec
+        return tf.TensorSpec(images.shape[1:], self.compute_dtype)
+
+    # TODO(lukewood): promote to user facing API if needed
+    def _compute_bounding_box_signature(self, bounding_boxes):
```
ianstenbit, line 167: Do I understand correctly that this is forcing bounding boxes to be ragged? If so -- is that too coercive of an API?

keras_cv/layers/preprocessing/random_flip.py
```diff
@@ -152,6 +155,9 @@ def augment_bounding_boxes(
                 "`RandomFlip(bounding_box_format='xyxy')`"
             )
 
+        if isinstance(bounding_boxes, tf.RaggedTensor):
```
ianstenbit, line 158: Why does this need to be made dense here?

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -59,6 +59,8 @@ class Resizing(BaseImageAugmentationLayer):
             largest possible resize of the image (of size `(height, width)`) that
             matches the target aspect ratio. By default
             (`pad_to_aspect_ratio=False`), aspect ratio may not be preserved.
+        pad_only: If True, resize the images without aspect ratio distortion only using
```
ianstenbit, line 62: Maybe this API change should be a separate PR?

keras_cv/layers/preprocessing/maybe_apply.py
```diff
@@ -29,6 +29,8 @@ class MaybeApply(BaseImageAugmentationLayer):
         rate: controls the frequency of applying the layer. 1.0 means all elements in
             a batch will be modified. 0.0 means no elements will be modified.
             Defaults to 0.5.
+        batchwise: (Optional) bool, whether or not to pass entire batches to the
```
ianstenbit, line 32: Is the expected behavior here that if batchwise=true, we decide on a per-batch level whether to apply a layer?

If so, might be worth documenting that.

ianstenbit (changes requested):

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
LukeWood: So the reason you don't want to do this as a constructor argument is that there is NO user use case to tweak this argument.

A given layer ALWAYS has this set, or not set.  That's why we made `auto_vectorize` an attribute as well.  Basically the goal is to fully hide this from users.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
LukeWood: Yeah, so right now this is always making bounding boxes Ragged.

I'm on the fence - on one hand we can minimize API drift by requiring bounding boxes to be RaggedTensors.  on the other hand it is a bit more coercive.  

We can also internally always process raggeds, and then handle outputs being converted to Dense tensors in `format_output` and preserve the input format.  The issue with this is that if we just keep converting back and forth we're not netting ANY performance gains; its a performance hit.  

I say we submit this as is, and if users request dense bounding box tensors we implement it then.   Thoughts?

keras_cv/layers/preprocessing/maybe_apply.py
LukeWood: Updated the docstring, updated the behavior, etc

keras_cv/layers/preprocessing/random_flip.py
LukeWood: Getting the shape of a RaggedTensor is inconsistent, and it doesn't work if we don't make the tensor dense here.  The tensor is always dense at this point.

In a previous iteration this was required - but I just tested without it and it works now.  The base layer handles this now.

keras_cv/layers/preprocessing/resizing.py
LukeWood: Removed - yeah good point its probably not the right move

LukeWood:
Ok @ianstenbit all comments addressed or justification provided

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
ianstenbit: I'm really not sure what the right answer here is. Maybe @martin-gorner has an opinion?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -339,16 +466,27 @@ def _format_inputs(self, inputs):
...
 
+        valid_keys = [
+            IMAGES,
+            LABELS,
+            TARGETS,
+            BOUNDING_BOXES,
+            KEYPOINTS,
+            SEGMENTATION_MASKS,
+        ]
+        if not all([x in valid_keys for x in inputs.keys()]):
```
ianstenbit, line 477: I think this is a departure for the current functionality where we just pass-through all the keys we don't know about. What's the value of raising an error here?

Is the idea to alert users of a typo if e.g. they had a key like "buonding_boxes"?

I worry a bit that this is too coercive, but at the same time I can imagine that using a KPL and not having the boxes augmented because of a typo in your dict keys would be very annoying.

keras_cv/layers/preprocessing/ragged_image_test.py
```diff
@@ -0,0 +1,135 @@
...
+        {
+            "height": 224,
+            "width": 224,
+            "zoom_factor": (0.8, 1.0),
+            "aspect_ratio_factor": (3 / 4, 4 / 3),
+        },
+    ),
+]
+
+RAGGED_OUTPUT_TEST_CONFIGURATIONS = []
```
ianstenbit, line 109: Is there a missing test case for these configurations / should this be removed?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
LukeWood: Reverted

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
tanzhenyu: +1 -- how does user know which layers should force ragged output, and which should be false?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
tanzhenyu, line 167: any reason why this is private method, but `compute_image_signature` is public method?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -129,12 +145,101 @@ def __init__(self):
...
+
+        return fn_output_signature
+
+    @staticmethod
+    def _any_ragged(inputs):
+        if isinstance(inputs[IMAGES], tf.RaggedTensor):
+            return True
+        if BOUNDING_BOXES in inputs:
+            return True
+        # TODO(lukewood): exhaustively check cases
```
tanzhenyu, line 227: do we have to check for keypoints as well?

keras_cv/layers/preprocessing/grayscale.py
```diff
@@ -53,6 +53,18 @@ def __init__(self, output_channels=1, **kwargs):
         # This layer may raise an error when running on GPU using auto_vectorize
         self.auto_vectorize = False
 
+    def compute_image_signature(self, images):
```
tanzhenyu, line 56: maybe comment on why overriding `compute_image_signature` is required?

keras_cv/layers/preprocessing/mosaic.py
```diff
@@ -234,10 +238,12 @@ def _update_bounding_box(
             images=images,
             dtype=self.compute_dtype,
         )
-
         boxes_for_mosaic = tf.gather(bounding_boxes, permutation_order[index])
+        if isinstance(boxes_for_mosaic, tf.RaggedTensor):
+            boxes_for_mosaic = boxes_for_mosaic.to_tensor(-1, shape=[None, 5])
+        # boxes_for_mosaic = boxes_for_mosaic.to_tensor(-1)
```
tanzhenyu, line 244: remove if not needed?

keras_cv/layers/preprocessing/mosaic_test.py
```diff
@@ -80,7 +80,7 @@ def test_return_shapes(self):
 
         self.assertEqual(xs.shape, [2, 512, 512, 3])
         self.assertEqual(ys_labels.shape, [2, 10])
-        self.assertEqual(ys_bounding_boxes.shape, [2, 12, 5])
+        self.assertEqual(ys_bounding_boxes.shape, [2, None, 5])
```
tanzhenyu, line 83: any reason this is changed to None?

tanzhenyu (changes requested):
Thanks for the PR!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
LukeWood: Right now I see no use case to customize the bounding box output signature

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
LukeWood: Done

keras_cv/layers/preprocessing/mosaic_test.py
LukeWood: It outputs a ragged instead of a padded dense now after the API change - this is an expected change.

LukeWood: (we could call to_tensor() in the test if we want)

LukeWood:
Final round of comments addressed, @ianstenbit and @tanzhenyu PTAL and LGTM if it is good after the previous batch of comments.

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
LukeWood: Lets add an API for this if users ask for it.  We'll be processing Raggeds internally no matter what.

ianstenbit (approved):
Thanks Luke!

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -182,6 +194,7 @@ def resize_single_with_pad_to_aspect(x):
             inputs["images"] = image
 
             if boxes is not None:
+                boxes = keras_cv.bounding_box.filter_sentinels(boxes)
```
tanzhenyu, line 197: unrelated to this PR: filter_empty_boxes?

keras_cv/layers/preprocessing/randomly_zoomed_crop.py
```diff
@@ -79,7 +79,7 @@ def __init__(
         )
 
         self._check_class_arguments(height, width, zoom_factor, aspect_ratio_factor)
-
+        self.force_output_dense_images = True
```
tanzhenyu, line 82: this seems to be introducing more user cognitive load, is there any other way to reduce it?

PR #1054, opened 2022-11-28
(OD input pipeline Step 3/4) Add RandomAspectRatio layer

LukeWood:
This PR adds a RandomAspectRatio layer that takes a Dense image tensor and always outputs a ragged tensor of aspect-ratio distorted images.

JitteredResize does not tweak aspect ratios, so we can use this layer to fill all aspect ratio distortion based needs across tasks.


LukeWood:
ok @ianstenbit @tanzhenyu @qlzh727 I believe this is prepared!

keras_cv/layers/preprocessing/random_aspect_ratio.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def compute_image_signature(self, images):
+        return tf.RaggedTensorSpec(
+            shape=(None, None, images.shape[-1]),
+            ragged_rank=1,
+            dtype=self.compute_dtype,
+        )
+
+    def augment_bounding_boxes(self, bounding_boxes, transformation, image, **kwargs):
+        if self.bounding_box_format is None:
```
tanzhenyu, line 67: Can this be raised during init?

tanzhenyu (commented):
Thanks for the PR!
Implementation wise looks good to me.
Can you add a description on why we need this? (AFAIK, the OD models are expected to train well on JitteredResize layer?)

tanzhenyu (approved):

PR #1059, opened 2022-11-28
[Semantic Segmentation] Add first segmentation weights (credit @tanzhenyu)

ianstenbit:
This also updates our training history updater to handle segmentation logs, and updates the deeplab implementation to load weights (also updates docstrings for deeplab)


DavidLandup0:
Do you think this might be slightly improved with #968?
A 30-epoch run did 58% in validation mean IoU. I'm doing a 100-epoch test right now with dice to see if we can hopefully hit 70%, based on @tanzhenyu's DeepLabV3 training script 🤔

keras_cv/models/segmentation/deeplab.py
```diff
@@ -56,13 +64,19 @@ def __init__(
         classes,
         include_rescaling,
         backbone,
-        weights,
+        backbone_weights=None,
+        weights=None,
```
tanzhenyu, line 68: do we want to provide default voc weights?

keras_cv/models/segmentation/deeplab.py
```diff
@@ -136,6 +150,15 @@ def __init__(
...
         self.segmentation_head = segmentation_head
 
+        weights = parse_weights(weights, False, "deeplabv3")
+        if weights and not tf.io.gfile.exists(weights):
+            raise ValueError(
+                "The `weights` argument should be either `None`, the path to the "
+                "weights file to be loaded, or a reference to pre-trained weights. "
+                f"Weights file not found at location: {weights}"
+            )
+        self.load_weights(weights)
```
tanzhenyu, line 160: interesting...does this work? Today all our models are subclassed models, meaning the variables are not created at this point, I wonder if `load_weights` will error out

shell/weights/update_training_history.py
```diff
@@ -75,13 +75,33 @@
...
 training_epochs = max(tensorboard_results[tensorboard_results.run == "train"].step)
-max_validation_accuracy = max(
-    tensorboard_results[
-        (tensorboard_results.run == "validation")
-        & (tensorboard_results.tag == "epoch_categorical_accuracy")
-    ].value
-)
-max_validation_accuracy = f"{max_validation_accuracy:.4f}"
+
+results_tags = tensorboard_results.tag.unique()
```
tanzhenyu, line 79: maybe comment the code?

tanzhenyu (commented):
Thanks for the PR!

keras_cv/models/segmentation/deeplab.py
ianstenbit: Ahh yes I forgot that we have to call `call` on a subclass model before this will load.

So perhaps we want to actually override `load_weights` for these models to support our custom weights?
For now, I've removed the `weights` param from this PR, as we should probably discuss how we'll offer pre-trained non-backbone models.

IIRC KerasNLP has come up with a `from_pretrained` API that we might want to consider?

shell/weights/update_training_history.py
ianstenbit: Good call -- added some comments about the val accuracy and mean iou

ianstenbit:
> Do you think this might be slightly improved with https://github.com/keras-team/keras-cv/pull/968?

I think that it's likely there is still lots of opportunity to improve our semantic segmentation training script. Right now, I'm focused on getting the infrastructure set up to offer segmentation weights, using Zhenyu's VOC weights as a starting point.

keras_cv/models/segmentation/deeplab.py
tanzhenyu: the `from_pretrained` doesn't help much on this case, because KerasNLP actually creates all the variables during init. I have an action item to make our models to be function subclass models though, after which adding `weights` would be easier.

keras_cv/models/segmentation/deeplab.py
tanzhenyu: I mean default to `voc`, given keras.application models default to `imagenet` - https://www.tensorflow.org/api_docs/python/tf/keras/applications/resnet50/ResNet50

tanzhenyu (changes requested):
see comments inline.

keras_cv/models/segmentation/deeplab.py
ianstenbit: sg -- for now would you like me to override `load_weights`?

keras_cv/models/segmentation/deeplab.py
ianstenbit: Gotcha. Currently, our backbones don't have default weights. I don't personally love the idea of default weights, but I haven't thought about it a lot. Maybe we should discuss this as a package-wide option in our KerasCV sync?

Regardless of whether we include default weights, I would think we should keep them consistent across our model offerings.

ianstenbit:
> see comments inline.

thanks!

keras_cv/models/segmentation/deeplab.py
tanzhenyu: it's ok not to override, I will make this complete before 0.4 anyway

keras_cv/models/segmentation/deeplab.py
tanzhenyu: sounds good to me -- we could make this change later

tanzhenyu (approved):
Thanks for the quick response!

PR #1060, opened 2022-11-29
[3D Object Detection] Make iou_3d a function

ianstenbit:


keras_cv/ops/iou_3d.py
```diff
@@ -17,32 +17,28 @@
...
-        )
-        self.iou_3d = pairwise_iou_op.pairwise_iou3d
-
-    def __call__(self, y_true, y_pred):
-        return self.iou_3d(y_true, y_pred)
+iou_3d = None
+try:
+    keras_cv_custom_ops = load_library.load_op_library(
+        resource_loader.get_path_to_datafile("../custom_ops/_keras_cv_custom_ops.so")
+    )
```
tanzhenyu, line 24: @bhack I believe you mentioned something about lazy loading?

ianstenbit: Thank you both! Switched to lazy loading, using the same LazySO strategy as TFA

tanzhenyu (approved):
Thank you!

PR #1063, opened 2022-11-30
Add pre-trained weights for ViTs

DavidLandup0:
# What does this PR do?

Converted to draft until #1014 is merged.

Since the last PR (#1014) is already getting large, and the description is verbose - separating the weights into a new PR.
Cloned from the ViT branch (once that's merged, this PR will be rebased and most of the changes in the files will disappear as they're already included in the previous PR).

Here are the official JAX weights, ported for KCV ViTs, as H5 files:

- Tiny16: https://drive.google.com/file/d/1fEo0l_zwibl3eJ6-xb9MTvHgIo8XySxY
- S16: https://drive.google.com/file/d/1R1tgxJScy0LK3HLGNYWwLG_UQp10ope2
- B16: https://drive.google.com/file/d/1M_zALLuEuoZ1fZNpWPqlD2Ahc9UPB9G4
- L16: https://drive.google.com/file/d/1QcPYoDZMbIaT-EdW6CjjqiwDaMbML9Zu
- S32: https://drive.google.com/file/d/1p7DCy3eukJFuzqbgdXWILNhJUx2e8Kzp
- B32: https://drive.google.com/file/d/1M_zALLuEuoZ1fZNpWPqlD2Ahc9UPB9G4

Updated from GCS to Google Drive to avoid bucket cost accumulation.

Inference on the ImageNetV2 validation set (ms/sep, top-1 and top-5 ACC):

Tiny16: 87ms/step - loss: 1.6895 - accuracy: 0.6067 - sparse_top_k_categorical_accuracy: 0.8285
S16:      178ms/step - loss: 1.2600 - accuracy: 0.6884 - sparse_top_k_categorical_accuracy: 0.8916
B16:      464ms/step - loss: 1.0631 - accuracy: 0.7293 - sparse_top_k_categorical_accuracy: 0.9199
L16:      1s/step - loss: 0.9622 - accuracy: 0.7545 - sparse_top_k_categorical_accuracy: 0.9321
S32:      54ms/step - loss: 1.6650 - accuracy: 0.6090 - sparse_top_k_categorical_accuracy: 0.8335
B32:      122ms/step - loss: 1.3684 - accuracy: 0.6650 - sparse_top_k_categorical_accuracy: 0.8781

The PR also updates `vit.py` to `parse_weights()` using their aliases - though, there are no hashes yet. Can you help with uploading them @ianstenbit?

### DIFF
--- a/examples/layers/preprocessing/classification/random_crop_demo.py
+++ b/examples/layers/preprocessing/classification/random_crop_demo.py
@@ -1,0 +1,34 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""random_crop_demo.py.py shows how to use the RandomCrop
+preprocessing layer. Operates on an image of elephant. In this script the image
+is loaded, then are passed through the preprocessing layers.
+Finally, they are shown using matplotlib.
+"""
+
+import demo_utils
+
+from keras_cv.layers.preprocessing import RandomCrop
+
+
+def main():
+    many_elephants = demo_utils.load_elephant_tensor(output_size=(300, 300))
+    layer = RandomCrop(100, 200)
+    augmented = layer(many_elephants)
+    demo_utils.gallery_show(augmented.numpy())
+
+
+if __name__ == "__main__":
+    main()
--- a/keras_cv/layers/__init__.py
+++ b/keras_cv/layers/__init__.py
@@ -14,7 +14,6 @@
 
 from tensorflow.keras.layers import CenterCrop
 from tensorflow.keras.layers import RandomBrightness
-from tensorflow.keras.layers import RandomCrop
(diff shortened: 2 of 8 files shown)


ianstenbit:
Thanks @DavidLandup0!

So far, all of our KerasCV weights are trained "in-house" (by our training scripts), and if possible we'd like to keep it that way (although I understand this may turn out to be very challenging for ViTs).

I'm going to sync with the team on this and get back to you -- it may be that offering the official weights for these models in particular is a good approach.

DavidLandup0:
I agree that we want to prefer reproducible weights with scripts hosted here. With a bit of back and forth in the comments of the last PR, we figured that porting them would probably be okay for these, at least for the larger ones 600M+ params).

I'm still working on the training script for ViTs so we have both a script for training *and* the free converted weights (eco friendly, yay!), but they're tricky to figure out as the divergence happens a couple of hours in, so every test run takes hours of good-looking curves followed by plateauing :(

As soon as I can compile a good script for training these, I'll submit it here as well :)

DavidLandup0:
P.S. These weights are pre-trained on ImageNet21k and fine-tuned on 1k, while our training script operates on 1k only currently.

DavidLandup0:
Agreed with @bhack. ViTs are tricky to train without a lot of data, so in their plain form, they're the most useful when using pre-trained official weights. 

There are few runs online to compare to (expensive + subpar accuracy without ImageNet21k pretraining), but there are pre-trained runs that we can compare to, on much more manageable datasets like CIFAR10 and CIFAR100 (happy to run a bunch of these on my local machine). Since users will be using them with pre-trained weights (otherwise, they can achieve better results with fewer flops using a CNN like EfficientNet), I think we should first validate the fine-tuning of this network with official weights, and then work on reproducing them using an in-house script later when that's more feasible.

Thoughts?

PR #1066, opened 2022-12-01
Add data transformer for Waymo Open Dataset

Guowang:


Guowang:
@ianstenbit @lengzq @tanzhenyu

keras_cv/datasets/waymo/transformer.py
```diff
@@ -0,0 +1,632 @@
...
+        label_box_meta=box_meta,
+        label_box_class=box_class,
+        label_box_density=box_density,
+        label_box_detection_difficulty=box_detection_difficulty,
+        label_box_mask=box_mask,
+        label_point_class=point_box_class,
+    )
+
+
+def point_vehicle_to_global(
```
tanzhenyu, line 484: this should be made private?

keras_cv/datasets/waymo/transformer.py
```diff
@@ -0,0 +1,632 @@
...
+    """
+    rot = sdc_pose[..., 0:3, 0:3]
+    loc = sdc_pose[..., 0:3, 3]
+    return (
+        tf.linalg.matmul(point_vehicle_xyz, rot, transpose_b=True)
+        + loc[..., tf.newaxis, :]
+    )
+
+
+def box_3d_vehicle_to_global(box_3d: tf.Tensor, sdc_pose: tf.Tensor) -> tf.Tensor:
```
tanzhenyu, line 504: same, private?

keras_cv/datasets/waymo/transformer.py
```diff
@@ -0,0 +1,632 @@
...
+        "label_box_class": point_label_tensors.label_box_class,
+        "label_box_density": point_label_tensors.label_box_density,
+        "label_box_detection_difficulty": point_label_tensors.label_box_detection_difficulty,
+        "label_box_mask": point_label_tensors.label_box_mask,
+        "label_point_class": point_label_tensors.label_point_class,
+        "label_point_nlz": point_tensors.label_point_nlz,
+    }
+
+
+def build_tensors_for_augmentation(
```
tanzhenyu, line 580: some offline discussion -- docstring that this is to convert range image to point cloud, and point users for a way to use range image directly?
(Do we need a parse range image function in the future?)

tanzhenyu (approved):
LGTM, with some minor comments

keras_cv/datasets/waymo/transformer.py
Guowang: Done

keras_cv/datasets/waymo/transformer.py
Guowang: Done

keras_cv/datasets/waymo/transformer.py
Guowang: Done, added docstring to build_tensors_from_wod_frame. The range image parsing function is in frame_utils from waymo_open_dataset.utils.

keras_cv/datasets/waymo/load.py
```diff
@@ -16,24 +16,29 @@
...
             frame.ParseFromString(record)
             yield transformer(frame)
 
     return _generator
 
 
-def load(tfrecord_path, transformer, output_signature):
+def load(
+    tfrecord_path,
+    transformer=transformer.build_tensors_from_wod_frame,
```
ianstenbit, line 36: Let's also update the docstring to reflect this default transformer and output signature
Other than that LGTM.

Thank you!

ianstenbit (approved):

keras_cv/datasets/waymo/load.py
Guowang: Done, thanks Ian!

keras_cv/datasets/waymo/transformer.py
```diff
@@ -0,0 +1,635 @@
...
+"""Transformer to convert Waymo Open Dataset proto to model inputs."""
+
+from typing import Dict
+from typing import List
+from typing import Sequence
+from typing import Tuple
+
+import numpy as np
+import tensorflow as tf
+from waymo_open_dataset import dataset_pb2
```
ianstenbit, line 23: If we're going to make WOD an optional dependency, we need to wrap our imports in a try/catch to allow importing the package without throwing an error due to missing deps.

(Here and in load.py)

ianstenbit (changes requested):

keras_cv/datasets/waymo/transformer.py
Guowang: Added try/catch to __init__.py.

ianstenbit (approved):

PR #1070, opened 2022-12-01
Switch to one-configuration-per-architecture GCB testing for backbones

ianstenbit:
This dramatically cuts down on model testing time while still keeping good coverage of backbones. WDYT?


tanzhenyu (commented):
Would this mean that we have to manually test those commented models if relevant changes are made?

ianstenbit:
> Would this mean that we have to manually test those commented models if relevant changes are made?

More or less, yes. This would mean that when a backbone is updated we need to manually run its tests if we want to check every configuration of the model. Given the infrequency with which we update backbones, I think this is reasonable. 

The backbones will _still be tested_ by CI, just not _every_ configuration of every backbone, which is arguably overkill even when changes are made to the implementation.

tanzhenyu (approved):
LGTM! Thanks for fixing it temporarily!

PR #1071, opened 2022-12-01
Re-introduce evaluate_train_step_metrics API

LukeWood:
When I train a RetinaNet in Colab I like to evaluate metrics on a tiny eval set:

```
eval_ds.take(20)
```

but not over the training set.  Evaluating over the training set causes a 10x slowdown (1hr epoch time from 3 mins).  This actually gives a strong approximation (usually within 0.05~ of the true value for recall and MaP), but is very cheap to do.  

Lets introduce this under compile() so users can set it multiple times, and default it to false so users don't have to tinker with it.

### DIFF
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -277,7 +277,13 @@ class RetinaNet(ObjectDetectionBaseModel):
         )
 
     def compile(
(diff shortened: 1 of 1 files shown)


LukeWood:
https://colab.research.google.com/drive/1FXF4kT6WNymY5IvBBhkamdFF5NG5C0K3?usp=sharing

bhack:
It is private

divyashreepathihalli (approved):

PR #1076, opened 2022-12-01
Fix patch that somehow got dropped

LukeWood:
Required to work with MixUp(), Mosaic(), and other ragged output layers.


ianstenbit:
Should this have been caught by a test?

(Also -- looks like you need to run the linter)

LukeWood:
> Should this have been caught by a test?
> 
> (Also -- looks like you need to run the linter)

Honestly its a little hard to repro in a test case, I'd prefer to just fix as a workaround for now - eventually I'd like Mosaic and MixUp to just preserve the last dimension but for now this is the simplest workaround.

ianstenbit (approved):

PR #1079, opened 2022-12-02
[Object Detection] Fix SmoothL1 Loss Calculation

### LINKED ISSUE #920: smooth l1 loss does not match formula
Here's our implementation:
https://github.com/keras-team/keras-cv/blame/master/keras_cv/losses/smooth_l1.py#L39-L42

Here's a definition used by PyTorch:
https://pytorch.org/docs/stable/generated/torch.nn.SmoothL1Loss.html

A beta denominator is missing. It still works fine if the cutoff is 1, but anything else would be different.
We should also add more unit test to it while fixing it

DavidLandup0:
# What does this PR do?

- Updates the SmoothF1 Loss to correctly account for the `beta`.
- Renamed `f1_cutoff` to `beta` for consistency with paper terminology and refactored `f1_cutoff` usage.
- Added numerical tests for the `beta` argument.

Fixes #920

## Who can review?

Tagging original authors (@LukeWood @fchollet) and @tanzhenyu for opening the issue.

### DIFF
--- a/keras_cv/losses/smooth_l1.py
+++ b/keras_cv/losses/smooth_l1.py
@@ -23,12 +23,15 @@ class SmoothL1Loss(tf.keras.losses.Loss):
     contribute to the overall loss based on their squared difference, and values greater
     than `beta` contribute based on their raw difference.
 
-    As `beta` approaches 1 - the loss approaches L2 loss.
-    As `beta` approaches 0 - the loss approaches L1 loss.
+    The function is quadratic for small values, and linear for large values,
+    balancing the positives of L1 and L2 losses.
+
+    L2 loss is used on targets between [0...beta] (near zero).
+    L1 loss is used on targets between [beta...+inf] (large values).
 
     Args:
-        beta: differences between y_true and y_pred that are larger than `beta` are
-            treated as `L1` values.
+        beta: default 1.0, the threshold at which the loss behaves as L1 or L2, in
+        the range of [0..1].
     """
 
     def __init__(self, beta=1.0, **kwargs):
--- a/keras_cv/losses/smooth_l1.py
+++ b/keras_cv/losses/smooth_l1.py
@@ -19,33 +19,33 @@ import tensorflow as tf
 class SmoothL1Loss(tf.keras.losses.Loss):
     """Implements Smooth L1 loss.
 
-    SmoothL1Loss implements the SmoothL1 function, where values less than `l1_cutoff`
+    SmoothL1Loss implements the SmoothL1 function, where values less than `beta`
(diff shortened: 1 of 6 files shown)


LukeWood:
Hey @DavidLandup0 Thanks for the PR!  Whats the motivation for the change?  

We picked the name `l1_cutoff` as it is more clear from a readability perspective than `beta`.  many readers can figure out what an l1 cutoff is without looking at the formula, beta you need to look at the formula.

keras_cv/losses/smooth_l1.py
```diff
@@ -19,33 +19,33 @@
...
     def call(self, y_true, y_pred):
         difference = y_true - y_pred
         absolute_difference = tf.abs(difference)
         squared_difference = difference**2
         loss = tf.where(
-            absolute_difference < self.l1_cutoff,
-            0.5 * squared_difference,
-            absolute_difference - 0.5,
+            absolute_difference < self.beta,
+            (0.5 * squared_difference) / self.beta,
```
LukeWood, line 41: ah I see that this actually changes the numerics

DavidLandup0: Yup - the original name was a clear use of the argument, since it was just a cutoff. It works perfectly the same as the change when the cutoff is 1.0 (default), but as soon as it changes, the loss was technically incorrect.
We can still keep the cutoff name for backwards compatability if you think that's a good idea, but I think we'd benefit more from having it change early on to a slightly more accurate term?

DavidLandup0: Reference formulas:

- https://pytorch.org/docs/stable/generated/torch.nn.SmoothL1Loss.html
- https://paperswithcode.com/method/self-adjusting-smooth-l1-loss

DavidLandup0: We should actually lose the parentheses in the second statement. Updating in a moment

DavidLandup0: Updated and updated docs to make the effect of `beta` clearer!

DavidLandup0: Pinging since I think that the fix for the loss func is relevant for the 0.4 release @LukeWood 
Not blocking by any means, but worth having if we're releasing OD/SemSeg

DavidLandup0:
Pinging again since this is a small change that might impact some of the workflows currently underway
@tanzhenyu

LukeWood:
Lets keep the naming convention.  In keras we prefer clarity over matching publications.  This is done as the ML landscape is ever-changing, and its impossible to keep track of every specific naming convention.   Better to give an indicative name, `cutoff` than `beta`

PR #1080, opened 2022-12-02
Faster RCNN: Replace keras.applications ResNet50 with KCV ResNet50

DavidLandup0:
# What does this PR do?

There's still a few TODOs in the Faster RCNN model, but this seemed like a quick/simple update, since `keras.applications` probably shouldn't be used in lieu of KCV's models.

## Who can review?

Tagging author: @tanzhenyu


DavidLandup0:
Edit: realized that all of the layer names are dependent on the model, so changing won't be as simple as plugging one out and the other one in. Delaying this for later.

PR #1086, opened 2022-12-04
fix label smoothing in focal loss

### LINKED ISSUE #672: Possibly wrong implementation of focal loss in regard of label smoothing

DavidLandup0:
# What does this PR do?

Introduced in #672, PR'd in #680 and re-opened due to seemingly going cold.
If @atuleu still wants to do this PR, feel free to ping here. Otherwise, this is a bug that should be fixed as the object detection API is being built.

Added numerical tests to verify that the loss is indeed `0.0` on a fully correct prediction, with label smoothing on, which was off in the original implementation.

Fixes #672

## Who can review?
@atuleu @LukeWood @tanzhenyu

### DIFF
--- a/keras_cv/losses/focal.py
+++ b/keras_cv/losses/focal.py
@@ -75,16 +75,17 @@ class FocalLoss(tf.keras.losses.Loss):
         y_pred = tf.convert_to_tensor(y_pred)
         y_true = tf.cast(y_true, y_pred.dtype)
 
-        if self.label_smoothing:
-            y_true = self._smooth_labels(y_true)
-
         if self.from_logits:
             y_pred = tf.nn.sigmoid(y_pred)
 
-        cross_entropy = K.binary_crossentropy(y_true, y_pred)
-
         alpha = tf.where(tf.equal(y_true, 1.0), self._alpha, (1.0 - self._alpha))
         pt = y_true * y_pred + (1.0 - y_true) * (1.0 - y_pred)
+
+        if self.label_smoothing:
+            y_true = self._smooth_labels(y_true)
+
+        cross_entropy = K.binary_crossentropy(y_true, y_pred)
+
         loss = alpha * tf.pow(1.0 - pt, self._gamma) * cross_entropy
         # In most losses you mean over the final axis to achieve a scalar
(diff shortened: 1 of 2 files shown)


LukeWood (approved):

ianstenbit:
@DavidLandup0 lmk if you'd like to rebase and merge this, otherwise I'll close it.

Thanks for the PRs, and sorry that many of these got buried.

ianstenbit:
Oops -- didn't mean to close it just yet

DavidLandup0:
Yep, I'll rebase this :)
Thanks for taking another look! It changes the numerics of the loss function at `beta` != 1. Should we use `1` as the default to ensure backward compatability, unless we already do?

ianstenbit:
> Yep, I'll rebase this :) Thanks for taking another look! It changes the numerics of the loss function at `beta` != 1. Should we use `1` as the default to ensure backward compatability, unless we already do?

I'm not super familiar with this loss, so I'm not sure.

It seems like the only delta here is that we do the label smoothing after computing `pt` instead of before -- is there some other nuance I'm missing?

I don't think there is wide usage of label smoothing with this, so if the old version is not quite numerically correct I'm not concerned with making that reproducible, so backwards compat isn't a huge issue unless it's an API breakage.

PR #1089, opened 2022-12-06
transform records.

tanzhenyu:
This PR transform WOD into dict of Tensors first, and load directly.

### DIFF
--- a/examples/training/object_detection_3d/waymo/serialize_records.py
+++ b/examples/training/object_detection_3d/waymo/serialize_records.py
@@ -1,0 +1,95 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+import time
+
+import tensorflow as tf
+import tensorflow_datasets as tfds
+from waymo_open_dataset import dataset_pb2
+
+from keras_cv.datasets.waymo import build_tensors_for_augmentation
+from keras_cv.datasets.waymo import load
+from keras_cv.datasets.waymo import transformer
+
+TRAINING_RECORD_PATH = "/home/overflow/code/wod_records"  # "gs://waymo_open_dataset_v_1_0_0_individual_files/training"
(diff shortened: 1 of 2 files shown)


examples/training/object_detection_3d/waymo/serialize_records.py
```diff
@@ -0,0 +1,95 @@
...
+    return example_proto.SerializeToString()
+
+
+# Load the training dataset
+filenames = tf.data.TFRecordDataset.list_files(
+    os.path.join(TRAINING_RECORD_PATH, "*.tfrecord")
+)
+
+
+def convert_frame_generator(inputs):
```
ianstenbit, line 57: Looks like this is unused

examples/training/object_detection_3d/waymo/train_pillars.py
```diff
@@ -11,29 +11,70 @@
...
 
+global_batch = 1
+
+features_dict = {
+    "point_clouds": tf.io.RaggedFeature(dtype=tf.float32),
+    "bounding_boxes": tf.io.RaggedFeature(dtype=tf.float32),
+}
+
+
+def build_tensors(x):
```
ianstenbit, line 33: (not for this PR, just a thought): it might be nice to eventually bundle this in with the loader?

The loader could have two public methods: one for extracting point clouds / bboxes and serializing them into the denser tfrecords that we'll use for training, and the other for loading the dense records

ianstenbit (changes requested):
Thanks! I really like this approach.

I'm curious -- what is the ratio of file size for the serialized records vs the raw frame records?

lengzq (approved):

ianstenbit:
Examples-only PR -- we can skip our GCBRun

Guowang (approved):

tanzhenyu:
> Examples-only PR -- we can skip our GCBRun

Is this ready to merge?

ianstenbit:
> > Examples-only PR -- we can skip our GCBRun
> 
> Is this ready to merge?

Yes, lg on my end

tanzhenyu:
> > > Examples-only PR -- we can skip our GCBRun
> > 
> > 
> > Is this ready to merge?
> 
> Yes, lg on my end

What was the issue with serialization slow down?

ianstenbit:
> What was the issue with serialization slow down?

I wasn't able to figure out what the issue is. I'm seeing 100% cpu utilization (single core) on my GCP VM when trying to run this, but that's true both with and without the change to use `.load`.

I will try some more performance testing on a Colab instance Monday, but for now I think as long as the performance issue is isolated to my machine it's not an issue with the transormer code.

ianstenbit (approved):

tanzhenyu:
> > What was the issue with serialization slow down?
> 
> I wasn't able to figure out what the issue is. I'm seeing 100% cpu utilization (single core) on my GCP VM when trying to run this, but that's true both with and without the change to use `.load`.
> 
> I will try some more performance testing on a Colab instance Monday, but for now I think as long as the performance issue is isolated to my machine it's not an issue with the transormer code.

I tested it again with the latest commit, converting each file (~1GB) takes around 100 seconds.

tanzhenyu:
skipping test since this only touches example scripts

PR #1092, opened 2022-12-07
[Mixed Precision Support] MixUp Layer

DavidLandup0:
# What does this PR do?

A few weeks ago, most of the issues with missing support for mixed precision were fixed, but MixUp seems to have been overlooked.

```
tf.keras.mixed_precision.set_global_policy('mixed_float16')

(images, labels), _ = tf.keras.datasets.cifar10.load_data()
images = images[:10]
labels = labels[:10]
labels = tf.cast(tf.one_hot(labels, 10), tf.float32)

mixup = keras_cv.layers.preprocessing.MixUp(10)
mixup({'images': images, 'labels': labels})
```

```
...
InvalidArgumentError: Exception encountered when calling layer "mix_up" "                 f"(type MixUp).

cannot compute Mul as input #1(zero-based) was expected to be a float tensor but is a half tensor [Op:Mul]

Call arguments received by layer "mix_up" "                 f"(type MixUp):
  • inputs={'images': 'tf.Tensor(shape=(10, 32, 32, 3), dtype=float16)', 'labels': 'tf.Tensor(shape=(10, 1, 10), dtype=float16)'}
  • training=True
```

This PR also updates the example to cast the input labels into floats.
Should we add dtype support tests for all preprocessing layers?

@LukeWood @tanzhenyu

### DIFF
--- a/keras_cv/layers/preprocessing/mix_up.py
+++ b/keras_cv/layers/preprocessing/mix_up.py
@@ -36,6 +36,8 @@ class MixUp(BaseImageAugmentationLayer):
     Sample usage:
     ```python
     (images, labels), _ = tf.keras.datasets.cifar10.load_data()
+    # Labels must be floating-point and one-hot encoded
+    labels = tf.cast(tf.one_hot(labels, 10), tf.float32)
     mixup = keras_cv.layers.preprocessing.MixUp(10)
     augmented_images, updated_labels = mixup({'images': images, 'labels': labels})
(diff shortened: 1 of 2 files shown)


DavidLandup0:
Should we add dtype support tests for all preprocessing layers?

bhack:
> Should we add dtype support tests for all preprocessing layers?

👍

LukeWood:
can with add an entry to `with_mixed_precision_test.py`?

DavidLandup0:
Done

PR #1095, opened 2022-12-07
(Step 4/5) Adds a random ragged crop augmentation layer

LukeWood:
This Pr adds a layer that randomly crops ragged portions of images from a batch of dense images.  This is done by sampling a height and width % from the original, and cropping out that portion of the image from the parent image randomly.


LukeWood:
JitteredResize now fills this need.

Issue #1075, opened 2022-12-01
RetinaNet doesn't train on TPU Strategy


tanzhenyu:
/keras-cv/keras_cv/models/object_detection/object_detection_base_model.py", line 68, in fit
    return super().fit(x=dataset, validation_data=validation_data, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/keras/utils/traceback_utils.py", line 67, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/usr/local/lib/python3.8/dist-packages/tensorflow/python/distribute/tpu_strategy.py", line 982, in _check_spec
    raise ValueError(
ValueError: Found tensor 1/0 with spec <class 'tensorflow.python.ops.ragged.ragged_tensor.RaggedTensorSpec'>. TPUStrategy does not support distributed datasets with device prefetch when using sparse or ragged tensors. If you intend to use sparse or ragged tensors, please pass a tf.distribute.InputOptions object with experimental_fetch_to_device set to False to your dataset distribution function.

PR #1113, opened 2022-12-08
Fix RetinaNet training on TPU.

tanzhenyu:
This gives mAP ~34.0

Fixes https://github.com/keras-team/keras-cv/issues/1075

### DIFF
--- a/examples/training/object_detection/pascal_voc/retina_net.py
+++ b/examples/training/object_detection/pascal_voc/retina_net.py
@@ -327,7 +327,7 @@ eval_ds = eval_ds.map(
     num_parallel_calls=tf.data.AUTOTUNE,
 )
 eval_ds = eval_ds.apply(
-    tf.data.experimental.dense_to_ragged_batch(BATCH_SIZE, drop_remainder=True)
+    tf.data.experimental.dense_to_ragged_batch(GLOBAL_BATCH_SIZE, drop_remainder=True)
 )
 eval_ds = eval_ds.map(pad_fn, num_parallel_calls=tf.data.AUTOTUNE)
 eval_ds = eval_ds.prefetch(2)
--- a/examples/training/object_detection/pascal_voc/retina_net.py
+++ b/examples/training/object_detection/pascal_voc/retina_net.py
@@ -332,8 +332,6 @@ eval_ds = eval_ds.map(pad_fn, num_parallel_calls=tf.data.AUTOTUNE)
 eval_ds = eval_ds.prefetch(2)
 
 
-# eval_ds = eval_ds.map(unpackage_dict, num_parallel_calls=tf.data.AUTOTUNE)
-
 """
 Our data pipeline is now complete.  We can now move on to model creation and training.
 """
--- a/examples/training/object_detection/pascal_voc/retina_net.py
+++ b/examples/training/object_detection/pascal_voc/retina_net.py
@@ -1,15 +1,31 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import resource
+import sys
 
 import tensorflow as tf
 import tensorflow_datasets as tfds
+from absl import flags
 from tensorflow import keras
 
 import keras_cv
+from keras_cv.callbacks import PyCOCOCallback
 
 low, high = resource.getrlimit(resource.RLIMIT_NOFILE)
 resource.setrlimit(resource.RLIMIT_NOFILE, (high, high))
 
-BATCH_SIZE = 8
 EPOCHS = 100
 CHECKPOINT_PATH = "checkpoint/"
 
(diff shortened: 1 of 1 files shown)


examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -1,15 +1,37 @@
...
+Date created: 2022/09/27
+Last modified: 2022/12/08
+Description: Use KerasCV to train a RetinaNet on Pascal VOC 2007.
+"""
 import resource
+import sys
 
 import tensorflow as tf
 import tensorflow_datasets as tfds
+from absl import flags
```
qlzh727, line 26: order wise, the absl should appear before tf.

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -38,9 +60,38 @@
 ]
 class_mapping = dict(zip(range(len(class_ids)), class_ids))
 
+flags.DEFINE_string(
+    "weights_path",
+    "weights_{epoch:02d}.h5",
```
qlzh727, line 65: It seems that this is not just a path, but a formatted string for the file name. Should we make it clear in the flag description?

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -38,9 +60,38 @@
...
+    "Directory which will be used to store weight checkpoints.",
```
qlzh727, line 66: h5 seems to be a file, not a directory.

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -38,9 +60,38 @@
...
+)
+flags.DEFINE_string(
+    "tensorboard_path",
+    "logs",
```
qlzh727, line 70: The default value here is bit weird, and should it have a relative location?

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -38,9 +60,38 @@
...
+    "Directory which will be used to store tensorboard logs.",
+)
+FLAGS = flags.FLAGS
+FLAGS(sys.argv)
+
+# parameters from RetinaNet [paper](https://arxiv.org/abs/1708.02002)
```
qlzh727, line 76: What's the "parameters" here?

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -202,41 +253,85 @@ def apply(inputs):
...
+
+def pad_fn(image, boxes):
+    return image, boxes.to_tensor(default_value=-1.0)
+
+
 train_ds = train_ds.map(
     proc_train_fn("xywh", image_size), num_parallel_calls=tf.data.AUTOTUNE
 )
 train_ds = train_ds.apply(
     tf.data.experimental.dense_to_ragged_batch(BATCH_SIZE, drop_remainder=True)
```
qlzh727, line 317: Shouldn't we use the global batch size here for the original dataset?

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -202,41 +253,85 @@ def apply(inputs):
...
 )
+
+train_ds = train_ds.map(pad_fn, num_parallel_calls=tf.data.AUTOTUNE)
+train_ds = train_ds.shuffle(8)
```
qlzh727, line 321: Should this value change based on the batch size?

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -202,41 +253,85 @@ def apply(inputs):
...
+        raw_image = inputs["image"]
+        raw_image = tf.cast(raw_image, tf.float32)
+
+        img_size = tf.shape(raw_image)
+        height = img_size[0]
+        width = img_size[1]
+
+        target_height = tf.cond(
+            height > width,
+            lambda: 640.0,
```
qlzh727, line 272: Probably lift this 640 out as a constant.

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -202,41 +253,85 @@ def apply(inputs):
...
+            images=image,
+            source="xyxy",
+            target=bounding_box_format,
+        )
+        gt_classes = tf.cast(inputs["objects"]["label"], tf.float32)
+        gt_classes = tf.expand_dims(gt_classes, axis=-1)
+
+        bounding_boxes = tf.concat([gt_boxes, gt_classes], axis=-1)
+
+        return image, bounding_boxes
```
qlzh727, line 304: Should we return dict (same as the input), or just the tuple here.

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -250,190 +345,62 @@ def unpackage_dict(inputs):
...
+        # pixel range (0, 255) or if you have already rescaled your inputs to the range
+        # (0, 1).  In our case, we feed our model images with inputs in the range (0, 255).
+        include_rescaling=True,
+    )
+    # Fine-tuning a RetinaNet is as simple as setting backbone.trainable = False
+    model.backbone.trainable = False
+
+    metrics = [
+        keras_cv.metrics.COCOMeanAveragePrecision(
+            class_ids=range(21),
```
qlzh727, line 370: maybe we should change 21 to len(class_ids)

qlzh727 (changes requested):

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: Actually no longer needed. Removed.

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: the input needs to be tuple. The boxes will be splitted into dict after RetinaNet supports that.

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: Done.

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: Good catch. Done

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: Good catch. Done

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: Everything from batch size, learning rate, to image size, etc.

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: By default it is using relative path

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: Changing to `weights_name` (compare to file_name)

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: Changing to `weights_name`, compare to file_name

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: Done.

qlzh727: This doesn't seems to be updated.

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -202,41 +254,85 @@ def apply(inputs):
...
+train_ds = train_ds.shuffle(8 * strategy.num_replicas_in_sync)
+train_ds = train_ds.prefetch(2)
+
 eval_ds = eval_ds.map(
-    proc_train_fn(bounding_box_format="xywh", img_size=image_size),
+    proc_eval_fn(bounding_box_format="xywh", target_size=image_size),
     num_parallel_calls=tf.data.AUTOTUNE,
 )
 eval_ds = eval_ds.apply(
     tf.data.experimental.dense_to_ragged_batch(BATCH_SIZE, drop_remainder=True)
```
qlzh727, line 330: same here, it should be the global batch size.

qlzh727 (approved):

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: So that seems to be `shell/format.sh` auto-correcting it

examples/training/object_detection/pascal_voc/retina_net.py
tanzhenyu: Done.

PR #1116, opened 2022-12-09
Delete .so files

ianstenbit:


tanzhenyu (approved):

bhack:
Can you add `.so` file type in https://github.com/keras-team/keras-cv/blob/master/.gitignore ?

ianstenbit:
> Can you add `.so` file type in https://github.com/keras-team/keras-cv/blob/master/.gitignore ?

Good call -- done

bhack:
I suppose the size overhead is small. If not we could remove these from the history with `git filter-branch`.

ianstenbit:
> I suppose the size overhead is small. If not we could remove these from the history with `git filter-branch`.

Total between the two they were 108kb, so I think it's probably not an issue

PR #740, opened 2022-08-29
Porting EfficientNets to KerasCV

AdityaKane2001:
@sayakpaul @LukeWood 

Added Efficientnets as discussed in #625.

/auto Closes https://github.com/keras-team/keras-cv/issues/551

PR #1119, opened 2022-12-12
[0.4 Release SemSeg] Refactor Deeplab to be a functional subclassing model

### LINKED ISSUE #1081: Segmentation Models for Non-RGB Image
**Short Description**

I think, from current implementation, it can be used for **RGB** image but for **GrayScale** image and also for **Non-RGB Multi-Channel** image, it fails to build the model. 

```python
def get_arch():
    model = DeepLabV3(
        classes=21,
        backbone="resnet50_v2",
        include_rescaling=True,
        backbone_weights=None
    )
    return model

model = get_arch()
rgb_input = tf.ones(shape=(1, 100, 100, 3))
model(rgb_input ).shape # OK

model = get_arch()
gray_input = tf.ones(shape=(1, 100, 100, 1))
model(gray_input ).shape # NO

model = get_arch()
multi_channel_input = tf.ones(shape=(1, 100, 100, 4))
model(multi_channel_input ).shape # NO
```

**Existing Implementations**

1. For **gray image**, maybe we can do 

```python
input = tf.ones(shape=(1, 100, 100, 1))

tf.repeat(input, repeats=3, axis=-1) # (non-trainablae)
<tf.Tensor: shape=(1, 100, 100, 3), dtype=float32
or,
tf.keras.layers.Conv2D(3, 1, 1)(input) # (trainable)
<tf.Tensor: shape=(1, 100, 100, 3), dtype=float32
```

Integreate any of the above in the model if `input.shape[-1] == 1`.

2. For **non-rgb multi-channel** input tensor, maybe we can do,

```python
def get_arch(encoder_wg='imagenet'):
    model = DeepLabV3(
        classes=21,
        backbone="resnet50_v2",
        include_rescaling=True,
        backbone_weights=encoder_wg
    )
    return model

model = get_arch(encoder_wg=None)
multi_channel_input = tf.ones(shape=(1, 100, 100, 4))
model( multi_channel_input ).shape # should build
``` 

Or, we can use 

```python
def get_arch(encoder_wg='imagenet'):
    model = DeepLabV3(
         # for input.shape[-1] > 3
         tf.keras.layers.Conv2D(3, 1, 1)(input) 
         # ....
    )
    return model

multi_channel_input = tf.ones(shape=(1, 100, 100, 4))
model = get_arch(encoder_wg='imagenet')
model( multi_channel_input ).shape # should build
```

@qlzh727 @tanzhenyu

### LINKED ISSUE #1104: Refactor Deeplab to be a functional subclassing model
as title

DavidLandup0:
# What does this PR do?

Refactors the DeepLabV3 model to use the hybrid functional subclassing API, and to use internal components instead of redefining them. Seems to train the same (tested first half of the first epoch on both the original and this refactor).

- Refactored to functional subclassing API
- Updated tests
- ~~Changed proprietary segmentation head in lieu of the `__internal__` `SegmentationHead`.~~ `__internal__` segmentation head is now part of DeepLab
- Added `output_activation` for the segmentation head which optionally applies an activation layer to the output (defaults to `softmax`), but was hardcoded before which would fail for logit outputs
- ~~Extracted a verbose backbone init to a helper method for readability~~ Backbone is now a KCV model
- The `Resizing` layer seems to throw an error if we don't specify the input shape upfront.
- Allows the use of greyscale images for DeepLabV3

While I figure out if we can avoid having the input shape as a mandatory argument - the rest is ready for review.

Fixes #1104 
Fixes #1081

## Before submitting
      Pull Request section?
- [x] Was this discussed/approved via a Github issue? Please add a link #1104 
      to it if that's the case.

## Who can review?

@LukeWood @tanzhenyu


keras_cv/models/segmentation/deeplab.py
```diff
@@ -56,136 +71,102 @@ class DeepLabV3(tf.keras.models.Model):
...
+            ),
+            tf.keras.layers.BatchNormalization(),
+            tf.keras.layers.Activation("relu"),
+            tf.keras.layers.Dropout(0.2),
+            tf.keras.layers.Conv2D(
+                filters=classes,
+                kernel_size=(1, 1),
+                padding="same",
+                use_bias=False,
+                activation="softmax",
```
innat, line 151: Is activation type hard coded? The loss function with `from_logits=True` might fail. 

cc. @qlzh727

keras_cv/models/segmentation/deeplab.py
```diff
@@ -56,136 +71,102 @@ class DeepLabV3(tf.keras.models.Model):
...
-        base_config = super().get_config()
-        return dict(list(base_config.items()) + list(config.items()))
+    inputs = utils.parse_model_inputs(input_shape, input_tensor)
+    height = input_shape[0]
+    width = input_shape[1]
+
+    x = inputs
+
+    if include_rescaling:
+        x = layers.Rescaling(1 / 255.0)(x)
```
innat, line 89: Just to inform, model from `keras_cv.models` already take include_recaling parameters and compute it internally. Models like EfficientNet-V2 has two different rescaling approach. So, this could be an issue.

keras_cv/models/segmentation/deeplab.py
```diff
@@ -56,136 +71,102 @@ class DeepLabV3(tf.keras.models.Model):
...
+    model = tf.keras.Model(inputs, output, name=name, **kwargs)
+
+    if backbone_weights is not None:
+        backbone.load_weights(backbone_weights)
+
+    return model
+
+
+def get_segmentation_head(classes):
+    return tf.keras.Sequential(
```
innat, line 135: Can you set name for the segmentation model? It will be convinient to visualize with `plot_model` function. ex,

```python
Sequential(
    layers=None, name='segmentation head'
)
```

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Good catch. I just noticed that we have a `SegmentationHead` in `__internal__`. Updating to use it, as it doesn't have an activation in the final layer. We can then pass the output of the head through an activation layer if it's requested in DeepLabV3.

DavidLandup0: Updated!

keras_cv/models/segmentation/deeplab.py
DavidLandup0: The `include_rescaling` flag is also passed to the backbone model, so they're synced. We might want to check whether a custom backbone has a rescaling layer and raise a warning message or error if it doesn't, since it'll be easy to miss this on the user's end.

Could you clarify what you mean by EffNets having a different rescaling approach? Afaik, they just either have or don't have the layer at the start :thinking:

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Exchanged the Sequential seg head for the modular component, which is a single layer now.
I'll add the name to the layer

DavidLandup0: image

keras_cv/models/segmentation/deeplab.py
innat: About effnet-v2, [check](https://github.com/keras-team/keras/blob/db667716ac6be59a8495b58e0d2e246e9c031599/keras/applications/efficientnet_v2.py#L960-L972) in keras.applications. But it looks like, it didn't follow in [keras-cv.models](https://github.com/keras-team/keras-cv/blob/b0039e2b213474f6fca8db6c3827f0ff6d6ca6df/keras_cv/models/efficientnet_v2.py#L895-L896)

cc. @sebastian-sz

DavidLandup0: Oh, yes. The normalization was dropped while porting them: https://github.com/keras-team/keras-cv/pull/740#discussion_r960951275

sebastian-sz: Context: The normalization/rescaling in keras application's EfficientNetV2 is coupled with the weights ported from the original repository - they expect this kind of preprocessing.

In keras-cv the goal is to train networks from scratch so there was no need to strictly follow the same preprocessing.

tanzhenyu (changes requested):
Thanks for the PR!
Have you tested with the training script?

DavidLandup0:
Yup! Seems to train the same on the first epoch. It's worth testing for more than that.
I'll do a 20-epoch run for validation?

DavidLandup0:
I updated the `SegmentationHead` to be customizable in the call, rather than hardcoded. With it, it was adjusted to match the segmentation head manually created in `DeepLabV3` before, but now it uses the modular component rather than manually implement it again.

The only difference is that the original head from `DeepLabV3` had set `use_bias` false for the classification layer, which should've probably been `True` instead. The new `SegmentationHead` has an adjustable `use_bias` for the first Conv2D (no need for it because it's followed by BN) but has bias for the second Conv2D (no BN after it, has use). Therefore, the param counts are different accounting for the bias:

Original:

```
Total params: 39,174,912
Trainable params: 39,125,888
Non-trainable params: 49,024
```

New:

```
Total params: 39,174,933
Trainable params: 39,125,909
Non-trainable params: 49,024
```

That being said - it's totally equivalent to the original implementation, but significantly cleaner and uses KCV components now. Running a longer training sequence, but the first epoch looks the exact same as the original one so far, which is to be expected.

DavidLandup0:
image
For reference - the KCV training script, with 2 GPUs, and subsequently a batch size of 8 and 0.0035 base LR, after just a few epochs

innat:
@DavidLandup0 Is it possible to address this issue https://github.com/keras-team/keras-cv/issues/1081 in this PR or is it out of scope for now?

DavidLandup0:
The issue seems to be in the Spatial Pyramid Pooling component. Since it's a separate component - maybe it's better to update it later individually in a separate PR 🤔 
The updated `SegmentationHead` is only updated to remove hardcoded values to replace them with optionals with the same defaults, but this might require more than that, so IMO, it's out of scope for this PR

DavidLandup0:
Actually, nvm - it's updated to allow for greyscale images! Moving the convo to that issue

DavidLandup0:
Epoch 11: val_mean_io_u: 0.5723

DavidLandup0:
Just pinging since this should be clean, and was marked as priority a couple of weeks ago

@tanzhenyu

keras_cv/models/segmentation/__internal__/segmentation_head.py
```diff
@@ -61,15 +61,30 @@ def __init__(
...
-        """"""
+        """
+        Args:
+            classes: the number of possible classes for the segmentation map
+            convs: default 2; the number of conv blocks to use in the head (conv2d-batch_norm-activation blocks)
+            filters: default 256; the number of filters in each Conv2D layer
+            activations: default 'relu'; the activation to apply in conv blocks
+            output_scale_factor: default None; the scale to apply in the UpSampling call before the output
+            dropout: default 0.0; the dropout to apply between each conv block
+            **kwargs:
```
tanzhenyu, line 77: add docstring for `kernel_size` and `use_bias`?

keras_cv/models/segmentation/__internal__/segmentation_head.py
```diff
@@ -126,6 +145,8 @@ def get_config(self):
             "filters": self.filters,
             "activations": self.activations,
             "output_scale_factor": self.output_scale_factor,
+            "dropout": self.dropout,
+            "kernel_size": self.kernel_size,
```
tanzhenyu, line 149: `use_bias`?

keras_cv/models/segmentation/deeplab.py
```diff
@@ -13,9 +13,13 @@
 # limitations under the License.
 
 import tensorflow as tf
+from tensorflow import keras
+from tensorflow.keras import layers
 
 from keras_cv.layers.spatial_pyramid import SpatialPyramidPooling
+from keras_cv.models import utils
 from keras_cv.models.resnet_v2 import ResNetV2
+from keras_cv.models.segmentation.__internal__ import SegmentationHead
```
tanzhenyu, line 22: See comment below, we should simply remove the __internal__ implementation

keras_cv/models/segmentation/deeplab.py
```diff
@@ -67,12 +73,21 @@ def __init__(
...
-        self.classes = classes
-        # ================== Backbone and weights. ==================
+        inputs = utils.parse_model_inputs(input_shape, input_tensor)
+        height = input_shape[0]
+        width = input_shape[1]
+
+        x = inputs
+
+        if include_rescaling:
+            x = layers.Rescaling(1 / 255.0)(x)
```
tanzhenyu, line 89: just a nit that we're moving away from `backbone` as a string and always accepting `backbone` as a `keras.Model`, so we no longer need `backbone_weights` and `include_rescaling` here

keras_cv/models/segmentation/deeplab.py
```diff
@@ -82,110 +97,95 @@ def __init__(
...
+                        "The input shape is set up for greyscale images with one channel, but backbone weights are trained on colored images and cannot be loaded."
+                    )
+                backbone = get_resnet_backbone(
+                    backbone_weights, include_rescaling, input_shape, **kwargs
                 )
 
         else:
             # TODO(scottzhu): Might need to do more assertion about the model
+            # What else do we want to test for? Shapes? This feels like too little, but
+            # more assertions feel like they'd be limiting.
```
tanzhenyu, line 113: remove this comment?

keras_cv/models/segmentation/deeplab.py
```diff
@@ -82,110 +97,95 @@ def __init__(
...
-                        kernel_size=(1, 1),
-                        padding="same",
-                        use_bias=False,
-                        activation="softmax",
-                        # Force the dtype of the classification head to float32 to avoid the NAN loss
-                        # issue when used with mixed precision API.
-                        dtype=tf.float32,
-                    ),
-                ]
+            segmentation_head = SegmentationHead(
```
tanzhenyu, line 131: I think I intentionally moved away from __internal__.segmentation_head.py, because the implementation here works better (in terms of model quality).
Would you mind simply extracting this out, but still keep the SegmentationHead inside the file? Just similar to class RCNNHead inside faster_rcnn.py

keras_cv/models/segmentation/deeplab.py
```diff
@@ -67,12 +73,21 @@ def __init__(
         backbone_weights=None,
         spatial_pyramid_pooling=None,
         segmentation_head=None,
+        segmentation_head_activation="softmax",
```
tanzhenyu, line 76: I think the `segmentation_head` should be responsible for the activation

keras_cv/models/segmentation/deeplab.py
```diff
@@ -82,110 +97,95 @@ def __init__(
...
-                    tf.keras.layers.BatchNormalization(),
-                    tf.keras.layers.Activation("relu"),
-                    tf.keras.layers.Dropout(0.2),
-                    tf.keras.layers.Conv2D(
-                        filters=classes,
-                        kernel_size=(1, 1),
-                        padding="same",
-                        use_bias=False,
-                        activation="softmax",
-                        # Force the dtype of the classification head to float32 to avoid the NAN loss
```
tanzhenyu, line 139: This "forcing to float32" part needs to be kept.

tanzhenyu (changes requested):
Thanks for the PR! Definitely getting cleaner

tanzhenyu:
> Epoch 11: val_mean_io_u: 0.5723
> 
> Edit: Replicates the curve as shown in the TB logs. Haven't run for the whole 99 epochs, but the majority of the learning happens in the first 30, and the first 15 were replicated fully.
> 
> Confident that it trains the same :)

I remember it should get to 69.0 meanIOU around 60 epochs

tanzhenyu:
> I updated the `SegmentationHead` to be customizable in the call, rather than hardcoded. With it, it was adjusted to match the segmentation head manually created in `DeepLabV3` before, but now it uses the modular component rather than manually implement it again.
> 
> The only difference is that the original head from `DeepLabV3` had set `use_bias` false for the classification layer, which should've probably been `True` instead. The new `SegmentationHead` has an adjustable `use_bias` for the first Conv2D (no need for it because it's followed by BN) but has bias for the second Conv2D (no BN after it, has use). Therefore, the param counts are different accounting for the bias:
> 
> Original:
> 
> ```
> Total params: 39,174,912
> Trainable params: 39,125,888
> Non-trainable params: 49,024
> ```
> 
> New:
> 
> ```
> Total params: 39,174,933
> Trainable params: 39,125,909
> Non-trainable params: 49,024
> ```
> 
> That being said - it's totally equivalent to the original implementation, but significantly cleaner and uses KCV components now. Running a longer training sequence, but the first epoch looks the exact same as the original one so far, which is to be expected. Can you take a look? @tanzhenyu

Also, would you mind researching a little on other implementation / original paper to see what is the expected #params?

keras_cv/models/segmentation/__internal__/segmentation_head.py
DavidLandup0: Done

keras_cv/models/segmentation/__internal__/segmentation_head.py
DavidLandup0: Good catch!

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Removing

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Yes - the segmentation head in DeepLab was different than the `SegmentationHead`, and seemed to be tailored to the model.

The reason I updated the `SegmentationHead` layer with new arguments was to make sure that it can be used to simulate the exact internal head you built in the DeepLab class. Then, I used the updated head and customized it to create an exact replica of the tailored head, sans the `use_bias` arguments (which lead to a few unnecessary params).

Generalizing the segmentation head and making it customizable is IMO a really important part of shipping sem seg in KCV.

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Makes sense, updating

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Yup - I don't remember why I separated them, but it seemed sensible at the time... Updating

DavidLandup0: Oh I remember - I wanted to separate them to allow a backbone to use logits instead of a softmax, in which case we could turn off the softmax for the segmentation head.

Updated to have the `output_activation` of the head optional within the semantic segmentation head class (output can be logits or softmaxed probability distribution).

keras_cv/models/segmentation/deeplab.py
DavidLandup0: It's still there, but at a different place in the file

DavidLandup0:
> Also, would you mind researching a little on other implementation / original paper to see what is the expected #params?

The papers didn't release param numbers, sadly. The official code repo relies on TF1 and might be tricky to inspect.

- The `torchvision` equivalent has 42,004,074 params and a 66 mIOU on Pascal VOC.
- Ours has 39M params and ~68 mIOU on Pascal VOC.

DavidLandup0:
> I remember it should get to 69.0 meanIOU around 60 epochs. Maybe Ian already published a tensorboard log and weights?

Yes - that's the run I was referencing. I can do another longer run and share TB logs, figured the first 15 would be enough.
I'm currently doing another run on the Kaggle enviro (sem seg script testing), but as soon as that's done, I'll do one for this

keras_cv/models/segmentation/deeplab.py
```diff
@@ -82,110 +95,88 @@ def __init__(
...
-                    stackwise_dilations=BACKBONE_CONFIG["ResNet50V2"][
-                        "stackwise_dilations"
-                    ],
-                    include_rescaling=include_rescaling,
-                    include_top=False,
-                    name="resnet50v2",
-                    weights=parse_weights(backbone_weights, False, "resnet50v2"),
-                    pooling=None,
-                    **kwargs,
+                if backbone_weights and input_shape[-1] == 1:
```
tanzhenyu, line 100: what if input_shape = None?
when we use `inputs=parse_model_inputs`, just get the channels info as well.

keras_cv/models/segmentation/deeplab.py
```diff
@@ -82,110 +95,88 @@ def __init__(
...
-        feature_map = self.backbone(inputs, training=training)
-        output = self.aspp(feature_map, training=training)
-        output = self.up_layer(output, training=training)
-        output = self.segmentation_head(output, training=training)
-        return output
-
-    # TODO(tanzhenyu): consolidate how regularization should be applied to KerasCV.
-    def compile(self, weight_decay=0.0001, **kwargs):
-        self.weight_decay = weight_decay
-        super().compile(**kwargs)
```
tanzhenyu, line 166: I think the model will heavily overfit without weight regularization, so we need to keep `compile` and `train_step` here.

keras_cv/models/segmentation/deeplab.py
tanzhenyu: I'm fine with making it more generic. However do note that we're not offering segmentation head as a public API, so:
1) please extract this out from the model init section, but still keep the SegmentationHead in the same file
2) remove __internal__/segmentation_head.py file.

The reason we're doing it this way is to give users a comprehensive view of how they should build each component to work with the meta architecture. This especially applies to head where it's only a mixture of conv and dense layers.

tanzhenyu (changes requested):

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Hmm, I see. I misunderstood that the seghead file will eventually become a public API.
Makes sense, refactoring. Thanks for the clarification!

DavidLandup0: Do I remove the `segmentation_head_test.py` as well?

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Removed

keras_cv/models/segmentation/deeplab.py
DavidLandup0: If the the input shape isn't known, the initialization fails, since the Upsampling layers needs the size during instantiation. Can we have a `build()` method besides `__init__` for this, while doing functional subclassing? If so, we could avoid this

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Adding them back in!
Since these are on new lines, GH might treat them as mine. How can I add you as the author of the methods? 😓

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Did a `git mv keras_cv/models/segmentation/__internal__/segmentation_head_test.py keras_cv/models/segmentation/segmentation_head_test.py` and replaced the import so the test still works. Though, GH now treats the tests as my LOC instead of Scott's...

Looking into how I can preserve the right credit

DavidLandup0:
Updated with your comments in mind. I'll run a training job to verify it still trains the same and share the TB logs.
If it ends up being too much for a Kaggle environment, might I ask you to run it on the team's GCP with 4 accelerators?

keras_cv/models/segmentation/deeplab.py
tanzhenyu: you can call `build` inside `__init__`

keras_cv/models/segmentation/deeplab.py
tanzhenyu: I think it's also possible for us to completely rely on `SGD(weight_decay)` -- as long as we can verify this works well on TF 2.10, because the training script requires at least 2.10 as `keras.metrics.MeanIOU` is fixed after 2.10.

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Awesome! I'll update it to use the input size and see if the model can be built with unknown sizes upfront. The upsampling blocks are making it a bit more difficult

tanzhenyu:
> Updated with your comments in mind. I'll run a training job to verify it still trains the same and share the TB logs. If it ends up being too much for a Kaggle environment, might I ask you to run it on the team's GC with 4 accelerators?
> 
> Edit: for both the original implementation, and the refactor, the Kaggle environment takes 1s/step (both the single-GPU and double-GPU setups). I'll train until I hit the limit, but if we can do the full validation test on the team's GC environmennt, it'd be great. @tanzhenyu

Before this is ready for merge, I will run it again on my end to make sure this works fine. However in the future as we're getting more contributors working on it, we need to find a way that is more scalable, allowing contributors to use more GPUs

keras_cv/layers/serialization_test.py
```diff
@@ -19,7 +19,7 @@
 from keras_cv import core
 from keras_cv import layers as cv_layers
 from keras_cv.layers.vit_layers import PatchingAndEmbedding
-from keras_cv.models.segmentation.__internal__ import SegmentationHead
+from keras_cv.models.segmentation.deeplab import SegmentationHead
```
tanzhenyu, line 22: no need to add serialization for this head

keras_cv/models/segmentation/deeplab.py
tanzhenyu: No need to test for the head.

keras_cv/models/segmentation/segmentation_head_test.py
```diff
@@ -14,7 +14,7 @@
 
 import tensorflow as tf
 
-from keras_cv.models.segmentation.__internal__ import SegmentationHead
+from keras_cv.models.segmentation.deeplab import SegmentationHead
```
tanzhenyu, line 17: you can simply remove this file

tanzhenyu (changes requested):

keras_cv/models/segmentation/segmentation_head_test.py
DavidLandup0: Sure, removing head-related tests

keras_cv/models/segmentation/deeplab.py
DavidLandup0: It would be great if we can do that. 
Should I try it out in the next run or will you? My runs take longer and run for fewer epochs, which makes experiments a bit harder. If we have the time, I don't mind running them.

DavidLandup0:
> Before this is ready for merge, I will run it again on my end to make sure this works fine. However in the future as we're getting more contributors working on it, we need to find a way that is more scalable, allowing contributors to use more GPUs

Thanks! Sorry to dump it on you. Any form of access to GPUs would surely help with the development, but the resource management seems like it might be a difficult hurdle to solve.

tanzhenyu:
> > Before this is ready for merge, I will run it again on my end to make sure this works fine. However in the future as we're getting more contributors working on it, we need to find a way that is more scalable, allowing contributors to use more GPUs
> 
> Thanks! Sorry to dump it on you. Any form of access to GPUs would surely help with the development, but the resource management seems like it might be a difficult hurdle to solve.

It's hard, and also it's not necessary to have this at the early stage of development. The focus of keras-team is trying to get it running first (detection / segmentation), so contributors don't need to worry about data ingestion, preprocessing, metrics, losses, etc, but only focus on the modeling part (which is the current situation). After that, we should improve the process so contributors can build backbones / meta architectures easier

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
```diff
@@ -129,12 +136,13 @@ def proc_eval_fn(examples):
...
-        classes=21,
-        backbone="resnet50_v2",
+    backbone = ResNet50V2(
         include_rescaling=True,
-        backbone_weights="imagenet",
+        input_shape=(512, 512, 3),
+        include_top=False,
+        weights="imagenet",
     )
+    model = DeepLabV3(classes=21, backbone=backbone, input_shape=(512, 512, 3))
```
tanzhenyu, line 145: nit -- can DeepLab use `input_shape=None` by default and infer the input shape from backbone, but error out if neither `input_shape` is provided or backbone input_shape is also unknown?

keras_cv/models/segmentation/deeplab.py
```diff
@@ -180,12 +127,163 @@ def train_step(self, data):
...
+    def __init__(
+        self,
+        classes,
+        convs=2,
+        filters=256,
+        activations="relu",
+        output_scale_factor=None,
+        dropout=0.0,
+        kernel_size=3,
+        output_activation="softmax",
```
tanzhenyu, line 193: just say `activation`

keras_cv/models/segmentation/deeplab.py
```diff
@@ -180,12 +127,163 @@ def train_step(self, data):
...
+    ```
+    """
+
+    def __init__(
+        self,
+        classes,
+        convs=2,
+        filters=256,
+        activations="relu",
+        output_scale_factor=None,
```
tanzhenyu, line 190: I don't feel `output_scale_factor` is necessary for now, remove it?

keras_cv/models/segmentation/deeplab.py
```diff
@@ -180,12 +127,163 @@ def train_step(self, data):
...
+        lowest_level = next(iter(sorted(inputs)))
+        x = inputs[lowest_level]
+        for conv_layer, bn_layer in zip(self._conv_layers, self._bn_layers):
+            x = conv_layer(x)
+            x = bn_layer(x)
+            x = tf.keras.activations.get(self.activations)(x)
+            if self.dropout:
+                x = self.dropout_layer(x)
+
+        if self.output_scale_factor is not None:
```
tanzhenyu, line 272: same - no need for it. Creating layer inside `call` is troublesome anyway

tanzhenyu (changes requested):
Dropped several minor comments.
The PR is in a good shape now!

tanzhenyu:
also need to run `sh shell/format.sh&lint.sh`

DavidLandup0:
This now works! The checks allow for a channel of 1, as per a previous request:

```
backbone = keras_cv.models.ResNet50V2(include_rescaling=True, 
                                      input_shape=(256, 256, 3), 
                                      include_top=False, 
                                      weights='imagenet')

model = keras_cv.models.DeepLabV3(classes=25,
                  backbone=backbone)
```

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
DavidLandup0: Done

keras_cv/models/segmentation/deeplab.py
DavidLandup0: It's necessary in DeepLabV3+, which is next in line, and uses the same type of segmentation head.
Since this'll become proprietary for DLV3 and isn't reused, we can remove it.

DavidLandup0:
I think I updated according to all points you raised.
One thing we may want to revisit some time down the line is whether the segmentation head should be public or not - DeepLabV3Plus uses it too, with an additional upscaling layer. I'll update the other PR to reflect all of the feedback and design choices from here.

Thanks for the detailed feedback! Let me know if I should update anything else.
@tanzhenyu

keras_cv/models/segmentation/deeplab.py
DavidLandup0: > when we use inputs=parse_model_inputs, just get the channels info as well.

Got the channels, but they're never used in the arch explicitly. The linter is complaining about an unused variable

keras_cv/models/segmentation/deeplab.py
```diff
@@ -13,152 +13,116 @@
...
-from keras_cv.models.resnet_v2 import ResNetV2
-from keras_cv.models.weights import parse_weights
+from keras_cv.models import utils
 
-BACKBONE_CONFIG = {
-    "ResNet50V2": {
-        "stackwise_filters": [64, 128, 256, 512],
-        "stackwise_blocks": [3, 4, 6, 3],
-        "stackwise_strides": [1, 2, 2, 2],
-        "stackwise_dilations": [1, 1, 1, 2],
```
tanzhenyu, line 26: Did `stackwise_dilations` got removed?

tanzhenyu:
With the latest commit, the meanIOU achieves 65.6% at 60 epochs. The tensorboard with head achieves ~68.0% at 60 epochs. I suspect there's some modeling difference between the two.

DavidLandup0:
Hmm, that's a significant drop. I think it came from the fact that I omitted the dilations in the backbone initialization in the training script. Added it back in now:

```
    backbone = ResNet50V2(
        include_rescaling=True,
        stackwise_dilations=[1, 1, 1, 2],
        input_shape=(512, 512, 3),
        include_top=False,
        weights="imagenet",
    )
```

The DeepLabV3 model itself has the same number of params as before:

```
Total params: 39,174,912
Trainable params: 39,125,888
Non-trainable params: 49,024
```

The last classification layer was reverted to not use bias, as in the original implementation. It's for a grand total of 10 parameters in difference, so that couldn't have been the reason for the drop, but I've matched it for good measure. Could you run it again?
Sorry for the omission on my end!

DavidLandup0:
To validate the structure, I've added weight-loading to the model:

```
backbone = keras_cv.models.ResNet50V2(include_rescaling=True,
                                      stackwise_dilations=[1, 1, 1, 2],
                                      input_shape=(512, 512, 3), 
                                      include_top=False, 
                                      weights='imagenet')

model = DeepLabV3(classes=21,
                  backbone=backbone,
                  weights='voc')
```

keras_cv/models/segmentation/deeplab.py
DavidLandup0: Yup - we don't need the config anymore since the backbone is being supplied by the user instead of being built within DeepLabV3.
We'll want the user to supply dilations as well. 

I forgot to add them in the training script after removing here - that's updated now

tanzhenyu:
> To validate the structure, I've added weight-loading to the model:
> 
> ```
> backbone = keras_cv.models.ResNet50V2(include_rescaling=True,
>                                       stackwise_dilations=[1, 1, 1, 2],
>                                       input_shape=(512, 512, 3), 
>                                       include_top=False, 
>                                       weights='imagenet')
> 
> model = DeepLabV3(classes=21,
>                   backbone=backbone,
>                   weights='voc')
> ```

Yes that was my guess too. Kicking off another run

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
```diff
@@ -129,12 +136,14 @@ def proc_eval_fn(examples):
         boundaries=[30000 * 16 / global_batch],
         values=[base_lr, 0.1 * base_lr],
     )
-    model = DeepLabV3(
-        classes=21,
-        backbone="resnet50_v2",
+    backbone = ResNet50V2(
         include_rescaling=True,
-        backbone_weights="imagenet",
+        stackwise_dilations=[1, 1, 1, 2],
```
tanzhenyu, line 141: Please also add a comment here, 
"the stackwise_dilation config improves the meanIOU by 2 points"

tanzhenyu (approved):

examples/training/semantic_segmentation/pascal_voc/deeplab_v3.py
tanzhenyu: Given I have verified the difference is indeed cause by the config. We should document it here so future users wouldn't make the same mistake.
Also we should start to think how to pass backbone configs into a generic basic_training script.
@DavidLandup0

DavidLandup0: Added!
Yeah, I was thinking about the same thing. Doing the backbone args might prove to be a pain if we want to do them in a user friendly way.
I suppose we could do a list arg, and unpack it but that feels a bit ragged.

I'll post some proposals in the general script once this and DLV3Plus are done

DavidLandup0:
SimCLR seems to have issues with mixed precision training during /gcbrun?
Was this present earlier?

tanzhenyu:
The test doesn't seem related to the PR. Manually merging
Thanks for the contribution!

DavidLandup0:
Awesome, thanks!
Glad to have worked on this - thank you for assigning it to me.

I just pushed some new updates for DeepLabV3+. It seems to train slightly better than DeepLabV3, but we'd need to run a longer test. :)

PR #1121, opened 2022-12-12
Update bounding box format to a dictionary.

LukeWood:
This PR migrates the KerasCV bounding box format to be a dictionary.  This allows us to separate classes, boxes, and confidence scores.  There are a few benefits to doing this:

- code organization is easier
- classes will be able to be either one hot encoded or in index format
- additional keys ensure future-proofing
- we can make  separate the masking out from classes to a new field in the dictionary: `mask`

---

Remaining steps:

- [x] convert function
- [x] filter boxes
- [x] RetinaNet model
- [x] MaskRCNN model
- [x] KPLs
- [x] training script (RetinaNet)
- [x] Object detection layers
- [x] COCO Metrics
- [x] Pycococallback
- [x] training script (RCNN)
- [x] unit test validate
- [x] unit test to_dense
- [x] unit test to_ragged
- [x] verify MaP of RetinaNet after change
- [x] verify Recall of MaskRCNN


examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -69,9 +67,9 @@
 train_ds = tfds.load(
     "voc/2007", split="train+validation", with_info=False, shuffle_files=True
 )
-train_ds = train_ds.concatenate(
-    tfds.load("voc/2012", split="train+validation", with_info=False, shuffle_files=True)
-)
+# train_ds = train_ds.concatenate(
+#     tfds.load("voc/2012", split="train+validation", with_info=False, shuffle_files=True)
+# )
```
tanzhenyu, line 72: why change?

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -127,7 +126,7 @@ def apply(inputs):
             scale_factor=(0.8, 1.25),
             bounding_box_format="xywh",
         ),
-        layers.MaybeApply(layers.MixUp(), rate=0.5, batchwise=True),
+        # layers.MaybeApply(layers.MixUp(), rate=0.5, batchwise=True),
```
tanzhenyu, line 129: why comment out?

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -153,6 +152,7 @@ def unpackage_inputs(data):
 train_ds = train_ds.map(unpackage_inputs, num_parallel_calls=tf.data.AUTOTUNE)
 eval_ds = eval_ds.map(unpackage_inputs, num_parallel_calls=tf.data.AUTOTUNE)
 
+<<<<<<< HEAD
```
tanzhenyu, line 155: seems a merging conflict?

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: not ready for review, still WIP

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -194,19 +189,20 @@ def pad_fn(image, boxes):
...
     keras.callbacks.ModelCheckpoint(CHECKPOINT_PATH, save_weights_only=True),
-    PyCOCOCallback(eval_ds, "xywh"),
+    keras_cv.callbacks.PyCOCOCallback(eval_ds, "xywh"),
 ]
 
 history = model.fit(
     train_ds,
-    validation_data=eval_ds,
-    epochs=35,
+    epochs=50,
```
ianstenbit, line 206: Is change of epoch count intentional?

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -194,19 +189,20 @@ def pad_fn(image, boxes):
     classification_loss="focal",
     box_loss="smoothl1",
     optimizer=optimizer,
+    metrics=[],
```
ianstenbit, line 192: Why is this needed?

keras_cv/bounding_box/converters.py
```diff
@@ -170,99 +159,112 @@ def _xyxy_no_op(boxes, images=None, image_shape=None):
...
+        [(left + right) / 2.0, (top + bottom) / 2.0, right - left, bottom - top],
         axis=-1,
     )
 
 
 def _rel_xyxy_to_xyxy(boxes, images=None, image_shape=None):
     image_height, image_width = _image_shape(images, image_shape, boxes)
-    left, top, right, bottom, rest = tf.split(
-        boxes, [1, 1, 1, 1, boxes.shape[-1] - 4], axis=-1
+    left, top, right, bottom = tf.split(
```
ianstenbit, line 193: Did the linter expand this to a ton of lines? Seems like it doesn't need to be broken up like this (same for line 213)

keras_cv/bounding_box/to_dense.py
```diff
@@ -0,0 +1,69 @@
...
+            compatible.
+    """
+    info = validate.validate(bounding_boxes)
+
+    # Already running in masked mode
+    if not info["ragged"]:
+        return bounding_boxes
+
+    if isinstance(bounding_boxes["classes"], tf.RaggedTensor):
+        print("classes", tf.shape(bounding_boxes["classes"]))
```
ianstenbit, line 53: rm

keras_cv/bounding_box/to_ragged.py
```diff
@@ -0,0 +1,69 @@
...
+        boxes.
+    """
+    info = validate.validate(bounding_boxes)
+
+    if info["ragged"]:
+        return bounding_boxes
+
+    boxes = bounding_boxes.get("boxes")
+    classes = bounding_boxes.get("classes")
+    mask = classes != -1
```
ianstenbit, line 55: Let's make this sentinel value an optional kwarg that defaults to -1

keras_cv/bounding_box/utils.py
```diff
@@ -107,12 +99,20 @@ def clip_to_image(bounding_boxes, bounding_box_format, images=None, image_shape=
...
+    classes = tf.where(nan_indices, tf.constant(-1, classes.dtype), classes)
+
+    # TODO update dict and return
+    clipped_bounding_boxes, classes = _format_outputs(
+        clipped_bounding_boxes, classes, squeeze
     )
-    clipped_bounding_boxes = _format_outputs(clipped_bounding_boxes, squeeze)
-    return clipped_bounding_boxes
+
+    result = bounding_boxes.copy()
```
ianstenbit, line 112: Why do we need to copy this?

keras_cv/bounding_box/validate.py
```diff
@@ -0,0 +1,87 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+
+def validate(bounding_boxes):
+    """validates that a given set of bounding boxes complies to KerasCV format.
```
ianstenbit, line 18: nit: complies with

keras_cv/bounding_box/validate_test.py
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+
+class ValidateTest(tf.test.TestCase):
+    pass
```
ianstenbit, line 18: Looks like these tests are missing still?

keras_cv/layers/object_detection/non_max_suppression_test.py
```diff
@@ -23,126 +23,136 @@ def test_return_shapes(self):
...
+        self.assertEqual(outputs["confidence"].shape, [3, None])
 
     def test_ragged_output_with_differing_shapes(self):
         layer = NonMaxSuppression(8, "xywh", iou_threshold=0.1)
         images = tf.ones((2, 480, 480, 3))
 
-        predictions = tf.convert_to_tensor(
-            [
+        predictions = {
+            "boxes": tf.convert_to_tensor(
```
ianstenbit, line 76: This also looks like it got sorta nuked by the autoformatter

keras_cv/metrics/coco/mean_average_precision_test.py
```diff
@@ -22,7 +22,7 @@
 
 
 class COCOMeanAveragePrecisionTest(tf.test.TestCase):
-    def test_runs_inside_model(self):
+    def DISABLE_test_runs_inside_model(self):
```
ianstenbit, line 25: Why are we disabling all these?

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -439,12 +439,17 @@ def compile(
...
         **kwargs,
     ):
         # TODO(tanzhenyu): Add metrics support once COCOMap issue is addressed.
         # https://github.com/keras-team/keras-cv/issues/915
-        if "metrics" in kwargs.keys():
-            raise ValueError("currently metrics support is not supported intentionally")
+        if metrics is not None and len(metrics) != 0:
+            raise ValueError(
+                "`RetinaNet` does not currently support the use of "
+                "`metrics` due to performance and distribution concerns. Please us the "
```
ianstenbit, line 450: nit: s/us/use

ianstenbit (changes requested):
Thanks Luke!

keras_cv/bounding_box/converters.py
tanzhenyu: Same

keras_cv/bounding_box/to_dense.py
```diff
@@ -0,0 +1,69 @@
...
+
+def _classes_shape(batched, classes_shape, max_boxes):
+    if max_boxes is None:
+        return None
+    if batched:
+        return [None, max_boxes] + classes_shape[:2]
+    return [max_boxes] + classes_shape[:2]
+
+
+def to_dense(bounding_boxes, max_boxes=None):
```
tanzhenyu, line 37: what's the difference between this and `to_tensor`?

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: yeah this is intentional, sometimes it takes this long to converge

keras_cv/bounding_box/converters.py
LukeWood: Unfortunately yes

keras_cv/metrics/coco/mean_average_precision_test.py
LukeWood: COCO metrics don't work at all after this PR.  I can make them a private API if we'd like - but I figure until the numerics are worked out we should just rely on the Callback. 

Lets focus on getting the release out first.

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -157,18 +157,13 @@ def unpackage_inputs(data):
...
 # TODO(tanzhenyu): consider remove padding while reduce function tracing.
 def pad_fn(image, boxes):
-    boxes = boxes.to_tensor(default_value=-1.0, shape=[GLOBAL_BATCH_SIZE, 32, 5])
-    gt_boxes = boxes[..., :4]
-    gt_classes = boxes[..., 4]
-    return image, {
-        "boxes": gt_boxes,
-        "classes": gt_classes,
-    }
+    return image, bounding_box.to_dense(boxes, max_boxes=32)
```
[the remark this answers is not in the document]
LukeWood: to_dense() converts the underlying tensors to dense tensors.

ex:
```
{
"boxes": tf.RaggedTensorSpec(shape=(None, RaggedNone, 4),
"classes": tf.RaggedTensorSpec(shape=(None, RaggedNone),
}
```
=>

```
{
"boxes": tf.TensorSpec(shape=(None, 32, 4),
"classes": tf.TensorSpec(shape=(None, 32),
}
```

keras_cv/bounding_box/to_dense.py
LukeWood: this operates on the dictionary format.  It converts the underlying components to dense tensors.

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -194,19 +189,20 @@ def pad_fn(image, boxes):
...
+
 callbacks = [
     keras.callbacks.TensorBoard(log_dir="logs"),
-    keras.callbacks.ReduceLROnPlateau(patience=5),
-    keras.callbacks.EarlyStopping(patience=10),
+    keras.callbacks.ReduceLROnPlateau(monitor="loss", patience=5),
+    keras.callbacks.EarlyStopping(monitor="loss", patience=10),
     keras.callbacks.ModelCheckpoint(CHECKPOINT_PATH, save_weights_only=True),
-    PyCOCOCallback(eval_ds, "xywh"),
+    keras_cv.callbacks.PyCOCOCallback(eval_ds, "xywh"),
```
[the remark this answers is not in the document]
LukeWood: This is just an import change to match Keras style - no opinion on making this a public API.  I think it is  nice for us to offer a way to evaluate with PyCOCO tools - even when we have our own metrics well ironed out.

examples/training/object_detection/pascal_voc/faster_rcnn.py
```diff
@@ -63,206 +66,87 @@
...
-def pad_fn(examples):
-    gt_boxes = examples.pop("gt_boxes")
-    gt_classes = examples.pop("gt_classes")
-    gt_boxes = gt_boxes.to_tensor(default_value=-1.0, shape=[global_batch, 32, 4])
-    gt_classes = gt_classes.to_tensor(default_value=-1.0, shape=[global_batch, 32, 1])
-    return examples["images"], {
-        "boxes": gt_boxes,
-        "classes": gt_classes,
-    }
+def dict_to_tuple(inputs):
```
tanzhenyu, line 90: can we modify pad_fn, and remove `dict_to_tuple`?

examples/training/object_detection/pascal_voc/faster_rcnn.py
```diff
@@ -289,4 +173,4 @@ def pad_fn(examples):
     rpn_box_loss="Huber",
     rpn_classification_loss="BinaryCrossentropy",
 )
-model.fit(train_ds, epochs=18, validation_data=eval_ds, callbacks=callbacks)
```
tanzhenyu, line 292: why remove validation_data? I think they're here to calculate `val_loss`

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -81,15 +80,16 @@ def apply(inputs):
...
             gt_boxes,
             images=image,
             source="rel_yxyx",
             target=bounding_box_format,
         )
-        bounding_boxes = tf.concat([gt_boxes, gt_classes], axis=-1)
-        return {"images": image, "bounding_boxes": bounding_boxes}
+        return {
+            "images": image,
+            "bounding_boxes": {"boxes": gt_boxes, "classes": gt_classes},
```
tanzhenyu, line 91: Shouldn't the output be:
return (images, {"boxes": gt_boxes, "classes": gt_classes})?

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -157,18 +157,13 @@ def unpackage_inputs(data):
...
 # TODO(tanzhenyu): consider remove padding while reduce function tracing.
 def pad_fn(image, boxes):
-    boxes = boxes.to_tensor(default_value=-1.0, shape=[GLOBAL_BATCH_SIZE, 32, 5])
-    gt_boxes = boxes[..., :4]
-    gt_classes = boxes[..., 4]
-    return image, {
-        "boxes": gt_boxes,
-        "classes": gt_classes,
-    }
+    return image, bounding_box.to_dense(boxes)
```
tanzhenyu, line 160: My suggestion is don't offer a last-minute API before 0.4 release, instead just keep the padding function as-is, given this is a training example script

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -194,19 +189,20 @@ def pad_fn(image, boxes):
...
     optimizer=optimizer,
+    metrics=[],
 )
 
+
 callbacks = [
     keras.callbacks.TensorBoard(log_dir="logs"),
-    keras.callbacks.ReduceLROnPlateau(patience=5),
-    keras.callbacks.EarlyStopping(patience=10),
+    keras.callbacks.ReduceLROnPlateau(monitor="loss", patience=5),
```
tanzhenyu, line 198: Shouldn't we use `val_loss` here?

keras_cv/__init__.py
```diff
@@ -18,6 +18,7 @@
 version_check.check_tf_version()
 # isort:on
 
+from keras_cv import callbacks
```
tanzhenyu, line 21: From the example training script, it sounds like we're just re-import the callback, but from here it sounds like we're offering it as a public API?
@ianstenbit WDYT?

keras_cv/bounding_box/__init__.py
```diff
@@ -23,8 +23,8 @@
 from keras_cv.bounding_box.formats import YXYX
 from keras_cv.bounding_box.iou import compute_iou
 from keras_cv.bounding_box.pad_batch_to_shape import pad_batch_to_shape
-from keras_cv.bounding_box.utils import add_class_id
+from keras_cv.bounding_box.to_dense import to_dense
+from keras_cv.bounding_box.to_ragged import to_ragged
```
tanzhenyu, line 27: Same comments above -- prefer not to add new APIs before 0.4

keras_cv/bounding_box/converters.py
```diff
@@ -112,54 +112,43 @@ def decode_single_level(anchor, box_delta):
 
 
 def _center_yxhw_to_xyxy(boxes, images=None, image_shape=None):
-    y, x, height, width, rest = tf.split(
-        boxes, [1, 1, 1, 1, boxes.shape[-1] - 4], axis=-1
-    )
+    y, x, height, width = tf.split(boxes, [1, 1, 1, 1], axis=-1)
```
tanzhenyu, line 115: nit: the original implement should handle both dim=4 and dim>4, so it's not strictly required to make this before 0.4, I guess.
But leaving it up to you

keras_cv/bounding_box/converters.py
tanzhenyu: If that's the case, can we make [1, 1, 1, 1] as a constant in the header and re-use it here?

keras_cv/bounding_box/converters.py
```diff
@@ -354,6 +356,22 @@ def convert_format(
         dtype: the data type to use when transforming the boxes.  Defaults to
             `tf.float32`.
     """
+    if isinstance(boxes, dict):
+        boxes["boxes"] = convert_format(
```
tanzhenyu, line 360: Do we want the API to accept dict?
it's not bad to do `convert_format(y["boxes"], ...)` TBH.
Also if we do want to make the behavior change, then need to update the docstring.

keras_cv/bounding_box/formats.py
```diff
@@ -27,19 +27,12 @@ class XYXY:
...
-    - CLASS: class of the object contained in the bounding box
-    - CONFIDENCE: confidence that the box is valid, used in predictions
     """
 
     LEFT = 0
     TOP = 1
     RIGHT = 2
     BOTTOM = 3
-    CLASS = 4
-    CONFIDENCE = 5
```
tanzhenyu, line 42: Finally, yay!

keras_cv/bounding_box/validate.py
```diff
@@ -0,0 +1,87 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+
+def validate(bounding_boxes):
+    """validates that a given set of bounding boxes complies with KerasCV format.
+
+    For a set of bounding boxes to be valid it must satisfy the following conditions:
+    - `bounding_boxes` must be a dictionary
```
tanzhenyu, line 21: This seems a little suboptimal -- can we always pass around dict and assume it contains "boxes" and "classes"?

keras_cv/callbacks/pycoco_callback.py
```diff
@@ -44,6 +44,12 @@ def __init__(self, validation_data, bounding_box_format, cache=True, **kwargs):
         self.bounding_box_format = bounding_box_format
         super().__init__(**kwargs)
 
+    def _num_detections(self, bounding_boxes):
+        boxes = bounding_boxes["boxes"]
+        if isinstance(boxes, tf.RaggedTensor):
+            return boxes.row_lengths(axis=1)
+        raise ValueError("Unimplemented")
```
tanzhenyu, line 51: I think the consensus is the nms decoder always return "num_detections"?

keras_cv/callbacks/pycoco_callback.py
```diff
@@ -95,20 +92,24 @@ def boxes_only(images, boxes):
...
+        ground_truth["num_detections"] = [boxes.row_lengths(axis=1)]
+        ground_truth["boxes"] = [boxes.to_tensor(-1)]
+        ground_truth["classes"] = [classes.to_tensor(-1)]
+
+        y_pred = bounding_box.convert_format(
+            y_pred, source=self.bounding_box_format, target="yxyx"
         )
 
-        predictions = {}
+        y_pred = bounding_box.to_ragged(y_pred)
```
tanzhenyu, line 103: Same as above -- we always return dense tensors with `num_detections`, so we don't need to run any ragged conversion here

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -112,10 +107,9 @@ def call(self, predictions, images=None):
         )
 
         # preparing the predictions for TF NMS op
-        boxes = tf.expand_dims(predictions[..., :4], axis=2)
-        class_predictions = tf.cast(predictions[..., 4], tf.int32)
-        scores = predictions[..., 5]
-
+        boxes = tf.expand_dims(predictions["boxes"], axis=2)
+        class_predictions = tf.cast(predictions["classes"], tf.int32)
+        scores = predictions["confidence"]
```
tanzhenyu, line 112: Just return whatever is returned from tf.image.combined_nms without converting to ragged and TensorArray operations

keras_cv/layers/object_detection/non_max_suppression_test.py
tanzhenyu: Yeah..I don't understand why it didn't format this file previously, but the linter chooses to run such aggressive formating in this PR

keras_cv/metrics/coco/utils_test.py
```diff
@@ -21,19 +21,19 @@
 
 
 class UtilTest(tf.test.TestCase):
-    def test_filter_bounding_boxes_empty(self):
+    def DISABLE_test_filter_bounding_boxes_empty(self):
```
tanzhenyu, line 24: Why do we have to disable those tests?

keras_cv/metrics/coco/recall_test.py
```diff
@@ -21,7 +21,7 @@
 
 
 class COCORecallTest(tf.test.TestCase):
-    def test_runs_inside_model(self):
+    def DISABLE_test_runs_inside_model(self):
```
tanzhenyu, line 24: Do we have to disable those tests?

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -279,6 +279,7 @@ def compile(
         box_loss=None,
         classification_loss=None,
         loss=None,
+        metrics=None,
```
tanzhenyu, line 282: if this is not supported, then no need to expose this to user

tanzhenyu (changes requested):
Thanks for the PR!
This is a gigantic PR, I understand it's difficult to break this down into smaller ones.
However, can we try to break this down into different steps?
Step 1 -- only make KPL changes, to return dict outputs, without caring about RetinaNet training script. Without additional `to_ragged`, `to_dense`, `validate`, etc
Step 2 -- only make NMS decoder changes to work with PyCOCOCallbacks
Step 3 -- only make RetinaNet changes to integrate with Step 1.
Those 3 steps seems fairly independent with each other IMHO.

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: the padding function doesn't work with the dictionary format.   We need this API no matter what internally - this is a very common conversion.  It replaces filter_sentinels and pad_with_sentinels.

keras_cv/callbacks/pycoco_callback.py
LukeWood: I coerce to ragged in line 67 by filtering -1s.

keras_cv/layers/object_detection/non_max_suppression_test.py
LukeWood: Its the dictionary nesting - I guess the formatter handles it differently.  This sucks, let me try to fix.

LukeWood: No luck, a fix would be welcome but I couldn't get it to stay nice.

keras_cv/bounding_box/converters.py
LukeWood: lets update the docstring, I personally think its trivial to support so we may as well.  Its a nice convenience.

keras_cv/bounding_box/validate.py
LukeWood: Can you clarify this comment?  Thats basically what I'm hoping to do here, ensure we always have boxes and classes.  We can change our libraries assumptions, but at the moment all components assume classes to be provided.

keras_cv/callbacks/pycoco_callback.py
LukeWood: I don't think theres a consensus here personally;  Is there any reason to not output ragged tensors from predict(), or at least let users decide which way?

I have supported num_detections as well now, but I don't think we should totally abandon outputting ragged boxes.

keras_cv/callbacks/pycoco_callback.py
LukeWood: Doing this makes us support both Ragged/Dense predictions, and it makes them consistent as well.  Callbacks also run on CPU so no harm done to TPU compat.

keras_cv/metrics/coco/utils_test.py
LukeWood: COCO metrics are entirely disabled in this PR as there is:

- no way to make them work with the current implementation
- they still have numerical issues.

I figured it is better to keep this PR as small as possible and just disable COCO metrics, then re-enable them when they are fixed.

keras_cv/metrics/coco/recall_test.py
LukeWood: COCO metrics are entirely disabled in this PR as there is:

- no way to make them work with the current implementation
- they still have numerical issues.

I figured it is better to keep this PR as small as possible and just disable COCO metrics, then re-enable them when they are fixed.

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: the difference is that metrics=[] raised an error before, now it doesn't.  Just a rough edge off the API.

examples/training/object_detection/pascal_voc/faster_rcnn.py
LukeWood: Updated and rerverted to track val_loss

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: no, because the KPLs rely on `"images"` being in a key.  KPLs accept a single dictionary call argument, not tuple.

examples/training/object_detection/pascal_voc/faster_rcnn.py
LukeWood: Done

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: Updated

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: Removed

keras_cv/__init__.py
LukeWood: any reason to not export it?  No matter what our users will want a way to evaluate COCO metrics using PyCOCOTools.  Some people care about exact numerics.

keras_cv/bounding_box/converters.py
LukeWood: Done

keras_cv/bounding_box/utils.py
LukeWood: tf.function considerations.  KPLS call tf.cond() and if you dont copy the dictionary autograph complains that you are relying on inter-dependent branches.  copy() prevents this; 

BTW to be super clear, it only copies the dictionary not the tensors.

keras_cv/metrics/coco/recall_test.py
bhack: Is still like if we could isolate a new test under jit_compile with `@unittest.expectedFailure` of the metrics as its support status it was becoming sometime quite"mystical"

It will be easier to monitor what happens in the compiler side with an expecting failing test on the component.

keras_cv/bounding_box/__init__.py
LukeWood: I would view this as a substitution of an API (lines 28-29) rather than an addition of an API:

see:
```
from keras_cv.bounding_box.utils import filter_sentinels
from keras_cv.bounding_box.utils import pad_with_sentinels
```

has been removed, these replace it.

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -147,28 +147,12 @@ def apply(inputs):
 
 
 def unpackage_inputs(data):
```
tanzhenyu, line 149: you could merge this with previous `unpackage_inputs` function to a similar output in the rcnn training script

keras_cv/__init__.py
tanzhenyu: Isn't the consensus to replace the callback with PyMetrics?
@ianstenbit will know more

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: Actually not possible, one format is for the KPLs, one is for the model.  They're distinct; (images, y) => {"images": images, "bounding_boxes": bounding_boxes}

LukeWood: (not possible at this time, we could make an OD helper function that unpackages dictionaries?)

tanzhenyu: I'm a little confused....the two training scripts should have unified data input and data output, right?

keras_cv/bounding_box/validate.py
tanzhenyu: ok maybe rename it? like `extract_box_info`. Otherwise `validate` sounds like it's not returning anything

tanzhenyu (changes requested):

keras_cv/layers/object_detection/non_max_suppression.py
tanzhenyu, line 112: scores -> confidence

keras_cv/layers/object_detection/non_max_suppression_test.py
```diff
@@ -110,23 +107,70 @@ def test_ragged_output_with_differing_shapes(self):
         layer = NonMaxSuppression(8, "xywh", iou_threshold=0.1)
         images = tf.ones((2, 480, 480, 3))
 
-        predictions = tf.convert_to_tensor(
+        boxes = [
```
tanzhenyu, line 110: I wonder if this can be prevented from `# pylint: disable=no-member`?

keras_cv/layers/object_detection/retina_net_label_encoder.py
```diff
@@ -73,9 +73,8 @@ def __init__(
         )
         self.built = True
 
-    def _encode_sample(self, gt_boxes, gt_classes, anchor_boxes):
+    def _encode_sample(self, bounding_boxes, anchor_boxes):
```
tanzhenyu, line 76: bounding_boxes -> box_labels

keras_cv/layers/object_detection/roi_sampler.py
```diff
@@ -87,7 +87,7 @@ def call(
         Args:
           rois: [batch_size, num_rois, 4]
           gt_boxes: [batch_size, num_gt, 4]
-          gt_classes: [batch_size, num_gt, 1]
```
tanzhenyu, line 90: I somehow remember that if it's without the last dimension, some weird broadcast will happen and the training result would be bad, just make sure you have run both OD models end to end before final merge

keras_cv/layers/preprocessing/jittered_resize.py
```diff
@@ -193,7 +193,7 @@ def augment_bounding_boxes(self, bounding_boxes, transformation, **kwargs):
                 "Please provide a `bounding_box_format` when augmenting "
                 "bounding boxes with `JitteredResize()`."
             )
-
+        result = bounding_boxes.copy()
```
tanzhenyu, line 196: instead of doing copy, can we just say result = bounding_boxes, then later:
`
boxes = bounding_boxes.pop("boxes")
boxes = convert_format(boxes, ...)
result["boxes"] = boxes
`
?

keras_cv/layers/preprocessing/mosaic.py
```diff
@@ -151,10 +151,8 @@ def _batch_augment(self, inputs):
                 axis=-1,
             )
 
-            if isinstance(bounding_boxes, tf.RaggedTensor):
-                bounding_boxes = bounding_boxes.to_tensor(-1)
-
-            bounding_boxes = tf.vectorized_map(
+            bounding_boxes = bounding_box.to_dense(bounding_boxes)
+            bounding_boxes = tf.map_fn(
```
tanzhenyu, line 155: Why changing from vectorized_map to map_fn?

keras_cv/layers/preprocessing/random_flip.py
```diff
@@ -111,38 +111,35 @@ def _flip_image(image, transformation):
...
                 1 - x1,
                 x4,
-                rest,
             ],
             axis=-1,
         )
-        output = tf.squeeze(output, axis=1)
-        return output
+        bounding_boxes = bounding_boxes.copy()
+        bounding_boxes["boxes"] = tf.squeeze(output, axis=1)
```
tanzhenyu, line 125: If user run this eagerly, this will actually copy the tensor, I think?

LukeWood: the tf.squeeze() or bounding_boxes.copy()?  

In any case, do we care at all about eager performance?  From my understanding keras does not care about eager performance as it is already really bad.

keras_cv/layers/object_detection/retina_net_label_encoder.py
LukeWood: updated

keras_cv/layers/object_detection/roi_sampler.py
LukeWood: reverted

keras_cv/layers/preprocessing/jittered_resize.py
LukeWood: .copy() only copies the dictionary - we aren't copying a ton of Tensors, we're copying a single dictionary during trace time.   I don't think this is a concern performance wise.

keras_cv/layers/preprocessing/mosaic.py
LukeWood: map_fn is able to output a different output structure than input structure using fn_output_signature, whereas vectorized map does not expose an argument to specify output signature.  

So this is required to output the dictionary format.

bhack:
Is there a specific advantage to not introduce a Bbox class?

LukeWood:
> Is there a specific advantage to not introduce a Bbox class?

Dictionaries work better with Keras/TensorFlow unfortunately.

bhack:
> > Is there a specific advantage to not introduce a Bbox class?
> 
> Dictionaries work better with Keras/TensorFlow unfortunately.

I meant a `dictable` class:
https://docs.python.org/3/library/stdtypes.html#object.__dict__

bhack:
Or extension types:
https://www.tensorflow.org/guide/extension_type

keras_cv/bounding_box/validate.py
```diff
@@ -0,0 +1,87 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+
+def validate(bounding_boxes):
```
ianstenbit, line 17: What do you think about "validate_format"?

keras_cv/bounding_box/to_ragged_test.py
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+from keras_cv import bounding_box
+
+
+class ToRaggedTest(tf.test.TestCase):
+    def test_converts_to_ragged(self):
```
ianstenbit, line 20: Maybe a round-trip test with randomly initialized boxes would be good?

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -0,0 +1,154 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+from keras_cv import bounding_box
+
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class NonMaxSuppression(tf.keras.layers.Layer):
```
ianstenbit, line 20: If we're creating a brand new NonMaxSupression layer, shouldn't the NmsPredictionDecoder be updated to use this internally?

keras_cv/metrics/coco/numerical_tests/mean_average_precision_test.py
```diff
@@ -39,7 +39,7 @@ class MeanAveragePrecisionTest(tf.test.TestCase):
     Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.610
     """
 
-    def test_mean_average_precision_correctness_default(self):
+    def DISABLE_test_mean_average_precision_correctness_default(self):
```
ianstenbit, line 42: If we're disabling all of these tests, we should also mark the API as private or at least add a docstring that it isn't stable

ianstenbit (changes requested):

keras_cv/bounding_box/validate.py
LukeWood: Done, I think this is a good name.

keras_cv/metrics/coco/numerical_tests/mean_average_precision_test.py
LukeWood: Done

keras_cv/bounding_box/to_ragged_test.py
LukeWood: Done

LukeWood:
ok @ianstenbit, all tests fixed

ianstenbit (approved):
Thanks Luke!

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -388,12 +388,17 @@ def compile(
         rpn_classification_loss=None,
         weight_decay=0.0001,
         loss=None,
+        metrics=None,
```
tanzhenyu, line 391: If we currently don't support it, then no need to expose it.

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -486,7 +491,7 @@ def train_step(self, data):
         if sample_weight is not None:
             raise ValueError("`sample_weight` is currently not supported.")
         gt_boxes = y["boxes"]
-        gt_classes = y["classes"]
+        gt_classes = tf.expand_dims(y["classes"], axis=-1)
```
tanzhenyu, line 494: Can you add a TODO for me to remove this hack?

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -504,7 +509,7 @@ def test_step(self, data):
...
```
tanzhenyu, line 512: Same

keras_cv/models/object_detection/retina_net/retina_net.py
tanzhenyu: Then let's check that through `if kwargs['metrics']`

keras_cv/layers/object_detection/retina_net_label_encoder.py
```diff
@@ -169,12 +171,16 @@ def call(self, images, boxes, classes):
                 "support RaggedTensor inputs for the `images` argument.  Received "
                 f"`type(images)={type(images)}`."
             )
-        gt_boxes = tf.cast(boxes, self.dtype)
-        gt_classes = tf.cast(classes, self.dtype)
 
-        gt_boxes = bounding_box.convert_format(
-            gt_boxes, source=self.bounding_box_format, target="xywh", images=images
+        box_labels = bounding_box.to_dense(box_labels)
```
tanzhenyu, line 175: why need `to_dense` here? I think the input is already dense

keras_cv/layers/object_detection/retina_net_label_encoder.py
```diff
@@ -153,9 +155,9 @@ def _encode_sample(self, gt_boxes, gt_classes, anchor_boxes):
             ),
             tf.cast(matches, tf.int32),
         )
-        return label
+        return result
 
-    def call(self, images, boxes, classes):
+    def call(self, images, box_labels):
         """Creates box and classification targets for a batch
```
tanzhenyu, line 161: Need to update the docstring -- what does box_labels contain

keras_cv/layers/object_detection/retina_net_label_encoder.py
```diff
@@ -184,19 +190,12 @@ def call(self, images, boxes, classes):
...
-
-        result = self._encode_sample(gt_boxes, gt_classes, anchor_boxes)
+        result = self._encode_sample(box_labels, anchor_boxes)
         result = bounding_box.convert_format(
             result, source="xywh", target=self.bounding_box_format, images=images
         )
-        encoded_box_targets = result[..., :4]
-        class_targets = result[..., 4]
+        encoded_box_targets = result["boxes"]
+        class_targets = result["classes"]
```
tanzhenyu, line 198: It looks like now we're in a divergence -- previously every component accept tuples and returns tuples (same for model), but now some components accept dict input but return tuples.
Should we unify this? @ianstenbit

keras_cv/layers/preprocessing/jittered_resize.py
```diff
@@ -206,24 +206,24 @@ def augment_bounding_boxes(self, bounding_boxes, transformation, **kwargs):
...
-        bounding_boxes = keras_cv.bounding_box.filter_sentinels(bounding_boxes)
-        return keras_cv.bounding_box.convert_format(
-            bounding_boxes,
+        result = keras_cv.bounding_box.convert_format(
+            result,
             image_shape=self.target_size + (3,),
             source="yxyx",
             target=self.bounding_box_format,
         )
+        result = keras_cv.bounding_box.to_ragged(result, dtype=self.compute_dtype)
```
tanzhenyu, line 225: BaseAugmentationLayer already runs `to_ragged` during `_augment`, why do we need to run it again here?

keras_cv/layers/preprocessing/mix_up.py
```diff
@@ -110,27 +110,40 @@ def _update_labels(self, labels, lambda_sample, permutation_order):
...
+                f"Got: inputs = {inputs}."
             )
+
         if labels is not None and not labels.dtype.is_floating:
             raise ValueError(
                 f"MixUp received labels with type {labels.dtype}. "
                 "Labels must be of type float."
             )
 
+        if bounding_boxes is not None:
```
tanzhenyu, line 139: Should this check be around for all layers?

keras_cv/layers/preprocessing/mosaic.py
```diff
@@ -164,8 +162,18 @@ def _batch_augment(self, inputs):
                     index,
                 ),
                 tf.range(batch_size),
+                fn_output_signature={
+                    "boxes": tf.RaggedTensorSpec(
+                        shape=[None, 4],
```
tanzhenyu, line 167: This shape seems too restrictive -- what if we want to support higher dimension?

keras_cv/layers/preprocessing/random_flip.py
tanzhenyu: Ok, we could revisit this if memory issue comes up

keras_cv/losses/iou_loss.py
```diff
@@ -82,7 +82,9 @@ def call(self, y_true, y_pred):
                 f"the bounding boxes. Received y_true.shape[-1]={y_true.shape[-1]}."
             )
 
-        ious = bounding_box.compute_iou(y_true, y_pred, self.bounding_box_format)
+        ious = bounding_box.compute_iou(
+            y_true[..., :4], y_pred[..., :4], self.bounding_box_format
```
tanzhenyu, line 86: why is it concatenated here? Shouldn't we expect dict tensors, or simply single Tensor at this point?

keras_cv/metrics/coco/mean_average_precision_test.py
tanzhenyu: I'm fine with completely hiding COCOMetrics
However, add TODO at top of the file explaining why we did this, so we won't get lost later

keras_cv/models/object_detection/faster_rcnn.py
tanzhenyu: Also, given this is backward breaking, we should have a safety-check here: if the input doesn't have it, expand the dimension, otherwise keep as-is

tanzhenyu (changes requested):
Thanks! Getting closer now!

bhack:
Any comment on the extension type solution before merge?

Thanks

tanzhenyu:
> https://www.tensorflow.org/guide/extension_type

We are looking at supporting dataclass, which seems to be similar to the `ExtensionType` you mentioned above. I think it'd be a natural fit for box related stuff, however it's not exactly a collection of Tensors, but some would be enums and python natives

keras_cv/bounding_box/__init__.py
LukeWood: can we wait until we move everything to a stable API to worry too much about this?  We will need to use this in training scripts too anyways, so I'd like to avoid the effort of changing all of the imports + replacing the statements if I'm just going to undo it in the next PR.

keras_cv/layers/object_detection/retina_net_label_encoder.py
LukeWood: the input can be either ragged or Dense.  If the input is already dense the function simply returns the original, so this basically ensures we are ALWAYS operating in dense mode.

keras_cv/layers/object_detection/retina_net_label_encoder.py
LukeWood: lets solve in follow up

keras_cv/layers/preprocessing/jittered_resize.py
LukeWood: Removed, correct!

keras_cv/layers/preprocessing/mix_up.py
LukeWood: I guess `bounding_box.validate_format()` handles this now, so we don't need this anymore - updated to use the utility.

keras_cv/layers/preprocessing/mosaic.py
LukeWood: Lets solve that if a need arises, but I think storing them in a different key makes sense for now?  I'd opt to follow the YAGNI principle here; it sounds like it could be useful, but lets wait until theres a strong need.

keras_cv/models/object_detection/faster_rcnn.py
LukeWood: I did this so that `metrics=[]` can pass

LukeWood: updated to remove

keras_cv/models/object_detection/faster_rcnn.py
LukeWood: Added a safety check raising a value error.  Lets require proper inputs instead of trying to massage inputs to be properly formatted.

LukeWood: added todo

keras_cv/losses/iou_loss.py
LukeWood: Good point, updated to conform

LukeWood:
/gcbrun/gcbrun

bhack:
> Yeah I'd also prefer a dataclass to an extension type for now, but I think we should revisit.  This is definitely an improvement over the concatted box format.

What kind of data type dataclass support?

bhack:
I don't know if this assumption it is still valid today https://stackoverflow.com/questions/73446860/can-tensorflows-tf-function-be-used-with-methods-of-dataclasses

LukeWood:
> I don't know if this assumption is still valid today https://stackoverflow.com/questions/73446860/can-tensorflows-tf-function-be-used-with-methods-of-dataclasses

@bhack 
you still get that same error with Dataclasses

bhack:
> @bhack
you still get that same error with Dataclasses

But ExtensionType work in graph mode.

LukeWood:
> > @bhack
> > you still get that same error with Dataclasses
> 
> But ExtensionType work in graph mode.

ah yes - ExtensionTypes may be the right fix

PR #1124, opened 2022-12-12
[Backbones] Update basic_training to support weight decay and ema

ianstenbit:
These are both Keras 2.11 features, so I've also updated the script to reference this dependency.


examples/training/classification/imagenet/basic_training.py
```diff
@@ -74,6 +75,11 @@
     False,
     "Whether or not to use FP16 mixed precision for training.",
 )
+flags.DEFINE_boolean(
```
tanzhenyu, line 78: wanna also fix the random generator shape mismatch in this PR?

ianstenbit: Good call -- done

tanzhenyu:
Doesn't look like code formatting is related to the PR. skipping the test

Issue #1052, opened 2022-11-24
StableDiffusion v2.0


LukeWood:
We should also request Gus conversion script, and add it to the repo.

mfidabel:
Would like to help on this feature also

kamalkraj:
Interested

tanzhenyu:
Closed as 2.1 is supported

PR #1130, opened 2022-12-13
Add support for Stable Diffusion 2.1

Jobayer-000:
# What does this PR do?
This PR configures text encoder and diffusion model for Stable Diffusion v2 support and adds sdv2.1 weights for the corresponding models.
Updates:
changed f_c to fully_connected,
use_q_gelu to use_quick_gelu,
version to weights; weights as a dictionary containing file origin and file_hash so that other weights version can be used externally by changing the config. SD2.0 for example.
and defaults to version v2.

Updates:
-Separated text encoder and diffusion model as TextEncoderV2 and DiffusionModelV2.
-No version argument but separate StableDiffusionV2 class as a subclass of StableDiffusion .
 
Updates:
- changed importing in stable_diffusion.__init__ file  the way keras_cv do. Formatting error should resolve now.
- I thought that it will be better if we have a StableDIffusionBase class and StableDiffusion, StableDiffusionV2 subclassing it. 

## Before submitting
      Pull Request section?
      to it if that's the case.https://github.com/keras-team/keras-cv/issues/1052

## Who can review?


keras_cv/models/stable_diffusion/stable_diffusion.py
```diff
@@ -35,8 +35,53 @@
...
+            'max_length': 77
+            },
+        'diffusion_model': {
+            'model_channels': 320,
+            'channel_mult': [ 1, 2, 4, 4 ],
+            'num_heads': 8,
+            'head_dim': -1,
+            'context_length':768,
+            'max_length': 77,
+            'f_c': False,
```
ianstenbit, line 57: Maybe let's rename f_c to something a bit more descriptive?

ianstenbit (changes requested):
Thank you for the PR -- this is **so awesome**! 🎉 🎊 

I've left a few quick comments, and I'm going to sync with @LukeWood about what is the right way to handle the `config`.

We want to make sure that custom workflows like TextualInversion still work correctly without having to wrangle with the config too much, so that might mean changing the config structure a bit.

In the meantime, do you have a Colab or example script using SD V2 after this change?

bhack:
Are we still missing weights conversion scripts?

bhack:
> In the meantime, do you have a Colab or example script using SD V2 after this change?

Do we have an inference tests using the same seed/input with pre-computed output from the original implementation?

ianstenbit:
Here's a colab showing 2.1 working with this PR:
https://colab.sandbox.google.com/gist/ianstenbit/ce4a1c1f4230db23a8185486cc1c9128/sd-2-1.ipynb

ianstenbit:
> > In the meantime, do you have a Colab or example script using SD V2 after this change?
> 
> Do we have an inference tests using the same seed/input with pre-computed output from the original implementation?

I am going to do some A/B benchmarking in a Colab and I'll share it here

keras_cv/models/stable_diffusion/text_encoder.py
```diff
@@ -18,47 +18,59 @@
...
-                file_hash="4789e63e07c0e54d6a34a29b45ce81ece27060c499a709d556c7755b42bb0dc4",
-            )
+            if config['version']=='v1':
+                text_encoder_weights_fpath = keras.utils.get_file(
+                    origin="https://huggingface.co/fchollet/stable-diffusion/resolve/main/kcv_encoder.h5",
+                    file_hash="4789e63e07c0e54d6a34a29b45ce81ece27060c499a709d556c7755b42bb0dc4",
+                )
+            else:
+                text_encoder_weights_fpath = keras.utils.get_file(
+                    origin="https://huggingface.co/Jobayer/stable_diffusion_v2/resolve/main/text_encoder_v2_1.h5",
```
tanzhenyu, line 44: Does OpenClip share the same model architecture?

tanzhenyu:
> Are we still missing weights conversion scripts?

Good point, I think we should add that in this PR

keras_cv/models/stable_diffusion/stable_diffusion.py
```diff
@@ -87,14 +132,15 @@ def __init__(
         self,
         img_height=512,
         img_width=512,
+        version='v1',
```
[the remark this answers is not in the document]
LukeWood: actually, because this is a new architecture we might want to make this a different model:

```
class StableDiffusionv2
```

That's what we do for ResNet, MobileNet, etc

keras_cv/models/stable_diffusion/text_encoder.py
```diff
@@ -67,7 +79,7 @@ def call(self, inputs):
         residual = x
         x = self.layer_norm2(x)
         x = self.fc1(x)
-        x = x * tf.sigmoid(x * 1.702)  # Quick gelu
+        x = self.gelu(x)
```
LukeWood, line 82: looks like a behavioral change that will break SD 1.0, right?

keras_cv/models/stable_diffusion/text_encoder.py
```diff
@@ -18,47 +18,59 @@
...
         tokens = self.token_embedding(tokens)
         positions = self.position_embedding(positions)
         return tokens + positions
 
+def quick_gelu(x):
+    return x * tf.sigmoid(x * 1.702)
 
 class CLIPEncoderLayer(keras.layers.Layer):
-    def __init__(self, **kwargs):
+    def __init__(self, embed_dim, num_heads, use_q_gelu=True, **kwargs):
```
LukeWood, line 66: use_q_gelu => use_quick_gelu

keras_cv/models/stable_diffusion/text_encoder.py
LukeWood: yeah it is open clip

keras_cv/models/stable_diffusion/text_encoder.py
```diff
@@ -18,47 +18,59 @@
...
+                )(x)
         embedded = keras.layers.LayerNormalization(epsilon=1e-5)(x)
         super().__init__([tokens, positions], embedded, name=name)
 
         if download_weights:
-            text_encoder_weights_fpath = keras.utils.get_file(
-                origin="https://huggingface.co/fchollet/stable-diffusion/resolve/main/kcv_encoder.h5",
-                file_hash="4789e63e07c0e54d6a34a29b45ce81ece27060c499a709d556c7755b42bb0dc4",
-            )
+            if config['version']=='v1':
```
LukeWood, line 37: I think we should prefer a `weights` argument

keras_cv/models/stable_diffusion/text_encoder.py
Jobayer-000: I didn't found any issue running SD 1.0 in my case.

ianstenbit: Yeah this shouldn't be an issue, since in the SD 1 case "gelu" is set to `x * tf.sigmoid(x * 1.702)`

ianstenbit:
Hey @Jobayer-000 -- thanks again for this awesome contribution!

We think it would be best to separate out StableDiffusion 2.1 as a separate class called StableDiffusionV2 instead of adding a `version` parameter to the existing StableDiffusion class.

Then, instead of having TextEncoder + DiffusionModel accept a config, we'd just have a separate TextEncoderV2 and a DiffusionModelV2.

Would you be willing to update your PR to switch to this approach?

If not, I would be happy to send a few commits to do the refactoring. Just let me know :)

keras_cv/models/stable_diffusion/stable_diffusion.py
```diff
@@ -508,3 +484,81 @@ def _get_initial_diffusion_noise(self, batch_size, seed):
...
+        img_width: Width of the images to generate, in pixel. Note that only
+            multiples of 128 are supported; the value provided will be rounded
+            to the nearest valid value. Default: 512.
+        jit_compile: Whether to compile the underlying models to XLA.
+            This can lead to a significant speedup on some systems. Default: False.
+    Example:
+    ```python
+    from keras_cv.models import StableDiffusion
+    from PIL import Image
+    model = StableDiffusion(img_height=512, img_width=512, jit_compile=True)
```
ianstenbit, line 512: In the V2 example let's say `model = StableDiffusionV2`

keras_cv/models/stable_diffusion/stable_diffusion.py
```diff
@@ -31,25 +28,22 @@
...
-from keras_cv.models.stable_diffusion.text_encoder import TextEncoder
+from keras_cv.models.stable_diffusion.text_encoder import TextEncoder, TextEncoderV2
 
 MAX_PROMPT_LENGTH = 77
 
 
 class StableDiffusion:
     """Keras implementation of Stable Diffusion.
-
     Note that the StableDiffusion API, as well as the APIs of the sub-components
```
ianstenbit, line 40: It looks like your auto-formatter has removed all of these blank lines in the docs -- would you mind re-adding them?

ianstenbit (approved):
This looks good to me!

I've left a few minor comments to address before we merge.
Thank you!

keras_cv/models/stable_diffusion/diffusion_model.py
```diff
@@ -151,23 +240,29 @@ def call(self, inputs):
...
 
 class SpatialTransformer(keras.layers.Layer):
-    def __init__(self, num_heads, head_size, **kwargs):
+    def __init__(self, num_heads, head_size, fully_connected=False, **kwargs):
         super().__init__(**kwargs)
         self.norm = GroupNormalization(epsilon=1e-5)
         channels = num_heads * head_size
-        self.conv1 = PaddedConv2D(num_heads * head_size, 1)
+        if fully_connected:
+            self.proj1 = keras.layers.Dense(num_heads * head_size) # sdv2 uses dense layer rather than conv
```
tanzhenyu, line 248: nit: stable diffusion V2 instead of sdv2

keras_cv/models/stable_diffusion/stable_diffusion.py
```diff
@@ -280,7 +279,7 @@ def inpaint(
     ):
         """Inpaints a masked section of the provided image based on the provided prompt.
         Note that this currently does not support mixed precision.
-
+
```
tanzhenyu, line 282: needs formating

keras_cv/models/stable_diffusion/text_encoder.py
```diff
@@ -49,16 +69,18 @@ def call(self, inputs):
         positions = self.position_embedding(positions)
         return tokens + positions
 
+def quick_gelu(x):
+    return x * tf.sigmoid(x * 1.702)
 
 class CLIPEncoderLayer(keras.layers.Layer):
-    def __init__(self, **kwargs):
+    def __init__(self, embed_dim, num_heads, use_quick_gelu=True, **kwargs):
```
tanzhenyu, line 76: instead of `use_quick_gelu`, can we pass in an activation function, and defaults to None?

keras_cv/models/stable_diffusion/text_encoder.py
```diff
@@ -37,6 +37,26 @@ def __init__(self, max_length, vocab_size=49408, name=None, download_weights=Tru
...
+        x = CLIPEmbedding(vocab_size, 1024, max_length)([tokens, positions])
+        for _ in range(23):
+            x = CLIPEncoderLayer(1024, 16, use_quick_gelu=False)(x)
+        embedded = keras.layers.LayerNormalization(epsilon=1e-5)(x)
+        super().__init__([tokens, positions], embedded, name=name)
+
+        if download_weights:
+            text_encoder_weights_fpath = keras.utils.get_file(
+                origin="https://huggingface.co/Jobayer/stable_diffusion_v2/resolve/main/text_encoder_v2_1.h5",
+                file_hash="985002e68704e1c5c3549de332218e99c5b9b745db7171d5f31fcd9a6089f25b"
```
tanzhenyu, line 55: need an export weight from our public GCP bucket

tanzhenyu (dismissed):
Thanks for the contribution!

keras_cv/models/stable_diffusion/text_encoder.py
ianstenbit: I think instead of a GCS bucket we should just use HF weights to keep this separate from our corp infra (like what Francois' did with the original weights).

I will make a clone of these weights in a HF account that we own.

keras_cv/models/stable_diffusion/text_encoder.py
Jobayer-000: Updated. Can you see if it is ok now?

ianstenbit (approved):
I've re-run my A/B testing Colab after your changes and it still looks good.

Pending changing the source location for the weights downloads (and our formatting/testing passing), I will merge this

keras_cv/models/stable_diffusion/text_encoder.py
Jobayer-000: Updated.

LukeWood (dismissed):
LGTM, but it looks like the formatter needs to be run

ianstenbit:
@Jobayer-000 can you please run our autoformatter (located at shell/format.sh)?
Then I can merge

bhack:
> > Are we still missing weights conversion scripts?
> 
> Good point, I think we should add that in this PR

What about this?

ianstenbit:
> > > Are we still missing weights conversion scripts?
> > 
> > 
> > Good point, I think we should add that in this PR
> 
> What about this?

Can we do this in a follow-up PR @Jobayer-000

PR #1135, opened 2022-12-13
Fix bugs in frustum-based data augmentations

lengzq:
Use boolean mask to select valid points, then choose a frustum center from the valid points. Previous implementation will random choose a center from all points, which include none valid points.


tanzhenyu (commented):
add you add some test to catch it?

ianstenbit (approved):

tanzhenyu (approved):

tanzhenyu:
manual merging since gcp testing is broken

PR #1136, opened 2022-12-13
Migrate RandomZoom

divyashreepathihalli:


divyashreepathihalli:
random_zoom demo

keras_cv/layers/__init__.py
```diff
@@ -85,9 +83,11 @@
 from keras_cv.layers.preprocessing.random_gaussian_blur import RandomGaussianBlur
 from keras_cv.layers.preprocessing.random_hue import RandomHue
 from keras_cv.layers.preprocessing.random_jpeg_quality import RandomJpegQuality
+from keras_cv.layers.preprocessing.random_rotation import RandomRotation
```
ianstenbit, line 86: Was this missed in a previous PR?

keras_cv/layers/preprocessing/random_zoom.py
```diff
@@ -0,0 +1,235 @@
...
+from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
+    BaseImageAugmentationLayer,
+)
+from keras_cv.utils import preprocessing
+
+# In order to support both unbatched and batched inputs, the horizontal
+# and verticle axis is reverse indexed
+H_AXIS = -3
+W_AXIS = -2
+IMAGES = "images"
```
ianstenbit, line 28: Looks like these two strong constants are unused

keras_cv/layers/preprocessing/random_zoom.py
```diff
@@ -0,0 +1,235 @@
...
+
+@tf.keras.utils.register_keras_serializable(package="keras_cv")
+class RandomZoom(BaseImageAugmentationLayer):
+    """A preprocessing layer which randomly zooms images during training.
+
+    This layer will randomly zoom in or out on each axis of an image
+    independently, filling empty space according to `fill_mode`.
+
+    Input pixel values can be of any range (e.g. `[0., 1.)` or `[0, 255]`) and
+    of interger or floating point dtype. By default, the layer will output
```
ianstenbit, line 40: s/interger/integer

keras_cv/layers/preprocessing/random_zoom.py
```diff
@@ -0,0 +1,235 @@
...
+
+    Args:
+      height_factor: a float represented as fraction of value, or a tuple of
+        size 2 representing lower and upper bound for zooming vertically. When
+        represented as a single float, this value is used for both the upper and
+        lower bound. A positive value means zooming out, while a negative value
+        means zooming in. For instance, `height_factor=(0.2, 0.3)` result in an
+        output zoomed out by a random amount in the range `[+20%, +30%]`.
+        `height_factor=(-0.3, -0.2)` result in an output zoomed in by a random
+        amount in the range `[+20%, +30%]`.
```
ianstenbit, line 51: Seems like this range is wrong -- should it be [-30%, -20%]?

keras_cv/layers/preprocessing/random_zoom.py
```diff
@@ -0,0 +1,235 @@
...
+            interpolation=self.interpolation,
+        )
+        output = tf.squeeze(output, 0)
+        output.set_shape(original_shape)
+        return output
+
+    def augment_label(self, label, transformation, **kwargs):
+        return label
+
+    def compute_output_shape(self, input_shape):
```
ianstenbit, line 180: Don't think we need this

keras_cv/layers/preprocessing/random_zoom_test.py
```diff
@@ -0,0 +1,152 @@
...
+
+import numpy as np
+import tensorflow as tf
+from absl.testing import parameterized
+
+from keras_cv.layers.preprocessing.random_zoom import RandomZoom
+
+
+class RandomZoomTest(tf.test.TestCase, parameterized.TestCase):
+    def _run_test(self, height_factor, width_factor):
```
ianstenbit, line 23: Same as last PR -- let's rename this test_output_shape and parameterize it directly instead of calling it from other parameterized tests

keras_cv/layers/preprocessing/random_zoom.py
```diff
@@ -0,0 +1,235 @@
...
+
+    This layer will randomly zoom in or out on each axis of an image
+    independently, filling empty space according to `fill_mode`.
+
+    Input pixel values can be of any range (e.g. `[0., 1.)` or `[0, 255]`) and
+    of interger or floating point dtype. By default, the layer will output
+    floats.
+
+    Args:
+      height_factor: a float represented as fraction of value, or a tuple of
```
ianstenbit, line 44: Does passing `height_factor=0` result in no vertical zoom? If so, let's document that.

ianstenbit (changes requested):
Thanks!

keras_cv/layers/__init__.py
divyashreepathihalli: yeah, this was missed.

keras_cv/layers/preprocessing/random_zoom.py
divyashreepathihalli: done

keras_cv/layers/preprocessing/random_zoom.py
divyashreepathihalli: done

keras_cv/layers/preprocessing/random_zoom_test.py
divyashreepathihalli: done

keras_cv/layers/preprocessing/random_zoom.py
divyashreepathihalli: if height_factor is 0 and width_factor=None, then result images will have no zoom. I have added this to doc string.

keras_cv/layers/preprocessing/random_zoom.py
ianstenbit: Our BaseImageAugmentationLayer defines `_compute_output_signature` in lieu of the method `compute_output_signature` in base `keras.Layer`, and uses that instead, so in our KPLs we define `compute_image_signature` to define this, and we only need to do that if the output image shape is different from the input shape

ianstenbit (approved):
LGTM after my comment

Thank you!

keras_cv/layers/preprocessing/random_zoom.py
divyashreepathihalli: Got it! deleted compute_output_shape()

Issue #895, opened 2022-10-10
RetinaNet TPU compatibility


LukeWood:
With evaluate_train_time_metrics=False, .fit() training works, but not validation. Validation will try to XLA-compile the NMS layer which will fail. This is whether COCO metrics are defined as metrics in .compile() or not.

This can be fixed by checking if the metrics list is empty or not, and only tracing NMS when they are in fact empty.

bhack:
As I've contributed the `tf.bincout` bridge coverage in TF some of these issues could be solved by using python range instead of `tf.range`:
https://github.com/tensorflow/tensorflow/issues/56769

See @cheshire
https://github.com/tensorflow/tensorflow/commit/56b2e5a77fda3fa573c2eec86d38a92379508118

The residual issues is that, also with this change, we still have:

`pytest keras_cv/metrics/coco/mean_average_precision_test.py`

`bincount_op.cc:52 : UNIMPLEMENTED: Dynamic reshape is not implemented.
        while evaluating input 1 of Bincount operator as a compile-time constant.`

As the bincount input 1 is a `tf.gather_nd`. I don't know if @cheshire could comment on this:
https://github.com/keras-team/keras-cv/blob/e36de54a3b2ee82bed67858d48cad62409f50a64/keras_cv/metrics/coco/mean_average_precision.py#L250-L266

Instead in recall we have other XLA issues issues (@cheshire):
`pytest keras_cv/metrics/coco/recall_test.py`

```python
_REQUIRES failed at xla_ops.cc:296 : UNIMPLEMENTED: Dynamic degenerated dimension that's not most-minor nor most-major is not supported %reshape.37 = f32[1,1]{1,0} reshape(f32[1,1,1]{2,1,0} %slice.16), metadata={op_type="Unpack" op_name="while_1/while/unstack_1" source_file="/workspaces/keras-cv/keras_cv/bounding_box/iou.py" source_line=75}
```

bhack:
For `UNIMPLEMENTED: Dynamic reshape is not implemented` we can track https://github.com/tensorflow/tensorflow/issues/58051

PR #1148, opened 2022-12-15
(Step 4/5) Add KPLs to RetinaNet training script

LukeWood:
Notes:

- The checked in version of this script doesn't work at all
- This scores the same MaP as before, >0.32 every run
- This is much more readable
- This adds back in the COCO metrics, these are really useful for monitoring training on a GPU device.  We can disable them on TPU if we want, but I'd like to keep them for prototyping until we hit our final state
- I've removed the COCO evaluator callback until we move everything to native box format

---

Action items after this PR:

- [ ] swap to native box format
- [ ] add back the COCO callback
- [ ] fix the label encoder metric in MirroredStrategy mode

### DIFF
--- a/examples/training/object_detection/pascal_voc/retina_net.py
+++ b/examples/training/object_detection/pascal_voc/retina_net.py
@@ -169,12 +169,10 @@ model.compile(
     box_loss="smoothl1",
     optimizer=optimizer,
     metrics=[
-        keras_cv.metrics.COCORecall(
-            bounding_box_format="xywh", class_ids=range(20)
-        ),
+        keras_cv.metrics.COCORecall(bounding_box_format="xywh", class_ids=range(20)),
         keras_cv.metrics.COCOMeanAveragePrecision(
             bounding_box_format="xywh", class_ids=range(20)
-        )
+        ),
     ],
 )
 
--- a/examples/training/object_detection/pascal_voc/retina_net.py
+++ b/examples/training/object_detection/pascal_voc/retina_net.py
@@ -99,7 +99,6 @@ train_ds = train_ds.apply(
     tf.data.experimental.dense_to_ragged_batch(GLOBAL_BATCH_SIZE, drop_remainder=True)
 )
 
-train_ds = train_ds.map(pad_fn, num_parallel_calls=tf.data.AUTOTUNE)
 train_ds = train_ds.shuffle(8 * strategy.num_replicas_in_sync)
 train_ds = train_ds.prefetch(tf.data.AUTOTUNE)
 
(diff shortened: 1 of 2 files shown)


tanzhenyu (commented):
Interesting...I think COCOMetrics doesn't work on GPU as well (What's unique about TPU here?)

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -363,31 +170,33 @@ def pad_fn(image, boxes):
...
     keras.callbacks.ReduceLROnPlateau(patience=5),
     keras.callbacks.EarlyStopping(patience=10),
     keras.callbacks.ModelCheckpoint(CHECKPOINT_PATH, save_weights_only=True),
-    PyCOCOCallback(eval_ds.map(convert_to_dict), "xywh"),
 ]
 
 history = model.fit(
     train_ds,
-    epochs=100,
+    validation_data=eval_ds.take(20),
```
ianstenbit, line 199: Is the call to `.take` here intentional? I'd think we would want the whole eval DS?

ianstenbit:
> Interesting...I think COCOMetrics doesn't work on GPU as well (What's unique about TPU here?)

IIRC the root cause of COCOMetrics not working on TPU was the NMS issue. So in that case I'd expect it to work fine on GPU

ianstenbit (approved):
Looks a lot nicer!

bhack:
Wasn't it TPU only cause we cannot XLA compile for https://github.com/tensorflow/tensorflow/issues/56769#issuecomment-1184793361?

bhack:
Last time was https://github.com/keras-team/keras-cv/issues/895#issuecomment-1273430678

If was NMS it is TPU compatible with mlir bridge as we have tested in another ticket.

bhack:
The MMS test was at https://github.com/keras-team/keras-cv/issues/895#issuecomment-1273430678

tanzhenyu:
> > Interesting...I think COCOMetrics doesn't work on GPU as well (What's unique about TPU here?)
> 
> IIRC the root cause of COCOMetrics not working on TPU was the NMS issue. So in that case I'd expect it to work fine on GPU

That's a graph building time error. I mean realistically can TensorArray even work on GPU?

tanzhenyu:
> > > > Interesting...I think COCOMetrics doesn't work on GPU as well (What's unique about TPU here?)
> > > 
> > > 
> > > IIRC the root cause of COCOMetrics not working on TPU was the NMS issue. So in that case I'd expect it to work fine on GPU
> > 
> > 
> > That's a graph building time error. I mean realistically can TensorArray even work on GPU?
> 
> yeah, COCO metrics work on GPU. the only thing broken on GPU is the label encoder's % boxes caputured encoding tracker

I see. I haven't closely follow the implementation. Shared a colab link with you, it seems it doesn't work when we force everything to be placed on GPU.

LukeWood:
Lets keep this lightweight for now, I am mainly using this as a way to ensure we still get SOTA MaP.  When I migrate to dictionary format across all models I'll migrate this to use the callback.

bhack:
> Shared a colab link with you, it seems it doesn't work when we force everything to be placed on GPU.

Can you share the colab?

PR #1146, opened 2022-12-15
Separating MBConvBlock, FusedMBConvBlock and Refactoring EfficientNetV2

DavidLandup0:
# What does this PR do?

There's a design issue that's not aligned with KCV's principles. MBConv blocks, as well as well as Fused MBConv blocks are general-purpose blocks, introduced with MobileNets and then re-used in practically all mobile-oriented/efficient-desired architectures since. Currently, we only have them as part of EfficientNets, in which they're proprietary methods.

If we don't separate them, we'll accumulate tech debt that'll be only harder to recover from down the line.

Since MaxViT (currently WIP), MobileNets and various other future architectures will use them (and we'd want to expose these to the user as well as a common building block), I propose that we separate them into standalone layers, which is this PR.

I've refactored EfficientNetV2 to use layers instead of methods and they're identical in param counts, as validation that they behave the same. One issue to figure out is:

- Original effnets list each layer within each MBConvBlock
- New MBConvBlock layer encapsulates them within a single `MBConvBlock`

Because of this, the total number of layers differs (151 vs 26). We can:

1. Expose all layers within the `MBConvBlock` and `FusedMBConvBlock` and add them to the EffNet while it's being built (so as to be able to load weights as they are right now)
2. Keep layers encapsulated and port our own weights (or re-train effnets)

Here's the difference:

```
input_3 (InputLayer)        [(None, 224, 224, 3)]     0         
                                                                 
 rescaling_2 (Rescaling)     (None, 224, 224, 3)       0         
                                                                 
 stem_conv (Conv2D)          (None, 112, 112, 32)      864       
                                                                 
 stem_bn (BatchNormalization  (None, 112, 112, 32)     128       
 )                                                               
                                                                 
 stem_activation (Activation  (None, 112, 112, 32)     0         
 )                                                               
                                                                 
 block1a_ (FusedMBConvBlock)  (None, 112, 112, 16)     576       
                                                                 
 block2a_ (FusedMBConvBlock)  (None, 112, 112, 32)     3456      
...
```

VS:

```
input_1 (InputLayer)           [(None, 224, 224, 3  0           []                               
                                )]                                                                
                                                                                                  
 rescaling (Rescaling)          (None, 224, 224, 3)  0           ['input_1[0][0]']                
                                                                                                  
 stem_conv (Conv2D)             (None, 112, 112, 32  864         ['rescaling[0][0]']              
                                )                                                                 
                                                                                                  
[...]

PR #1151, opened 2022-12-16
Switch to new EfficientNet weights (post-PR #1146)

ianstenbit:
Draft for now.

Upon submitting this, I will also update the file names in GCS so that the new weights are what's loaded when loading EfficientNet weights.

Note that this _will_ break users who have an older version of KerasCV installed (before this PR) and are using EfficientNet weights, but from what I can tell based on our model weights download stats, that number is ~0


ianstenbit:
@DavidLandup0 @tanzhenyu 
Since #1146 is merged, we need to update our weights.
I have renamed the weights on GCS, and this PR fixes the hashes

tanzhenyu (approved):

PR #1152, opened 2022-12-16
[RetinaNet Refactor] Fix bug and Reuse box delta encoding and decoding logic.

tanzhenyu:
Fixing bug -- The original implementation uses "XYWH" for encoding and decoding. The paper uses "Center_XYWH" for encoding and decoding
The result is ~34.0mAP at 25 epochs.

### DIFF
--- a/keras_cv/bounding_box/converters.py
+++ b/keras_cv/bounding_box/converters.py
@@ -15,6 +15,7 @@
 
 from typing import List
 from typing import Optional
+from typing import Union
 
 import tensorflow as tf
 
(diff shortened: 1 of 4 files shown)


keras_cv/bounding_box/converters.py
```diff
@@ -30,11 +31,16 @@ def _encode_box_to_deltas(
     boxes: tf.Tensor,
     anchor_format: str,
     box_format: str,
-    variance: Optional[List[float]] = None,
+    variance: Optional[Union[List[float], tf.Tensor]] = None,
```
LukeWood, line 34: is this type annotation useful still or does it become overly burdensome?  No strong feelings, just might be worth dropping it

LukeWood (approved):
LGTM overall, nice that it boosts the MaP a bit!

keras_cv/bounding_box/converters.py
tanzhenyu: To me, it feels burdensome. But whenever I read someone else's code with type annotation, I still feel it's pretty useful. So it's really a double-edged sword IMHO

ianstenbit (approved):
Thanks!

PR #1154, opened 2022-12-16
Revert "(Step 4/5) Add KPLs to RetinaNet training script (#1148)"

tanzhenyu:
The original PR breaks TPU training by introducing ragged tensor

### DIFF
--- a/examples/training/object_detection/pascal_voc/retina_net.py
+++ b/examples/training/object_detection/pascal_voc/retina_net.py
@@ -27,13 +27,38 @@ from absl import flags
 from tensorflow import keras
 
 import keras_cv
-from keras_cv import layers
+from keras_cv.callbacks import PyCOCOCallback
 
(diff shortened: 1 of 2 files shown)


ianstenbit (commented):
Does it make sense to fix-forward here since we do want to switch to the KPLs? Or is that a lot more work?

LukeWood:
For what it’s worth I am re-adding the COCO callback and removing the metrics in the dict box PR.

Fine to proceed either way I will just reintroduce the changes in that PR if that’s easiest

tanzhenyu:
> For what it’s worth I am re-adding the COCO callback and removing the metrics in the dict box PR.
> 
> Fine to proceed either way I will just reintroduce the changes in that PR if that’s easiest

Oh in that case, feel free to go ahead with a forward fix then, I will close this PR

LukeWood (approved):

PR #1156, opened 2022-12-17
[Refactor RetinaNet] Use label dict in label encoder and base model.

tanzhenyu:


keras_cv/models/object_detection/object_detection_base_model.py
```diff
@@ -98,22 +98,20 @@ def evaluate(
...
-            y_training_target,
+        box_targets, class_targets = self.label_encoder(x, y)
+        box_targets = bounding_box.convert_format(
+            box_targets,
             source=self.label_encoder.bounding_box_format,
             target=self.bounding_box_format,
             images=x,
         )
-        return x, (y_for_metrics, y_training_target)
+        return x, {"gt_boxes": box_targets, "gt_classes": class_targets}
```
LukeWood, line 114: can we begin changing the names from "gt_" to just "boxes" and "classes"?

LukeWood, line 114: This is actually a deviation from the current behavior, we won't be able to use the COCO metrics anymore. 

This is fine if the callback works

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -457,11 +457,12 @@ def _track_regularization(self):
...
     def train_step(self, data):
         x, y = data
-        y_for_metrics, y_training_target = y
+        gt_boxes = y["gt_boxes"]
+        gt_classes = y["gt_classes"]
 
         with tf.GradientTape() as tape:
             y_pred = self(x, training=True)
-            loss = self._backward(y_training_target, y_pred)
+            loss = self._backward(gt_boxes, gt_classes, y_pred)
```
LukeWood, line 465: should we just entirely remove the metrics now?   We CANT evaluate them if we don't pass the raw boxes forward.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -471,18 +472,17 @@ def train_step(self, data):
...
         x, y = data
-        y_for_metrics, y_training_target = y
+        gt_boxes = y["gt_boxes"]
+        gt_classes = y["gt_classes"]
         y_pred = self(x, training=False)
-        _ = self._backward(y_training_target, y_pred)
+        _ = self._backward(gt_boxes, gt_classes, y_pred)
 
         # Early exit for no custom metrics
         if not self._includes_custom_metrics:
```
LukeWood, line 481: can remove this block if we're never decoding predictions

keras_cv/models/object_detection/object_detection_base_model.py
tanzhenyu: Done.

tanzhenyu: Yeah the callback works pretty well, and I really like the result output from pycoco

keras_cv/models/object_detection/retina_net/retina_net.py
tanzhenyu: Done

keras_cv/models/object_detection/retina_net/retina_net.py
tanzhenyu: Done

LukeWood (approved):

Issue #917, opened 2021-08-28
Add `ResNeXt [50, 101]` to keras.applications


innat:
**System information**.

TensorFlow version (you are using): 2.5
Are you willing to contribute it (Yes/No) : Yes

**Describe the feature and the current behavior/state**.

`ResNeXt` is a well-known classification model that is weirdly missing from the `keras.applications`. The idea that is presented [here](https://arxiv.org/abs/1611.05431) is great and I think it's fit for the model member in the keras package. 

**Will this change the current api? How?**
Yes. 

```
from tensorflow.keras.applications.resnext50 import ResNeXt50
from tensorflow.keras.applications.resnext101 import ResNeXt101
```

**Who will benefit from this feature?**
ML engineers and researcher who uses `tf.keras`. 

**Others**
Others implementation: https://github.com/qubvel/classification_models

PR #1157, opened 2022-12-17
[WIP] Adding ResNeXt

### LINKED ISSUE #905: ResNeXt Architecture
**Short Description**

ResNeXt is a pretty commonly used arch, had ~8k citations, and is oftentimes shown in comparison tables.
`torchvision` has ResNeXt as well. Might be a good fit.

**Other Information**

Signing up to work on it as well.

DavidLandup0:
# What does this PR do?

Adding the first draft for the ResNeXt architecture. Not fully finished, but opening it for a discussion.

# Open Questions

There are a couple of questions we'd want to figure out the answers to before we continue and finalize the proposal:

- ResNeXt is really close to ResNet. IMO, it's a special case of a ResNet. 95% of the code is the same - do we update the ResNetV2 class to include arguments for ResNeXt and add a few conditionals that make the difference?
- Names are ugly - ResNeXt50_32x4d, for example. There's also a ResNeXt50_32x8d so we can't collapse them by removing `Nd` from the end. `torchvision` uses the full names, but they don't look very readable IMO. Any suggestions?

# Todo
- Tests, docstrings, etc. No need to do these and other overhead until we figure out the questions above.

Fixes #905 and #917

## Before submitting

      Pull Request section?
      to it if that's the case. #905
- [] Did you write any new necessary tests?
- [] If this adds a new model, can you run a few training steps on TPU in Colab to ensure that no XLA incompatible OP are used?

## Who can review?

@bhack @tanzhenyu @LukeWood


IMvision12:
@tanzhenyu could review this PR, need answers to the given questions from David so that we can move forward with the PR and tests, and improve the code more...

bhack:
Do we have converted weights for these or we just plain to train from scratch?

tanzhenyu:
Thanks for the draft PR @DavidLandup0 @IMvision12 !
Re your questions
1. Yeah we should aim for reusability. However, do you know which part share the same code path with ResNet, and which part is unique to ResNeXt? That helps us judge how to run the refactoring.
2. Naming wise -- it's a little bit ugly, but I'm ok with it, as we might need to provide another layer of abstraction for users to initiate their models, for example `keras_cv.models.from_pretrained('ResNext', cardinality=32, bottleneck_width=4, image_size=[320, 320, 3])` so we don't need to spend too much time thinking about naming

DavidLandup0:
> Do we have converted weights for these or we just plan to train from scratch?

Planned to train from scratch at one point (there's a fair bit of models in the family, and some other families might take precedence).

We can convert the weights if the team agrees it's the best choice, though.

DavidLandup0:
> Thanks for the draft PR @DavidLandup0 @IMvision12 ! Re your questions
> 
> 1. Yeah we should aim for reusability. However, do you know which part share the same code path with ResNet, and which part is unique to ResNeXt? That helps us judge how to run the refactoring.
> 2. Naming wise -- it's a little bit ugly, but I'm ok with it, as we might need to provide another layer of abstraction for users to initiate their models, for example `keras_cv.models.from_pretrained('ResNext', cardinality=32, bottleneck_width=4, image_size=[320, 320, 3])` so we don't need to spend too much time thinking about naming

Thanks for the feedback! In terms of differences - it's got the same `Basic`, `Block` and `Stack` structure. The main difference is the usage of `groups` in the bottleneck layer. The rest of the code is practically the same. You can think of ResNeXt as ResNet with a new bottleneck layer.

IMO, we can add the block to the ResNet class and do a conditional:

```
if block_type == 'resnext':
   D = ...
   width = ...
   block = resnext_block_fn(D, width)
else:
  block = resnet_block_fn()
```

`D` and `width` are ResNeXt-only (the cardinality/width for the `groups` argument). The `block_type` would come from the `MODEL_CONFIGS` in which `ResNet` models would have one, and `ResNeXt` models would have the other type. This collapses the entire ResNeXt implementation into a single block (~20-30 lines) and new initializer methods for the public API

WDYT? @tanzhenyu

DavidLandup0:
ResNetV2 already has an interchangeable block function:

https://github.com/keras-team/keras-cv/blob/c278d87395c61d7300a90e4684aee674ba465e29/keras_cv/models/resnet_v2.py#L262

We'd just add another option for `block_fn()` that's used based on the `MODEL_CONFIGS`

IMvision12:
I believe we should follow the `one model, one file` principle, which states that each model should have its own file. This policy is also followed by huggingface, which has distinct files for each model even if there are numerous variants of certain models, such as vit, maxvit, and levit which are variants of vision transformer.

Also, updating resnet block for resnext and importing it as `from keras_cv.models.resnet.resnext_101` is not appropriate, feels wired, even though resnext is from resnet family and totally a different model, so i feel we should create a different files for

1. ResNet
2. ResNext
3. WideResNet
4. SE-Nets

What do you think?
@DavidLandup0 @tanzhenyu

bhack:
I think that a resnet bottleneck block it could be a shared assets in the resent family if we find and internal interface to cover the family needs.

As we are often exposed to incremental research in our filed it is ok to looking for invariants in the modularization effort of the library.

LukeWood:
Closing due to inactivity!  Feel free to re-open and discuss in an issue if you'd like to carry this forward!

IMvision12:
@LukeWood @ianstenbit are we accepting new backbones, since I am ready to add some new backbones that perform really well on imagenet and are not available in kcv, and @ianstenbit previously stated that we would begin accepting new backbones after some time.

ianstenbit:
@martin-gorner is ResNeXt a backbone that we want to offer?

@IMvision12 I think for now we probably want to wait until we've finished the backbone migration to the new API before we add new ones. IIRC ResNeXt is probably something we want to offer, but I'll see what Martin thinks

PR #1166, opened 2022-12-19
Remove support for Python 3.7 to align with keras-nightly

sampathweb:
* Keras Nightly & TensorFlow Nightly have dropped support for Python 3.7 so it can support numpy 1.23+
* Align setup.py to be similar to `keras-nlp` and `keras`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ setup(
     author_email="keras-cv@google.com",
     license="Apache License 2.0",
     install_requires=["packaging", "absl-py", "regex"],
-    python_requires=">=3.7",
+    python_requires=">=3.8",
     extras_require={
         "tests": [
             "flake8",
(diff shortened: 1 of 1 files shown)


qlzh727 (approved):
Please wait for the final approval from zhenyu.

LukeWood:
Some thoughts:

The upgrade from 3.7->3.8 is nontrivial for many; most pre-configured solutions use 3.7 and many offer no way to upgrade it.  A prime example of this is VertexAI; it is a huge pain to upgrade the python version - so much so that I actually just use my own server instead.

With those in mind, what is the specific reason we should align these version requirements?  If a user can TF 2.10 with KerasCV 0.4.0 and python 3.7 with both, I believe thats a usability win.

bhack:
> The upgrade from 3.7->3.8 is nontrivial for many; most pre-configured solutions use 3.7 and many offer no way to upgrade it. A prime example of this is VertexAI; it is a huge pain to upgrade the python version - so much so that I actually just use my own server instead.
With those in mind, what is the specific reason we should align these version requirements? If a user can TF 2.10 with KerasCV 0.4.0 and python 3.7 with both, I believe thats a usability win.

As we have introduced c++ custom ops in the repo we need to define our new " compatibility" policy. 
See my description at https://github.com/keras-team/keras-cv/pull/1170

LukeWood:
For now closing due to merge conflicts and inactivity.   Thanks for the PR @sampathweb - and will keep an eye on compatibility with respect to custom ops!

PR #1167, opened 2022-12-20
Add an internal UNet implementation

ianstenbit:
The modular UNet API that I had previously designed doesn't support the exact use case we have for our 3D OD model, so I started from scratch with an ultra-modular (and potentially very useless because it's so generic?) UNet design.

The default implementation of the blocks matches what we want for a starter CenterPillar model

### DIFF
--- a/keras_cv/models/__internal__/unet.py
+++ b/keras_cv/models/__internal__/unet.py
@@ -18,6 +18,11 @@ from tensorflow.keras import regularizers
 
 
 def Block(filters, downsample):
+    """A default block which serves as an example of the block interface.
+
+    This is the base block definition for a CenterPillar model.
+    """
+
     def apply(x):
         input_depth = x.shape.as_list()[-1]
         stride = 2 if downsample else 1
@@ -30,12 +35,9 @@ def Block(filters, downsample):
             stride,
             padding="same",
             kernel_initializer=initializers.VarianceScaling(),
-            kernel_regularizer=regularizers.L2(l2=1e-4),
         )(x)
         x = layers.BatchNormalization(
             synchronized=True,
-            beta_regularizer=regularizers.L2(l2=1e-8),
-            gamma_regularizer=regularizers.L2(l2=1e-8),
         )(x)
         x = layers.ReLU()(x)
 
@@ -45,12 +47,9 @@ def Block(filters, downsample):
             1,
             padding="same",
             kernel_initializer=initializers.VarianceScaling(),
-            kernel_regularizer=regularizers.L2(l2=1e-4),
         )(x)
         x = layers.BatchNormalization(
             synchronized=True,
-            beta_regularizer=regularizers.L2(l2=1e-8),
-            gamma_regularizer=regularizers.L2(l2=1e-8),
         )(x)
 
         if downsample:
@@ -65,7 +64,6 @@ def Block(filters, downsample):
                 1,
                 padding="same",
                 kernel_initializer=initializers.VarianceScaling(),
-                kernel_regularizer=regularizers.L2(l2=1e-4),
             )(residual)
 
         x = x + residual
@@ -83,12 +81,9 @@ def SkipBlock(filters):
             1,
             1,
             kernel_initializer=initializers.VarianceScaling(),
-            kernel_regularizer=regularizers.L2(l2=1e-4),
         )(x)
         x = layers.BatchNormalization(
             synchronized=True,
-            beta_regularizer=regularizers.L2(l2=1e-8),
-            gamma_regularizer=regularizers.L2(l2=1e-8),
         )(x)
         x = layers.ReLU()(x)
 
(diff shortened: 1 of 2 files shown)


keras_cv/models/__internal__/unet.py
```diff
@@ -0,0 +1,172 @@
...
+        down_blocks: a list of (filter_count, num_blocks) tuples indicating the
+            number of filters and sub-blocks in each down block
+        up_blocks: a list of filter counts, one for each up block
+        down_block: a downsampling block
+        up_block: an upsampling block
+    """
+
+    def apply(x):
+        skip_connections = []
+        for filters, num_blocks in down_blocks:
```
tanzhenyu, line 163: maybe comment here what does `filters` and `num_blocks` represent

keras_cv/models/__internal__/unet.py
```diff
@@ -0,0 +1,172 @@
...
+            x = Block(filters, downsample=False)(x)
+
+        return x
+
+    return apply
+
+
+def UpSampleBlock(filters):
+    def apply(x, lateral_input):
+        x = layers.Conv2DTranspose(
```
tanzhenyu, line 114: Is the UNET paper using UpSampling2D or Conv2DTranspose? Can this be configurable?

keras_cv/models/__internal__/unet.py
```diff
@@ -0,0 +1,172 @@
...
+            filters,
+            3,
+            2,
+            padding="same",
+            kernel_initializer=initializers.VarianceScaling(),
+            kernel_regularizer=regularizers.L2(l2=1e-4),
```
tanzhenyu, line 120: I think in this codebase we usually don't put regularization during modeling part, but instead in training part. This makes modeling more generic and users can tune those hyperparameters without touch model parameters.

keras_cv/models/__internal__/unet.py
```diff
@@ -0,0 +1,172 @@
...
+        )(x)
+        x = layers.BatchNormalization(
+            synchronized=True,
```
tanzhenyu, line 123: Maybe we should ask whether we need to make this configurable or not. @Guowang @lengzq

keras_cv/models/__internal__/unet.py
```diff
@@ -0,0 +1,172 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from tensorflow.keras import initializers
+from tensorflow.keras import layers
+from tensorflow.keras import regularizers
+
+
+def Block(filters, downsample):
```
tanzhenyu, line 20: Consider adding docstring (even though it's not public API)

tanzhenyu (commented):
Thanks for the PR!

keras_cv/models/__internal__/unet.py
lengzq: You meant synchronized batchnorm? Sure.

keras_cv/models/__internal__/unet.py
lengzq: Bilinear is slower than ConvTranspose on TPU. We could support bilinear up sampling once we verify the unet performance.

ianstenbit: The idea here is that this UpSampleBlock (as well as the DownSampleBlock) are specific to this UNet.

When this is a public API, we could example call `UNet(...., down_block=MyCustomDownBlock)`
In that future, these particular up/down blocks would be available as e.g. a CenterPillarDownBlock

keras_cv/models/__internal__/unet.py
ianstenbit: Does this need to be configured in the final OSS model?

If possible, I'd like to keep config to a minimum unless necessary

keras_cv/models/__internal__/unet.py
ianstenbit: Good call -- removing them throughout.

keras_cv/models/__internal__/unet.py
ianstenbit: done

keras_cv/models/__internal__/unet.py
```diff
@@ -0,0 +1,172 @@
...
+    All up and down blocks scale by a factor of two. Skip connections are
+    included.
+
+    All function parameters require curried functions as inputs which return a
+    function that acts on tensors as inputs.
+
+    Args:
+        down_blocks: a list of (filter_count, num_blocks) tuples indicating the
+            number of filters and sub-blocks in each down block
+        up_blocks: a list of filter counts, one for each up block
```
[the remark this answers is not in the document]
ianstenbit: done (same for up_blocks)

keras_cv/models/__internal__/unet.py
lengzq: Let's always use synchronized batch norm.

keras_cv/models/__internal__/unet.py
lengzq: Sounds good!

lengzq (approved):

keras_cv/models/__internal__/unet.py
tanzhenyu: Ah sounds good.

tanzhenyu (approved):

PR #1173, opened 2022-12-21
[Refactor RetinaNet] Step 4 Encode labels in model forward pass.

tanzhenyu:
# What does this PR do?
This significantly reduces the training time from 280ms/step to 180ms/step on TPU. This also helps the OD models to get rid of overriding fit and evaluate.


keras_cv/models/object_detection/object_detection_base_model.py
```diff
@@ -37,84 +33,6 @@ def __init__(self, bounding_box_format, label_encoder, **kwargs):
         self.bounding_box_format = bounding_box_format
         self.label_encoder = label_encoder
```
LukeWood, line 35: We can remove the constructor now too, right?

LukeWood: Ah wait, we can remove the encoder portion of the constructor but not the decoder portion.

ianstenbit: It seems like we should be able to remove the ODBaseModel entirely in the next step, right?

ianstenbit (approved):
LGTM -- thanks!

keras_cv/models/object_detection/object_detection_base_model.py
tanzhenyu: Re Luke -- Done.
Re Ian -- yeah that's the expected next PR :-)

quantumalaviya: Does this imply that I should refactor my YoloX implementation to directly inherit from `keras.Model` or is a new OD base model being worked on?

tanzhenyu: We're aiming for the ideal state where OD model implementers only need to inherit from `keras.Model`. The OD base model still exist in this PR because it provides a customized way to run `predict`. The next PR will remove OD base model.
So yeah, it'd be great if you can refactor YOLOX to reflect that

quantumalaviya: Sounds good, thanks for the overview!

tanzhenyu:
Manually merging as the gcp run failing due to custom op build

PR #1175, opened 2022-12-21
[Refactor RetinaNet] Step 5 remove the inheritance from ObjectDetectionBaseModel.

tanzhenyu:
The prediction function as a helper will be re-used for FasterRCNN in the next PR.


keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -189,6 +187,9 @@ def __init__(
                 f"`label_encoder.box_variance={label_encoder.box_variance}`."
             )
 
+    def make_predict_function(self, force=False):
```
ianstenbit, line 190: This LGTM if we _are_ planning to make .predict return decoded boxes, but I was under the impression we might be changing this?

ianstenbit (approved):

keras_cv/models/object_detection/retina_net/retina_net.py
tanzhenyu: Yeah that's the consensus -- we still like `predict` to return decoded boxes, but don't like doing so with OD Base Model. So here's the workaround

tanzhenyu:
Skipping gcbrun since custom ops build fails

PR #1182, opened 2022-12-22
[Refactor RetinaNet] Step 6 Unify FasterRCNN to decode during predict.

tanzhenyu:
The PyCOCOCallback no longer need to take `input_nms`.

### DIFF
--- a/keras_cv/models/object_detection/faster_rcnn.py
+++ b/keras_cv/models/object_detection/faster_rcnn.py
@@ -388,7 +388,9 @@ class FasterRCNN(tf.keras.Model):
             samples_per_image=256,
             positive_fraction=0.5,
         )
-        self._prediction_decoder = NMSDecoder(bounding_box_format=bounding_box_format)
+        self._prediction_decoder = prediction_decoder or NMSDecoder(
+            bounding_box_format=bounding_box_format
+        )
 
     def _call_rpn(self, images, anchors, training=None):
         image_shape = tf.shape(images[0])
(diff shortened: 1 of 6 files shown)


keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -382,7 +388,7 @@ def __init__(
             samples_per_image=256,
             positive_fraction=0.5,
         )
-        self.nms_decoder = nms_decoder or NMSDecoder()
+        self._prediction_decoder = NMSDecoder(bounding_box_format=bounding_box_format)
```
ianstenbit, line 391: Shouldn't this continue to use `prediction_decoder` instead of the default, when specified?

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -554,11 +560,45 @@ def test_step(self, data):
...
+
+    @property
+    def prediction_decoder(self):
+        return self._prediction_decoder
+
+    @prediction_decoder.setter
+    def prediction_decoder(self, prediction_decoder):
+        self._prediction_decoder = prediction_decoder
+        self.make_predict_function(force=True)
+        self.make_test_function(force=True)
```
ianstenbit, line 579: I think that both here and RetinaNet we only need to re-build the predict function in this setter, as the test and train functions don't call the decoder, right?

ianstenbit (changes requested):
LGTM after a few little comments -- thank you!

keras_cv/models/object_detection/faster_rcnn.py
tanzhenyu: Good catch. Done

keras_cv/models/object_detection/faster_rcnn.py
tanzhenyu: Good point. Done

ianstenbit (approved):
Thank you!

PR #1128, opened 2022-12-13
[0.4 Release] Adding DeepLabV3Plus Architecture

DavidLandup0:
# What does this PR do?

While not strictly necessary, shipping V3Plus besides V3 makes sense since they're very similar.
Adds the DeepLabV3Plus architecture. Trains on the segmentation script here. Just replace:

```
 model = DeepLabV3(classes=21, backbone=backbone)
```

With:

```
 model = DeepLabV3Plus(classes=21, backbone=backbone)
```

Currently supports ResNet101V2 and ResNet50V2 as the backbone with the default parameters (101 is common, 50 has weights). The authors also used Xception. As it's not currently in KCV, we're currently only supporting ResNet. It also allows for a custom backbone, with an optional argument for the low-level layer name. The low-level layer has a default for ResNets.

Written as a functional subclassing model, just like #1119.

### Performance

Some initial runs and comparison (on 224x224 images for testing speed, which turned out to have similar reported mIOUs as 512x512 images):

DeepLabV3 with ResNet50V2 (random weights):

```
Epoch 1/10
2124/2124 [==============================] - 368s 156ms/step - loss: 1.3275 - sparse_categorical_crossentropy: 1.3275 - mean_io_u: 0.0380 - sparse_categorical_accuracy: 0.6898 - val_loss: 1.2368 - val_sparse_categorical_crossentropy: 1.2368 - val_mean_io_u: 0.0421 - val_sparse_categorical_accuracy: 0.7045
Epoch 2/10
2124/2124 [==============================] - 319s 150ms/step - loss: 1.1916 - sparse_categorical_crossentropy: 1.1916 - mean_io_u: 0.0498 - sparse_categorical_accuracy: 0.7020 - val_loss: 1.1088 - val_sparse_categorical_crossentropy: 1.1088 - val_mean_io_u: 0.0594 - val_sparse_categorical_accuracy: 0.7071
Epoch 3/10
2124/2124 [==============================] - 323s 152ms/step - loss: 1.1254 - sparse_categorical_crossentropy: 1.1254 - mean_io_u: 0.0665 - sparse_categorical_accuracy: 0.7087 - val_loss: 1.0990 - val_sparse_categorical_crossentropy: 1.0989 - val_mean_io_u: 0.0730 - val_sparse_categorical_accuracy: 0.7133
Epoch 4/10
[...]

PR #1183, opened 2022-12-22
Basic Semantic Segmentation Training Script

DavidLandup0:
# What does this PR do?
As we're working towards introducing more semantic segmentation architectures:

- DeepLabV3 refactor #1119
- DeepLabV3Plus #1128
- FCN #1176

We should have a generalized semantic segmentation training script, similar to the `basic_training.py` for image classification. The general outline for the DeepLabV3 script is great, and we can plug and play models into it, as long as we can customize some aspects of it such as the batch size, initial learning rate, etc.

This script is a generalized form of the existing DeepLabV3 training script meant for any given semantic segmentation architecture on Pascal VOC, in the style of @ianstenbit's `basic_training.py` script. It's been refactored to simplify methods and calls, use KPLs, and use Ian's addition for EMA and weight decay. (currently commented out because they're from an unsupported Keras release afaik)

IMO, having a central script to test these will also force us to ensure that all semantic segmentation models have the same input/output, and that they follow a standardized form that users can change by simply changing the model's name. It's better to introduce this earlier rather than later. Also, this will prevent us having multiple training scripts that are largely the same.

### TODO
A few edge cases left to tweak and introduce, but the majority of the script is there.

## Before submitting
      Pull Request section?
- [] Did you write any new necessary tests?
- [] If this adds a new model, can you run a few training steps on TPU in Colab to ensure that no XLA incompatible OP are used?

## Who can review?

@tanzhenyu @ianstenbit @LukeWood


examples/training/semantic_segmentation/pascal_voc/basic_segmentation_training.py
```diff
@@ -0,0 +1,186 @@
...
+
+def preprocess_image(img, cls_seg):
+    img = tf.keras.layers.Resizing(512, 512, interpolation="nearest")(img)
+    inputs = {"images": img, "segmentation_masks": cls_seg}
+    inputs = keras_cv.layers.preprocessing.RandomFlip("horizontal")(inputs)
+    return inputs["images"], inputs["segmentation_masks"]
+
+
+def proc_train_fn(examples):
+    image = examples.pop("image")
```
DavidLandup0, line 110: Maybe flip and then resize instead? 
Since we're getting a dict, popping, turning into a dict again to do `RandomFlip` and then returning separate outputs again.

The former might be more computationally expensive though, in which case this approach is better?

examples/training/semantic_segmentation/pascal_voc/basic_segmentation_training.py
```diff
@@ -0,0 +1,184 @@
...
+
+    image, cls_seg = preprocess_image(image, cls_seg, flip=True)
+
+    sample_weight = tf.equal(cls_seg, 255)
+    zeros = tf.zeros_like(cls_seg)
+    cls_seg = tf.where(sample_weight, zeros, cls_seg)
+    return image, cls_seg
+
+
+def proc_eval_fn(examples):
```
DavidLandup0, line 121: Since the only difference between `proc_train_fn` and `proc_eval_fn` is whether we're flipping the images as well or not (i.e. do augs or not) - can we collapse the two into a single func and pass an 'augment=boolean' flag?

Anticipating other augmentations that KCV now supports with segmentation masks.

DavidLandup0:
Btw, the preprocessing call works as expected:

DavidLandup0:
Loving the KPLs that support segmentation maps! Added a few placeholder augmentations to test. We should standardize which augmentations are used and to which degree:

image

examples/training/semantic_segmentation/pascal_voc/basic_segmentation_training.py
```diff
@@ -0,0 +1,194 @@
...
+        inputs = preprocessing.RandomGaussianBlur(kernel_size=3, factor=0.3)(inputs)
+        inputs = preprocessing.RandomRotation(factor=0.1, segmentation_classes=21)(
+            inputs
+        )
+
+    return inputs["images"], inputs["segmentation_masks"]
+
+
+def proc_train_fn(examples):
+    image = examples.pop("image")
```
DavidLandup0, line 120: Maybe we should just pass `examples` here and pop in the preprocessing function instead?

DavidLandup0:
Doing a 30-epoch run for validation with:

```
!python keras-cv/examples/training/semantic_segmentation/pascal_voc/basic_segmentation_training.py --model_name=DeepLabV3 --backbone=resnet50_v2 --use_ema=False --weights_path='/weights' --tensorboard_path='/tensorboard'
```

tanzhenyu (changes requested):
Thanks for the PR!
Can you help us build incremental changes, i.e.
1) git mv `examples/training/semantic_segmentation/pascal_voc/deeplab.py` `examples/training/semantic_semgnetation/pascal_voc/basic_training.py`
2) only add `model_name` flags, without adding any other KPLs / flags
3) start adding KPLs in next PR, but always run the model end to end to verify before and after effect

DavidLandup0:
We can do step-by-step, but through a series of PRs then, we'd have DeepLab-tuned parameters being used for a general script. Is that okay? Since PRs can take some time to run/verify/merge, won't we create a bottleneck for testing segmentation models with a series of merges?

If this ends up running >= original script, would that be enough validation, to hopefully avoid a bottleneck?

tanzhenyu:
> We can do step-by-step, but through a series of PRs then, we'd have DeepLab-tuned parameters being used for a general script. Is that okay? Since PRs can take some time to run/verify/merge, won't we create a bottleneck for testing segmentation models with a series of merges?
> 
> If this ends up running >= original script, would that be enough validation, to hopefully avoid a bottleneck? I'll go with your gut feeling.

which part of it is specific to deeplab?

DavidLandup0:
> which part of it is specific to deeplab?

The epochs to train for, LR, image sizes, augmentations, batch size, backbone and LR decay are dependent on or tuned to work with DeepLab, and likely won't translate well to other architectures. Similar to how each model trained with `basic_training.py` used *some* different params for training, for any other model here, we'll probably want to set different params.

If we set them all as customizable, and re-run the same configuration DeepLab was trained on (set using flags instead of hardcoded), would that serve as validation? 🤔

tanzhenyu:
> > which part of it is specific to deeplab?
> 
> The epochs to train for, LR, image sizes, augmentations, batch size, backbone and LR decay are dependent on or tuned to work with DeepLab, and likely won't translate well to other architectures. Similar to how each model trained with `basic_training.py` used _some_ different params for training, for any other model here, we'll probably want to set different params.
> 
> If we set them all as customizable, and re-run the same configuration DeepLab was trained on (set using flags instead of hardcoded), would that serve as validation? 🤔

I think image size / batch size / weight decay are usually using common hyperparameters, they might be able to share. For epochs, we could just set for 100 epochs with early stopping.
In the long run, we want to keep training script as generic as possible, and all hyperparameters per model somewhere else, so the training will look like:
`.../basic_training.py`
run_deeplab_train.sh -- call basic_training with its own hyperparameter
run_fcn_train.sh -- call basic_training with its own hyperparameter
etc.

tanzhenyu:
Let's get this PR going by simply scoping it to add `model_name` in the script, and git mv from deeplab_v3.py to basic_training.py, so we could more easily track the diff.
If you wouldn't be able to run git mv to show the diff, feel free to simply open another PR and close this one

DavidLandup0:
I'll likely draft this one and create a new [1/X step] PR with just model_name

DavidLandup0:
Saved changes on a new branch and will repurpose this one for step-by-step

DavidLandup0:
Added model_name flag and kwargs @tanzhenyu

examples/training/semantic_segmentation/pascal_voc/basic_training.py
```diff
@@ -136,15 +148,16 @@ def proc_eval_fn(examples):
...
+    backbone = models.ResNet50V2(
         include_rescaling=True,
         # This argument gives a 2% mIoU increase
         stackwise_dilations=[1, 1, 1, 2],
         input_shape=(512, 512, 3),
         include_top=False,
         weights="imagenet",
     )
-    model = DeepLabV3(classes=21, backbone=backbone)
+    model = segmentation.__dict__[FLAGS.model_name]
```
tanzhenyu, line 159: this can be retrieved from keras_cv.models.__dict__, no?

tanzhenyu (changes requested):

examples/training/semantic_segmentation/pascal_voc/basic_training.py
DavidLandup0: Good catch! They weren't originally in `models`, but have been moved since

DavidLandup0: Updated

tanzhenyu (approved):

DavidLandup0:
Opening next PR as soon as this is merged. Here's the idea for updates:

- Support TPU with mixed precision (currently only works on GPUs)
- Support backbones via flags
- Change preprocessing functions to use KCV layers and simplify the calls 
- Allow changing loss functions
- EMA and weight decay (2.11)
 
Thoughts?

tanzhenyu:
> Opening next PR as soon as this is merged. Here's the idea for updates order:
> 
> * Epochs flag
> * Support TPU with mixed precision (currently only works on GPUs)
> * Support backbones via flags [requires running a test]
> * Change preprocessing functions to use KCV layers and simplify the calls [requires running a test]
> * Allow changing loss functions
> * EMA and weight decay (2.11)
> 
> Thoughts?

Out of all of them, supporting TPU is the most prioritized one. I have test on TPU, it works extremely slow. It might be due to dataset problem or upsampling problem, which I haven't had the chance to figure it out

tanzhenyu:
No need to run gcp test since this only touches training script

DavidLandup0:
> Out of all of them, supporting TPU is the most prioritized one. I have test on TPU, it works extremely slow. It might be due to dataset problem or upsampling problem, which I haven't had the chance to figure it out

Could it be the missing bfloat16? Pushing that change now so we can test whether that helps

tanzhenyu:
> > Out of all of them, supporting TPU is the most prioritized one. I have test on TPU, it works extremely slow. It might be due to dataset problem or upsampling problem, which I haven't had the chance to figure it out
> 
> Could it be the missing bfloat16? Pushing that change now so we can test whether that helps

That's not the reason. TPU will use bfloat by default

PR #1186, opened 2022-12-22
add voxelization layer.

tanzhenyu:


tanzhenyu:
For discussing the API and implementation, before adding unit test

keras_cv/layers/object_detection3d/voxelization.py
```diff
@@ -0,0 +1,267 @@
...
+    Returns:
+      point_voxel_id: [B, N] unique ID of each voxel.
+    """
+    batch_size, _, dim = point_voxel_xyz.shape.as_list()
+    if batch_size is None:
+        batch_size = tf.shape(point_voxel_xyz)[0]
+    assert dim == len(voxel_spatial_size), f"{point_voxel_xyz.shape}"
+
+    voxel_spatial_size_prod = [
+        np.prod(voxel_spatial_size[i:]).item() for i in range(dim)
```
bhack, line 70: Do we need to use numpy here?

keras_cv/layers/object_detection3d/voxelization.py
```diff
@@ -0,0 +1,267 @@
...
+        self._voxel_size = voxel_size
+        self._spatial_size = spatial_size
+        self._voxel_spatial_size = compute_voxel_spatial_size(
+            spatial_size, self._voxel_size
+        )
+        self._voxel_spatial_size_volume = np.prod(self._voxel_spatial_size).item()
+
+    def call(
+        self,
+        point_xyz: tf.Tensor,
```
ianstenbit, line 211: Since only the first positional input to call is used to compute shape metadata in Keras, maybe the xyz+feature inputs should be passed in as a tuple instead of as different params?

keras_cv/layers/object_detection3d/voxelization.py
```diff
@@ -0,0 +1,267 @@
...
+        point_feature: tf.Tensor,
+        point_mask: tf.Tensor,
+        training: bool,
+    ) -> tf.Tensor:
+        """Voxelizes and learns voxel features with a point net.
+
+        B: batch_size.
+        N: number of points.
+        dim: the input dimension.
```
ianstenbit, line 220: Isn't dim implied to be 3 basedon the `xyz` notation?

keras_cv/layers/object_detection3d/voxelization.py
```diff
@@ -0,0 +1,267 @@
...
+        point_feature = point_feature * point_mask_float
+        new_dim = point_feature.shape.as_list()[-1]
+        point_feature = tf.reshape(point_feature, [-1, new_dim])
+        point_voxel_id = tf.reshape(point_voxel_id, [-1])
+        # [B * num_voxels, new_dim]
+        voxel_feature = tf.math.unsorted_segment_max(
+            point_feature, point_voxel_id, batch_size * self._voxel_spatial_size_volume
+        )
+        # unsorted_segment_max sets empty values to -inf(float).
+        voxel_feature_valid_mask = voxel_feature > -1000
```
ianstenbit, line 259: Maybe -1000 should be a static constant like epsilon?

keras_cv/layers/object_detection3d/voxelization.py
tanzhenyu: Yeah that'd be my proposal as well, I think this could potentially make `point_feature` more generic, it doesn't require a specific concatenation (much like our box concatenation over box / class / score today)

keras_cv/layers/object_detection3d/voxelization.py
tanzhenyu: yeah, to correct it -- should be the dimension after point net

keras_cv/layers/object_detection3d/voxelization.py
tanzhenyu: will remove. Thanks

keras_cv/layers/object_detection3d/voxelization.py
tanzhenyu: Good point. Done

Guowang (approved):

keras_cv/layers/object_detection3d/voxelization.py
```diff
@@ -0,0 +1,267 @@
...
+        point_voxel_id = compute_point_voxel_id(
+            point_voxel_xyz, self._voxel_spatial_size
+        )
+        # [B, N]
+        point_voxel_id = point_voxel_id * point_voxel_mask_int
+
+        return point_voxel_feature, point_voxel_id, point_voxel_mask
+
+
+class Voxelization(tf.keras.layers.Layer):
```
tanzhenyu, line 186: DynamicVoxelization?

bhack:
Do we have a clear docstring of what this layers is doing?

keras_cv/layers/object_detection3d/voxelization.py
```diff
@@ -0,0 +1,267 @@
...
+    """Dynamic voxelization and pool layer."""
```
lengzq, line 187: """Point to voxel feature encoding using dynamic voxelization."""

lengzq (approved):

tanzhenyu:
> Do we have a clear docstring of what this layers is doing?

Added docstring

PR #1170, opened 2022-12-20
Upgrade  to the last TF stable version

bhack:
# What does this PR do?

This  a recurrent PR, until automated (see https://github.com/keras-team/keras-cv/pull/837)

This time we need also to handle a new policy as we have introduced c++ custom ops in the repository.

The TF Addons policy was to only support custom ops for the last stable TF version.

Now you need to choice your policy.

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [X] Did you read the [contributor guideline](https://github.com/keras-team/keras-cv/blob/master/.github/CONTRIBUTING.md),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue? Please add a link
      to it if that's the case.
- [ ] Did you write any new necessary tests?
- [ ] If this adds a new model, can you run a few training steps on TPU in Colab to ensure that no XLA incompatible OP are used?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #1188, opened 2022-12-22
Update release process to build wheels with custom ops

ianstenbit:
Still need to verify that the Linux wheels are manylinux compatible

Also need to build an additional wheel that has no custom ops that has universal compatibility


bhack:
Are we going to produce precompiled custom ops only for the last stable TF release?

ianstenbit:
> Are we going to produce precompiled custom ops only for the last stable TF release?

Currently trying to build wheels for all major TF versions starting with 2.9

bhack:
> Currently trying to build wheels for all major TF versions starting with 2.9

Are you shure that we have ABI compatibility with custom OPS.
As we don't enforce a tensorflow version in the wheel are we going to broke ABI compatibility?

https://github.com/keras-team/keras-cv/blob/master/setup.py

ianstenbit:
> > Currently trying to build wheels for all major TF versions starting with 2.9
> 
> Are you shure that we have ABI compatibility with custom OPS. As we don't enforce a tensorflow version in the wheel are we going to broke ABI compatibility?
> 
> https://github.com/keras-team/keras-cv/blob/master/setup.py

Ahh right -- I always forget about ABI compat issues.

I think we should just build custom ops for TF 2.11 in that case.

bhack:
> I think we should just build custom ops for TF 2.11 in that case.

Thia is what we have done historically in TF Add-ons but if you don't enforce TF version with custom ops in the wheels it is still a problem. 

I don't know if you have copied also the warning logic when you have taken the lazy loader:

https://github.com/tensorflow/addons/blob/14af819a7dbbb857c6a210dcaa38120d1c55e312/tensorflow_addons/utils/resource_loader.py#L73

But you can also use something more strictly then a warning.

bhack:
P.s. It was also the topic introduced in my PR few days ago:

https://github.com/keras-team/keras-cv/pull/1170

tanzhenyu (approved):

bhack:
It seems that the check is missing:

https://github.com/keras-team/keras-cv/blob/master/keras_cv/utils/resource_loader.py

ianstenbit:
> It seems that the check is missing:
> 
> https://github.com/keras-team/keras-cv/blob/master/keras_cv/utils/resource_loader.py

I will add this in a separate PR

PR #1195, opened 2022-12-23
Improve FlipY

lengzq:
Previous FlipY implementation flips along the Y axis, which perform worse than flips the Y coordinate (along the X axis).


keras_cv/layers/preprocessing3d/global_random_flip_y.py
```diff
@@ -53,26 +53,24 @@ def get_config(self):
     def augment_point_clouds_bounding_boxes(
         self, point_clouds, bounding_boxes, transformation, **kwargs
     ):
-        point_clouds_x = -point_clouds[..., 0:1]
+        point_clouds_y = -point_clouds[..., 1:2]
```
tanzhenyu, line 56: Should we rename the layer to FlipX?
And the docstring says "flips on y axis"

tanzhenyu (changes requested):

keras_cv/layers/preprocessing3d/global_random_flip_y.py
lengzq: Done. We could upgrade this function in a future PR to support Flip XY

tanzhenyu (approved):

PR #1199, opened 2022-12-27
Fix gpu tests after pinning bazel version in #1188

ianstenbit:


cloudbuild/README.md
```diff
@@ -30,14 +30,14 @@ To add a dependency for GPU tests:
 - Have a Keras team member update the Docker image for GPU tests by running the remaining steps
 - Create a `Dockerfile` with the following contents:
 ```
-FROM tensorflow/tensorflow:2.10.0-gpu
+FROM tensorflow/tensorflow:2.11.0-gpu
```
bhack, line 33: If you are going to merge this in the Cloud build CI check  https://github.com/keras-team/keras-cv/pull/1170 for alignment.

ianstenbit: Ack -- that LGTM. Merged

tanzhenyu: Do we want 2.11?

ianstenbit: I think yes -- we should be testing against the latest stable release of TF, and offering custom ops compiled against the latest version of TF

tanzhenyu (approved):

cloudbuild/README.md
bhack: See our thread starting at https://github.com/keras-team/keras-cv/pull/1188#issuecomment-1363413573

We are no more a python only library. So we need to start to reason about versioning also at c++ level and probably at runtime that it is still missing but available in TF Addons.

PR #1201, opened 2022-12-27
Add ABI warning to custom ops loader

ianstenbit:
This is similar to the [TFA implementation](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/utils/resource_loader.py) and will raise warnings if users have an incompatible TF version with KerasCV's custom ops

### DIFF
--- a/keras_cv/version_check_test.py
+++ b/keras_cv/version_check_test.py
@@ -13,12 +13,21 @@
 # limitations under the License.
 
 import pytest
+import tensorflow as tf
 
 from keras_cv import version_check
 
+@pytest.fixture(autouse=True)
+def cleanup_tf_version():
+    actual_tf_version = tf.__version__
+    # Tests will be run after yield
+    yield
+
+    # Cleanup
+    tf.__version__ = actual_tf_version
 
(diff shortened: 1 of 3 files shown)


ianstenbit:
@bhack

bhack (approved):

keras_cv/version_check_test.py
```diff
@@ -28,11 +39,11 @@ def test_check_tf_version_error():
...
     # should pass
-    version_check.tf.__version__ = "2.9.1rc2"
+    tf.__version__ = "2.9.1rc2"
     version_check.check_tf_version()
 
 
 def test_check_tf_version_passes_nightly():
     # should pass
-    version_check.tf.__version__ = "2.10.0-dev20220419"
+    tf.__version__ = "2.10.0-dev20220419"
```
bhack, line 48: Where are these and RC used?

ianstenbit: `tf.__version__` is checked in version_check.py, and this is just verifying that it works for nightly and rc builds

tanzhenyu (approved):

PR #1204, opened 2022-12-28
[Refactor RetinaNet] Step 10, Unify call and loss reduction for Retin…

tanzhenyu:
…aNet.

This makes model.call return decoded boxes, use SUM reduction across OD models.

# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -73,9 +73,9 @@ class RetinaNet(tf.keras.Model):
             `scales=[2**x for x in [0, 1 / 3, 2 / 3]]`,
             `sizes=[32.0, 64.0, 128.0, 256.0, 512.0]`,
             and `aspect_ratios=[0.5, 1.0, 2.0]`.
-        label_encoder: (Optional) a keras.Layer that accepts an image Tensor and a
-            bounding box Tensor to its `call()` method, and returns RetinaNet training
-            targets.  By default, a KerasCV standard LabelEncoder is created and used.
+        label_encoder: (Optional) a keras.Layer that accepts an image Tensor, a
+            bounding box Tensor and a bounding box class Tensor to its `call()` method,
+            and returns RetinaNet training targets.  By default, a KerasCV standard LabelEncoder is created and used.
             Results of this `call()` method are passed to the `loss` object passed into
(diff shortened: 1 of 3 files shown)


keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -167,15 +167,6 @@ def __init__(
             output_filters=9 * 4, bias_initializer="zeros"
         )
 
-        self.loss_metric = tf.keras.metrics.Mean(name="loss")
-        self.classification_loss_metric = tf.keras.metrics.Mean(
-            name="classification_loss"
-        )
-        self.box_loss_metric = tf.keras.metrics.Mean(name="box_loss")
-        self.regularization_loss_metric = tf.keras.metrics.Mean(
```
ianstenbit, line 175: It looks like the regularization loss and the overall sum loss are not reintroduced anywhere -- am I missing something or does this PR remove those metrics?

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -245,20 +217,24 @@ def call(self, images, training=False):
 
         return box_pred, cls_pred
 
+    def call(self, images, training=None):
+        box_pred, cls_pred = self._forward(images, training=training)
+        if not training:
```
ianstenbit, line 222: I'm not certain how this will work with functional subclassing -- it seems like we'll want to get rid of `_forward` in the functional subclass model (that's how I've done it in my implementation of RetinaNet as a functional subclass -- PR pending #1203 being merged)

I suppose we can override call in the functional subclass model to call super.call and then decode the deltas to real boxes?

ianstenbit (changes requested):

keras_cv/models/object_detection/retina_net/retina_net.py
tanzhenyu: It's re-introduced in `compile(...weight_decay)`, just like RCNN models

keras_cv/models/object_detection/retina_net/retina_net.py
tanzhenyu: Discussed offline, let's try this in your PR

ianstenbit (approved):

PR #1205, opened 2022-12-28
[Object Detection] Use functional subclass paradigm for RetinaNet

ianstenbit:


keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -308,7 +314,7 @@ def compile(
         super().compile(loss=losses, **kwargs)
 
     def compute_loss(self, images, gt_boxes, gt_classes, training):
-        box_pred, cls_pred = self._forward(images, training=training)
+        box_pred, cls_pred = self.call(images, training=training)
```
tanzhenyu, line 317: if you use `call` here, would the validation loss be out of order (compared to the train loss)?

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -411,6 +417,19 @@ def test_step(self, data):
 
         return self.compute_metrics(x, {}, {}, sample_weight={})
 
+    def get_config(self):
+        return {
+            "classes": self.classes,
+            "bounding_box_format": self.bounding_box_format,
+            "backbone": self.backbone,
```
tanzhenyu, line 424: don't we need to call `self.backbone.get_config` here?

tanzhenyu (commented):
Thanks for the PR!

keras_cv/models/object_detection/retina_net/retina_net.py
tanzhenyu: (as the validation_data would use `test_step` which calls `call` with training=False, so the decoded box is used to compute validation loss?

keras_cv/models/object_detection/retina_net/retina_net.py
ianstenbit: Since the backbone is a Keras concept, config can be loaded recursively (same thing we do in RetinaNetLabelEncoder with the AnchorGenerator).

But we were missing serialization registration -- Done

keras_cv/models/object_detection/retina_net/retina_net.py
ianstenbit: Good catch -- I think this should be calling `super().call` (which should be identical to `_forward`)

I updated it to call super().call and the validation loss looks correct.
Without the change, the validation loss is super high (as one might expect)

tanzhenyu (approved):
Please verify the result is >32 mAP after training 30 epochs

ianstenbit:
> Please verify the result is >32 mAP after training 30 epochs

SG -- training now

tanzhenyu:
> > Please verify the result is >32 mAP after training 30 epochs
> 
> SG -- training now

FWIW, on TPU v3-8 it's training at 175ms/step, and >32 mAP. Let me know if you need me to share the training loss / tensorboard

ianstenbit:
> > > Please verify the result is >32 mAP after training 30 epochs
> > 
> > 
> > SG -- training now
> 
> FWIW, on TPU v3-8 it's training at 175ms/step, and >32 mAP. Let me know if you need me to share the training loss / tensorboard

On a TPUv3-8 with this change I am seeing 175ms/step and an MAP of 30 after 20 epochs. If it hits 32mAP by 30 epochs, I will merge this PR.

PR #1211, opened 2022-12-29
Add pretrained weights for ResNet50

### LINKED ISSUE #1193: Add imagenet pretrained weights for ResNet50 v1
**Short Description**
Though ResNet50 v2 performs better, many past papers that serve as benchmarks/baselines use a ResNet50 v1. It would be great to be able to load a pretrained ResNet50 v1 model in Keras CV to make comparisons to those existing models.

ianstenbit:
These weights score 75.5% top-1 imagenet validation accuracy, which beats the Keras applications weights (74.9%)

Fixes #1193

### DIFF
--- a/examples/training/classification/imagenet/training_history.json
+++ b/examples/training/classification/imagenet/training_history.json
@@ -153,6 +153,25 @@
             "validation_accuracy": "0.8010"
         }
     },
+    "resnet50": {
+        "v0": {
+            "accelerators": 8,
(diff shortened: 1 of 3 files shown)


tanzhenyu (approved):
Nice!!!
In the future, we might consider logging training history with multiple metrics, (such as top-5 accuracy in this case), so each new update may not be strictly better than a previous one

PR #1220, opened 2022-12-30
Use keras application backbone for better result.

tanzhenyu:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/examples/training/object_detection/pascal_voc/faster_rcnn.py
+++ b/examples/training/object_detection/pascal_voc/faster_rcnn.py
@@ -64,9 +64,31 @@ train_ds = train_ds.concatenate(
 eval_ds = tfds.load("voc/2007", split="test", with_info=False)
 
 with strategy.scope():
-    backbone = keras_cv.models.ResNet50(
-        include_top=False, weights="imagenet", include_rescaling=True
-    ).as_backbone()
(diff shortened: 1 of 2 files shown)


examples/training/object_detection/pascal_voc/faster_rcnn.py
```diff
@@ -64,9 +64,31 @@
 eval_ds = tfds.load("voc/2007", split="test", with_info=False)
 
 with strategy.scope():
-    backbone = keras_cv.models.ResNet50(
-        include_top=False, weights="imagenet", include_rescaling=True
-    ).as_backbone()
+    inputs = tf.keras.layers.Input(shape=image_size)
```
ianstenbit, line 67: Let’s make this a helper function so it doesn’t crowd out the more relevant code here. (Same for the other script)

ianstenbit (approved):
LGTM if we put building the backbone into a helper method in both cases

examples/training/object_detection/pascal_voc/faster_rcnn.py
tanzhenyu: Actually the 2 scripts are using different levels of outputs from backbone, so we'd have to create a slightly more generic helper function to return different backbones in this case

PR #1228, opened 2023-01-03
MaxViT Components

DavidLandup0:
# What does this PR do?

Adds MaxViT components, done with @ayulockin. Since the official MaxViT is implemented in TF/Keras, some elements were ported from the repo with appropriate credit, while the rest was written from scratch or utilize existing KerasCV components such as MBConvBlocks.

Here, we add:

- `WindowPartitioning` layer
- `UnWindowPartitioning` layer
- `GridPartitioning` layer
- `UnGridPartitioning` layer
- `RelativeMultiHeadAttention` layer
- `MaxViTStem` layer (Convs)
- `MaxViTBlock` layer (Performs MBConv -> Block-Attention (Block-SA+FFN) -> Grid-Attention (Grid-SA+FFN))

MaxViT itself is:

> MaxViTStem -> repeated MaxViTBlock -> pooling -> fc head

image

The grid/ungrid/window/unwindow layers are used within each MaxViTBlock, and are omitted from the image. We can make them internal, if need be.

### Considerations

The official repository has a `MaxViTTransformerEncoder`, which was never used in the code in lieu of manually building these blocks with `RelativeMultiHeadAttention`. We could (and should) modularize this further into a `MaxViTTransformerEncoder` in the `transformer_encoder.py` alongside the ViT encoder. Originally in a previous discussion, we thought about generalizing the existing `TransformerEncoder` to encompass both, but doing so would mean entangling them and adding 3 new arguments, which would be less user-friendly. I.e. we'd have the default initialization default to the generic transformer encoder (as used in ViTs), but setting `attn_type='relative_multihead` and `attn_input_form='block'` or `attn_input_form='grid'` to create any of the two transformer encoder types for `MaxViTTransformerEncoder`. Thoughts?

Additionally, the official repo uses a custom `TrailDense` layer, which has been introduced as `EinsumDense` in Keras 2.11. This PR uses `EinsumDense` instead. This *might* make weight porting harder, if the implementations differ under the hood, but MaxViTs seem to be trainable as easily as CNNs, so we'd be training it with `basic_training.py` instead, so that likely won't make a difference.

## Who can review?

@tanzhenyu @LukeWood @ianstenbit @bhack


DavidLandup0:
Added layer tests

keras_cv/layers/__init__.py
```diff
@@ -19,6 +19,13 @@
 
 from keras_cv.layers.feature_pyramid import FeaturePyramid
 from keras_cv.layers.fusedmbconv import FusedMBConvBlock
+from keras_cv.layers.maxvit_layers import GridPartitioning
+from keras_cv.layers.maxvit_layers import MaxViTBlock
+from keras_cv.layers.maxvit_layers import MaxViTStem
+from keras_cv.layers.maxvit_layers import RelativeMultiHeadAttention
+from keras_cv.layers.maxvit_layers import UnGridPartitioning
+from keras_cv.layers.maxvit_layers import UnWindowPartitioning
+from keras_cv.layers.maxvit_layers import WindowPartitioning
```
innat, line 28: Does window partitioning appear in swin transformer? If so, should they place as a separate layers?

DavidLandup0: It does, but afaik, it's a different operation so we couldn't reuse this one potentially.
I'll look into how they differ (if the output differs), since if the output is the same we can use them between them without issues?

PR #1232, opened 2023-01-03
Added CSPDarkNet variants and as_backbone as a util

### LINKED ISSUE #1171: CSPDarkNet has weights only for when depth and width multiplier are 1.0
This makes the current weights compatible only with YoloX_l. 

Would it make sense to have CSPDarknet_tiny, _s, _m, _l and _x so that the imagenet weights can be trained for these separately?

cc. @ianstenbit @LukeWood

quantumalaviya:
# What does this PR do?
This PR fixes the issue discussed in #1171 and also migrates as_backbone to be a model util

Fixes #1171

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LukeWood

### DIFF
--- a/keras_cv/models/csp_darknet.py
+++ b/keras_cv/models/csp_darknet.py
@@ -19,6 +19,7 @@ Reference:
     - [YoloX Paper](https://arxiv.org/abs/2107.08430)
     - [YoloX implementation](https://github.com/ultralytics/yolov3)
 """
+import types
 
 import tensorflow as tf
 from tensorflow import keras
@@ -34,10 +35,10 @@ from keras_cv.models.weights import parse_weights
 
 
 def CSPDarkNet(
+    depth_multiplier,
+    width_multiplier,
     include_rescaling,
     include_top,
-    depth_multiplier=1.0,
-    width_multiplier=1.0,
     use_depthwise=False,
     classes=None,
     weights=None,
@@ -45,7 +46,7 @@ def CSPDarkNet(
     input_tensor=None,
     pooling=None,
     classifier_activation="softmax",
-    name=None,
+    name="CSPDarkNet",
     **kwargs,
 ):
     """Instantiates the CSPDarkNet architecture.
@@ -63,14 +64,14 @@ def CSPDarkNet(
     [guide to transfer learning & fine-tuning](https://keras.io/guides/transfer_learning/).
 
     Args:
+        depth_multiplier: A float value used to calculate the base depth of the model
+            this changes based the detection model being used.
+        width_multiplier: A float value used to calculate the base width of the model
+            this changes based the detection model being used.
         include_rescaling: whether or not to Rescale the inputs.If set to True,
             inputs will be passed through a `Rescaling(1/255.0)` layer.
         include_top: whether to include the fully-connected layer at the top of
             the network.  If provided, `classes` must be provided.
-        depth_multiplier: A float value used to calculate the base depth of the model
-            this changes based the detection model being used. Defaults to 1.0.
-        width_multiplier: A float value used to calculate the base width of the model
-            this changes based the detection model being used. Defaults to 1.0.
         use_depthwise: a boolean value used to decide whether a depthwise conv block
             should be used over a regular darknet block. Defaults to False
         classes: optional number of classes to classify images into, only to be
@@ -92,13 +93,10 @@ def CSPDarkNet(
         classifier_activation: A `str` or callable. The activation function to use
             on the "top" layer. Ignored unless `include_top=True`. Set
             `classifier_activation=None` to return the logits of the "top" layer.
-
-        name: (Optional) name to pass to the model.  Defaults to "DarkNet".
+        name: (Optional) name to pass to the model.  Defaults to "CSPDarkNet".
     Returns:
         A `keras.Model` instance.
     """
-    weights = parse_weights(weights, include_top, "cspdarknet")
-
     if weights and not tf.io.gfile.exists(weights):
         raise ValueError(
             "The `weights` argument should be either `None` or the path to the "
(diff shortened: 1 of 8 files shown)


quantumalaviya:
@ianstenbit When you get a chance, can you review this and train weights for these variants after merge? To start with, weights for `CSPDarkNetTiny` would be enough for me to continue experiments on YoloX.

quantumalaviya:
@qlzh727 (tagging because of the todo note) @tanzhenyu I have attempted to add `as_backbone` as a util here. Please take a look and see if this works.

keras_cv/models/utils_test.py
```diff
@@ -28,3 +28,40 @@ def test_parse_model_inputs(self):
 
         input_tensor = layers.Input(shape=input_shape)
         self.assertIs(utils.parse_model_inputs(input_shape, input_tensor), input_tensor)
+
+    def test_as_backbone_util_exception(self):
```
ianstenbit, line 32: Could you rename this test to test_as_backbone_missing_backbone_level_outputs

keras_cv/models/csp_darknet.py
```diff
@@ -177,7 +180,230 @@ def CSPDarkNet(
...
+        input_shape=input_shape,
+        input_tensor=input_tensor,
+        pooling=pooling,
+        classifier_activation=classifier_activation,
+        name=name,
+        **kwargs,
+    )
+
+
+def CSPDarkNetX(
```
ianstenbit, line 375: Should this be XL instead of X, or is X a convention for CSP Darket

ianstenbit (commented):
Looks good -- just a few nits

DavidLandup0:
Thanks for the PR!
In a previous PR, I was trying to use ResNetV2's `as_backbone()` but the issue was that I could not select the specific layers (only a range). The range had two issues:

- Unused layers (only needed the first and last, middle ones added params but weren't used)
- The final output had 2048 channels (while the paper used a few layers earlier to limit the channels to 256 to reduce the effect of the feature maps when added to another stage)

It's *great* that we can have `as_backbone()` as a general util instead of a method in each model. To make it more globally usable (and the default approach to exposing backbones) - should we maybe allow for `layer_names=('l1', 'l2', ...)` instead of a range?

keras_cv/models/csp_darknet.py
quantumalaviya: The YoloX model corresponding to this is called YoloX-x. Hence, the name.

quantumalaviya:
I agree with David. I think the API should be updated to something like this:

```python
as_backbone(levels=None)
```

Where `levels` is one of:
- a tuple of size 2 indictating a range which would work as it does currently. `None` at any one of the indices would imply the limit for that level. For example, (None, 5) would mean levels 3, 4, 5 for a backbone where 2 is the minimum level.
- a string indicating the name of the layer to output.
- a list of strings that have all the layer names required in the output.

Defaults to None which includes all levels from `_backbone_level_outputs`.

With this, all the models can potentially support `as_backbone`. `_backbone_level_outputs` will be required only for a model to support the range tuple. Moreover, this way, users don't necessarily have to know what the internal layer names are.

Ofc, this API would need some refining but maybe its a better alternative? Would love your opinions on this @LukeWood @ianstenbit @tanzhenyu

ianstenbit:
> I agree with David. I think the API should be updated to something like this:
> 
> ```python
> as_backbone(levels=None)
> ```
> 
> Where `levels` is one of:
> 
> * a tuple of size 2 indictating a range which would work as it does currently. `None` at any one of the indices would imply the limit for that level. For example, (None, 5) would mean levels 3, 4, 5 for a backbone where 2 is the minimum level.
> * a string indicating the name of the layer to output.
> * a list of strings that have all the layer names required in the output.
> 
> Defaults to None which includes all levels from `_backbone_level_outputs`.
> 
> With this, all the models can potentially support `as_backbone`. `_backbone_level_outputs` will be required only for a model to support the range tuple. Moreover, this way, users don't necessarily have to know what the internal layer names are.
> 
> Ofc, this API would need some refining but maybe its a better alternative? I also wonder if this util is should even be exposed or kept as an internal util. Would love your opinions on this @LukeWood @ianstenbit @tanzhenyu

I like the idea of adding some more flexibility to the `as_backbone` API -- can you open up an issue to discuss this?

ianstenbit (approved):
LGTM -- thanks!

As long as the GPU tests pass, I'll merge this and kick off training of the CSPDarkNetTiny to unblock you on YoloX

tanzhenyu:
> I agree with David. I think the API should be updated to something like this:
> 
> ```python
> as_backbone(levels=None)
> ```
> 
> Where `levels` is one of:
> 
> * a tuple of size 2 indictating a range which would work as it does currently. `None` at any one of the indices would imply the limit for that level. For example, (None, 5) would mean levels 3, 4, 5 for a backbone where 2 is the minimum level.
> * a string indicating the name of the layer to output.
> * a list of strings that have all the layer names required in the output.
> 
> Defaults to None which includes all levels from `_backbone_level_outputs`.
> 
> With this, all the models can potentially support `as_backbone`. `_backbone_level_outputs` will be required only for a model to support the range tuple. Moreover, this way, users don't necessarily have to know what the internal layer names are.
> 
> Ofc, this API would need some refining but maybe its a better alternative? I also wonder if this util is should even be exposed or kept as an internal util. Would love your opinions on this @LukeWood @ianstenbit @tanzhenyu

What we returned here is dict: int -> output_tensor. If we do support string names, I have two questions:
1) how do expect users to find those layer names -- this can sometimes be trivial, but sometimes quite some manual work
2) would the output be string -> output_tensor, and how does that interact with our FPN / Detection / Segmentation models which today only accept int?

PR #1234, opened 2023-01-03
Add ConvNeXtV2

### LINKED ISSUE #1233: ConvNeXt V2
**Short Description**
Just released - ConvNeXt with a new internal layer.

**Papers**
https://arxiv.org/abs/2301.00808

**Existing Implementations**
https://github.com/facebookresearch/ConvNeXt-V2

**Other Information**
If accepted, sign me up for the PR!

DavidLandup0:
# What does this PR do?

WIP - added `GlobalResponseNormalization` layer, working on V2 now.
Pinging when done.

Fixes #1233 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


DavidLandup0:
P.S. Current ConvNeXt is TPU-incompatible. V2 bypasses dtype issues, but on Kaggle's TPUs, runs into a:

```
 Invalid argument: {{function_node __inference_train_function_50061}} Compilation failure: Input 1 to node `convnext_atto/convnext_atto_stage_0_block_0/global_response_normalization_1/norm/ListDiff` with op ListDiff must be a compile-time constant.
```

I moved the projection dims from `build()` to `_init_()` because of this, but the same issue remains. This *may* be an outdated TF version issue, since Kaggle TPUs use TF 2.4.2. Could someone test this on a new environment?

edwardyehuang:
Please check my implementation of ConvNeXt and ConvNeXtV2.
https://github.com/edwardyehuang/iSeg/blob/master/backbones/convnext.py
https://github.com/edwardyehuang/iSeg/blob/master/backbones/convnext_v2.py

They are tested on TPU and tested on downstream tasks.

Note that I did not use tf.norm in GRN of ConvNeXtV2 because it has many issues on TensorFlow when ord is set to 2.

edwardyehuang:
Also, cast to float32 for both inputs and gamma/beta weights is required to avoid NaN on mixed precision (it seems GRN was very numerically sensitive based on my testing yesterday). Make sure to convert back before returning the result.

DavidLandup0:
Hey @edwardyehuang - thanks for weighing in!
The TPU issue seems to be tied to the `tf.norm` call. I figured it's still worth testing, but it seems to cause other problems as well, such as extremely slow training on GPUs. Might be worth opening an issue in the core TF repo, since it'd be useful to have it as part of layers during training...

Thank you for linking your implementations! You're blazingly fast 😅
Adding a reference to your comment in the docstrings. Seems to be training fine right now.

DavidLandup0:
> Also, cast to float32 for both inputs and gamma/beta weights is required to avoid NaN on mixed precision (it seems GRN was very numerically sensitive based on my testing yesterday). Make sure to convert back before returning the result.

Noted, thanks! This'll save a bunch of debugging.

DavidLandup0:
> They are tested on TPU and tested on downstream tasks.

If you have a few minutes - did you train them using one of the general scripts from iSeg or did you implement a masked autoencoder framework for training? If you already have a structure in place, KerasCV is a great place if you want to share scripts for reproducibility and transparency when it comes to training procedures. :)

edwardyehuang:
> 

Hi. I only train downstream tasks using iSeg. I did not implement pre-training for ConvNeXtV2.

DavidLandup0:
> Hi. I only train downstream tasks using iSeg. I did not implement pre-training for ConvNeXtV2.

Thanks for taking the time! We'll do a training script here then.

tanzhenyu:
https://github.com/keras-team/keras-cv/issues/1233

DavidLandup0:
Thanks for the comments! Do we continue with this one without a specific training script (masked autoencoder) like we did for ConvNeXtV1, or do we put it off for now?
The architectural difference is small, and the crux of it lies in the GRN layer and it's use in the general block. It's fairly easily verifiable if that works or not if we don't want to put a lot of time into this.

tanzhenyu:
> Thanks for the comments! Do we continue with this one without a specific training script (masked autoencoder) like we did for ConvNeXtV1, or do we put it off for now? The architectural difference is small, and the crux of it lies in the GRN layer and it's use in the general block. It's fairly easily verifiable if that works or not if we don't want to put a lot of time into this.

What was the reason we did have training script for V1? Did we port weights for that model?

DavidLandup0:
The weights were ported, but I don't see them in weights.py.
I can port them for V2 if need be.

DavidLandup0:
We should probably also host the existing weights here as well

tanzhenyu:
> We should probably also host the existing weights here as well

Yes let's port the weights then, can you include the porting script in this PR as well?

DavidLandup0:
Sure! I'll add weight porting, and imagenetv2 results in this PR, and files in the next one

DavidLandup0:
Weight porting script: https://colab.research.google.com/drive/1-fDt3XDm1WmUGVo-29Ap9-BIiTH8sZrQ?usp=sharing

Weights load but imagenet performance implies that something is wrong. Going line-by-line, it might not be the model, but the eval pipeline. I can't find any info on the expected input range, except for V1, which is uint8. Looking into it

PR #1238, opened 2023-01-04
Minor updates to 3D preprocessing layers

lengzq:
1. frustum_random_dropping_points.py: Change the random_point_mask shape to [1, points, 1] to align with the global drop point logic.

2. swap_background.py: Convert the masks used in boolean_mask from float tensors to bool tensors.


keras_cv/layers/preprocessing3d/frustum_random_dropping_points.py
```diff
@@ -102,7 +102,7 @@ def get_random_transformation(self, point_clouds, **kwargs):
         frustum_mask = tf.concat(frustum_mask, axis=0)
         random_point_mask = (
             self._random_generator.random_uniform(
-                [num_frames, num_points, 1], minval=0.0, maxval=1
+                [1, num_points, 1], minval=0.0, maxval=1
```
tanzhenyu, line 105: any reason for this? Seems we're making it less generic

tanzhenyu: If we need to make this, can you add a comment to why this is the case

lengzq: Similar to global drop point. We use the same point mask for all frames.

ianstenbit (approved):
Could you please update the PR description to something like 

"Minor updates to 3D preprocessing layers"

Thanks!

keras_cv/layers/preprocessing3d/frustum_random_dropping_points.py
lengzq: Done.

tanzhenyu (approved):

PR #1240, opened 2023-01-05
Add new transform functions for converting to vehicle frame and padding.

Guowang:


keras_cv/layers/object_detection3d/utils.py
```diff
@@ -0,0 +1,72 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+
+def HasRank(tensor, expected_rank):
```
ianstenbit, line 18: Let's move these utils to keras_cv/utils/tensor_utils.py and use lowercase_underscore naming convention (to match other utils).

It seems like these methods aren't 3D OD specific, so a more generic codepath for them probably makes more sense

keras_cv/layers/object_detection3d/utils_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+import tensorflow as tf
+
+from keras_cv.layers.object_detection3d.utils import PadOrTrimTo
+
+
+class PadOrTrimToTest(tf.test.TestCase):
+    """Tests for PadOrTrimTo, branched from https://github.com/tensorflow/lingvo/blob/master/lingvo/core/py_utils_test.py."""
+
+    def test2DConstantShapePad(self):
```
ianstenbit, line 23: nit: lowercase_underscore naming for these test methods too

ianstenbit (changes requested):
Thank you!

keras_cv/layers/object_detection3d/utils.py
tanzhenyu: Is it possible to make this private utils for now, before we consider how to generalize it and move to keras_cv/utils?

keras_cv/datasets/waymo/transformer.py
```diff
@@ -481,6 +482,26 @@ def _get_point_label(
     )
 
 
+def _inv_loc(rot: tf.Tensor, loc: tf.Tensor) -> tf.Tensor:
```
tanzhenyu, line 485: I think this function is already provided in voxel_utils.py, can you re-use that one?

tanzhenyu (changes requested):
Thanks for the PR!

keras_cv/datasets/waymo/transformer.py
Guowang: Seems no functions for invert location? I can see get_yaw_rotation but not fit here.

keras_cv/layers/object_detection3d/utils.py
Guowang: Add '_' prefix to indicate it is internal and add comments to docstring.

keras_cv/layers/object_detection3d/utils_test.py
Guowang: done

keras_cv/datasets/waymo/transformer.py
tanzhenyu: Oops, sorry, should be here:
https://github.com/keras-team/keras-cv/blob/master/keras_cv/layers/object_detection3d/center_net_label_encoder.py#L55
Can we extract this out into the voxel_utils.py?

tanzhenyu (approved):
LGTM after re-using `inv_loc`

keras_cv/datasets/waymo/transformer.py
Guowang: Sure, also move internal utils to voxel_utils.

PR #1251, opened 2023-01-06
Add pretrained weights for CSPDarkNetTiny

ianstenbit:
Top-1 ImageNet accuracy is ~62%.

This seems low to me but I can't find an appropriate benchmark. @quantumalaviya is this around what you'd expect?


LukeWood (approved):
Looks good to me, @quantumalaviya do you know if these scores are ok for Tiny?

quantumalaviya:
Does seem a bit low but I can't find a benchmark either. CSPDarkNets are usually used only for detection tasks.

ianstenbit:
@quantumalaviya I'm going to merge this for now then -- we can always add new weights if these ones turn out to be sub-par

PR #1261, opened 2023-01-11
API polish on NMS layer

LukeWood:
This PR:

- changes some argument names in `call()` to remove unnecessary abbreviations
- uses `preserve_rel` to drop the need for `images`, this preserves identical behavior.
- changes the name to NonMaxSuppression.  I think this is more indicative of the layers' current functionality.

### DIFF
--- a/keras_cv/layers/object_detection/nms_prediction_decoder.py
+++ b/keras_cv/layers/object_detection/nms_prediction_decoder.py
@@ -19,7 +19,7 @@ from keras_cv import bounding_box
 
 # TODO(tanzhenyu): provide a TPU compatible NMS decoder.
 @tf.keras.utils.register_keras_serializable(package="keras_cv")
-class NmsDecoder(tf.keras.layers.Layer):
+class NonMaxSuppression(tf.keras.layers.Layer):
     """A Keras layer that decodes predictions of an object detection model.
 
     Arguments:
@@ -57,32 +57,35 @@ class NmsDecoder(tf.keras.layers.Layer):
         self.max_detections_per_class = max_detections_per_class
         self.built = True
 
-    def call(self, box_pred, confidence_pred):
+    def call(self, box_prediction, confidence_prediction):
         """Accepts images and raw predictions, and returns bounding box predictions.
 
         Args:
-            images: Tensor of shape [batch, height, width, channels].
-            box_pred: Dense Tensor of shape [batch, boxes, 4] in the
+            box_prediction: Dense Tensor of shape [batch, boxes, 4] in the
                 `bounding_box_format` specified in the constructor.
-            confidence_pred: Dense Tensor of shape [batch, boxes, num_classes].
+            confidence_prediction: Dense Tensor of shape [batch, boxes, num_classes].
         """
-        box_pred = bounding_box.convert_format(
-            box_pred,
+        target_format = bounding_box.preserve_rel(
+            target_bounding_box_format="yxyx",
+            bounding_box_format=self.bounding_box_format,
+        )
+        box_prediction = bounding_box.convert_format(
+            box_prediction,
             source=self.bounding_box_format,
-            target="yxyx",
+            target=target_format,
         )
         if self.from_logits:
             confidence_pred = tf.nn.softmax(confidence_pred)
 
-        box_pred = tf.expand_dims(box_pred, axis=-2)
+        box_prediction = tf.expand_dims(box_prediction, axis=-2)
         (
             box_pred,
             confidence_pred,
             class_pred,
             valid_det,
         ) = tf.image.combined_non_max_suppression(
-            boxes=box_pred,
-            scores=confidence_pred,
+            boxes=box_prediction,
+            scores=confidence_prediction,
             max_output_size_per_class=self.max_detections_per_class,
             max_total_size=self.max_detections,
             score_threshold=self.confidence_threshold,
@@ -91,7 +94,7 @@ class NmsDecoder(tf.keras.layers.Layer):
         )
         box_pred = bounding_box.convert_format(
             box_pred,
-            source="yxyx",
+            source=target_format,
             target=self.bounding_box_format,
         )
         return {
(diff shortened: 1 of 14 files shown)


keras_cv/layers/object_detection/nms_prediction_decoder.py
```diff
@@ -57,32 +57,35 @@ def __init__(
...
-            images: Tensor of shape [batch, height, width, channels].
-            box_pred: Dense Tensor of shape [batch, boxes, 4] in the
+            box_prediction: Dense Tensor of shape [batch, boxes, 4] in the
                 `bounding_box_format` specified in the constructor.
-            confidence_pred: Dense Tensor of shape [batch, boxes, num_classes].
+            confidence_prediction: Dense Tensor of shape [batch, boxes, num_classes].
         """
-        box_pred = bounding_box.convert_format(
-            box_pred,
+        target_format = bounding_box.preserve_rel(
```
tanzhenyu, line 68: Doesn't this require `image_shape` or `image` later during the conversion? I'm not sure why we prefer converting to relative format

keras_cv/layers/object_detection/nms_prediction_decoder.py
```diff
@@ -19,7 +19,7 @@
 
 # TODO(tanzhenyu): provide a TPU compatible NMS decoder.
 @tf.keras.utils.register_keras_serializable(package="keras_cv")
-class NmsDecoder(tf.keras.layers.Layer):
+class NonMaxSuppression(tf.keras.layers.Layer):
```
ianstenbit, line 22: Let's also rename this file (and corresponding test file) to match this

tanzhenyu, line 22: I'm fine with either NMS or NonMaxSuppression., but I prefer it be more specific, for example, MultiClassNMS or BatchedNMS.

tanzhenyu (changes requested):

keras_cv/layers/object_detection/nms_prediction_decoder.py
ianstenbit, line 68: Just a thought: reading this code, it's really not obvious what `preserve_rel` does. Should we consider renaming it?

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -146,7 +146,7 @@ def __init__(
             ).as_backbone()
         )
 
-        self._prediction_decoder = prediction_decoder or cv_layers.NmsDecoder(
+        self._prediction_decoder = prediction_decoder or cv_layers.NonMaxSuppression(
```
ianstenbit, line 149: IMO it's become a bit confusing that we call this a "decoder" since it's not doing the relative-to-absolute coordinates transformation step.

Would it make sense to rename this concept in the OD models?

ianstenbit (changes requested):

keras_cv/layers/object_detection/nms_prediction_decoder.py
LukeWood: preserve_rel makes it so you don't need image shape.

it matches the input's relative/nonrelative status, so you can use it for operations where relative/nonrelative doesn't matter.

LukeWood: (IoU comptuation/NMS falls in that category)

keras_cv/layers/object_detection/nms_prediction_decoder.py
LukeWood: Sounds good - I'll think about a more descriptive name.

keras_cv/layers/object_detection/nms_prediction_decoder.py
LukeWood: lets rename this because @tanzhenyu raised a similar question above.  Lets do this in a separate PR though.

https://github.com/keras-team/keras-cv/issues/1262

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: Fair enough, again lets do that in a separate PR though.

keras_cv/layers/object_detection/nms_prediction_decoder.py
LukeWood: Updated to MultiClassNMS

LukeWood: Done

LukeWood:
@tanzhenyu @ianstenbit all changes addressed

keras_cv/layers/object_detection/nms_prediction_decoder.py
tanzhenyu: But how does the box gets converted to a relative format without an image shape?
Also what's the benefit here? Given this layer is running on CPU today, I'd rather minimize the ops we need to call

tanzhenyu (changes requested):
Left 1 more comment.

keras_cv/layers/object_detection/nms_prediction_decoder.py
LukeWood: it doesn't.  There's no extra ops.  preserve_rel() matches the relative/nonrelative format of the source format in the target format.

It is still a single conversion - check out the code of preserve_rel().  What it does is make it so if you're in yxyx you stay in yxyx, but if you're in rel_yxyx you stay in rel_yxyx.

This makes it so you don't need images to apply NMS, but your outputs are the same.   Its strictly better - and theres no performance cost.

LukeWood: (preserve_rel is purely a python space construct.  It returns a format name based on the target format to prevent calls that require image shapes)

tanzhenyu: Oh....that's a really confusing name then....

tanzhenyu (approved):

ianstenbit (approved):

PR #1265, opened 2023-01-12
add utility function to mask invalid detections

LukeWood:
Fixes the output of the NMS layer to comply with the KerasCV bounding box format.
Right now, the returned boxes from NMS include all boxes, not just those that pass through NMS.  This causes confusing behavior where there are many more boxes included in the result Tensor than there should be.

---

    """masks out invalid detections with -1s.

    This utility is mainly used on the output of `tf.image.combined_non_max_suppression`
    operations.  The output of `tf.image.combined_non_max_suppression` contains padding
    values of 0 instead of -1.  KerasCV expects all bounding boxes to be padded with
    -1s instead of 0s.  This function uses the value of `num_detections` to mask out
    invalid boxes with -1s.

    Args:
        bounding_boxes: a dictionary complying with KerasCV bounding box format.  In
            addition to the normal required keys, these boxes are also expected to have
            a `num_detections` key.

---

XLA testing was done to ensure TPU compat:

```
def test_in_xla():
  num_detections = tf.constant([5, 4, 2, 3])
  boxes = tf.random.uniform((4, 100, 4))
  classes = tf.random.uniform((4, 100))
  mask = tf.repeat(tf.expand_dims(tf.range(boxes.shape[1]), axis=0), repeats=boxes.shape[0], axis=0)
  mask = mask < num_detections[:, None]
  mask = tf.expand_dims(mask, axis=-1)
  mask = tf.repeat(mask, boxes.shape[-1], axis=-1)
  result = tf.where(mask, boxes, -tf.ones_like(boxes))
  return result
```

---

Todo:

- [x] add unit tests

### DIFF
--- a/keras_cv/bounding_box/mask_invalid_detections.py
+++ b/keras_cv/bounding_box/mask_invalid_detections.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import tensorflow as tf
+
 from keras_cv.bounding_box.to_dense import to_dense
 from keras_cv.bounding_box.to_ragged import to_ragged
 from keras_cv.bounding_box.validate_format import validate_format
@@ -43,6 +45,7 @@ def mask_invalid_detections(bounding_boxes):
 
     boxes = bounding_boxes.get("boxes")
     classes = bounding_boxes.get("classes")
+    num_detections = bounding_boxes.get("num_detections")
 
     # Create a mask to select only the first N boxes from each batch
     mask = tf.repeat(
@@ -56,6 +59,10 @@ def mask_invalid_detections(bounding_boxes):
     mask = tf.expand_dims(mask, axis=-1)
     mask = tf.repeat(mask, repeats=boxes.shape[-1], axis=-1)
     boxes = tf.where(mask, boxes, -tf.ones_like(boxes))
+
+    result = bounding_boxes.copy()
+    result["boxes"] = boxes
+    result["classes"] = classes
 
     if info["ragged"]:
         return to_ragged(result)
--- a/keras_cv/bounding_box/mask_invalid_detections.py
+++ b/keras_cv/bounding_box/mask_invalid_detections.py
@@ -1,0 +1,63 @@
+# Copyright 2022 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_cv.bounding_box.to_dense import to_dense
+from keras_cv.bounding_box.to_ragged import to_ragged
+from keras_cv.bounding_box.validate_format import validate_format
+
+
+def mask_invalid_detections(bounding_boxes):
+    """masks out invalid detections with -1s.
+
+    This utility is mainly used on the output of `tf.image.combined_non_max_suppression`
+    operations.  The output of `tf.image.combined_non_max_suppression` contains padding
+    values of 0 instead of -1.  KerasCV expects all bounding boxes to be padded with
+    -1s instead of 0s.  This function uses the value of `num_detections` to mask out
+    invalid boxes with -1s.
+
+    Args:
+        bounding_boxes: a dictionary complying with KerasCV bounding box format.  In
+            addition to the normal required keys, these boxes are also expected to have
+            a `num_detections` key.
+    """
+    # ensure we are complying with KerasCV bounding box format.
+    info = validate_format(bounding_boxes)
+    if "num_detections" not in bounding_boxes:
+        raise ValueError(
+            "`bounding_boxes` must have key 'num_detections' "
+            "to be used with `bounding_box.mask_invalid_detections()`."
+        )
+
+    # required to use tf.where()
+    bounding_boxes = to_dense(bounding_boxes)
+
+    boxes = bounding_boxes.get("boxes")
+    classes = bounding_boxes.get("classes")
+
+    # Create a mask to select only the first N boxes from each batch
+    mask = tf.repeat(
+        tf.expand_dims(tf.range(boxes.shape[1]), axis=0), repeats=boxes.shape[0], axis=0
+    )
+    mask = mask < num_detections[:, None]
+
+    classes = tf.where(mask, classes, -tf.ones_like(classes))
+
+    # resuse mask for boxes
+    mask = tf.expand_dims(mask, axis=-1)
+    mask = tf.repeat(mask, repeats=boxes.shape[-1], axis=-1)
+    boxes = tf.where(mask, boxes, -tf.ones_like(boxes))
+
+    if info["ragged"]:
+        return to_ragged(result)
+
+    return result
--- a/keras_cv/bounding_box/mask_invalid_detections.py
+++ b/keras_cv/bounding_box/mask_invalid_detections.py
@@ -31,6 +31,13 @@ def mask_invalid_detections(bounding_boxes):
         bounding_boxes: a dictionary complying with KerasCV bounding box format.  In
             addition to the normal required keys, these boxes are also expected to have
             a `num_detections` key.
+
+    Returns:
+        bounding boxes with proper masking of the boxes according to `num_detections`.
+        This allows proper interop with `tf.image.combined_non_max_suppression`.
+        Returned boxes match the specification fed to the function, so if the bounding
+        box tensor uses `tf.RaggedTensor` to represent boxes the returned value will
+        also return `tf.RaggedTensor` representations.
     """
     # ensure we are complying with KerasCV bounding box format.
     info = validate_format(bounding_boxes)
--- a/keras_cv/bounding_box/mask_invalid_detections.py
+++ b/keras_cv/bounding_box/mask_invalid_detections.py
@@ -61,7 +61,9 @@ def mask_invalid_detections(bounding_boxes, output_ragged=False):
 
     # Create a mask to select only the first N boxes from each batch
     mask = tf.repeat(
-        tf.expand_dims(tf.range(boxes.shape[1]), axis=0), repeats=boxes.shape[0], axis=0
+        tf.expand_dims(tf.range(tf.shape(boxes)[1]), axis=0),
+        repeats=tf.shape(boxes)[0],
+        axis=0,
     )
     mask = mask < num_detections[:, None]
 
--- a/keras_cv/bounding_box/mask_invalid_detections.py
+++ b/keras_cv/bounding_box/mask_invalid_detections.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 import tensorflow as tf
 
-from keras_cv.bounding_box.to_dense import to_dense
 from keras_cv.bounding_box.to_ragged import to_ragged
 from keras_cv.bounding_box.validate_format import validate_format
 
--- a/keras_cv/bounding_box/mask_invalid_detections.py
+++ b/keras_cv/bounding_box/mask_invalid_detections.py
@@ -18,20 +18,23 @@ from keras_cv.bounding_box.to_ragged import to_ragged
 from keras_cv.bounding_box.validate_format import validate_format
 
 
-def mask_invalid_detections(bounding_boxes):
+def mask_invalid_detections(bounding_boxes, output_ragged=False):
     """masks out invalid detections with -1s.
 
     This utility is mainly used on the output of `tf.image.combined_non_max_suppression`
-    operations.  The output of `tf.image.combined_non_max_suppression` contains padding
-    values of 0 instead of -1.  KerasCV expects all bounding boxes to be padded with
-    -1s instead of 0s.  This function uses the value of `num_detections` to mask out
+    operations.  The output of `tf.image.combined_non_max_suppression` contains all of
+    the detections, even invalid ones.  Users are expected to use `num_detections` to
+    determine how many boxes are in each image.
+
+    In contrast, KerasCV expects all bounding boxes to be padded with -1s.
+    This function uses the value of `num_detections` to mask out
     invalid boxes with -1s.
 
     Args:
         bounding_boxes: a dictionary complying with KerasCV bounding box format.  In
             addition to the normal required keys, these boxes are also expected to have
             a `num_detections` key.
-
+        output_ragged: whether or not to output RaggedTensor based bounding boxes.
     Returns:
         bounding boxes with proper masking of the boxes according to `num_detections`.
         This allows proper interop with `tf.image.combined_non_max_suppression`.
@@ -41,14 +44,17 @@ def mask_invalid_detections(bounding_boxes):
     """
     # ensure we are complying with KerasCV bounding box format.
     info = validate_format(bounding_boxes)
+    if info["ragged"]:
+        raise ValueError(
+            "`bounding_box.mask_invalid_detections()` requires inputs to be Dense "
+            "tensors.  Please call `bounding_box.to_dense(bounding_boxes)` before "
+            "passing your boxes to `bounding_box.mask_invalid_detections()`."
+        )
     if "num_detections" not in bounding_boxes:
         raise ValueError(
             "`bounding_boxes` must have key 'num_detections' "
             "to be used with `bounding_box.mask_invalid_detections()`."
         )
-
-    # required to use tf.where()
-    bounding_boxes = to_dense(bounding_boxes)
 
     boxes = bounding_boxes.get("boxes")
     classes = bounding_boxes.get("classes")
@@ -71,7 +77,7 @@ def mask_invalid_detections(bounding_boxes):
     result["boxes"] = boxes
     result["classes"] = classes
 
-    if info["ragged"]:
+    if output_ragged:
         return to_ragged(result)
 
     return result
--- a/keras_cv/bounding_box/mask_invalid_detections_test.py
+++ b/keras_cv/bounding_box/mask_invalid_detections_test.py
@@ -39,8 +39,10 @@ class MaskInvalidDetectionsTest(tf.test.TestCase):
                 [tf.random.uniform((5, 4)), tf.random.uniform((10, 4))]
             ),
             "num_detections": tf.constant([2, 3]),
-            "classes": tf.ragged.stack([tf.random.uniform((5,)), tf.random.uniform((5,))]),
+            "classes": tf.ragged.stack([tf.random.uniform((5,)), tf.random.uniform((10,))]),
         }
 
         result = bounding_box.mask_invalid_detections(bounding_boxes)
-        self.assertTrue(isinstance(result, tf.RaggedTensor))
+        self.assertTrue(isinstance(result['boxes'], tf.RaggedTensor))
+        self.assertEqual(result['boxes'][0].shape[0], 2)
+        self.assertEqual(result['boxes'][1].shape[0], 3)
--- a/keras_cv/bounding_box/mask_invalid_detections_test.py
+++ b/keras_cv/bounding_box/mask_invalid_detections_test.py
@@ -39,7 +39,7 @@ class MaskInvalidDetectionsTest(tf.test.TestCase):
                 [tf.random.uniform((5, 4)), tf.random.uniform((10, 4))]
             ),
             "num_detections": tf.constant([2, 3]),
-            "classes": tf.ragged.stack([tf.random.uniform((5)), tf.random.uniform((5))]),
+            "classes": tf.ragged.stack([tf.random.uniform((5,)), tf.random.uniform((5,))]),
         }
 
         result = bounding_box.mask_invalid_detections(bounding_boxes)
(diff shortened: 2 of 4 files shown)


keras_cv/bounding_box/mask_invalid_detections.py
```diff
@@ -0,0 +1,70 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+
+from keras_cv.bounding_box.to_dense import to_dense
+from keras_cv.bounding_box.to_ragged import to_ragged
+from keras_cv.bounding_box.validate_format import validate_format
+
+
```
bhack, line 20: Can we test this in graph mode (tf.function)?

LukeWood: Good idea - will add now.

keras_cv/bounding_box/mask_invalid_detections_test.py
```diff
@@ -0,0 +1,75 @@
...
+        self.assertAllClose(boxes_from_image_3, bounding_boxes["boxes"][2, :4, :])
+
+    def test_correctly_masks_based_on_max_dets_in_graph(self):
+        bounding_boxes = {
+            "boxes": tf.random.uniform((4, 100, 4)),
+            "num_detections": tf.constant([2, 3, 4, 2]),
+            "classes": tf.random.uniform((4, 100)),
+        }
+
+        @tf.function()
```
bhack, line 46: Can we reuse Keras decorators https://github.com/keras-team/keras/blob/master/keras/feature_column/dense_features_v2_test.py#L39-L40

It will very useful also for keras-cv tests.

keras_cv/bounding_box/mask_invalid_detections.py
```diff
@@ -0,0 +1,77 @@
...
+    This utility is mainly used on the output of `tf.image.combined_non_max_suppression`
+    operations.  The output of `tf.image.combined_non_max_suppression` contains padding
+    values of 0 instead of -1.  KerasCV expects all bounding boxes to be padded with
+    -1s instead of 0s.  This function uses the value of `num_detections` to mask out
+    invalid boxes with -1s.
+
+    Args:
+        bounding_boxes: a dictionary complying with KerasCV bounding box format.  In
+            addition to the normal required keys, these boxes are also expected to have
+            a `num_detections` key.
```
ianstenbit, line 33: What is the benefit of this structure over making num_detections a separate parameter to this method?

keras_cv/bounding_box/mask_invalid_detections.py
```diff
@@ -0,0 +1,77 @@
...
+    # ensure we are complying with KerasCV bounding box format.
+    info = validate_format(bounding_boxes)
+    if "num_detections" not in bounding_boxes:
+        raise ValueError(
+            "`bounding_boxes` must have key 'num_detections' "
+            "to be used with `bounding_box.mask_invalid_detections()`."
+        )
+
+    # required to use tf.where()
+    bounding_boxes = to_dense(bounding_boxes)
```
ianstenbit, line 51: Shouldn't we expect the boxes to be dense as a precondition, since if they were ragged we wouldn't need to mask them this way?

ianstenbit (changes requested):

keras_cv/bounding_box/mask_invalid_detections.py
tanzhenyu: Let's try *NOT* to do this before we figured out the TPU compatible NMS layer?

tanzhenyu (changes requested):

keras_cv/bounding_box/mask_invalid_detections.py
LukeWood: Mainly to match the tf.image API & our existing APIs (coco callback, etc).  

I think keeping num_detections is better to minimize the delta of this PR, but we can remove the concept of num_detections in a follow up if we think thats better

keras_cv/bounding_box/mask_invalid_detections.py
LukeWood: sure, updated

keras_cv/bounding_box/mask_invalid_detections_test.py
LukeWood: I don't think these are exported in the Keras API as of now.  We can use them if we eventually export them but from now we are avoiding the `import keras` pattern.

bhack: Yes it is a `from keras.testing_infra import test_combinations`
If they don't want to expose these we could duplicate in our test utils.

LukeWood:
@tanzhenyu this is TPU compat: https://colab.research.google.com/drive/191C05Elf33CvmPWcU1dG8ag84J2g9trJ?usp=sharing

bhack:
Can you make the colab public?

LukeWood:
> Can you make the colab public?

no, but here's the code:

```
def test_in_xla():
  num_detections = tf.constant([5, 4, 2, 3])
  boxes = tf.random.uniform((4, 100, 4))
  classes = tf.random.uniform((4, 100))
  mask = tf.repeat(tf.expand_dims(tf.range(boxes.shape[1]), axis=0), repeats=boxes.shape[0], axis=0)
  mask = mask < num_detections[:, None]
  mask = tf.expand_dims(mask, axis=-1)
  mask = tf.repeat(mask, boxes.shape[-1], axis=-1)
  result = tf.where(mask, boxes, -tf.ones_like(boxes))
  return result
```

bhack:
> > Can you make the colab public?
> 
> no, but here's the code:
> 
> ```
> def test_in_xla():
>   num_detections = tf.constant([5, 4, 2, 3])
>   boxes = tf.random.uniform((4, 100, 4))
>   classes = tf.random.uniform((4, 100))
>   mask = tf.repeat(tf.expand_dims(tf.range(boxes.shape[1]), axis=0), repeats=boxes.shape[0], axis=0)
>   mask = mask < num_detections[:, None]
>   mask = tf.expand_dims(mask, axis=-1)
>   mask = tf.repeat(mask, boxes.shape[-1], axis=-1)
>   result = tf.where(mask, boxes, -tf.ones_like(boxes))
>   return result
> ```

> Let's try NOT to do this before we figured out the TPU compatible NMS layer?

Where is the NMS layer in your gist?

bhack:
> Let's try NOT to do this before we figured out the TPU compatible NMS layer?

Are we talking about this https://github.com/keras-team/keras-cv/pull/1021#issuecomment-1317925675?

tanzhenyu:
> @tanzhenyu this is TPU compat: https://colab.research.google.com/drive/191C05Elf33CvmPWcU1dG8ag84J2g9trJ?usp=sharing

But it's using several tf.shape which might become dynamic in certain cases.
Can you simply provide a colab that:
1) reproduces this layer exactly
2) turn soft device placement off?

tanzhenyu:
> > @tanzhenyu this is TPU compat: https://colab.research.google.com/drive/191C05Elf33CvmPWcU1dG8ag84J2g9trJ?usp=sharing
> 
> But it's using several tf.shape which might become dynamic in certain cases. Can you simply provide a colab that:
> 
> 1. reproduces this layer exactly
> 2. turn soft device placement off?

We need to make sure this is actually running on TPU

LukeWood:
this layer is already not TPU compatible.  This is a bug fix.  TPU compatibility is a separate issue.  

output format is wrong.  I'm fixing a bug.

tanzhenyu:
> this layer is already not TPU compatible. This is a bug fix. TPU compatibility is a separate issue.
> 
> output format is wrong. I'm fixing a bug.

We need to make sure we're not introducing more incompatible changes here

LukeWood:
> > this layer is already not TPU compatible. This is a bug fix. TPU compatibility is a separate issue.
> > output format is wrong. I'm fixing a bug.
> 
> We need to make sure we're not introducing more incompatible changes here

if this is the case, you can remove it when you make it TPU compatible

LukeWood:
> > this layer is already not TPU compatible. This is a bug fix. TPU compatibility is a separate issue.
> > output format is wrong. I'm fixing a bug.
> 
> We need to make sure we're not introducing more incompatible changes here

Did we do the same level of TPU testing for https://github.com/keras-team/keras-cv/commit/b6c48f85e5ad1f5d2c7a1a5feb5b3589050d1f64 ?

tanzhenyu:
> > > this layer is already not TPU compatible. This is a bug fix. TPU compatibility is a separate issue.
> > > output format is wrong. I'm fixing a bug.
> > 
> > 
> > We need to make sure we're not introducing more incompatible changes here
> 
> if this is the case, we can remove it when we make it TPU compatible...
> 
> we have a temporary API workaround for NMS being non-TPU compatible in the first place. why wouldn't we want it to at least output correct values?
> 
> This is a one off bug fix PR - TPU compatibility discussion is out of scope for this PR. When we make it TPU compatible, the output format should still be KerasCV native bounding box format. If getting to there requires removing this method, or modifying it then great! that's fine and a discussion for the PR that enables TPU compatibility.

I think you misunderstood the point:
1) I try to make sure the added code here is still TPU-compatible, this is irrelevant to what we have today (which is unfortunately also not TPU-compatible), so when we replace tf.image.combined_nms, this will not create extra trouble.
2) However, even with 1), there's still a risk, what if we cannot exactly replace tf.image.combined_nms, but we have to change the output so that it cannot be padded with custom value (-1)? In that case we will have to break backward compatibility once again and this work is wasted.

In any sense, I think it's better if we have a clear story for TPU-compatible combined NMS first, so we don't go back-and-forth on this problem -- the NMS layer was [introduced](https://github.com/keras-team/keras-cv/pull/576) in the first place with 2 layers of indirection (NMSPredictionDecoder layer + NonMaxSuppression layer) and its TPU-compatibility is unknown, I spent lots of time debugging, unifying and cleaning this up. I **REALLY** hope the next PR we make on this, we fix this problem once for all.

LukeWood:
> > > > this layer is already not TPU compatible. This is a bug fix. TPU compatibility is a separate issue.
> > > > output format is wrong. I'm fixing a bug.
> > > 
> > > 
> > > We need to make sure we're not introducing more incompatible changes here
> > 
> > 
> > if this is the case, we can remove it when we make it TPU compatible...
> > we have a temporary API workaround for NMS being non-TPU compatible in the first place. why wouldn't we want it to at least output correct values?
> > This is a one off bug fix PR - TPU compatibility discussion is out of scope for this PR. When we make it TPU compatible, the output format should still be KerasCV native bounding box format. If getting to there requires removing this method, or modifying it then great! that's fine and a discussion for the PR that enables TPU compatibility.
> 
> I think you misunderstood the point:
> 
> 1. I try to make sure the added code here is still TPU-compatible, this is irrelevant to what we have today (which is unfortunately also not TPU-compatible), so when we replace tf.image.combined_nms, this will not create extra trouble.
> 2. However, even with 1), there's still a risk, what if we cannot exactly replace tf.image.combined_nms, but we have to change the output so that it cannot be padded with custom value (-1)? In that case we will have to break backward compatibility once again and this work is wasted.
> 
> In any sense, I think it's better if we have a clear story for TPU-compatible combined NMS first, so we don't go back-and-forth on this problem -- the NMS layer was [introduced](https://github.com/keras-team/keras-cv/pull/576) in the first place with 2 layers of indirection (NMSPredictionDecoder layer + NonMaxSuppression layer) and its TPU-compatibility is unknown, I spent lots of time debugging, unifying and cleaning this up. I **REALLY** hope the next PR we make on this, we fix this problem once for all.

what backwards compatibility would we break?   Lets make the utility private then?   Its only used in one place.  

what we have right now is broken and not TPU compatible.  I completely fail to see why you don't want me to merge this.  The existing behavior is wrong.  This fixes it.

tanzhenyu:
> > > > > this layer is already not TPU compatible. This is a bug fix. TPU compatibility is a separate issue.
> > > > > output format is wrong. I'm fixing a bug.
> > > > 
> > > > 
> > > > We need to make sure we're not introducing more incompatible changes here
> > > 
> > > 
> > > if this is the case, we can remove it when we make it TPU compatible...
> > > we have a temporary API workaround for NMS being non-TPU compatible in the first place. why wouldn't we want it to at least output correct values?
> > > This is a one off bug fix PR - TPU compatibility discussion is out of scope for this PR. When we make it TPU compatible, the output format should still be KerasCV native bounding box format. If getting to there requires removing this method, or modifying it then great! that's fine and a discussion for the PR that enables TPU compatibility.
> > 
> > 
> > I think you misunderstood the point:
> > 
> > 1. I try to make sure the added code here is still TPU-compatible, this is irrelevant to what we have today (which is unfortunately also not TPU-compatible), so when we replace tf.image.combined_nms, this will not create extra trouble.
> > 2. However, even with 1), there's still a risk, what if we cannot exactly replace tf.image.combined_nms, but we have to change the output so that it cannot be padded with custom value (-1)? In that case we will have to break backward compatibility once again and this work is wasted.
> > 
> > In any sense, I think it's better if we have a clear story for TPU-compatible combined NMS first, so we don't go back-and-forth on this problem -- the NMS layer was [introduced](https://github.com/keras-team/keras-cv/pull/576) in the first place with 2 layers of indirection (NMSPredictionDecoder layer + NonMaxSuppression layer) and its TPU-compatibility is unknown, I spent lots of time debugging, unifying and cleaning this up. I **REALLY** hope the next PR we make on this, we fix this problem once for all.
> 
> what backwards compatibility would we break? Lets make the utility private then? Its only used in one place.
> 
> I appreciate all of the effort that went into making this unified, but what we have right now is broken in terms of output format AND not TPU compatible. This is not a big change, this is not a one way change, and nothing is set in stone.
> 
> I completely fail to see why you don't want me to merge this. The existing behavior is wrong. This fixes it. It doesn't devalue the work done to unify, or add work if you want to make it TPU compatible. You can comment out or remove the last line of code when its time - lets make the utility private? Would that make you more comfortable with this change?

If you make this layer as a private `_MultiClassNMS`, then I'm fine with merging this. But you're still missing the point -- I'm saying what if TPU-compatible layer means we cannot pad with -1? Do we change input box format then? (Obviously I can try my best to avoid that, but I cannot guarantee)

LukeWood:
I'd like to get more eyes on this... from my perspective this is a clear bug fix and not a regression of any sort.  Right now we are outputting INVALID boxes to users with no indication that we deviate from the existing KerasCV format.  All underling ops used are TPU compatible.  I've tested in XLA.

perhaps more opinions would help: should we merge this?

LukeWood:
> > > > > > this layer is already not TPU compatible. This is a bug fix. TPU compatibility is a separate issue.
> > > > > > output format is wrong. I'm fixing a bug.
> > > > > 
> > > > > 
> > > > > We need to make sure we're not introducing more incompatible changes here
> > > > 
> > > > 
> > > > if this is the case, we can remove it when we make it TPU compatible...
> > > > we have a temporary API workaround for NMS being non-TPU compatible in the first place. why wouldn't we want it to at least output correct values?
> > > > This is a one off bug fix PR - TPU compatibility discussion is out of scope for this PR. When we make it TPU compatible, the output format should still be KerasCV native bounding box format. If getting to there requires removing this method, or modifying it then great! that's fine and a discussion for the PR that enables TPU compatibility.
> > > 
> > > 
> > > I think you misunderstood the point:
> > > 
> > > 1. I try to make sure the added code here is still TPU-compatible, this is irrelevant to what we have today (which is unfortunately also not TPU-compatible), so when we replace tf.image.combined_nms, this will not create extra trouble.
> > > 2. However, even with 1), there's still a risk, what if we cannot exactly replace tf.image.combined_nms, but we have to change the output so that it cannot be padded with custom value (-1)? In that case we will have to break backward compatibility once again and this work is wasted.
> > > 
> > > In any sense, I think it's better if we have a clear story for TPU-compatible combined NMS first, so we don't go back-and-forth on this problem -- the NMS layer was [introduced](https://github.com/keras-team/keras-cv/pull/576) in the first place with 2 layers of indirection (NMSPredictionDecoder layer + NonMaxSuppression layer) and its TPU-compatibility is unknown, I spent lots of time debugging, unifying and cleaning this up. I **REALLY** hope the next PR we make on this, we fix this problem once for all.
> > 
> > 
> > what backwards compatibility would we break? Lets make the utility private then? Its only used in one place.
> > I appreciate all of the effort that went into making this unified, but what we have right now is broken in terms of output format AND not TPU compatible. This is not a big change, this is not a one way change, and nothing is set in stone.
> > I completely fail to see why you don't want me to merge this. The existing behavior is wrong. This fixes it. It doesn't devalue the work done to unify, or add work if you want to make it TPU compatible. You can comment out or remove the last line of code when its time - lets make the utility private? Would that make you more comfortable with this change?
> 
> If you make this layer as a private `_MultiClassNMS`, then I'm fine with merging this. But you're still missing the point -- I'm saying what if TPU-compatible layer means we cannot pad with -1? Do we change input box format then? (Obviously I can try my best to avoid that, but I cannot guarantee)

https://colab.research.google.com/drive/191C05Elf33CvmPWcU1dG8ag84J2g9trJ?usp=sharing

Works on TPU.

LukeWood:
for community members:

TPU test:
```
import tensorflow as tf
print("Tensorflow version " + tf.__version__)

def test_in_xla():
  num_detections = tf.constant([5, 4, 2, 3])
  boxes = tf.random.uniform((4, 100, 4))
  classes = tf.random.uniform((4, 100))
  mask = tf.repeat(tf.expand_dims(tf.range(boxes.shape[1]), axis=0), repeats=boxes.shape[0], axis=0)
  mask = mask < num_detections[:, None]
  mask = tf.expand_dims(mask, axis=-1)
  mask = tf.repeat(mask, boxes.shape[-1], axis=-1)
  result = tf.where(mask, boxes, -tf.ones_like(boxes))
  return result

cluster_resolver = tf.distribute.cluster_resolver.TPUClusterResolver()
tf.config.experimental_connect_to_cluster(cluster_resolver)
tf.tpu.experimental.initialize_tpu_system(cluster_resolver)
strategy = tf.distribute.TPUStrategy(cluster_resolver)

x = tf.constant(1.)
y = tf.constant(1.)
z = strategy.run(test_in_xla)
print(z)
```

bhack:
Honestly it will be a lot easier if we have explicit expected failing test committed in the repo.

This is why adding few decorator it will help to cover eager, graph, XLA in tests. Then TPU could be edge enough if at least compiles.

LukeWood:
> Honestly it will be a lot easier if we have explicit expected failing test committed in the repo.
> 
> This is why adding few decorator it will help to cover eager, graph, XLA in tests. Then TPU could be edge enough if at least compiles.

TPU test decorators for OSS would also be generically useful.

bhack:
Free TPU runtimes are not available on colab right now.

LukeWood:
> Free TPU runtimes are not available on colab right now.

my personal wish would be an emulator or utility that you can just do `tf.debugging.assert_tpu_compatible(my_function)` , but thats probably a pipe dream 😢

bhack:
> TPU test decorators for OSS would also be generically useful.

Sure but I the specific TPU case where we run these test? Currently we don't have TPU in the OSS CI. 
So we need to tedious copy and paste tests manually on Colab where often TPU are unavailable (like now) and the trst decorator module could be not available in the public API like in Keras (in Colab keras-cv installed wheel). so we need to execute pytest from Keras-CV git clone on Colab if and when a TPU runtime is finally available .

Not the devenv we are dreaming.

It is why it easier to cover the other cases at least with decorators and expected failing test. It is better then where we are now and we could better interact with the bridge team on TF or experiment with the new bridge.

LukeWood:
> > TPU test decorators for OSS would also be generically useful.
> 
> Sure but in the specific TPU case where we run these decorated tests? Currently we don't have TPU in the OSS CI. So we need to tedious copy and paste tests manually on Colab where often TPU are unavailable (like now) and the decorator module could be not available in the public API like in Keras (in Colab keras-cv installed wheel). so we need to execute pytest from Keras-CV git clone on Colab (on the PR) if and when a TPU runtime is finally available .
> 
> Not the devenv we are dreaming.
> 
> It is why it easier to cover the other cases at least with decorators and expected failing test. It is better then where we are now and we could better interact with the bridge team on TF or experiment with the new bridge.

Happy to support you in a push for tooling to improve that..  I agree its important - but unsure who would be creating it :(. Probably out of scope for us given our resource allocation

bhack:
> my personal wish would be an emulator or utility that you can just do tf.debugging.assert_tpu_compatible(my_function) , but thats probably a pipe dream

I think that it could more easy then that. So not exactly a pipe dream it is just that more people need to push in the right direction:
https://github.com/openxla/stablehlo/issues/216#issuecomment-1263615220

But sometimes i feel like that I am the only one to try to cross interact between teams.

LukeWood:
> > my personal wish would be an emulator or utility that you can just do tf.debugging.assert_tpu_compatible(my_function) , but thats probably a pipe dream
> 
> I think that it could more easy then that. So not exactly a pipe dream it is just that more people need to push in the right direction: [openxla/stablehlo#216 (comment)](https://github.com/openxla/stablehlo/issues/216#issuecomment-1263615220)
> 
> But sometimes i feel like that I am the only one to try to cross interact between teams.

happy to push on this internally in any way I can.  I'll start a document and try to circulate it.

bhack:
> Happy to support you in a push for tooling to improve that.. I agree its important - but unsure who would be creating it :(. Probably out of scope for us given our resource allocation

You can talk with @martin-gorner as I could help if you are not too extremely constrained on resource allocation.

LukeWood:
> > > I'd like to get more eyes on this... from my perspective this is a clear bug fix and not a regression of any sort. Right now we are outputting INVALID boxes to users with no indication that we deviate from the existing KerasCV format. All underling ops used are TPU compatible. I've tested in XLA.
> > > perhaps more opinions would help: should we merge this?
> > 
> > 
> > I already said in earlier comments:
> > 
> > 1. How to make sure it’s running on TPU by disabling soft device placement
> > 2. I’m fine with merging it if we make the layer private.
> >    I hope the comments are useful
> 
> I mean't make the utility private, users need a way to customize decoding. and ok, I will run it with soft device placement.

It works with tf.config.set_soft_device_placement(False).

tanzhenyu (approved):
Thanks for the patience. I love seeing getting things right than fast

Issue #1269, opened 2023-01-13
Add XLA expected failing test for NMS


bhack:
@tanzhenyu @LukeWood As this Colab is working https://github.com/keras-team/keras-cv/pull/1021#issuecomment-1317925675 could you try to add/share an expected failing test for you NMS case?

tanzhenyu:
Thanks for reporting it!

PR #1270, opened 2023-01-13
Test compilation of cv_layers.MultiClassNonMaxSuppression

bhack:
# What does this PR do?
It will try to test `MultiClassNonMaxSuppression` compilation as a proxy TPU test. 

Fixes https://github.com/keras-team/keras-cv/issues/1269

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@tanzhenyu @LukeWood @ianstenbit

### DIFF
--- a/keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
+++ b/keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
@@ -46,6 +46,7 @@ class NmsPredictionDecoderTest(tf.test.TestCase):
         self.assertEqual(result["classes"].shape, [8, 100])
         self.assertEqual(result["confidence"].shape, [8, 100])
 
+
 @unittest.expectedFailure
 class NmsPredictionDecoderTestWithXLA(tf.test.TestCase):
     def test_decode_predictions_output_shapes(self):
@@ -54,7 +55,8 @@ class NmsPredictionDecoderTestWithXLA(tf.test.TestCase):
         self.assertEqual(result["boxes"].shape, [8, 100, 4])
         self.assertEqual(result["classes"].shape, [8, 100])
         self.assertEqual(result["confidence"].shape, [8, 100])
-        
+
+
 class NmsPredictionDecoderTestWithXLAMlirBridge(tf.test.TestCase):
     def setUp(self):
         tf.config.experimental.enable_mlir_bridge()
--- a/keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
+++ b/keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
@@ -23,8 +23,8 @@ def decode_predictions_output_shapes():
     classes = 10
     predictions_shape = (8, 98208, 4 + classes)
 
-    predictions = tf.random.uniform(
-        shape=predictions_shape, minval=0.0, maxval=1.0, dtype=tf.float32
+    predictions = tf.random.stateless_uniform(
+        shape=predictions_shape, seed=(2, 3), minval=0.0, maxval=1.0, dtype=tf.float32
     )
     box_pred = predictions[..., :4]
     confidence_pred = predictions[..., 4:]
@@ -46,7 +46,6 @@ class NmsPredictionDecoderTest(tf.test.TestCase):
         self.assertEqual(result["classes"].shape, [8, 100])
         self.assertEqual(result["confidence"].shape, [8, 100])
 
-
 @unittest.expectedFailure
 class NmsPredictionDecoderTestWithXLA(tf.test.TestCase):
     def test_decode_predictions_output_shapes(self):
@@ -55,8 +54,7 @@ class NmsPredictionDecoderTestWithXLA(tf.test.TestCase):
         self.assertEqual(result["boxes"].shape, [8, 100, 4])
         self.assertEqual(result["classes"].shape, [8, 100])
         self.assertEqual(result["confidence"].shape, [8, 100])
-
-
+        
 class NmsPredictionDecoderTestWithXLAMlirBridge(tf.test.TestCase):
     def setUp(self):
         tf.config.experimental.enable_mlir_bridge()
--- a/keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
+++ b/keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
@@ -13,11 +13,65 @@
 # limitations under the License.
 
 import tensorflow as tf
+import unittest
 
 from keras_cv import layers as cv_layers
 
-
 class NmsPredictionDecoderTest(tf.test.TestCase):
+    def test_decode_predictions_output_shapes(self):
+        classes = 10
+        predictions_shape = (8, 98208, 4 + classes)
+
+        predictions = tf.random.uniform(
+            shape=predictions_shape, minval=0.0, maxval=1.0, dtype=tf.float32
+        )
+        box_pred = predictions[..., :4]
+        confidence_pred = predictions[..., 4:]
+
+        layer = cv_layers.MultiClassNonMaxSuppression(
+            bounding_box_format="xyxy",
+            from_logits=True,
+            max_detections=100,
+        )
+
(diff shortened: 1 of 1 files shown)


keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
```diff
@@ -39,3 +39,57 @@ def test_decode_predictions_output_shapes(self):
...
+        self.assertEqual(result["boxes"].shape, [8, 100, 4])
+        self.assertEqual(result["classes"].shape, [8, 100])
+        self.assertEqual(result["confidence"].shape, [8, 100])
+
+@unittest.expectedFailure
+class XlaNmsPredictionDecoderTest(tf.test.TestCase):
+    # TODO This is not failing as it seems if uncommented it will globally 
+    # disable the MLIR bridge. 
+    # Are we compiling in the same context?
+    # tf.config.experimental.disable_mlir_bridge()
```
bhack, line 73: @smit-hinsu @paynecl Is there a way to control the MLIR bridge for a single test?

jpienaar: Note I see other tests doing the equivalent of

```python
if (not tf_test_utils.is_mlir_bridge_enabled()):
  self.skipTest("TPU tests require MLIR bridge")
```

If this were running with both new bridge enabled and disabled then you'd test both paths without needing to disable yourself. But @paynecl is on rotation and would know best.

bhack: We have also `@test_util.disable_mlir_bridge` but I don't that we have these symbols exported. The problem is more about if and how it is possible to control the bridge activation/disactivation for a single test.

Here I am more going to test/confirm the expected failure with the "old" bridge and the test pass with the new MLIR bridge as we are not sure that we could already enable the new bridge. But controlling the bridge in single test it is still useful for test coverage.

keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
```diff
@@ -39,3 +41,60 @@ def test_decode_predictions_output_shapes(self):
         self.assertEqual(result["boxes"].shape, [8, 100, 4])
         self.assertEqual(result["classes"].shape, [8, 100])
         self.assertEqual(result["confidence"].shape, [8, 100])
+
+
+class XlaMlirBridgeNmsPredictionDecoderTest(tf.test.TestCase):
+    tf.config.experimental.enable_mlir_bridge()
```
LukeWood, line 47: @bhack is this more similar to TPU compilation than normal XLA?  (Just curious)

bhack: No, this about forcing to use the new MLIR bridge without fallback on the old one. 
I guess that almost all the development resources are landing on the new MLIR bridge instead of the old one. 

It is why sometimes we could pass the compilation with the new bridge but not with the old one.
Also I suppose that many type of BUG/FR in TF are not going to receive too much attention if submitted on the old bridge (but I could be wrong here).

As you can see TF nightly removed the [experimental status of this bridge](https://www.tensorflow.org/api_docs/python/tf/config/experimental/enable_mlir_bridge?version=nightly) but it is still not fully enabled by default.

bhack: Instead for an improved feedback to developers on what it is supported and what not when we are going to compile a function see our threads with @paynecl at:
https://github.com/openxla/stablehlo/issues/216#issuecomment-1263615220
https://github.com/tensorflow/tensorflow/pull/56510

LukeWood, line 47: @bhack you can use pytest fixtures to accomplish this!

Call:
https://www.tensorflow.org/api_docs/python/tf/config/experimental/disable_mlir_bridge

after a `yield` statement in the fixture to disabe.

bhack: Yes I've already tested that yesterday but the teardown it seems to not work correctly. 
I've pushed so you could check the code.

keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
```diff
@@ -39,3 +42,28 @@ def test_decode_predictions_output_shapes(self):
         self.assertEqual(result["boxes"].shape, [8, 100, 4])
         self.assertEqual(result["classes"].shape, [8, 100])
         self.assertEqual(result["confidence"].shape, [8, 100])
+
+    @unittest.expectedFailure
+    def test_with_xla(self):
+        # TODO if the teardown is working correctly the next line is not required
```
bhack, line 48: @LukeWood Here we need to explicit disable the new bridge again.

bhack:
@tanzhenyu Can you check that the test is really representative of how we are going to use the layer? Just to check that the compilation test has the real use case coverage.

bhack:
I've refactored the tests to not mix our `tf.test.TestCase`, as it is an `unittest`, with pytest fixtures. 
I don't think that we have specific policy for test but eventually we could still mix these two modules but we need to handle it with care (https://docs.pytest.org/en/7.1.x/how-to/unittest.html). 

Now,It is a bit strange that with the last formulation also the MLIR bridge is failing as it was working with the decorator and with the previous formulation:

```python
 OP_REQUIRES failed at xla_ops.cc:362 : INVALID_ARGUMENT: TF to XLA legalization failed: <unknown>:0: error: The following operations cannot be legalized: tf.CombinedNonMaxSuppression (count: 1). These legalization failure(s) may be due to missing TF to HLO lowerings and/or unsupported attributes, etc.
```
See the full CI log for failure details.

keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
```diff
@@ -12,30 +12,62 @@
...
+    confidence_pred = predictions[..., 4:]
+
+    layer = cv_layers.MultiClassNonMaxSuppression(
+        bounding_box_format="xyxy",
+        from_logits=True,
+        max_detections=100,
+    )
+
+    result = layer(box_prediction=box_pred, confidence_prediction=confidence_pred)
+    return result
```
bhack, line 39: The only difference now is that we expect to return the output from the compiled function. This is triggered the failure with the MLIR bridge.

tanzhenyu (approved):
The general method here is fine with me

keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
bhack: https://www.tensorflow.org/xla/known_issues?hl=en#random_number_generation_ignores_tf_seed

bhack:
Do we need something else here?

LukeWood:
> Do we need something else here?

Let me check it out on my work computer - then can merge if all looks good.

keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
```diff
@@ -12,30 +12,62 @@
...
 # limitations under the License.
 
+import unittest
+
 import tensorflow as tf
 
 from keras_cv import layers as cv_layers
 
 
+def decode_predictions_output_shapes():
```
LukeWood, line 22: any reason to make this a global helper function?  

Can you define it inside of the unit test just to keep the definition local to the test?  So define it in the test body, then compile it.

LukeWood (approved):
Approved with a single Nit.

keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
bhack: Where do you want to move?
As It is shared between 3 test classes.

LukeWood: ah ok - then lets just leave it.

LukeWood:
Thank you @bhack !   Really appreciate this change!

Issue #1277, opened 2023-01-15
Getting Import Error


IMvision12:
I was running some test files locally but I am facing this error
ImportError: cannot import name 'compute_pycoco_metrics' from 'keras_cv.metrics.coco' (keras-cv\keras_cv\metrics\coco\__init__.py)

does the imports in pycoco_callback.py `from keras_cv.metrics.coco import compute_pycoco_metrics` is wrong?
@tanzhenyu

LukeWood:
Thanks for the bug report, I'll check it out!

arnavrneo:
@LukeWood Don't you think an import statement (`from keras_cv.metrics.coco.pycoco_wrapper import compute_pycoco_metrics`) is missing in `__init__.py` of `metrics` directory.

bhack:
Why the CI tests are not broken?

tomrtk:
> Why the CI tests are not broken?

looks to be because in CI it installs the `test` extras, where `pycocotools` is pulled in. 

https://github.com/keras-team/keras-cv/blob/742fa2caa34b2f5772510955a22ec5e0ceeda2be/.github/workflows/actions.yml#L36

Installing `pycocotools` fixes the import errors, so it should maybe be added as a dependencie in `setup.py`?

bhack:
Or we need to mark some tests as requiring extra.

bhack:
@ianstenbit Is the contribution guide aligned with the CI receipt?

ianstenbit:
> @ianstenbit Is the contribution guide aligned with the CI receipt?

@bhack I'm not sure I understand the question -- can you elaborate? By my understanding, #1279 should fix this

bhack:
I meant are we executing these tests subset in the CI without these dependencies installed? Do we supposed that the contributor devenv could have these not installed following our Contributing makrdown?

PR #1279, opened 2023-01-16
Fix callback import error

LukeWood:
Before this:

```
>>> import keras_cv
You do not have pycocotools installed, so KerasCV pycoco metrics are not available. Please run `pip install pycocotools`.
  File "<stdin>", line 1, in <module>
  File "/Users/yeti/workspace/keras-cv/keras_cv/__init__.py", line 21, in <module>
    from keras_cv import callbacks
  File "/Users/yeti/workspace/keras-cv/keras_cv/callbacks/__init__.py", line 14, in <module>
    from keras_cv.callbacks.pycoco_callback import PyCOCOCallback
  File "/Users/yeti/workspace/keras-cv/keras_cv/callbacks/pycoco_callback.py", line 18, in <module>
    from keras_cv.metrics.coco import compute_pycoco_metrics
ImportError: cannot import name 'compute_pycoco_metrics' from 'keras_cv.metrics.coco' (/Users/yeti/workspace/keras-cv/keras_cv
/metrics/coco/__init__.py)
```

fixxes #1277

### DIFF
--- a/keras_cv/callbacks/__init__.py
+++ b/keras_cv/callbacks/__init__.py
@@ -11,7 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from keras_cv.callbacks.pycoco_callback import PyCOCOCallback
+try:
+    from keras_cv.callbacks.pycoco_callback import PyCOCOCallback
+except ImportError:
+    print(
+        "You do not have pyococotools installed, so the `PyCOCOCallback` API is not available."
+    )
 
 try:
     from keras_cv.callbacks.waymo_evaluation_callback import WaymoEvaluationCallback


LukeWood:
@tanzhenyu @ianstenbit feel free to merge if this looks good to you.

keras_cv/callbacks/__init__.py
```diff
@@ -11,7 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from keras_cv.callbacks.pycoco_callback import PyCOCOCallback
+try:
+    from keras_cv.callbacks.pycoco_callback import PyCOCOCallback
+except ImportError:
+    print(
```
bhack, line 17: Probably we could use `ImportWarning` instead of a `print`.

bhack:
Are we always going to emit these warning just for importing keras-cv or when we import the specific callbacks with missing modules dependencies?

LukeWood:
> Are we always going to emit these warning just for importing keras-cv or when we import the specific callbacks with missing modules dependencies?

it would be great to only print when users import specific modules or symbols but the reality is that’s probably a pretty big engineering problem and I think the maintanence burden would likely exceed the value.

keras_cv/callbacks/__init__.py
ianstenbit: this sgtm -- Luke wdyt?

LukeWood: lets just keep it simple for now: https://github.com/keras-team/keras-cv/blob/master/keras_cv/metrics/coco/__init__.py#L21

I'd rather fix this across the board in one PR than deviate between the import files.  Lets keep this focused to fixing master for now.

ianstenbit (approved):

PR #1289, opened 2023-01-18
Fix the pascal VOC loader to use the dictionary format

LukeWood:
also makes some other small API updates:

- remove img_size argument, this should be done by users using their own Resizing layer
- add support for 2012 split so we can replicate SOTA results

### DIFF
--- a/keras_cv/datasets/pascal_voc/load.py
+++ b/keras_cv/datasets/pascal_voc/load.py
@@ -16,7 +16,6 @@ import tensorflow as tf
 import tensorflow_datasets as tfds
 
 from keras_cv import bounding_box
-from keras_cv import layers as cv_layers
 
 
 def curry_map_function(bounding_box_format):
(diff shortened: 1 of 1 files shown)


keras_cv/datasets/pascal_voc/load.py
```diff
@@ -79,20 +69,26 @@ def load(
         shuffle: whether or not to shuffle the dataset.  Defaults to True.
         shuffle_buffer: the size of the buffer to use in shuffling.
         shuffle_files: (Optional) whether or not to shuffle files, defaults to True.
-        img_size: (Optional) size to resize the images to, if not provided image batches
-            will be of type `tf.RaggedTensor`.
+        dataset: (Optional) the PascalVOC dataset to load from.  Should be either
```
ianstenbit, line 72: let's mention what the default is

ianstenbit (approved):
Do affected example scripts need to be updated?

PR #1291, opened 2023-01-18
Make `include_rescaling` and `include_top`  required positionals

LukeWood:
to quickly describe why one is better:

```
ResNet50V2(True, False)
# OR
ResNet50V2(include_rescaling=True, include_top=False)
```

### DIFF
--- a/keras_cv/models/convnext.py
+++ b/keras_cv/models/convnext.py
@@ -386,6 +386,7 @@ def ConvNeXt(
 
 
 def ConvNeXtTiny(
+    *,
     include_rescaling,
     include_top,
     drop_path_rate,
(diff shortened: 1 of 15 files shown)


keras_cv/models/convnext.py
```diff
@@ -386,6 +386,7 @@ def ConvNeXt(
 
 
 def ConvNeXtTiny(
+    *,
```
ianstenbit, line 389: Part of me wonders if, given the infrequency with which this language feature seems to be used, we should include a comment?

ianstenbit (approved):
In the description: s/positionals/kwargs

Are there really no tests or examples broken by this change? If so, great.
I could have sworn that I had written a test at some point where I didn't include the names of the include_* args in a backbone

LukeWood:
@ianstenbit I'll manually run model tests now to verify!

LukeWood:
(FYI @mattdangerw and @jbischof - no action needed, but thought you guys may be interested to see that we are doing this)

PR #1293, opened 2023-01-19
Implement RepeatedAugmentation as a KerasCV API

LukeWood:
more reading and fixes https://github.com/keras-team/keras-cv/issues/372

@sayakpaul @altaal

---

TODOS:

- [ ] name the api (RepeatedAugmentation I'm not sold on)

### DIFF
--- a/keras_cv/layers/preprocessing/repeated_augmentation.py
+++ b/keras_cv/layers/preprocessing/repeated_augmentation.py
@@ -35,6 +35,21 @@ class RepeatedAugmentation(BaseImageAugmentationLayer):
             shuffle: whether or not to shuffle the result.  Essential when using an
                 asynchronous distribution strategy such as ParameterServerStrategy.
 
+        Usage:
+        ```python
+        repeated_augment = cv_layers.RepeatedAugmentation(
+            augmenters=[
+                cv_layers.RandAugment(value_range=(0, 255)),
+                cv_layers.RandomFlip(),
+            ]
+        )
+        inputs = {
+            "images": tf.ones((8, 512, 512, 3)),
+            "labels": tf.ones((8,)),
+        }
+        outputs = repeated_augment(inputs)
+        ```
+
         References:
         - [DEIT implementaton](https://github.com/facebookresearch/deit/blob/ee8893c8063f6937fec7096e47ba324c206e22b9/samplers.py#L8
     )
--- a/keras_cv/layers/preprocessing/repeated_augmentation.py
+++ b/keras_cv/layers/preprocessing/repeated_augmentation.py
@@ -11,8 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from tensorflow import keras
 import tensorflow as tf
+from tensorflow import keras
 
 from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
     BaseImageAugmentationLayer,
--- a/keras_cv/layers/preprocessing/repeated_augmentation.py
+++ b/keras_cv/layers/preprocessing/repeated_augmentation.py
@@ -29,6 +29,8 @@ class RepeatedAugmentation(BaseImageAugmentationLayer):
     using smaller gradients.  RepeatedAugmentation handles this by re-using the same
     image multiple times within a batch creating correlated samples.
 
+    This layer increases your batch size by a factor of `len(augmenters)`.
+
     Args:
         augmenters: the augmenters to use to augment the image
         shuffle: whether or not to shuffle the result.  Essential when using an
@@ -38,7 +40,6 @@ class RepeatedAugmentation(BaseImageAugmentationLayer):
 
     List of identical augmenters:
     ```python
-    ```python
     repeated_augment = cv_layers.RepeatedAugmentation(
         augmenters=[cv_layers.RandAugment(value_range=(0, 255))] * 8
     )
@@ -47,6 +48,7 @@ class RepeatedAugmentation(BaseImageAugmentationLayer):
         "labels": tf.ones((8,)),
     }
     outputs = repeated_augment(inputs)
+    # outputs now has a batch size of 64 because there are 8 augmenters
     ```
 
     List of distinct augmenters:
--- a/keras_cv/layers/preprocessing/repeated_augmentation.py
+++ b/keras_cv/layers/preprocessing/repeated_augmentation.py
@@ -78,7 +78,7 @@ class RepeatedAugmentation(BaseImageAugmentationLayer):
         self.shuffle = shuffle
 
     def _batch_augment(self, inputs):
-        if 'bounding_boxes' in inputs:
+        if "bounding_boxes" in inputs:
             raise ValueError(
                 "RepeatedAugmentation() does not yet support bounding box labels."
             )
--- a/keras_cv/layers/preprocessing/repeated_augmentation.py
+++ b/keras_cv/layers/preprocessing/repeated_augmentation.py
@@ -30,6 +30,11 @@ class RepeatedAugmentation(BaseImageAugmentationLayer):
         using smaller gradients.  RepeatedAugmentation handles this by re-using the same
         image multiple times within a batch creating correlated samples.
 
+        Args:
+            augmenters: the augmenters to use to augment the image
+            shuffle: whether or not to shuffle the result.  Essential when using an
+                asynchronous distribution strategy such as ParameterServerStrategy.
+
         References:
         - [DEIT implementaton](https://github.com/facebookresearch/deit/blob/ee8893c8063f6937fec7096e47ba324c206e22b9/samplers.py#L8
     )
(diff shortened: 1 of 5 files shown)


LukeWood:
@ianstenbit @fchollet @sayakpaul I'd appreciate any input on names for this component, but if nothing better I think "RepeatedAugmentation" at least gets colloquial usage so its not too bad.

bhack:
For the performance aspect of the original ticket does this mean that we still need to randomize each element of the batch with different sampled random params?
Or could we apply the same transformation param grouping different images?

LukeWood:
> For the performance aspect of the original ticket does this mean that we still need to randomize each element of the batch with different sampled random params?

As of now every item is randomized - and then concatenated.  This is the most logical approach I believe - as it just relies on the behavior of a KerasCV augmentation layer under the hood (i.e. RandAugment) to handle this.

Do we have any specific concerns with that approach?

bhack:
If we can use the same random parameters for sub-batches of different images we could mitigate a bit the heavy and well known performance overhead we have in many preprocessing layers (for the well documented  tf.image API design and well documented vectorized map fallbacks).

So the point is: if we repeat images in the batch why we cannot repeat transformations params on different images?

keras_cv/layers/preprocessing/repeated_augmentation.py
```diff
@@ -0,0 +1,115 @@
...
+
+        Args:
+            augmenters: the augmenters to use to augment the image
+            shuffle: whether or not to shuffle the result.  Essential when using an
+                asynchronous distribution strategy such as ParameterServerStrategy.
+
+        Usage:
+        ```python
+        repeated_augment = cv_layers.RepeatedAugmentation(
+            augmenters=[
```
ianstenbit, line 40: I would expect with this kind of functionality that it's somewhat common for the N augmenters to be identical -- is that true?

If so, we should strive to make the API support that without having to pass a list of N of the same object

keras_cv/layers/preprocessing/repeated_augmentation.py
```diff
@@ -0,0 +1,115 @@
...
+            raise ValueError(
+                "RepeatedAugmentation expects inputs in a dictionary with format "
+                '{"images": images, "labels": labels}.'
+                f"Got: inputs = {inputs}"
+            )
+
+        image_results = []
+        labels_results = []
+
+        for augmenter in self.augmenters:
```
ianstenbit, line 84: I may be misunderstanding here, but can't we do away with some of the special-casing keys here with something like

```
augmenter_outputs = [augmenter(inputs) for augmenter in self.augmenters]

outputs = {}
for k in input.keys():
    outputs[k] = tf.concat([batch[k] for batch in augmenter_outputs, axis=0)
```

ianstenbit (changes requested):
Do I understand correctly from reading the code that if I input a batch of N images and I am using 3 augmenters, the resulting batch has 3N images in it?

If so, I wonder if this should be renamed to include the word "Duplicate" in some fashion? Or perhaps to just explain this behavior in the docstring.

bhack:
@martin-gorner
What do you think here about our thread at https://github.com/tensorflow/tensorflow/issues/55639?

As we are repeating the same images with different augmentations in the same batch is it quite different from repeating the same augmentation param with different images?

LukeWood:
> As we are repeating the same images with different augmentations in the same batch is it quite different from repeating the same augmentation param with different images?

Whats the benefit of repeating the same augmentation parameters?  I think that a specialized layer is probably the right choice for this, directly using `get_random_transformation()` and the `augment_*` methods.  It would be a bit unintuitive to reuse random samples across the batch for the augmentations.

keras_cv/layers/preprocessing/repeated_augmentation.py
LukeWood: Good point, my thought for this was:

```python
[cv_layers.RandAugment(value_range=(0, 255)] * 8
```

Which works.  but maybe a specialized API surface would be better?  Open to feedback/ideas on this for sure - it is absolutely a common use case.

keras_cv/layers/preprocessing/repeated_augmentation.py
LukeWood: doesn't work for bounding boxes or any other custom datatypes:

```
tf.concat(batch['bounding_boxes'])
# => cannot concatenate a dictionary object
```

Perhaps we ship it with special casing for bounding boxes but other than that just use:

```
outputs = {}
for k in input.keys():
    outputs[k] = tf.concat([batch[k] for batch in augmenter_outputs, axis=0)
```
?

LukeWood:
I agree this is an important issue, but I'm not sure I understand the issue with respect to this specific layer in that this layer will invoke each augmentation on the batch axis, and the sub-layers would be the issue primarily

bhack:
> My though process is that in this layer, we will invoke each augmentation on the batch axis, and the sub-layers would be the issue primarily

The perimeter here is that these sub-layers could be not be aware anymore about the sample "identity" as it is lost. Right?

So it will be eventually impossible to organize/rearrange "virtual" sub-batches using the sample identity.

I hope it is clear now.

LukeWood:
> > Do I understand correctly from reading the code that if I input a batch of N images and I am using 3 augmenters, the resulting batch has 3N images in it?
> > If so, I wonder if this should be renamed to include the word "Duplicate" in some fashion? Or perhaps to just explain this behavior in the docstring.
> 
> Bumping this comment -- @LukeWood

lets take the docstring approach.

LukeWood:
> > > Do I understand correctly from reading the code that if I input a batch of N images and I am using 3 augmenters, the resulting batch has 3N images in it?
> > > If so, I wonder if this should be renamed to include the word "Duplicate" in some fashion? Or perhaps to just explain this behavior in the docstring.
> > 
> > 
> > Bumping this comment -- @LukeWood
> 
> lets take the docstring approach.

updated

keras_cv/layers/preprocessing/repeated_augmentation.py
```diff
@@ -0,0 +1,113 @@
...
+        self.augmenters = augmenters
+        self.shuffle = shuffle
+
+    def _batch_augment(self, inputs):
+        if 'bounding_boxes' in inputs:
+            raise ValueError(
+                "RepeatedAugmentation() does not yet support bounding box labels."
+            )
+
+        augmenter_outputs = [augmenter(inputs) for augmenter in self.augmenters]
```
bhack, line 86: Ok here eventually we have the opportunity to tell to the augmenter to not randomize the augmentation param within the batch.

keras_cv/layers/preprocessing/repeated_augmentation.py
```diff
@@ -0,0 +1,113 @@
...
+        if "bounding_boxes" in inputs:
+            raise ValueError(
+                "RepeatedAugmentation() does not yet support bounding box labels."
+            )
+
+        augmenter_outputs = [augmenter(inputs) for augmenter in self.augmenters]
+
+        outputs = {}
+        for k in inputs.keys():
+            outputs[k] = tf.concat([batch[k] for batch in augmenter_outputs], axis=0)
```
altaal, line 90: Does this line concatenate images in batches of an epoch concatenated?
I dont get the concatenation. May I ask if you can add a comment for this part?

altaal: In the original implementations of the repeated aug, there is a hyper parameter of number of repeats. Basically, how many times an augmented version of an image would exist inside a batch. I dont see that in this implementations.

More information:
Ignoring repeated aug, Augmentations are two parts:
primary augmentations that happen to all images (https://github.com/facebookresearch/deit/blob/ee8893c8063f6937fec7096e47ba324c206e22b9/augment.py#L98).
secondary augmentations that has a randomness inside. This will ensure that augmentations of an image are not the same (https://github.com/facebookresearch/deit/blob/ee8893c8063f6937fec7096e47ba324c206e22b9/augment.py#L111).

LukeWood: An update from chat: we will meet on Tuesday to discuss more details of how RepeatedAugmentation is used and will figure out how that impacts the API.

bhack: > An update from chat: we will meet on Tuesday to discuss more details of how RepeatedAugmentation is used and will figure out how that impacts the API.

As it seems they are using torch vision APJ are they going to apply the same transformation param within the batch? 
See the referecend doc and the Kornia option:
https://discuss.pytorch.org/t/random-batch-transforms/134173

keras_cv/layers/preprocessing/repeated_augmentation.py
bhack: Check my comment at https://github.com/keras-team/keras-cv/pull/1293#discussion_r1083533391

keras_cv/layers/preprocessing/repeated_augmentation.py
LukeWood, line 90: batch is bad name - update to output

LukeWood: > As it seems they are using torch vision API are they going to apply the same transformation param within the batch?

Makes sense, but I do not have the engineering resources at this time to iron this out for this layer.  In the future I agree important - but I'll sort out the high priority KPL performances first.

bhack: > but I'll sort out the high priority KPL performances first.

What do you mean?

LukeWood:
@ianstenbit should be ready!

altaal (approved):

ianstenbit (approved):
Thanks Luke!

Issue #1010, opened 2022-11-14
Reorganize [ops] and [custom_ops]


LukeWood:
Currently, we have some tech debt here: ops contains some custom layers as well as some ops.  We should really fix this and put the components that are like layers under layers instead of ops.  Then we can merge custom_ops into the ops directory.

Its confusing to have both.

bhack:
Do you have any example?

LukeWood:
`custom_ops`: https://github.com/keras-team/keras-cv/tree/master/keras_cv/custom_ops

should live under `ops` because it contains tensorflow ops.

ops on the other hand contains a ton of stuff (BoxMatcher, etc), but also IoU3D which is a custom TF op: 

https://github.com/keras-team/keras-cv/tree/master/keras_cv/ops

innat:
Why [tensorflow ops](https://github.com/keras-team/keras-cv/tree/master/keras_cv/custom_ops) is here in the first place? Also it looks [bazel](https://github.com/keras-team/keras-cv#installing-custom-ops-from-source) is reuqired to build custom ops. It looks weird to see these in here as well.

bhack:
> Why [tensorflow ops](https://github.com/keras-team/keras-cv/tree/master/keras_cv/custom_ops) is here in the first place? Also it looks [bazel](https://github.com/keras-team/keras-cv#installing-custom-ops-from-source) is reuqired to build custom ops. It looks weird to see these in here as well.

What do you mean?

innat:
I mean, why do we need to write C code for some custom operations (ie. `PairwiseIou3D`?

LukeWood:
Still a few utils to fix

PR #1299, opened 2023-01-23
[1/3~] Begin cleaning up ops namespace

LukeWood:
Begin moving some things out of ops namespace so that the custom ops namespace can be converged into one (either `op` or `custom_ops`).  Maybe we need a nice home for shared utilities (target_gather) - but for now I'm just leaving it in keras_cv/layers/bounding_box/

Fixes https://github.com/keras-team/keras-cv/issues/1010


keras_cv/layers/object_detection/target_gather.py
```diff
@@ -12,16 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional
-
 import tensorflow as tf
 
 
 def _target_gather(
```
ianstenbit, line 18: I don't think this belongs in `layers` -- maybe `utils`?

ianstenbit (changes requested):

keras_cv/layers/object_detection/target_gather.py
LukeWood: keras_cv/utils/ for now?  I had it in here because  it is only used by OD layers - perhaps a utils file in `keras_cv/layers/object_detection/utils.py`?

ianstenbit: We already have keras_cv/utils/...
Maybe we should put it there?

LukeWood: moved - done

ianstenbit (approved):
thanks Luke!

PR #1303, opened 2023-01-23
[3D Object Detection] Rename GlobalRandomFlipY -> GlobalRandomFlip

ianstenbit:
The layer still only supports Y-flipping, but now has a more consistent API surface compared to similar 2D layers

### DIFF
--- a/keras_cv/layers/preprocessing_3d/global_random_flip.py
+++ b/keras_cv/layers/preprocessing_3d/global_random_flip.py
@@ -23,10 +23,11 @@ BOUNDING_BOXES = base_augmentation_layer_3d.BOUNDING_BOXES
 
 
 @tf.keras.utils.register_keras_serializable(package="keras_cv")
-class GlobalRandomFlipY(base_augmentation_layer_3d.BaseAugmentationLayer3D):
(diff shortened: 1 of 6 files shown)


LukeWood:
Here's one question: do you think "Global" is a good name or should it be RandomFlip3D?

LukeWood (approved):

keras_cv/layers/preprocessing_3d/global_random_flip.py
```diff
@@ -44,11 +45,15 @@ class GlobalRandomFlipY(base_augmentation_layer_3d.BaseAugmentationLayer3D):
 
     """
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(self, axis=1, **kwargs):
```
LukeWood, line 48: Instead of axis why don't we use `y=True, x=False, z=False` and raise on x and Z?  thoughts?

ianstenbit: sgtm -- I've called them flip_x/y/z instead though. wdyt?

LukeWood (approved):

PR #1313, opened 2023-01-25
Update RetinaNet + RCNN to use 'classification' instead of cls

LukeWood:


keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -350,15 +350,15 @@ def compute_loss(self, images, boxes, classes, training):
         box_weights = positive_mask / normalizer
         y_true = {
             "box": boxes,
-            "cls": cls_labels,
+            "classification": cls_labels,
```
ianstenbit, line 353: should we also s/cls/class for variable names if we're making this change?

LukeWood: lets do it, but lets keep it self contained for now as this change is more important.  This change shows up in user facing APIs - variable names don't.

ianstenbit: Where does this show up in APIs? I assume it shows up in logs as a loss name?

ianstenbit (approved):

PR #1318, opened 2023-01-26
Update multi_class_non_max_suppression.py

LukeWood:
This will give much better results aesthetically to our users.

### DIFF
--- a/keras_cv/layers/object_detection/multi_class_non_max_suppression.py
+++ b/keras_cv/layers/object_detection/multi_class_non_max_suppression.py
@@ -43,7 +43,7 @@ class MultiClassNonMaxSuppression(tf.keras.layers.Layer):
         bounding_box_format,
         from_logits,
         iou_threshold=0.5,
-        confidence_threshold=0.5,
+        confidence_threshold=0.9,
         max_detections=100,
         max_detections_per_class=100,
         **kwargs,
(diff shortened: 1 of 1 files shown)


keras_cv/layers/object_detection/multi_class_non_max_suppression.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         bounding_box_format,
         from_logits,
         iou_threshold=0.5,
-        confidence_threshold=0.5,
+        confidence_threshold=0.9,
```
ianstenbit, line 46: Is this default pretty specific to RetinaNet, or is it also a sensible default for FasterRCNN + potentially other models?

LukeWood: From what I saw in @quantumalaviya 's work he ALSO tunes this to 0.9~ or so.  I think in YOLOx it may even be 0.99 actually.

In short: its probably not generically true, but its better than 0.5

ianstenbit: okay. Please update the docstring to reflect the new default.

ianstenbit (approved):
LG after updating docs

jbischof (approved):
Thanks @LukeWood! Do you have any updated colabs or samples to demonstrate the new results?

LukeWood:
> Thanks @LukeWood! Do you have any updated colabs or samples to demonstrate the new results?

https://github.com/keras-team/keras-io/pull/1211

PR #1322, opened 2023-01-27
Remove extra models from CI tests for ViT

ianstenbit:
This is consistent with how we're doing CI for other models to avoid running tons of expensive and redundant tests


LukeWood (approved):
LGTM thanks Ian

PR #1324, opened 2023-01-27
Fix typo in FasterRCNN compile error message

ianstenbit:


bhack:
Can we add a test on the critical section like in https://github.com/keras-team/keras-cv/pull/1270

LukeWood (approved):

LukeWood:
LGTM thanks

ianstenbit:
> Can we add a test on the critical section like in #1270

@bhack I'm not sure I follow -- do you mean a test asserting the contents of this error more specifically?

bhack:
I meant the critical section related to the metric. Is it failing under XLA compile?

ianstenbit:
> I meant the critical section related to the metric. Is it failing under XLA compile?

There isn't really a specific metric here to be tested. We're currently disallowing the use of all custom Metrics with OD models due to performance+compatability.

Issue #1256, opened 2023-01-09
Pre-release CI import test


LukeWood:
this job should:

- create a fresh environment
- pip install keras-cv
- run python, import keras_cv

If this fails the job succeeds.  This is basically a dependency test.

bhack:
As have this new ticket what about the dependency on Tensorflow?

Can you expose a bit why we check TF only at runtime?

Issue #455, opened 2022-05-25
Update performance benchmarking script to include a layer with ops that rely on fallbacks for autovectorization


LukeWood:
https://github.com/tensorflow/tensorflow/issues/56242

bhack:
https://github.com/tensorflow/tensorflow/issues/56242 was not going to importing keras-cv so it is related to Keras only.

I  think some preprocessing layers in Keras with 2.9 was refactored, like KLP layers in Keras-cv, on the new base class acquiring a randomization policy within the batch + vectorized_map (randomize the transformation for every single element in the batch).

Other then for missing converters  I think that the problem often is in that native ops that doesn't support "a batch" of different parameters:
https://github.com/tensorflow/tensorflow/issues/55639

As already exposed with  https://github.com/keras-team/keras-cv/issues/291 on nightly we have the new signature that will expose the root cause of the fallback: 
```python 
def vectorized_map(fn, elems, fallback_to_while_loop=True, warn=True):
```

If you want to really benchmark the performance gap here an in Keras refactored preprocessing layers I suppose what you really want to test is the batch augmentation (with fixed parameter) vs within batch augmentation.

Related:
https://github.com/keras-team/keras-cv/issues/372#issuecomment-1110827256

bhack:
P.s. just to make the history more consistent, for who is landing on this ticket directly, we started the discussion with @qlzh727 3 months ago at https://github.com/keras-team/keras-cv/pull/146.

Issue #291, opened 2022-04-10
vectorize KPLS: `tf.vectorized_map` fallback cause


bhack:
This is the current list of fallback we have on master with the related cause for our KPLs ~(it needs tensorflow/tensorflow#55562 or tf-nightly when merged)~:

`pytest -rP keras_cv/layers/preprocessing/ | grep while_loop | sort | uniq`

```python
WARNING  tensorflow:pfor.py:1082 Using a while_loop for converting AdjustHue cause Input "delta" of op 'AdjustHue' expected to be loop invariant.
WARNING  tensorflow:pfor.py:1082 Using a while_loop for converting AdjustSaturation cause Input "scale" of op 'AdjustSaturation' expected to be loop invariant.
WARNING  tensorflow:pfor.py:1082 Using a while_loop for converting DepthwiseConv2dNative cause Input "filter" of op 'DepthwiseConv2dNative' expected to be not loop invariant.
WARNING  tensorflow:pfor.py:1082 Using a while_loop for converting DynamicPartition cause there is no register converter for this op.
WARNING  tensorflow:pfor.py:1082 Using a while_loop for converting DynamicStitch cause there is no register converter for this op.
WARNING  tensorflow:pfor.py:1082 Using a while_loop for converting HistogramFixedWidth cause there is no register converter for this op.
WARNING  tensorflow:pfor.py:1082 Using a while_loop for converting ImageProjectiveTransformV3 cause there is no register converter for this op.
WARNING  tensorflow:pfor.py:1082 Using a while_loop for converting RandomShuffle cause there is no register converter for this op.
WARNING  tensorflow:pfor.py:1082 Using a while_loop for converting StridedSlice cause Input "input" of op 'StridedSlice' expected to be not loop invariant.
```

You can find exactly which test is involed running:
`pytest -rP keras_cv/layers/preprocessing/`  

Some of these could be fixed with a refactor in our code, others need a contribution (trivial or not) for a registered converter (e.g. See https://github.com/keras-team/keras-cv/issues/259#issuecomment-1094140155)

I think that we could care about this:
- monitor how many fallback we are accumulating directly in the CI, with a Github Action, so that we could check this at the PR stage
- contribute the (trivial?) converters that we care about in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/parallel_for/pfor.py#L920-L997
[...]

PR #1331, opened 2023-01-30
Implement vectorized base image augmentation layer w/ `Grayscale()` example

LukeWood:
We have a ton of performance issues with KPLs due to auto vectorization not working with tf.image ops.

Basically my proposed approach is as follows:

- replace KPLs that can be vectorized fully with  VectorizedBaseImageAugmentation subclasses

Fixes: https://github.com/keras-team/keras-cv/issues/581

---

Caveats of this approach:
- a bit less flexibility (i.e. in the case image augmentation could be vectorized, but bounding box augmentation could be not)
- loss of ragged support
- perhaps a bit convoluted implementation wise to have two base layers

### DIFF
--- a/keras_cv/layers/preprocessing/batched_base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/batched_base_image_augmentation_layer.py
@@ -38,14 +38,21 @@ class BatchedBaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomL
     def __init__(self, seed=None, **kwargs):
         super().__init__(seed=seed, **kwargs)
 
+    def augment_ragged_image(self, image, transformations, **kwargs):
+        """Augment a single image when ragged images are passed during training.
+
+        Args:
+            image:
+        """
+
     def augment_images(self, images, transformations, **kwargs):
         """Augment a batch of images during training.
 
         Args:
-          image: 3D image input tensor to the layer. Forwarded from
+          image: 4D image input tensor to the layer. Forwarded from
             `layer.call()`.
-          transformation: The transformation object produced by
-            `get_random_transformation`. Used to coordinate the randomness
+          transformations: The transformations object produced by
+            `get_random_transformations`. Used to coordinate the randomness
             between image, label, bounding box, keypoints, and segmentation mask.
 
         Returns:
@@ -53,17 +60,17 @@ class BatchedBaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomL
         """
         raise NotImplementedError()
 
-    def augment_labels(self, labels, transformation, **kwargs):
+    def augment_labels(self, labels, transformations, **kwargs):
         """Augment a batch of  labels during training.
 
         Args:
-          label: 1D label to the layer. Forwarded from `layer.call()`.
-          transformation: The transformation object produced by
-            `get_random_transformation`. Used to coordinate the randomness
+          label: 2D label to the layer. Forwarded from `layer.call()`.
+          transformations: The transformations object produced by
+            `get_random_transformations`. Used to coordinate the randomness
             between image, label, bounding box, keypoints, and segmentation mask.
 
         Returns:
-          output 1D tensor, which will be forward to `layer.call()`.
+          output 2D tensor, which will be forward to `layer.call()`.
         """
         raise NotImplementedError()
 
@@ -71,13 +78,13 @@ class BatchedBaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomL
         """Augment a batch of targets during training.
 
         Args:
-          target: 1D label to the layer. Forwarded from `layer.call()`.
-          transformation: The transformation object produced by
-            `get_random_transformation`. Used to coordinate the randomness
+          target: 2D label to the layer. Forwarded from `layer.call()`.
+          transformations: The transformations object produced by
+            `get_random_transformations`. Used to coordinate the randomness
             between image, label, bounding box, keypoints, and segmentation mask.
 
         Returns:
-          output 1D tensor, which will be forward to `layer.call()`.
+          output 2D tensor, which will be forward to `layer.call()`.
         """
         return self.augment_labels(targets, transformations)
 
@@ -89,12 +96,12 @@ class BatchedBaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomL
             `layer.call()`.
           bounding_boxes: 2D bounding boxes to the layer. Forwarded from
             `call()`.
-          transformation: The transformation object produced by
-            `get_random_transformation`. Used to coordinate the randomness
+          transformations: The transformations object produced by
+            `get_random_transformations`. Used to coordinate the randomness
             between image, label, bounding box, keypoints, and segmentation mask.
 
         Returns:
-          output 2D tensor, which will be forward to `layer.call()`.
+          output 3D tensor, which will be forward to `layer.call()`.
         """
         raise NotImplementedError()
 
@@ -102,14 +109,14 @@ class BatchedBaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomL
         """Augment a batch of keypoints for one image during training.
 
         Args:
-          keypoints: 2D keypoints input tensor to the layer. Forwarded from
+          keypoints: 3D keypoints input tensor to the layer. Forwarded from
             `layer.call()`.
-          transformation: The transformation object produced by
-            `get_random_transformation`. Used to coordinate the randomness
+          transformations: The transformations object produced by
+            `get_random_transformations`. Used to coordinate the randomness
             between image, label, bounding box, keypoints, and segmentation mask.
 
         Returns:
-          output 2D tensor, which will be forward to `layer.call()`.
+          output 3D tensor, which will be forward to `layer.call()`.
         """
         raise NotImplementedError()
 
@@ -120,8 +127,8 @@ class BatchedBaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomL
           segmentation_mask: 3D segmentation mask input tensor to the layer.
             This should generally have the shape [H, W, 1], or in some cases [H, W, C] for multilabeled data.
             Forwarded from `layer.call()`.
-          transformation: The transformation object produced by
-            `get_random_transformation`. Used to coordinate the randomness
+          transformations: The transformations object produced by
+            `get_random_transformations`. Used to coordinate the randomness
             between image, label, bounding box, keypoints, and segmentation mask.
 
         Returns:
@@ -138,13 +145,13 @@ class BatchedBaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomL
         keypoints=None,
         segmentation_masks=None,
     ):
-        """Produce random transformation config for a batch of inputs.
+        """Produce random transformations config for a batch of inputs.
 
         This is used to produce same randomness between
         image/label/bounding_box.
 
         Args:
-          batch_size: the batch size of transformation configuration to sample.
+          batch_size: the batch size of transformations configuration to sample.
           image: 3D image tensor from inputs.
           label: optional 1D label tensor from inputs.
           bounding_box: optional 2D bounding boxes tensor from inputs.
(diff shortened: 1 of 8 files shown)


bhack:
What about the product decision at https://github.com/tensorflow/tensorflow/issues/55639#issuecomment-1310150046?

LukeWood:
> What about the product decision at [tensorflow/tensorflow#55639 (comment)](https://github.com/tensorflow/tensorflow/issues/55639#issuecomment-1310150046)?

Curious @fchollet 's thoughts on this.  Its impossible to vectorize this one in TensorFlow and keep transformations independent from one another.  Do you think a slowdown is acceptable if this is the case?

bhack:
The bottleneck is with any API that doesn't have a batch dimension on param that you want to randomize or change x element.

So it is like if you need to loop over the batch calling the same op with a different parameter.  

I don't know if this could happen or not in the near future with the new XLA/MLIR birdge design and its genenerated kernels but right now not with vectorized map and I think not with the current bridge.

bhack:
https://github.com/tensorflow/tensorflow/pull/55335#discussion_r832778484

@jpienaar Can ops like this one be fused in a loop without breaking the raw ops API?

keras_cv/layers/preprocessing/vectorized_base_image_augmentation_layer.py
```diff
@@ -0,0 +1,304 @@
+# Copyright 2022 The KerasCV Authors
```
ianstenbit, line 1: nit: 2023

keras_cv/layers/preprocessing/vectorized_base_image_augmentation_layer.py
```diff
@@ -0,0 +1,304 @@
...
+BOUNDING_BOXES = "bounding_boxes"
+KEYPOINTS = "keypoints"
+SEGMENTATION_MASKS = "segmentation_masks"
+
+IS_DICT = "is_dict"
+BATCHED = "batched"
+USE_TARGETS = "use_targets"
+
+
+class VectorizedBaseImageAugmentationLayer(
```
ianstenbit, line 37: Is there a reason this can't subclass BaseImageAugmentationLayer?

keras_cv/layers/preprocessing/random_contrast.py
```diff
@@ -56,33 +55,44 @@ class RandomContrast(BaseImageAugmentationLayer):
...
-        return self.factor()
-
-    def augment_image(self, image, transformation, **kwargs):
-        contrast_factor = transformation
-        output = tf.image.adjust_contrast(image, contrast_factor=contrast_factor)
+    def get_random_transformation_batch(self, batch_size, **kwargs):
+        return
+
+    def augment_images(self, images, transformations, **kwargs):
+        # Note: we skip get_random_transformation_batch() in RandomContrast because the
```
ianstenbit, line 77: This comment seems like it belongs in the `get_random_transformation_batch` impl

ianstenbit (changes requested):

keras_cv/layers/preprocessing/vectorized_base_image_augmentation_layer.py
LukeWood: discussed a bit offline but to summarize for community members:

they barely share functionality in reality despite looking like they may.

LukeWood:
@jbischof @ianstenbit this is ready for a real review now.  We can incrementally migrate layers to vectorized layers using this approach.

bhack:
Before we merge I hope that we could clarify the point about the product decision at:

https://github.com/keras-team/keras-cv/issues/1256#issuecomment-1376134450

As It was since the first keras-cv commits that I was talking about the performance overhead of this design with `tf.image` API.

LukeWood:
> Before we merge I hope that we could clarify the point about the product decision at:
> 
> [#1256 (comment)](https://github.com/keras-team/keras-cv/issues/1256#issuecomment-1376134450)
> 
> As It was since the first keras-cv commits that I was talking about the performance overhead of this design with `tf.image` API.

What are we hoping to have clarified?  This change solves the performance issues of grayscale with no regression of functionality.  Item-wise augmentation is still deemed important.

bhack:
> What are we hoping to have clarified? This change solves the performance issues of grayscale with no regression of 

It is not only about grayscale as this related to many ops. An August not update list is at https://github.com/keras-team/keras-cv/issues/291#issuecomment-1230707124

LukeWood:
> > What are we hoping to have clarified? This change solves the performance issues of grayscale with no regression of
> 
> It is not only about grayscale as this is related to many ops. An August not update list is at [#291 (comment)](https://github.com/keras-team/keras-cv/issues/291#issuecomment-1230707124)

My goal in this PR is to provide an abstraction to allow us to optimize the performance of layers that are not being converted by pfor: https://github.com/keras-team/keras-cv/issues/291#issuecomment-1230707124

Does that answer the question?

bhack:
> Does that answer the question?

This was the same topic one year ago but we didn't care about the topic also if we was in a design phase:
https://github.com/keras-team/keras-cv/issues/455#issuecomment-1137111935

So what have changed in 1 year that instead we could not care in the design phase with all the explicit feedback and benchmarks that I've sent?

> not being converted by pfor

As you can read in many place and also in that comment it is not the only point for the failure/fallback pfor conversion as we have also missing converters.

As an extra note:
after all these months it seems that we don't have a tf.image API owners and probably also a pfor API owner as all the related tickets are without any other comet or action point.

LukeWood:
> > Does that answer the question?
> 
> This was the same topic one year ago but we didn't care about the topic also if we was in a design phase: [#455 (comment)](https://github.com/keras-team/keras-cv/issues/455#issuecomment-1137111935)
> 
> So what have changed in 1 year that instead we could not care in the design phase with all the explicit feedback and benchmarks that I've sent?
> 
> > not being converted by pfor
> 
> As you can read in many place and also in that comment it is not the only point for the failure/fallback of pfor conversion as we have also missing converters.
> 
> As an extra note: after all these months it seems that we don't have tf.image API owners and probably also pfor API owner as all the related tickets are without any other comment or action.

we do care about performance, and as you say there is no tf.image API owner or pfor API owner, so instead of for these issues to be fixed we'd like to offer vectorized implementations of these layers in the meantime.  Does that answer your question?

And to be clear, it was never that we didn't care about performance, it was simply that getting RandAugment working, OD augmentations working, and other product goals took priority.  Now that these are working I'm attempting to revisit these layers and optimize without major regressions in functionality.

bhack:
>  it was never that we didn't care about performance

Honestly the impression for one year it was that we did not care about performance. I have not collected any comment that we recognized the problem also if submitted in multiple ticket and that we just need to postpone the design for resource constrain.

In any case better later then never, we will see the residual fallback after this migration.

About the PR:
Do we need to subclass `tf.keras.__internal__.layers.BaseRandomLayer` for every new policy?

ianstenbit (changes requested):
Thanks Luke!

Lots of comments because this is a foundational piece and I want to make sure we get it really right.

keras_cv/layers/preprocessing/batched_base_image_augmentation_layer.py
```diff
@@ -0,0 +1,313 @@
...
+            result[SEGMENTATION_MASKS] = segmentation_masks
+
+        # preserve any additional inputs unmodified by this layer.
+        for key in inputs.keys() - result.keys():
+            result[key] = inputs[key]
+        return result
+
+    def call(self, inputs, training=True):
+        inputs = self._ensure_inputs_are_compute_dtype(inputs)
+        if training:
```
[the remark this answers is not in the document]
LukeWood: Good idea, done

keras_cv/layers/preprocessing/batched_base_image_augmentation_layer.py
```diff
@@ -0,0 +1,313 @@
...
+            else:
+                raise ValueError(
+                    "Image augmentation layers are expecting inputs to be "
+                    "rank 3 (HWC) or 4D (NHWC) tensors. Got shape: "
+                    f"{images.shape}"
+                )
+        else:
+            return inputs
+
+    def _format_inputs(self, inputs):
```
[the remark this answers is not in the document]
LukeWood: there are so many other layers on the base layer, and this will eventually replace the base layer if we finish migrating every layer.  

I don't think the case here is strong enough for this to be a subclass.

keras_cv/layers/preprocessing/batched_base_image_augmentation_layer.py
```diff
@@ -0,0 +1,313 @@
...
+        metadata = {IS_DICT: True, USE_TARGETS: False}
+        if tf.is_tensor(inputs):
+            # single image input tensor
+            metadata[IS_DICT] = False
+            inputs = {IMAGES: inputs}
+
+        metadata[BATCHED] = inputs["images"].shape.rank == 4
```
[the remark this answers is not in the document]
LukeWood: actually no, you can call any of these layers with a single input.  The "batched" quality of this layer refers to an implementation detail, not that it must be called on batches.

keras_cv/layers/preprocessing/batched_base_image_augmentation_layer.py
```diff
@@ -0,0 +1,313 @@
...
+        if inputs["images"].shape.rank == 3:
```
LukeWood, line 250: @ianstenbit see here, we expand dims if its operation in unbatched mode.

keras_cv/layers/preprocessing/batched_base_image_augmentation_layer.py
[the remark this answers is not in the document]
LukeWood: actually this has some different behavior in it too - in the expand_dims() section below.

LukeWood:
@ianstenbit all comments addressed.

Closed in favor of https://github.com/keras-team/keras-cv/pull/1373/files

LukeWood:
> > it was never that we didn't care about performance
> 
> Honestly the impression for one year was that we did not care about performance. I have not collected any comment that we recognized the problem also if it was submitted and benchmarked in multiple tickets and that we just need to postpone the design for resource constrain.
> 
> In any case better later then never, we will see the residual fallback after this migration.
> 
> About the PR: Do we need to subclass `tf.keras.__internal__.layers.BaseRandomLayer` for every new policy?

```
Honestly the impression for one year was that we did not care about performance. I have not collected any comment that we recognized the problem also if it was submitted and benchmarked in multiple tickets and that we just need to postpone the design for resource constrain.
```

Resource constraints are real and we were hoping pfor would either be:

- fixed
- sufficiently fast

Neither turned out so we are now prioritizing this.  Please keep in mind resource constraints are a real issue, and they don't mean issues aren't important; they simply mean we have other things that are more pressing at the time.

bhack:
> Neither turned out so we are now prioritizing this. Please keep in mind resource constraints are a real issue, and they don't mean issues aren't important; they simply mean we have other things that are more pressing at the time.

Generally I agree but it is a little bit different when we was still in the design phase and we had realtime feedback about how it will go to fail. 
If not it will be hard to interact clearly on the same page (mainly on PR) if we give priority/attention to f2f meeting.

Pfor could be not fixed if the underline design API of the ops we want to use in this library is not batch ready.

I think our approach was really out of design for Pfor from the early days. 

Then for missing converters it could be still fixed.

We cared about something (within the batch policy) with some theoretical validity but without an empirical verification about the faster convergence vs x epoch computational overhead.

Also as we have seen in another recent PR pytorch vision is not working within the batch so many reference models that we also train/replicare and the related reference papers were not trained with the within the batch policy at all.

LukeWood:
> We cared about something (within the batch policy) with some theoretical validity but without an empirical verification about the faster convergence vs x epoch computational overhead.

This already matters in contrastive learning and any case where samples are compared within a batch.  This is a case where a silent error could really waste a users time someday - and we should avoid this if possible.

In the current structure we will likely be able to have both performance and numerical correctness.  I think this is the best of all worlds

bhack:
> This already matters in contrastive learning and any case where samples are compared within a batch.  This is a case where a silent error could really waste a users time someday - and we should avoid this if possible.
> 
> In the current structure we will likely be able to have both performance and numerical correctness.  I think this is the best of all worlds

Yes but:

- We had only one contrastive network so by the priority/resource constrain point of view there is something wrong about what API to prioritize

- They never thought to fallback on vmap cause they known the design of the tf.image API. If you check again they have constrained the ops on CPU to try at least to use (large?) multicore machines for the preprocessing part of the computation:
https://github.com/tensorflow/similarity/blob/master/tensorflow_similarity/augmenters/simclr.py#L142:L143

PR #1336, opened 2023-02-01
Fix ConvMixer

### LINKED ISSUE #1334: Fix ConvMixer
Point Given by @DavidLandup0 :

- There's a typo in the `ConvMixer_Layer` (it's currently `CovnMixer_Layer`)
- This name should probably be `ConvMixerLayer` without the underscore to adhere to all other models we have
- We already have a public API layer for patching and embedding, while this model defines its own

Question : when we talk about existing PatchingAndEmbedding layers, are we referring about Vit's PatchingAndEmbedding Layer?

@DavidLandup0 @LukeWood

IMvision12:
Fix #1334
As far as the third question is concerned the working of Vit's PatchingAndEmbedding and ConvMixer's PatchEmbed is different.

@LukeWood


IMvision12:
I guess failing tests are not related to convmixer.py file

IMvision12:
@LukeWood i guess this PR is ready to merge!!!!

keras_cv/models/convmixer.py
```diff
@@ -87,7 +87,7 @@
 """
 
 
-def CovnMixer_Layer(dim, kernel_size):
+def ConvMixerLayer(dim, kernel_size):
     """CovnMixer Layer module.
```
ianstenbit, line 91: Can we fix the docstring too?

ianstenbit (changes requested):
Thanks!

keras_cv/models/convmixer.py
```diff
@@ -112,7 +112,7 @@ def apply(x):
     return apply
 
 
-def patch_embed(dim, patch_size):
+def PatchEmbed(dim, patch_size):
```
LukeWood, line 115: can we replace this with a class based layer component?

IMvision12: Done!

keras_cv/models/convmixer.py
IMvision12: Done!

IMvision12:
Improved docs too!!

keras_cv/models/convmixer.py
ianstenbit: I think this will probably break our pre-trained weights for ConvMixer (we only have 1).

@IMvision12 can you test if with this change you can still load the pre-trained weights?
If not, we'll need to convert them.

IMvision12: @ianstenbit I am getting this error on colab
`Exception: URL fetch failure on https://storage.googleapis.com/keras-cv/models/convmixer_512_16/imagenet/classification-v0.h5: 403 -- Forbidden`

IMvision12: using given code : 
```
from keras_cv.models.convmixer import ConvMixer_512_16
model = ConvMixer_512_16(include_rescaling=False, include_top=True, classes=1000, weights='imagenet')
```

IMvision12: Nope still getting same error!

ianstenbit: how about now?

IMvision12: So, the problem is solved but there's mismatch in layers :
`ValueError: Layer count mismatch when loading weights from file. Model expected 66 layers, found 67 saved layers.`

ianstenbit: Yeah, that's what I expected :/

We can either port these weights manually (for both the top and no-top versions), or we can skip on this part of the change for now.

I'm fine either way. Porting the weights isn't too complex. I'm not able to do it right now but can probably get to it late this week if needed.

IMvision12: @LukeWood what should we do?

IMvision12: @ianstenbit I guess we can skip this part for now, we can address this part when we will add weights for all other models!!

ianstenbit: That sgtm -- leaving this as a method, not a layer, is fine for now.

IMvision12: So, no more changes for me for the time being, right?

ianstenbit: Sorry I think I misunderstood you.
I don't think we should skip porting the weights. I meant let's skip making PatchEmbed a layer for now.

Currently, it's still a layer.

IMvision12: Okay I will change it back to function!!

IMvision12: Done!!

ianstenbit: sg -- thanks!

keras_cv/models/convmixer.py
```diff
@@ -112,29 +112,23 @@ def apply(x):
     return apply
 
 
-class PatchEmbed(layers.Layer):
+def PatchEmbed(dim, patch_size):
     """Implementation for Extracting Patch Embeddings.
     Args:
-        dim: integer, Number of filters for convolution layers.
+        inputs: Input tensor.
```
ianstenbit, line 118: It looks like this is now incorrect. Other than this, LGTM

IMvision12: Done!

ianstenbit (approved):
Thank you!

PR #1337, opened 2023-02-01
Reformat with black 23.1 to fix CI

ianstenbit:
@LukeWood


LukeWood (approved):

LukeWood:
Thanks @ianstenbit !!!

bhack:
If you are interested in the root cause of this is the we don't explicitly control black version:
https://github.com/keras-team/keras-cv/blob/e62a3fed37e89bc1c52777f3dfd6cd3581d40c65/setup.py#L58

So it is hard to recommend to the user a version if the CI could potentially change at every run.

ianstenbit:
> If you are interested in the root cause of this is the we don't explicitly control black version:
> 
> https://github.com/keras-team/keras-cv/blob/e62a3fed37e89bc1c52777f3dfd6cd3581d40c65/setup.py#L58
> 
> So it is hard to recommend to the user a version if the CI could potentially change at every run.

Is there a reason that we should pin a Black version? Afaik changes are generally backwards-compatible so I think we can just recommend that a user use the latest version of Black in general?

bhack:
> Is there a reason that we should pin a Black version? 

Is that we broke the format randomly at every CI run also on unrelated user PR (as it is already happened multiple time in this repo). 

I think that it is better to control when we want to upgrade also if we want to always use the last version.

ianstenbit:
> > Is there a reason that we should pin a Black version?
> 
> Is that we broke the format randomly at every CI run also on unrelated user PR (as it is already happened multiple time in this repo).
> 
> I think that it is better to control when we want to upgrade also if we want to always use the last version.

Ack that's fine by me. @LukeWood any issues on your end?
@bhack If you want to open a PR I'll approve -- otherwise I'll open one.

bhack:
@ianstenbit Please, talk internally with the team as we could find a compromise with  `~=` on the other deps. 

I think we will have less reproducibility but we will gain a bit in flexibility (and recurrent PR over `setup.py`) at the same time lowering the risk of randomly breaking the CI and local dev-envs.

PR #1340, opened 2023-02-02
Add gs:// path loading to weights

LukeWood:
Caveats:
- no file hash
- potential for bad caching

### DIFF
--- a/keras_cv/models/weights.py
+++ b/keras_cv/models/weights.py
@@ -15,14 +15,14 @@ from keras.utils import data_utils
 
 
 def parse_weights(weights, include_top, model_type):
+    if not weights or tf.io.gfile.exists(weights):
+        return weights
     if weights.startswith("gs://"):
         weights = weights.replace("gs://", "https://storage.googleapis.com/")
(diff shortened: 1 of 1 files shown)


keras_cv/models/weights.py
```diff
@@ -15,6 +15,12 @@
 
 
 def parse_weights(weights, include_top, model_type):
+    if weights.startswith("gs://"):
+        weights = weights.replace("gs://", "https://storage.googleapis.com/")
+        return data_utils.get_file(
+            origin=weights,
+            cache_subdir="custom_models",
```
ianstenbit, line 22: probably just "models" to be consistent with the equivalent on line 33

ianstenbit (approved):
Thanks Luke!

jbischof:
What's the motivation for this change?

LukeWood:
> What's the motivation for this change?

To support:
```
ResNet50(weights='gs://your-custom-path')
```

This makes loading weights in a sidecar evaluator or basically anywhere else that is not on the same host machine a lot easier.

PR #1348, opened 2023-02-02
Support pure-dictionary input format for OD models

### LINKED ISSUE #1342: Models should support pure-dictionary format
Currently, models expect inputs in the format of
`(images, {"boxes": ..., "labels": ...}`.

They should all support pure-dictionary inputs.

ianstenbit:
Fixes #1342

This also updates our GPU CI to actually run `INTEGRATION` tests, since our model tests are no longer so expensive that they time out the GPU tests. In order to do this, I had to add a separate `REGRESSION` control to two tests which are currently very expensive that don't belong in our CI.

### DIFF
--- a/keras_cv/models/object_detection/faster_rcnn.py
+++ b/keras_cv/models/object_detection/faster_rcnn.py
@@ -27,6 +27,7 @@ from keras_cv.layers.object_detection.roi_generator import ROIGenerator
 from keras_cv.layers.object_detection.roi_sampler import _ROISampler
 from keras_cv.layers.object_detection.rpn_label_encoder import _RpnLabelEncoder
 from keras_cv.models.object_detection import predict_utils
+from keras_cv.models.object_detection.__internal__ import unpack_input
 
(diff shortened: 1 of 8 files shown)


ianstenbit:
cc @martin-gorner

keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -487,9 +488,8 @@ def compute_loss(self, images, boxes, classes, training):
         )
 
     def train_step(self, data):
-        images, y, sample_weight = tf.keras.utils.unpack_x_y_sample_weight(data)
-        if sample_weight is not None:
-            raise ValueError("`sample_weight` is currently not supported.")
+        images, y = unpack_input(data)
```
LukeWood, line 491: One issue here: sample weight.  I guess we just don’t support it with the dict input?

ianstenbit: Correct. (Although this is not a regression -- we already don't support sample weight in the OD models)

LukeWood (approved):

PR #1355, opened 2023-02-03
Fixes weight loading for gs://

LukeWood:
tf.io.exists has a discrepency with the behavior of load_weights().  This causes an error when trying to load gs:// weights in some cases.

### DIFF
--- a/keras_cv/models/weights.py
+++ b/keras_cv/models/weights.py
@@ -15,7 +15,7 @@ from keras.utils import data_utils
 
 
 def parse_weights(weights, include_top, model_type):
-    if not weights or tf.io.gfile.exists(weights):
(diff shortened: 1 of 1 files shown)


ianstenbit (approved):
Can you patch this PR and make sure it works for your testing Colab?

Maybe we should add an OSS-only test that actually loads these weights.

LukeWood:
@ianstenbit  done and works: https://colab.research.google.com/drive/1tlkRq_V77EyQi1JcD46ispJkN9Jr3Mro?usp=sharing

LukeWood:
@ianstenbit let's do a unit test as a follow up so we can cut a release due to the OD guide being broken.

PR #1356, opened 2023-02-03
[Draft] Remove TensorFlow graph based COCO metrics

LukeWood:
removes the in graph COCO metrics due to:
- awful performance
- awful numerical stability
- soon to be obsoleted by Numpy versions

---

notes for the community:

The reason for this change is that we are introducing the concept of a "PyMetric", which runs  in the python runtime instead of the TensorFlow graph.  Due to the constraints of the tensorflow graph, variable size state is extremely slow.  Variable size state updates are quite important in COCO metric evaluation, and as such we will be shipping a version of COCO metrics (based on the TF implementations) that run in the python runtime.  In our benchmarks, this is significantly more performant and solves the issue of the need for a bucketization/approximation algorithm for MaP.  

----

A sad day for me indeed - but the correct move for the project.  In order to document the workflows used to make these work, I will still place these in their own repo at some point and do some writing to discuss the implementation as well as the issues we faced in making them "actually work".


bhack:
Can we log these metrics in the tensorboard on the eval step?

bhack:
Also as we had a technical paper on arxiv and also many hours invested in this I hope that we document why we cannot compile this kind of operations as they are very common in many networks.

LukeWood:
@bhack my understanding is that the issue lies in the inability of the TensorFlow graph to efficiently handle variable size state.

LukeWood:
> Can we log these metrics in the tensorboard on the eval step?

Yes.  These metrics will be indistinguishable from in-graph metrics to an end user, excluding a lack of TPU compat.

bhack:
> @bhack my understanding is that the issue lies in the inability of the TensorFlow graph to efficiently handle variable size state.


Yes but all the resources are working [hard on dynamism](https://github.com/openxla/stablehlo/labels/Dynamism)  it is why is better to comunicate what is not compiling.

bhack:
Se @burmako https://github.com/openxla/stablehlo/issues/8

LukeWood:
Instead of remove I will be incrementally migrating to PyMetric.

PR #1362, opened 2023-02-06
Migrate KerasCV WaymoOpendata components to use conditional imports

LukeWood:


LukeWood:
Looks like type annotations actually cause errors if packages are missing, making some updates.

keras_cv/utils/conditional_imports.py
```diff
@@ -0,0 +1,26 @@
...
+try:
+    import waymo_open_dataset
+except ImportError:
+    waymo_open_dataset = None
+
+
+def assert_waymo_opendata_installed(symbol_name):
+    if waymo_open_dataset is None:
+        raise ImportError(
+            f"{symbol_name} requires the `waymo-open-dataset-tf-2-6-0` package. "
```
ianstenbit, line 24: I don't think we should be recommending pip install here, since in many cases installing WOD requires installing from source so that one can use a higher TF version.

Since we depend on >=2.11, it seems misleading to recommend a pip install of an incompatible version of WOD

keras_cv/utils/conditional_imports.py
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+try:
+    import waymo_open_dataset
+except ImportError:
+    waymo_open_dataset = None
+
+
+def assert_waymo_opendata_installed(symbol_name):
```
ianstenbit, line 21: `assert_waymo_open_dataset_installed`?

keras_cv/datasets/waymo/load.py
```diff
@@ -67,6 +72,7 @@ def simple_transformer(frame):
     load("/path/to/tfrecords", simple_transformer, output_signature)
     ```
     """
+    assert_waymo_opendata_installed("keras_cv.datasets.waymo.load()")
```
ianstenbit, line 75: This is the only place where you include parens in the symbol name -- let's either exclude it here or include it everywhere

ianstenbit (changes requested):
Thank you Luke! This pattern lgtm, and we can do the same for PyCocoTools :)

keras_cv/datasets/waymo/load.py
LukeWood: Included everywhere

keras_cv/utils/conditional_imports.py
LukeWood: Done

keras_cv/utils/conditional_imports.py
LukeWood: renamed

LukeWood:
ok @ianstenbit - ready!

ianstenbit (approved):
Thanks Luke!

Issue #1343, opened 2023-02-02
Document input format for all KPLs + Models


ianstenbit:
Every model + KPL needs to include in its docstring information about the expected dictionary input format.

soma2000-lang:
@ianstenbit Is this issue available for community contributions?If yes then I would love to contribute

ianstenbit:
@soma2000-lang yes this is.

I will write up a docstring to describe our input format which can be included in Layers and models. I'll post it here this afternoon and then add the contribution-welcome label.

ianstenbit:
For community members who'd like to update docstrings for this issue, please use (or adapt) the language in the docstring update in #1363

soma2000-lang:
@ianstenbit Probably I can work on this one.

atharvapurdue:
Hi @soma2000-lang  @ianstenbit , is this issue open? Do you mind if I contribute to it as well? I am new to this repo and this has the 'good first issue' tag.  Thanks!

soma2000-lang:
@atharvapurdue I am working on adding doctrines on few of the KPLs ,but there are many where doctrings needs to

YimjuneLiu:
Hi @ianstenbit @soma2000-lang , Is this issue open? I am still a beginner, do you mind if I participate? Thanks!

ianstenbit:
Yes this issue is open for contributions -- it's a good issue for a beginner 👍 

Feel free to open a PR and tag this issue.

YimjuneLiu:
> Yes this issue is open for contributions -- it's a good issue for a beginner 👍 
> 
> Feel free to open a PR and tag this issue.

Thanks for your reply.

I see that some contributors have already done some work. Do I still need to do all the work? Or do part of it to prevent conflicts? If only part of it is done, which part specifically?

soma2000-lang:
@YimjuneLiu  Yes ,I am working on some of  the files .It would be great if you work on the other remaining KPL

YimjuneLiu:
> @YimjuneLiu  Yes ,I am working on some of  the files .It would be great if you work on the other remaining KPL

Ok, Can you tell me specifically which KPLs are included?

YimjuneLiu:
#1467

Anirban0011:
@ianstenbit Hi, is this issue still open to contributions? I would like to participate if it is the case. Thanks

YimjuneLiu:
[...]

PR #1363, opened 2023-02-06
Update docstring for base KPL

ianstenbit:
This is an example doc toward #1343

### DIFF
--- a/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
@@ -58,10 +58,10 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
     and bounding_boxes should be changed in the same way.
 
     The `call()` method supports two formats of inputs:
-    1. Single image tensor with shape (height, width, channels) or
+    1. A single image tensor with shape (height, width, channels) or
        (batch_size, height, width, channels)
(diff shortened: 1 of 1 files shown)


LukeWood (approved):
Thanks for the update Ian!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -57,14 +57,23 @@ class BaseImageAugmentationLayer(tf.keras.__internal__.layers.BaseRandomLayer):
...
-    2. A dict of tensors with stable keys. The supported keys are:
-      `"images"`, `"labels"` and `"bounding_boxes"` at the moment. We might add
-      more keys in future when we support more types of augmentation.
-
-    The output of the `call()` will be in two formats, which will be the same
-    structure as the inputs.
+    The `call()` method supports two formats of inputs:
+    1. A single image tensor with shape (height, width, channels) or
+       (batch_size, height, width, channels)
+    1. A dict of tensors with stable keys. The supported keys are:
```
jbischof, line 63: What does "stable" mean in this context?

jbischof (approved):
Nice!

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
ianstenbit: I don't think this was the right word choice -- I've updated the wording a bit.

PR #1365, opened 2023-02-06
Add `tensorflow` to `test` requirements

jbischof:
Current contributor setup instructions ((link)[https://github.com/jbischof/keras-cv/blob/master/.github/CONTRIBUTING.md#setup-environment]) do not work for running `pytest` in a fresh environment because `tensorflow` is not required. Making this an explicit requirement.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,6 +58,7 @@ setup(
             "black[jupyter]",
             "pytest",
             "pycocotools",
+            "tensorflow",
         ],
         "examples": ["tensorflow_datasets", "matplotlib"],
     },


LukeWood (approved):
looks good, thanks Jonathan!

bhack:
@jbischof Why only in extra/test if with the normal install we will only check/crash at runtime?

jbischof:
@bhack in truth I was only verifying the contributor flow. I assumed we were intentionally leaving `tensorflow` optional to make the package lightweight as we do for `tensorflow-text` in KerasNLP. Is this not the case?

bhack:
Do you think that we will increase the usability when we fail at runtime?

I think it is fine to fail at runtime with minor extra deps like waymo dataset or pycoco but I don't think we are increasing the usability with a core dependency like Tensorflow.

jbischof:
Yes that's what I'm asking @bhack! Any thoughts, @LukeWood?

PR #1374, opened 2023-02-09
Preparation for switch in Keras serialization format

nkovela1:
This PR refactors the deserialization of KerasCV layers in preparation for an upcoming switch in the serialization format of Keras objects. It adds `from_config` methods where necessary for deserialization and additionally enables the deserialization of lambdas where necessary.

Many of these changes are temporary until the addition of `keras.__internal__` namespace API support in the new serialization.

### DIFF
--- a/keras_cv/models/csp_darknet_test.py
+++ b/keras_cv/models/csp_darknet_test.py
@@ -29,6 +29,8 @@ MODEL_LIST = [
 class CSPDarkNetTest(ModelsTest, tf.test.TestCase, parameterized.TestCase):
     @parameterized.parameters(*MODEL_LIST)
     def test_application_base(self, app, _, args):
+        if hasattr(tf.keras.__internal__, "enable_unsafe_deserialization"):
+            tf.keras.__internal__.enable_unsafe_deserialization()
         super()._test_application_base(app, _, args)
 
     @parameterized.parameters(*MODEL_LIST)
--- a/keras_cv/utils/preprocessing.py
+++ b/keras_cv/utils/preprocessing.py
@@ -129,6 +129,7 @@ def parse_factor(param, min_value=0.0, max_value=1.0, param_name="factor", seed=
     if isinstance(param, dict):
         # For all classes missing a `from_config` implementation.
         # (RandomHue, RandomShear, etc.)
+        # To be removed with addition of `keras.__internal__` namespace support
         param = tf.keras.utils.deserialize_keras_object(param)
 
     if isinstance(param, core.FactorSampler):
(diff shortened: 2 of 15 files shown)


keras_cv/utils/preprocessing.py
```diff
@@ -126,6 +126,11 @@ def blend(image1: tf.Tensor, image2: tf.Tensor, factor: float) -> tf.Tensor:
 
 
 def parse_factor(param, min_value=0.0, max_value=1.0, param_name="factor", seed=None):
+    if isinstance(param, dict):
+        # For all classes missing a `from_config` implementation.
```
ianstenbit, line 130: Would it make more sense to just implement `from_config` for these layers?

ianstenbit (commented):
Do I understand correctly that the only layers that need this change are the ones whose serialization isn't automagically handled? (e.g. where FactorSamplers are used?)

keras_cv/models/csp_darknet_test.py
```diff
@@ -29,6 +29,8 @@
 class CSPDarkNetTest(ModelsTest, tf.test.TestCase, parameterized.TestCase):
     @parameterized.parameters(*MODEL_LIST)
     def test_application_base(self, app, _, args):
+        if hasattr(tf.keras.__internal__, "enable_unsafe_deserialization"):
```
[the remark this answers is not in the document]
nkovela1: This is a new internal API to allow the deserialization of lambdas (we are trying to actively discourage this, with exceptions such as where it's necessary in KerasCV). It is used in this test specifically because it is the only test in the suite that checks for serialization and deserialization.

The conditional is to ensure backwards-compatibility, since this new API only exists in Nightly now and will be released in the next stable release. We want to ensure it will work for users regardless of TF version.

keras_cv/utils/preprocessing.py
nkovela1: This is a more temporary overarching fix for these layers. With the new serialization format, the serialization of keras.__internal__ namespace is not supported yet, but that will be added in the near future. When that happens, this line can be removed as well as many of the `from config` methods in other layers.

keras_cv/models/csp_darknet_test.py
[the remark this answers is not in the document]
ianstenbit: Got it -- thanks Neel!

keras_cv/utils/preprocessing.py
ianstenbit: sg -- can you please add a comment here to mention that this is temporary, explaining when it'll be safe to remove?

(Also the same in the PR description)

ianstenbit (approved):
Thanks Neel! Just one little nit :)

PR #1379, opened 2023-02-09
Improve quickstart user experience

jbischof:
The quickstart is currently a bit awkward to use because you need the latest version of TF for the code to run and it is divided into many code blocks. Furthermore it seems to OOM on smaller GPUs (<30GB RAM).

Changes:
* Add installation guide
* Consolidate into one code block
* Reduce batch size to 16
* Add validation set
* Add prefetch

This version runs smoothly on the free version of colab (16GB RAM GPU).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,11 +31,25 @@ To learn more about the future project direction, please check the [roadmap](.gi
 - [Roadmap](.github/ROADMAP.md)
 - [API Design Guidelines](.github/API_DESIGN.md)
 
+## Installation
+
+To install the latest official release:
+
+```
+pip install keras-cv tensorflow --upgrade
+```
+
+To install the latest unreleased changes to the library, we recommend using
+pip to install directly from the master branch on github:
+
+```
+pip install git+https://github.com/keras-team/keras-cv.git tensorflow --upgrade
+```
+
 ## Quickstart
 
-Create a preprocessing pipeline:
-
 ```python
+# Create a preprocessing pipeline
 import keras_cv
 import tensorflow as tf
 from tensorflow import keras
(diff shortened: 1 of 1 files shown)


jbischof:
@LukeWood when I ran the quickstart example it OOM'd a regular colab and barely runs on a premium colab runtime. Why does it require >30GB of RAM to run? Is this a good "quick start"? 

Screenshot 2023-02-08 at 7 43 34 PM

LukeWood:
@jbischof perhaps we should also update keras.io alongside this to keep it in sync.

the file is in keras-io/templates/keras_cv.  LGTM

LukeWood (approved):

README.md
```diff
@@ -55,19 +69,13 @@ def augment_data(images, labels):
...
   outputs = augmenter(inputs)
   return outputs['images'], outputs['labels']
-```
-
-Augment a `tf.data.Dataset`:
 
-```python
+# Augment a `tf.data.Dataset`
 dataset = tfds.load('rock_paper_scissors', as_supervised=True, split='train')
 dataset = dataset.batch(64)
```
ianstenbit, line 75: re: OOM

Perhaps we should just decrease batch size here if this is crashing a Colab.

ianstenbit: We could also use a smaller model like MobileNetV3Small, althought DenseNet121 is already a relatively small backbone in terms of # of params.

README.md
```diff
@@ -55,19 +69,13 @@ def augment_data(images, labels):
...
 dataset = dataset.map(augment_data, num_parallel_calls=tf.data.AUTOTUNE)
```
ianstenbit, line 76: It's probably worth adding a `.prefetch(tf.data.AUTOTUNE)` to this dataset.

afaik this is pretty much always a worthwhile step to help optimize training speed, so it seems like something we should include in a quickstart.

ianstenbit (approved):
LGTM -- thanks!

My comments are mostly not related to your changes, just things we could touch up while we're looking at this anyway.

jbischof:
@ianstenbit do you want to take a look at the final version?

README.md
```diff
@@ -31,59 +31,73 @@ To learn more about the future project direction, please check the [roadmap](.gi
 - [Roadmap](.github/ROADMAP.md)
 - [API Design Guidelines](.github/API_DESIGN.md)
 
-## Quickstart
+## Installation
+
+To install the latest official release:
+
+```
+pip install keras-cv tensorflow --upgrade
```
bhack, line 39: Probably we could suggest a `venv` as we are implicitly breaking previous TensorFlow  and Keras installations with this command.

I understand that `--upgrade` is required to pass the runtime check over per-existing TF and Keras installation:
https://github.com/keras-team/keras-cv/blob/5227e52147343232a2183fefec1d31ed4c655079/keras_cv/version_check.py#L21-L31

jbischof: That's good advice. Personally I think it's common knowledge at this point that some containerization is required to have a good development and deployment experience in Python, so I don't think we're very special in this regard.

PR #1390, opened 2023-02-11
Replace closures with layers and models in resnetv2

jbischof:
WIP. NOT READY TO MERGE

This is a pilot to update our implementations as `Layers` and `Models` rather than a closure. This is part of our effort to standardize the KerasCV and KerasNLP APIs.

## Why this design?
* A subclassed `keras.Model` is needed to add a `from_preset` constructor for loading preset architectures and weights. In the long term this and other methods/properties will be maintained in a new `Backbone` base class.
* We initialize a functional `keras.Model` in the `__init__` to maintain the convenience of the `Functional` API.
* When using a subclassed `keras.Model` it is not possible to track weights in a closure-style submodel. Therefore we need to switch to `keras.layers.Layer` blocks. See @LukeWood's analysis ([link](https://lukewood.xyz/blog/subclass-model-pitfall)).

This should also improve the readability of `model.summary()`.

Steps:
* [x] Rewrite `BasicBlock` as `Layer`
* [x] Rewrite `Block` as `Layer`
* [x] Rewrite `Stack` as `Layer`
* [x] Rewrite `ResnetV2` as `Model`
* [ ] Convert weight files if incompatible

### DIFF
--- a/keras_cv/models/resnet_v2.py
+++ b/keras_cv/models/resnet_v2.py
@@ -394,35 +394,19 @@ class Stack(keras.layers.Layer):
         return super().from_config(config)
 
 
-def ResNetV2(
-    stackwise_filters,
-    stackwise_blocks,
-    stackwise_strides,
-    include_rescaling,
-    include_top,
-    stackwise_dilations=None,
-    name="ResNetV2",
-    weights=None,
-    input_shape=(None, None, 3),
-    input_tensor=None,
-    pooling=None,
-    classes=None,
-    classifier_activation="softmax",
-    block_fn=Block,
-    **kwargs,
-):
+@keras.utils.register_keras_serializable(package="keras_cv.models")
+class ResNetV2(keras.Model):
     """Instantiates the ResNetV2 architecture.
 
     Args:
-        stackwise_filters: number of filters for each stack in the model.
-        stackwise_blocks: number of blocks for each stack in the model.
-        stackwise_strides: stride for each stack in the model.
-        include_rescaling: whether or not to Rescale the inputs. If set to True,
+        stackwise_filters: int, number of filters for each stack in the model.
+        stackwise_blocks: int, number of blocks for each stack in the model.
+        stackwise_strides: int, stride for each stack in the model.
+        include_rescaling: bool, whether or not to Rescale the inputs. If set to True,
             inputs will be passed through a `Rescaling(1/255.0)` layer.
-            name: string, model name.
-        include_top: whether to include the fully-connected
+        include_top: bool, whether to include the fully-connected
             layer at the top of the network.
-        weights: one of `None` (random initialization),
+        weights: optional string, one of `None` (random initialization),
             or the path to the weights file to be loaded.
         input_shape: optional shape tuple, defaults to (None, None, 3).
         input_tensor: optional Keras tensor (i.e. output of `layers.Input()`)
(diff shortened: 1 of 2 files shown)


IMvision12:
So, Do we need to apply this changes to all other models?

jbischof:
Yes, @IMvision12 we will need significant help from the community once the pilot concludes! One thing I need to scope out is what changes (if any) we will to make to the weight checkpoints.

IMvision12:
> Yes, @IMvision12 we will need significant help from the community once the pilot concludes! One thing I need to scope out is what changes (if any) we will to make to the weight checkpoints.

Yeah, I'm willing to contribute. Also, about weights, when converting functions to classes an extra layers will be added so we need to map that extra layer

keras_cv/models/resnet_v2.py
```diff
@@ -325,90 +429,144 @@ def ResNetV2(
...
+        self.stackwise_blocks = stackwise_blocks
+        self.stackwise_strides = stackwise_strides
+        self.include_rescaling = include_rescaling
+        self.include_top = include_top
+        self.stackwise_dilations = stackwise_dilations
+        self.input_tensor = input_tensor
+        self.pooling = pooling
+        self.classes = classes
+        self.classifier_activation = classifier_activation
+        self.block_fn = Block
```
ianstenbit, line 545: self.block_fn = block_fn

keras_cv/models/resnet_v2.py
```diff
@@ -99,210 +101,312 @@
...
+        self.activation2 = layers.Activation("relu", name=self.name + "_2_relu")
+        self.conv3 = layers.Conv2D(
+            4 * filters, kernel_size=1, name=self.name + "_3_conv"
+        )
+        self.add = layers.Add(name=self.name + "_out")
+
+    def call(self, inputs):
+        x = self.bn_preactivation(inputs)
+        x = self.relu_preactivation(x)
+        shortcut = self.shortcut(x) if self.shortcut else x
```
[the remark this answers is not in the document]
ianstenbit: Ignore this, this was incorrect.

ianstenbit, line 292: After porting the weights and finding numerical inconsistencies, I discovered that this line is incorrect.

It should be:

```
if self.conv_shortcut:
          shortcut = self.shortcut(x)
        elif self.shortcut:
          shortcut = self.shortcut(inputs)
        else:
          shortcut = inputs
```

keras_cv/models/resnet_v2.py
```diff
@@ -99,210 +101,312 @@
...
     Args:
         filters: integer, filters of the basic layer.
-        kernel_size: default 3, kernel size of the bottleneck layer.
-        stride: default 1, stride of the first layer.
-        conv_shortcut: default False, use convolution shortcut if True,
+        kernel_size: integer, kernel size of the bottleneck layer. Defaults to 3.
+        stride: integer, stride of the first layer. Defaults to 1.
+        dilation: integer, the dilation rate to use for dilated convolution. Defaults to 1.
+        conv_shortcut: bool, uses convolution shortcut if `True`. Defaults to `False`.
           otherwise identity shortcut.
```
ianstenbit, line 114: "otherwise identity shortcut" isn't quite true -- really it defaults to identity or pooling, based on stride.
I don't think including all the detail is necessary in this docstring, but maybe we could say "otherwise identity or pooling shortcut, based on stride"

keras_cv/models/resnet_v2.py
```diff
@@ -99,210 +101,312 @@
...
+        self.filters = filters
+        self.kernel_size = kernel_size
+        self.stride = stride
+        self.dilation = dilation
+        self.stride2 = stride if dilation == 1 else 1
+        self.conv_shortcut = conv_shortcut
+        super().__init__(**kwargs)
+
+        self.bn_preactivation = layers.BatchNormalization(
+            axis=BN_AXIS, epsilon=BN_EPSILON, name=self.name + "_use_preactivation_bn"
```
ianstenbit, line 140: IMO this "use" prefix is kinda useless. Not sure why it was here in the first place, but I think this can just be _preactivation_bn (same for the relu below)

keras_cv/models/resnet_v2.py
ianstenbit: (same for BasicBlock btw)

keras_cv/models/resnet_v2.py
```diff
@@ -99,210 +101,312 @@
...
+
     Args:
         filters: integer, filters of the layer in a block.
         blocks: integer, blocks in the stacked blocks.
-        stride: default 2, stride of the first layer in the first block.
-        name: string, stack label.
+        stride: integer, stride of the first layer in the first block. Defaults to 2.
+        dilation: integer, the dilation rate to use for dilated convolution. Defaults to 1.
         block_fn: callable, `Block` or `BasicBlock`, the block function to stack.
         first_shortcut: default True, use convolution shortcut if True,
```
ianstenbit, line 326: boolean, whether to use a convolutional shortcut in the first block. Defaults to True.

keras_cv/models/resnet_v2.py
ianstenbit: (same for BasicBlock)

keras_cv/models/resnet_v2.py
ianstenbit: (same for BasicBlock)

ianstenbit (changes requested):
Thanks Jonathan!

keras_cv/models/resnet_v2.py
```diff
@@ -325,90 +429,144 @@ def ResNetV2(
...
+            "classifier_activation": self.classifier_activation,
+            "block_fn": keras.utils.get_registered_name(self.block_fn),
+            "name": self.name,
+            "trainable": self.trainable,
+        }
+
+    @classmethod
+    def from_config(cls, config):
+        block_class_name = config["block_fn"]
+        config["block_fn"] = keras.utils.get_registered_object(block_class_name)
```
LukeWood, line 568: nice - glad to see there's a way to do this.  I have actually not seen these utilities before.

keras_cv/models/resnet_v2.py
jbischof: Won't that mess up the existing checkpoints? Might be best to leave this alone for now

ianstenbit:
> @jbischof Can I start with other models after completion of 'resnetv2' by referring this PR?

Hey @IMvision12 -- we're currently comparing this approach to an alternative in #1401.
Once we've decided on an approach and merged the change for ResNetV2, we'll open up GitHub issues for the other models and folks can pick those up. Thanks for your enthusiasm! Hopefully we'll have a decision here soon

jbischof:
Abandoning this change in favor of #1401

PR #1392, opened 2023-02-13
Vectorized random saturation

### LINKED ISSUE #1386: Vectorize RandomSaturation
**Short Description**
I firstly dig into #1378 and accidentally get some outcomes of vectorized RandomSaturation 

`RandomSaturation` is current implementation
`RandomSaturationV2` is vectorized by `VectorizedBaseImageAugmentationLayer`

I tested the performance based on [benchmarks/vectorized_grayscale.py](https://github.com/keras-team/keras-cv/blob/24fe395229f73d94834b63b9b453e296d863a4af/benchmarks/vectorized_grayscale.py) with `factor=0.5`

The implementation looks like:
```python
class RandomSaturationV2(VectorizedBaseImageAugmentationLayer):
    def __init__(self, factor, seed=None, **kwargs):
        super().__init__(seed=seed, **kwargs)
        self.factor = preprocessing.parse_factor(
            factor,
            min_value=0.0,
            max_value=1.0,
        )
        self.seed = seed

    def get_random_transformation_batch(self, batch_size, **kwargs):
        return self.factor(shape=(batch_size,))

    def augment_images(self, images, transformations, **kwargs):
        transformations = tf.convert_to_tensor(transformations)
        adjust_factors = transformations / (1 - transformations)

        # rgb -> hsv -> mul at s channel with clipping -> rgb
        images = tf.image.rgb_to_hsv(images)
        s_channel_batch = tf.multiply(images[..., 1], adjust_factors[:, tf.newaxis, tf.newaxis])
        s_channel_batch = tf.clip_by_value(s_channel_batch, clip_value_min=0.0, clip_value_max=1.0)
        images = tf.stack([images[..., 0], s_channel_batch, images[..., 2]], axis=-1)
        images = tf.image.hsv_to_rgb(images)
        return images
```

but the outputs might be slightly slightly different with `factor=(0.5, 0.5)`:
I compared first 100 images from cifar 10 and there was 1 image failed to pass `tf.experimental.numpy.allclose`
(at `x_train[99]` if you want to know...)

should I contribute this implementation?

james77777778:
# What does this PR do?

Fixes #1386 

The vectorized version can pass all tests in old `keras_cv/layers/preprocessing/random_saturation_test.py`.

Then, as discussed in #1386, I copied the old random saturation layer into `keras_cv/layers/preprocessing/random_saturation_test.py` and wrote two tests to check the consistency.

This is the result:
- Input value range [0, 1]: the vectorized version can almost always pass the test with `atol=1e-5, rtol=1e-5`. I locally set batch size 1024 to test.
- Input value range [0, 255]: the vectorized version can only pass the test by setting higher `atol` that  `atol=1e-3, rtol=1e-5`.

I think the conversion between RGB and HSV might bring the error because the doc of `tf.image.rgb_to_hsv` says: [https://www.tensorflow.org/api_docs/python/tf/image/rgb_to_hsv](https://www.tensorflow.org/api_docs/python/tf/image/rgb_to_hsv)
>The output is only well defined if the value in images are in [0,1].

I add a benchmark script `benchmarks/vectorized_random_saturation.py` to show the improvement.

comparison
comparison_no_old_eager

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood

### DIFF
--- a/keras_cv/layers/preprocessing/random_saturation.py
+++ b/keras_cv/layers/preprocessing/random_saturation.py
@@ -61,6 +61,11 @@ class RandomSaturation(VectorizedBaseImageAugmentationLayer):
     def get_random_transformation_batch(self, batch_size, **kwargs):
         return self.factor(shape=(batch_size,))
 
+    def augment_ragged_image(self, image, transformation, **kwargs):
+        return self.augment_images(
+            images=image, transformations=transformation, **kwargs
+        )
+
     def augment_images(self, images, transformations, **kwargs):
         # Convert the factor range from [0, 1] to [0, +inf]. Note that the
         # tf.image.adjust_saturation is trying to apply the following math formula
@@ -79,24 +84,22 @@ class RandomSaturation(VectorizedBaseImageAugmentationLayer):
         adjust_factors = transformations / (1 - transformations)
 
         images = tf.image.rgb_to_hsv(images)
-        s_channel_batch = tf.multiply(
-            images[..., 1], adjust_factors[:, tf.newaxis, tf.newaxis]
+        s_channel = tf.multiply(
+            images[..., 1], adjust_factors[..., tf.newaxis, tf.newaxis]
         )
-        s_channel_batch = tf.clip_by_value(
-            s_channel_batch, clip_value_min=0.0, clip_value_max=1.0
-        )
-        images = tf.stack([images[..., 0], s_channel_batch, images[..., 2]], axis=-1)
+        s_channel = tf.clip_by_value(s_channel, clip_value_min=0.0, clip_value_max=1.0)
+        images = tf.stack([images[..., 0], s_channel, images[..., 2]], axis=-1)
         images = tf.image.hsv_to_rgb(images)
         return images
 
     def augment_bounding_boxes(self, bounding_boxes, transformation=None, **kwargs):
         return bounding_boxes
 
-    def augment_label(self, label, transformation=None, **kwargs):
-        return label
+    def augment_labels(self, labels, transformations=None, **kwargs):
+        return labels
 
-    def augment_segmentation_mask(self, segmentation_mask, transformation, **kwargs):
-        return segmentation_mask
+    def augment_segmentation_masks(self, segmentation_masks, transformations, **kwargs):
+        return segmentation_masks
 
     def get_config(self):
         config = {
--- a/keras_cv/layers/preprocessing/random_saturation.py
+++ b/keras_cv/layers/preprocessing/random_saturation.py
@@ -82,6 +82,7 @@ class RandomSaturation(VectorizedBaseImageAugmentationLayer):
         # it will be handled correctly when it is a one tensor.
         transformations = tf.convert_to_tensor(transformations)
         adjust_factors = transformations / (1 - transformations)
+        adjust_factors = tf.cast(adjust_factors, dtype=images.dtype)
 
         images = tf.image.rgb_to_hsv(images)
         s_channel = tf.multiply(
--- a/keras_cv/layers/preprocessing/random_saturation_test.py
+++ b/keras_cv/layers/preprocessing/random_saturation_test.py
@@ -21,7 +21,6 @@ from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
 from keras_cv.utils import preprocessing as preprocessing_utils
 
 
-@tf.keras.utils.register_keras_serializable(package="keras_cv")
 class OldRandomSaturation(BaseImageAugmentationLayer):
     """Randomly adjusts the saturation on given images.
 
--- a/keras_cv/layers/preprocessing/random_saturation_test.py
+++ b/keras_cv/layers/preprocessing/random_saturation_test.py
@@ -1,4 +1,4 @@
-# Copyright 2022 The KerasCV Authors
+# Copyright 2023 The KerasCV Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


keras_cv/layers/preprocessing/random_saturation_test.py
```diff
@@ -93,3 +173,29 @@ def test_config(self):
...
+        fixed_factor = (0.8, 0.8)
+        image = tf.random.uniform(shape=image_shape) * 255.0
+
+        layer = preprocessing.RandomSaturation(factor=fixed_factor)
+        old_layer = OldRandomSaturation(factor=fixed_factor)
+
+        output = layer(image)
+        old_output = old_layer(image)
+
+        self.assertAllClose(old_output, output, atol=1e-3, rtol=1e-5)
```
LukeWood, line 201: I think this is a fine error threshold.

Thanks!

keras_cv/layers/preprocessing/random_saturation.py
```diff
@@ -67,9 +75,19 @@ def augment_image(self, image, transformation=None, **kwargs):
...
         # transformation is 1.0, then it will result in to divide by zero error, but
         # it will be handled correctly when it is a one tensor.
-        transformation = tf.convert_to_tensor(transformation)
-        adjust_factor = transformation / (1 - transformation)
-        return tf.image.adjust_saturation(image, saturation_factor=adjust_factor)
+        transformations = tf.convert_to_tensor(transformations)
+        adjust_factors = transformations / (1 - transformations)
+
+        images = tf.image.rgb_to_hsv(images)
+        s_channel_batch = tf.multiply(
```
LukeWood, line 82: you can drop the `_batch` postfix here.

LukeWood (approved):
LGTM with a few small changes.

Great PR.  Thanks a lot for your hard work.

LukeWood:
@james77777778 thanks for the awesome contribution!

james77777778:
Hi @LukeWood 
Thanks for the review

I have updated the codes according to your comments and now can pass all relevant tests:

```bash
pytest keras_cv/layers/preprocessing/ragged_image_test.py
pytest keras_cv/layers/preprocessing/random_saturation_test.py
pytest keras_cv/layers/preprocessing/with_labels_test.py
pytest keras_cv/layers/preprocessing/with_segmentation_masks_test.py
pytest keras_cv/layers/preprocessing/with_mixed_precision_test.py
```

james77777778:
Once this PR successfully merged, I think I can take a look at `RandomBrightness`, `RandomContrast` and `RandomHue` to complete `RandomColorJitter` with similar approach?

keras_cv/layers/preprocessing/random_saturation_test.py
```diff
@@ -93,3 +172,29 @@ def test_config(self):
         self.assertTrue(isinstance(config["factor"], core.UniformFactorSampler))
         self.assertEqual(config["factor"].get_config()["lower"], 0.3)
         self.assertEqual(config["factor"].get_config()["upper"], 0.8)
+
+    def test_correctness_with_tf_adjust_saturation_normalized_range(self):
```
ianstenbit, line 176: @LukeWood I really like the pattern of A/B testing this before merging these changes, but I'm not sure it belongs in unit tests permanently. wdyt?

james77777778: Hi @ianstenbit 
I'm also wondering if there has better solution. I think the maintenance of old layers will be a problem if
1. `BaseImageAugmentationLayer` is modified or deprecated
2. It is implicit to have a class definition in unit tests

If the consistency of outputs is the major concern, maybe we can extract the computation part (`tf.multiply`, `tf.image`, ..., etc) instead of whole class definition for unit tests?

LukeWood: > @LukeWood I really like the pattern of A/B testing this before merging these changes, but I'm not sure it belongs in unit tests permanently. wdyt?

So, I see why it might not belong in unit tests permanently - but we do need a way to monitor the deltas here during code reviews.  We can't really just ask contributors "did you test this" - we could ask for a colab perhaps, but I think it fits a bit more cleanly in the unit tests.

Personally, I'd opt for include in unit tests for now, then we can remove them all at the end once we feel we have achieved stability.

bhack:
@LukeWood As we are accumulating benchmarks can with these PR can we add also a jit_compile version of in the benchmark?

If it fails the compilation we can wrap it with try/except (in benchmark) or if in test we could wrap it directly with `@unittest.expectedFailure` as in:
https://github.com/keras-team/keras-cv/blob/32701437b713050de951351529335f1fc725ceb0/keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py#L50

LukeWood:
> @LukeWood As we are accumulating benchmarks can with these PRs also add a jit_compile version in the benchmark?
> 
> If it fails the compilation we can wrap it with try/except (in benchmark) or if in test we could wrap it directly with `@unittest.expectedFailure` as in:
> 
> https://github.com/keras-team/keras-cv/blob/32701437b713050de951351529335f1fc725ceb0/keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py#L50

adding an XLA version in a try catch to the benchmarks sounds good to me!  Feel free to raise a PR and add it if you'd like

bhack:
> adding an XLA version in a try catch to the benchmarks sounds good to me! Feel free to raise a PR and add it if you'd like this contribution marked

Probably it is better to request this in tests coverage of the PR as at least they are runned also in CI so we could monitor tcontinuously the compilability over the time

Benchmarks are not runned by CI.

keras_cv/layers/preprocessing/random_saturation_test.py
ianstenbit: Okay that sgtm. Thank you Luke and thank you @james77777778 for your work on this!

ianstenbit: (One other option we could consider is a/b testing in the benchmark script).

Personally I find that a little bit nicer as then we don't have to have two copies of `OldRandomSaturation` wdyt?

ianstenbit (approved):
LGTM, but one question about where we should do numerical a/b testing (in comments)

keras_cv/layers/preprocessing/random_saturation_test.py
LukeWood: ok yeah I agree with @ianstenbit - let's put this in the benchmarks.

james77777778: Got it. I'm going to move a/b testing into `benchmarks/vectorized_random_saturation.py`.

Just one question:
During the migration of vectorizing preprocessing layers, the number of benchmark script might increase rapidly. Should I make new folder to put it?

LukeWood:
Thanks you @james77777778 for the great PR!

PR #1394, opened 2023-02-14
Added vectorized Auto Contrast implementation.

### LINKED ISSUE #1382: Vectorize Auto Contrast.
**Short Description**
Testing the water with this issue as https://github.com/keras-team/keras-cv/pull/1373 is merged.

How do you feel about migrating more layers to use vectorized approach? Below is a simple performance comparison of vectorized vs current. 

EDIT: misclick and submitted the issue too early...

Eager performance is better.  
Graph performance is the same.
Eager performance
Graph performance

sebastian-sz:
# What does this PR do?
Modifies `AutoContrast` to use `Vectorized` base image layer.

Fixes #1382 

Following #1392 I also added the benchmark and numerical check (compare old and new).
Benchmark results:
autocontrast_benchmark_2

autocontrast_benchmark_1

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood @ianstenbit

### DIFF
--- a/keras_cv/layers/preprocessing/auto_contrast_test.py
+++ b/keras_cv/layers/preprocessing/auto_contrast_test.py
@@ -14,7 +14,73 @@
 
 import tensorflow as tf
 
+from keras_cv.layers import BaseImageAugmentationLayer
 from keras_cv.layers import preprocessing
+from keras_cv.utils import preprocessing as utils_preprocessing
+
+
+class OldAutoContrast(BaseImageAugmentationLayer):
+    """Performs the AutoContrast operation on an image.
+
+    Auto contrast stretches the values of an image across the entire available
+    `value_range`.  This makes differences between pixels more obvious.  An example of
+    this is if an image only has values `[0, 1]` out of the range `[0, 255]`, auto
+    contrast will change the `1` values to be `255`.
+
+    Args:
+        value_range: the range of values the incoming images will have.
+            Represented as a two number tuple written [low, high].
+            This is typically either `[0, 1]` or `[0, 255]` depending
+            on how your preprocessing pipeline is setup.
+    """
+
+    def __init__(
+        self,
+        value_range,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.value_range = value_range
+
+    def augment_image(self, image, transformation=None, **kwargs):
+        original_image = image
+        image = utils_preprocessing.transform_value_range(
+            image,
+            original_range=self.value_range,
+            target_range=(0, 255),
+            dtype=self.compute_dtype,
+        )
+
+        low = tf.reduce_min(tf.reduce_min(image, axis=0), axis=0)
+        high = tf.reduce_max(tf.reduce_max(image, axis=0), axis=0)
+        scale = 255.0 / (high - low)
+        offset = -low * scale
+
+        image = image * scale[None, None] + offset[None, None]
+        result = tf.clip_by_value(image, 0.0, 255.0)
+        result = utils_preprocessing.transform_value_range(
+            result,
+            original_range=(0, 255),
+            target_range=self.value_range,
+            dtype=self.compute_dtype,
+        )
+        # don't process NaN channels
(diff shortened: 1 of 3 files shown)


LukeWood:
> # What does this PR do?
> Modifies `AutoContrast` to use `Vectorized` base image layer.
> 
> Fixes #1382
> 
> Following #1392 I also added the benchmark and numerical check (compare old and new). Benchmark results: autocontrast_benchmark_2
> 
> autocontrast_benchmark_1
> 
> ## Before submitting
> * [ ]  This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
> * [x]  Did you read the [contributor guideline](https://github.com/keras-team/keras-cv/blob/master/.github/CONTRIBUTING.md),
>   Pull Request section?
> * [x]  Was this discussed/approved via a Github issue? Please add a link
>   to it if that's the case.
> * [x]  Did you write any new necessary tests?
> * [ ]  If this adds a new model, can you run a few training steps on TPU in Colab to ensure that no XLA incompatible OP are used?
> 
> ## Who can review?
> @LukeWood @ianstenbit

@sebastian-sz any idea why the original is faster in some cases?  This is surprising to me, perhaps vectorized-map was performing quite well here?  

I assume this is still better when they pass bounding boxes in too though - as it would fallback to map_fn.

bhack:
> @sebastian-sz any idea why the original is faster in some cases? This is surprising to me, perhaps vectorized-map was performing quite well here?

Cause it was never in the fallback list for missing converters and it was not using tf.image OPS:
https://github.com/keras-team/keras-cv/issues/291#issuecomment-1424768222

LukeWood:
> > @sebastian-sz any idea why the original is faster in some cases? This is surprising to me, perhaps vectorized-map was performing quite well here?
> 
> Cause it was never in the fallback list for missing `converters` and it was not using `tf.image` OPS: [#291 (comment)](https://github.com/keras-team/keras-cv/issues/291#issuecomment-1424768222)
> 
> You could already run `AutoContrast` on master with disabling fallback and enabling warning:
> 
> ```python
>         if self.auto_vectorize:
>             return tf.vectorized_map(func, inputs, fallback_to_while_loop=False, warn=True)
> ```

that explains why its the same speed, but I'm still surprised that its *faster*.  Maybe its vectorizing something we didn't realize we could vectorize.

sebastian-sz:
@bhack @LukeWood 
I did inspect both layers, but unfortunately, I cannot pinpoint the place where the difference in execution time comes from.

These differences are not huge (~5ms), yet still, they exist.

LukeWood:
> @bhack @LukeWood I did inspect both layers, but unfortunately, I cannot pinpoint the place where the difference in execution time comes from.
> 
> These differences are not huge (~5ms), yet still, they exist.

yeah they're pretty tiny.  Also - the one that relies on vectorized_map will use `map_fn` when bounding boxes are provided, so yours will outperform in this case by a good margin.  I'm totally happy w/ this performance personally.

ianstenbit (approved):
LGTM. Thank you @sebastian-sz!

@LukeWood PTAL and let me know what you think about the a/b testing comment on #1392 as that applies here as well

keras_cv/layers/preprocessing/auto_contrast_test.py
```diff
@@ -90,3 +156,11 @@ def test_auto_contrast_properly_converts_value_range(self):
 
         self.assertTrue(tf.math.reduce_any(ys[0] == 0.0))
         self.assertTrue(tf.math.reduce_any(ys[0] == 1.0))
+
+    def test_is_consistent_with_non_vectorized_implementation(self):
```
ianstenbit, line 160: @sebastian-sz can you please move this test case to the benchmark, instead of this test case?

Then we can merge -- thanks!

sebastian-sz: @ianstenbit done.
1. Moved consistency test to benchmarks.
2. Added `get_random_transformation_batch` override as it's unused.

PR #1402, opened 2023-02-15
Enforce line length of 80 chars

jbischof:
Line length is currently set at 88 chars. Bring this back to 80 chars to agree with Keras and KerasNLP.

Note that this does not work for docstrings and comments. We have exempted them to allow for hyperlinks.


qlzh727 (approved):
Thanks for the fix.

shell/lint.sh
```diff
@@ -31,7 +31,7 @@ then
   exit 1
 fi
 [ $# -eq 0 ] && echo "no issues with flake8"
-black --check $files
+black --line-length 80 --check $files
```
bhack, line 34: I suggest to move this in the configuration file or we misaling IDE integrations.

LukeWood (approved):
fine by me as long as @haifeng-jin has no specific reason for 88.

we took 88 from keras-tuner, dont think it should be an issue as long as it is uniform.  LGTM

shell/lint.sh
jbischof: @bhack I like your suggestion but let's save this for a followup PR. We need to find a unified solution that works across all `keras-team` repos long term.

jbischof:
\gcbrun

PR #1404, opened 2023-02-16
Remove the need for a image input shape in StableDiffusion

LukeWood:
This is done by changing the call to `shape` with `tf.shape()`, and changing **-0.5 to 1/tf.sqrt(c).

Slightly better UX, output dim is constant so users will not need to resize input images anymore.


fchollet (approved):
LGTM, that's better

LukeWood:
@fchollet before merging I will run a few workflows.

jbischof:
Wait, does this change the SD API? I don't see that in the PR

LukeWood:
> Wait, does this change the SD API? I don't see that in the PR

It should only change the ImageEncoder, which our users get via a property on StableDiffusion.  API contract for constructing and using SD is unchanged; but image encoder works on any image size

PR #1409, opened 2023-02-16
Add box delta metric

LukeWood:
MeanBoxCountDelta counts the difference of counts of predicted and true boxes.

MeanBoxCountDelta looks at the number of boxes in the ground truth dataset
and counts the delta between that number and the number of boxes your object
detection model predicted for that image. This is primarily useful when
attempting to tune the confidence threshold of your
`MultiClassNonMaxSuppression` layer in an object detection model.  If this
metric is high, it indicates that your model is making too many or too few
predictions.  Ideally this metric will be zero.

Usage:

```python
mean_box_count_delta = keras_cv.metrics.MeanBoxCountDelta()
y_true = {
    "classes": [[0, -1, -1]],
    "boxes": [[[0, 0, 1, 1], [0, 0, 1, 1], [0, 0, 1, 1]]],
}
y_pred = {
    "classes": [[0, 1, 1, -1]],
    "boxes": [[[0, 0, 1, 1], [0, 0, 1, 1], [0, 0, 1, 1], [0, 0, 1, 1]]],
}
mean_box_count_delta.update_state(y_true, y_pred)
print(mean_box_count_delta.result())
# > 2.0

### CODE REVIEW

File: keras_cv/metrics/object_detection/__init__.py [hunk 1]
```diff
@@ -0,0 +1,13 @@
+# Copyright 2023 The KerasCV Authors
```
ianstenbit, line 1: Is object_detection the right dir for this? Do we plan to move COCO metrics here?

keras_cv/metrics/object_detection/mean_box_count_delta.py
```diff
@@ -0,0 +1,102 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+import tensorflow.keras as keras
+import tensorflow.keras.initializers as initializers
+
+from keras_cv import bounding_box
+
+
+class MeanBoxCountDelta(keras.metrics.Metric):
```
ianstenbit, line 21: For all other Mean metrics I'm aware of (except mAP), the "Mean" prefix is omitted. Should we do that here?

ianstenbit, line 21: Why not use `keras.metrics.MeanMetricWrapper`?

keras_cv/metrics/object_detection/mean_box_count_delta.py
```diff
@@ -0,0 +1,102 @@
...
+    """MeanBoxCountDelta counts the difference of counts of predicted and true boxes.
+
+    MeanBoxCountDelta looks at the number of boxes in the ground truth dataset
```
ianstenbit, line 24: the fact that it uses absolute value is critical to point out in the docstring imo

keras_cv/metrics/object_detection/mean_box_count_delta.py
```diff
@@ -0,0 +1,102 @@
...
+    and counts the delta between that number and the number of boxes your object
+    detection model predicted for that image. This is primarily useful when
+    attempting to tune the confidence threshold of your
+    `MultiClassNonMaxSuppression` layer in an object detection model.  If this
```
ianstenbit, line 28: This feels weird -- perhaps something other than `Mean` is appropriate so that it's more obvious whether there are too many or too few boxes?

keras_cv/bounding_box/ensure_tensor.py
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from keras_cv.utils import preprocessing
+
+
+def ensure_tensor(boxes, dtype=None):
```
ianstenbit, line 18: This same method name exists in keras_cv.utils with different functionality -- do we want that? Should this be renamed?

ianstenbit (changes requested):

keras_cv/metrics/object_detection/mean_box_count_delta.py
LukeWood: Yeah... I've thought about this too.  Sum doesn't work, as if you predict [-5] too few boxes one image, then [5] too many the next you get `0` which looks correct.  

Any better ideas?  Maybe this just means the name is bad?

keras_cv/bounding_box/ensure_tensor.py
LukeWood: I believe we do that for a ton of `bounding_box` functions.  From my perspective we should make bounding boxes behave as closely to tensors as possible, while still being dictionaries.

keras_cv/metrics/object_detection/mean_box_count_delta.py
LukeWood: Good point - updated to omit

keras_cv/metrics/object_detection/mean_box_count_delta.py
LukeWood: we may even want to include an option for both!  Updated.

keras_cv/metrics/object_detection/__init__.py
LukeWood: Yeah, I personally beleieve so.  not a user facing API - just an organizational thing.  

Personally I think task is a nice way to divide up metrics

keras_cv/metrics/object_detection/mean_box_count_delta.py
LukeWood: So, I did try this and unfortunately the `ensure_tensor` call errors out in the `call()` method.

keras_cv/metrics/object_detection/mean_box_count_delta.py
LukeWood: I added a `relative` mode to try to fix this issue.  You can have both metrics now: a relative one, and an absolute one.  This will help you run down issues in all cases.

keras_cv/metrics/object_detection/box_count_delta.py
```diff
@@ -0,0 +1,127 @@
...
+    BoxCountDelta looks at the number of boxes in the ground truth dataset
+    and counts the delta between that number and the number of boxes your object
+    detection model predicted for that image. This is primarily useful when
+    attempting to tune the confidence threshold of your
+    `MultiClassNonMaxSuppression` layer in an object detection model.  If this
+    metric is high, it indicates that your model is making too many or too few
+    predictions.  Ideally this metric will be zero.
+
+    Args:
+      mode: (Optional) either 'relative' or 'absolute'.  When set to 'absolute',
```
ianstenbit, line 33: Let's mention what relative does, and when that might be useful.

keras_cv/metrics/object_detection/box_count_delta.py
```diff
@@ -0,0 +1,127 @@
...
+class BoxCountDelta(keras.metrics.Metric):
+    """BoxCountDelta counts the difference of counts of predicted and true boxes.
+
+    BoxCountDelta looks at the number of boxes in the ground truth dataset
+    and counts the delta between that number and the number of boxes your object
+    detection model predicted for that image. This is primarily useful when
+    attempting to tune the confidence threshold of your
+    `MultiClassNonMaxSuppression` layer in an object detection model.  If this
+    metric is high, it indicates that your model is making too many or too few
+    predictions.  Ideally this metric will be zero.
```
ianstenbit, line 30: nit: maybe "close to zero"?

keras_cv/metrics/object_detection/box_count_delta.py
```diff
@@ -0,0 +1,127 @@
...
+
+        self.delta_sum.assign_add(
+            tf.cast(tf.math.reduce_sum(delta_sum), self.compute_dtype)
+        )
+        self.samples.assign_add(
+            tf.cast(tf.shape(y_pred["classes"])[0], self.compute_dtype)
+        )
+
+    def result(self):
+        return self.delta_sum / self.samples
```
ianstenbit, line 124: From a usefulness perspective, should we include stddev and mean as a dict here?

It seems like we could end up with a `mean` near `0` despite having a lot of extra and missing boxes, just in different images.

Your call, just an idea.

ianstenbit (approved):
This LGTM from an implementation perspective.

From a usefulness perspective, I'll defer to you and @martin-gorner

keras_cv/metrics/object_detection/box_count_delta.py
LukeWood: I really like the idea of outputting the standard deviation.  I'll see if I can introduce this.

keras_cv/metrics/object_detection/box_count_delta.py
LukeWood: I've added a paragraph on how to interpret this.

LukeWood:
Probably not super useful - maybe useful in niche cases.

at least not worth maintaining - as its easy to write your own in 15~ lines.

PR #1411, opened 2023-02-17
Revisit `value_range` in RandomContrast, RandomHue, RandomColorJitter, RandomBrightness

### LINKED ISSUE #1407: Add `value_range` to RandomContrast

james77777778:
# What does this PR do?

Fixes #1407 

1. `value_range` is introduced to `RandomContrast`
2. `value_range` is removed in `RandomHue`
3. `preprocessing_utils.transform_value_range` is removed in `RandomColorJitter`

I have checked that `tf.image.rgb_to_hsv` should have outputs with
- H and S channel in [0, 1]
- V channel in original value range

Ref:
[tf.image.rgb_to_hsv kernel](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/image/colorspace_op.h#L28)
[OpenCV RGB ↔ HSV](https://docs.opencv.org/4.x/de/d25/imgproc_color_conversions.html#color_convert_rgb_hsv)

So, the value range transform in RandomHue can be removed. I think the original implementation needs it because of requirement from `tf.image.adjust_hue`

RandomColorJitter can also skip value range transform because all sub-layers are handled it individually.

These modifications should simplify the process and keep numerical consistency with previous implementation.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood


james77777778:
I updated some unit tests:

1. `keras_cv/layers/preprocessing/ragged_image_test.py`
    - Add RandomBrightness and RandomContrast (they were missing before)
    - Update `value_range`
2. `keras_cv/layers/preprocessing/with_labels_test.py`
    - Update `value_range`
3. `keras_cv/layers/preprocessing/with_mixed_precision_test.py`
    - Add RandomContrast (it was missing before)
    - Update `value_range`
    - Previously, RandomSaturation, RandomColorJitter and RandomHue were tagged with `NO_CPU_FP16_KERNEL_LAYERS`, but now they all can pass the test
4. `keras_cv/layers/preprocessing/with_segmentation_masks_test.py`
    - Update `value_range`
 
In addition, I find the `rank` assertion in RandomBrightness would make `ragged_image_test.py` fail and we usually don't check the `rank` of input images in other preprocessing layers. So I simply removed it.

There are some big changes in unit tests and I hope I don't mess it up. Looking forward to your reviews!

keras_cv/layers/preprocessing/random_hue.py
```diff
@@ -53,12 +49,11 @@ class RandomHue(VectorizedBaseImageAugmentationLayer):
     ```
     """
 
-    def __init__(self, factor, value_range, seed=None, **kwargs):
```
LukeWood, line 56: Any reason to remove the value range here?  We use it in clipping,  so I believe it is valid.

benchmarks/vectorized_random_hue.py
```diff
@@ -51,7 +51,7 @@ class OldRandomHue(BaseImageAugmentationLayer):
 
     """
 
-    def __init__(self, factor, value_range, seed=None, **kwargs):
+    def __init__(self, factor, value_range=(0, 255), seed=None, **kwargs):
```
LukeWood, line 54: please remove default values for value_range

keras_cv/layers/preprocessing/random_color_jitter.py
```diff
@@ -66,51 +61,59 @@ class RandomColorJitter(VectorizedBaseImageAugmentationLayer):
             is used, a value between `0.0` and the passed float is sampled.
             In order to ensure the value is always the same, please pass a tuple
             with two identical floats: `(0.5, 0.5)`.
+        value_range: Optional list/tuple of 2 floats for the lower and upper limit
```
LukeWood, line 64: any reason to re-order this?  consistency I suppose?  Either way, this parameter needs to be required, not optional

keras_cv/layers/preprocessing/random_color_jitter.py
```diff
@@ -66,51 +61,59 @@ class RandomColorJitter(VectorizedBaseImageAugmentationLayer):
...
     """
 
     def __init__(
         self,
-        value_range,
         brightness_factor,
         contrast_factor,
         saturation_factor,
         hue_factor,
+        value_range=(0, 255),
```
LukeWood, line 90: remove default

LukeWood (changes requested):
Thanks for the PR!  Asked some questions, please address my comments in the PR!

keras_cv/layers/preprocessing/random_hue.py
james77777778: Because Hue channel is always at [0, 1]

We can validate this from https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/image/colorspace_op.h#L49-L57

For example, we have `value_range` in [a, b] indicates a is the min and b is the max
The value range of:
- `V`: [a, b]
- `range` : [0, b - a]
- `norm`: [0, (b-a)*(1/6)]
So the min of `H` is `0` if `R==V` and `G-B=0`
and the max of `H` occurs at `R==V` and `G-B` is slightly negative => value will be `H=H+1` at line57 and becomes very close to `1`

There is a colab to verify the behavior of this
https://colab.research.google.com/drive/1KBPQCL9pxI4Ar6KvnhfzncbTI8zT6kGq?usp=sharing

```bash
value_range: [0, 1]
  H channel: max= 1.00000;min= 0.00000
value_range: [0, 255]
  H channel: max= 1.00000;min= 0.00000
value_range: [0, 100]
  H channel: max= 1.00000;min= 0.00000
value_range: [-50, 100]
  H channel: max= 1.00000;min= 0.00000
```

In RandomHue, we add `adjust_factors` which is in range [-0.5, 0.5] and post-process by
```python
h_channel = tf.where(h_channel > 1.0, h_channel - 1.0, h_channel)
h_channel = tf.where(h_channel < 0.0, h_channel + 1.0, h_channel)
```
So the H channel is always at [0, 1] no matter what we do with `tf.image.rgb_to_hsv` and the operations of RandomHue

keras_cv/layers/preprocessing/random_color_jitter.py
james77777778: Yes, for consistency.
`value_range` is a bit orderless across preprocessing layers

I have a list of this:

## At 1st argument:
- AugMix
- Equalization
- Posterization
- RandAugment
- RandomChannelShift
- Solarization
- RandomColorJitter

## Not at 1st argument
- RandomBrightness
- RandomContrast
- RandomHue
- RandomSharpness

Should `value_range` be at 1st argument or next to `seed`/`**kwargs`?

I have no idea. I just make RandomColorJitter consistent with its sub-layers.

keras_cv/layers/preprocessing/random_color_jitter.py
james77777778: A following question:
Should I also remove default value (0, 255) in RandomBrightness, RandomContrast?

keras_cv/layers/preprocessing/random_color_jitter.py
james77777778: I'm going to make all `value_range` next to `seed` or `**kwargs` and remove default values.

This should improve user experience when using preprocessing layers.

benchmarks/vectorized_random_hue.py
james77777778: Done.

keras_cv/layers/preprocessing/random_color_jitter.py
james77777778: Done.
Also I change the position of `value_range` and remove default value.

Now only RandomSharpness has `value_range` after `factor`.

keras_cv/layers/preprocessing/random_color_jitter.py
james77777778: Done.

And I revert the change of `value_range` order. It is the same as previous implementation.

keras_cv/layers/preprocessing/random_hue.py
james77777778: @LukeWood 
There is no `value_range` in RandomSaturation for the same reason.
https://github.com/keras-team/keras-cv/pull/1411/files#r1112453229

To be more specific, the `value_range` in previous RandomHue is not for clipping ([previous implementation](https://github.com/keras-team/keras-cv/blob/15056344d08d234537fdf5b61d77c66c1fae691b/keras_cv/layers/preprocessing/random_hue.py#L81-L101)), but for value transformation (which is useless)

james77777778:
@LukeWood 
I have updated the PR:
- move `value_range` to the 1st argument in RandomBrightness, RandomContrast, RandomColorJitter. Now, only RandomSharpness has different position of `value_range`.
- fix benchmarks due to the changes

And the reason to remove `value_range` in RandomHue is at [comment](https://github.com/keras-team/keras-cv/pull/1411#discussion_r1112453229). In short, H channel is always in [0, 1] after `tf.image.rgb_to_hsv` with arbitrary input value range and the operations of RandomHue keep H channel in [0, 1] as well. So actually the clipping and value transform are useless in RandomHue.

james77777778:
I'm going to deal with these step by step, so the PR should be closed.

#1525

PR #1433, opened 2023-02-22
Vectorize ChannelShuffle

### LINKED ISSUE #1414: Vectorize ChannelShuffle
The PR contributing this should be based on:

https://github.com/keras-team/keras-cv/pull/1394

james77777778:
# What does this PR do?

Fixes #1414 

Benchmark results:

comparison
comparison_no_old_eager

Runtime gets high at some points due to retracing and some warning.
```
WARNING:tensorflow:Using a while_loop for converting RandomShuffle cause there is no registered converter for this op.
```
Here is the [log.txt](https://github.com/keras-team/keras-cv/files/10799632/log.txt)

Overall, I think vectorized implementation beats old one.

To check the consistency, I manually set batch_size=1 to make `tf.random.shuffle` acts the same for new and old implementation:
```python
class ChannelShuffleTest(tf.test.TestCase):
    def test_consistency_with_old_impl(self):
        # must set batch_size=1 due to randomness from
        # images = tf.random.shuffle(images, seed=self.seed)
        image_shape = (1, 32, 32, 3)
        fixed_seed = 2023
        image = tf.random.uniform(shape=image_shape) * 255.0

        layer = ChannelShuffle(groups=3, seed=fixed_seed)
        old_layer = OldChannelShuffle(groups=3, seed=fixed_seed)

        output = layer(image)
        old_output = old_layer(image)

        self.assertAllClose(old_output, output)
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood @ianstenbit


james77777778:
In this PR, I find that channel shuffle cannot process independently.
Need some time to survey but it could be difficult because `tf.random.shuffle` cannot apply on batched inputs.

LukeWood:
> In this PR, I find that channel shuffle cannot process independently. Need some time to survey but it could be difficult because `tf.random.shuffle` cannot apply on batched inputs.

can we roll our own version instead of using `tf.random.shuffle`?  I believe this could be done with `tf.gather_nd()` and `tf.random.uniform()`.

james77777778:
> > In this PR, I find that channel shuffle cannot process independently. Need some time to survey but it could be difficult because `tf.random.shuffle` cannot apply on batched inputs.
> 
> can we roll our own version instead of using `tf.random.shuffle`? I believe this could be done with `tf.gather_nd()` and `tf.random.uniform()`.

Thanks for the keywords.
I have tested tf.TensorArray (as a output list) with similar implementation from previous one but the performance decreased a lot...

I'm going to try `tf.gather_nd()` and `tf.random.uniform()`

bhack:
I think we can do:

```python
num_channels = x.shape.as_list()[-1]

    # Generate a random permutation of the channel indices
    shuffled_indices = tf.random.shuffle(tf.range(num_channels), seed=seed)

    # Permute the channels of the input tensor according to the random indices
    shuffled_x = tf.gather(x, shuffled_indices, axis=-1)
```

james77777778:
@LukeWood @bhack 
Thanks for your guidance

Much of the difficulty comes from how to independently shuffle indices in batch.

I adopt the `rand+argsort` trick based on
https://stackoverflow.com/a/55317373

```python
    def get_random_transformation_batch(self, batch_size, **kwargs):
        # get batched shuffled indices
        # for example: batch_size=2; self.group=5
        # indices = [
        #     [0, 2, 3, 4, 1],
        #     [4, 1, 0, 2, 3]
        # ]
        indices_distribution = tf.random.uniform(
            (batch_size, self.groups), seed=self.seed
        )
        indices = tf.argsort(indices_distribution, axis=-1)
        return indices
```
and then gather by
```python
    def augment_images(self, images, transformations, **kwargs):
        ......
        indices = transformations

        # append batch indexes next to shuffled indices
        batch_indexs = tf.reshape(
            tf.repeat(tf.range(batch_size), self.groups),
            (batch_size, self.groups),
        )
        indices = tf.stack([batch_indexs, indices], axis=-1)

        channels_per_group = num_channels // self.groups
        images = tf.reshape(
            images, [batch_size, height, width, self.groups, channels_per_group]
        )
        images = tf.transpose(images, perm=[0, 3, 1, 2, 4])
        images = tf.gather_nd(images, indices=indices)
        images = tf.transpose(images, perm=[0, 2, 3, 4, 1])
        images = tf.reshape(images, [batch_size, height, width, num_channels])
        return images
```

All unit tests passed but it is difficult to check the consistency with non-vectorized implementation because the randomness comes from different sources (`tf.random.shuffle`, `tf.random.uniform`).
I just removed the test in benchmark script.

Also a independence test is added.

The performance result:
comparison

bhack:
> Much of the difficulty comes from how to independently shuffle indices in batch.

If I remember correctly is `tf.random.experimental.stateless_shuffle`

james77777778:
> > Much of the difficulty comes from how to independently shuffle indices in batch.
> 
> If I remember correctly is `tf.random.experimental.stateless_shuffle`

I think `tf.random.experimental.stateless_shuffle` acts the same as `tf.random.shuffle`
So they can only shuffle tensor along 1 axis?

In ChannelShuffle, we want to shuffle channels independently in each image like:
idx=1 => shuffled_idx = [1, 2, 0]
idx=2 => shuffled_idx = [0, 2, 1]
...

It is achievable to loop over each image with the operations you mentioned.
Pseudo code like:
```python
outputs = tf.TensorArray(...)
for i in range(batch_size):
    num_channels = images[i].shape.as_list()[-1]
    shuffled_indices = tf.random.shuffle(tf.range(num_channels), seed=seed)
    shuffled_x = tf.gather(images[i], shuffled_indices, axis=-1)
    outputs.write(i, shuffled_x)
outputs.stack()
```
But it is too slow that it is even slower than non-vectorized version in eager mode.

I cannot figure out how to vectorize this with `tf.random.shuffle` or `tf.random.experimental.stateless_shuffle`. Please enlighten me if I missed something.

bhack:
> In ChannelShuffle, we want to shuffle channels independently in each image like:
idx=1 => shuffled_idx = [1, 2, 0]
idx=2 => shuffled_idx = [0, 2, 1]

Something like?
https://colab.research.google.com/gist/bhack/65760a7896ee9f6ec450ebbe8bd32ca2/exp_random_shuffle.ipynb

james77777778:
> > In ChannelShuffle, we want to shuffle channels independently in each image like:
> > idx=1 => shuffled_idx = [1, 2, 0]
> > idx=2 => shuffled_idx = [0, 2, 1]
> 
> Something like this? https://colab.research.google.com/gist/bhack/65760a7896ee9f6ec450ebbe8bd32ca2/exp_random_shuffle.ipynb

https://colab.research.google.com/drive/1b-tZErA-jCv6SNurdprdOYc6dLFr0ZSH?usp=sharing
I manually set the values to show that
- only 1 shuffled indices for all batch
- outputs are shuffled by same indices and not independently across batch

bhack:
Ok, so you meant within the batch. Sorry I was confused on what kind of randomization policy we finally selected for the new vectorizing refactoring (mandatory within the batch randomization or not)

james77777778:
Kindly ping @LukeWood @ianstenbit 

The idea of vectorizing please see https://github.com/keras-team/keras-cv/pull/1433#issuecomment-1441154826

The current implementation shuffles channel (by groups) independently for each image in the batch

ianstenbit:
@LukeWood PTAL -- what's the decision re: independent augmentation in batches?

LukeWood:
All augmentation should be done independently on a per-image basis.

LukeWood (approved):

LukeWood:
/gcbrun

Thanks you @james77777778 !!!

LukeWood:
Sorry one more thing:  Can add a test like the one in https://github.com/keras-team/keras-cv/pull/1480/files#diff-1c771135874332a25b437da4a5706faea85ff719331dfe7beb255403a1a66477

james77777778:
> Sorry one more thing: Can add a test like the one in https://github.com/keras-team/keras-cv/pull/1480/files#diff-1c771135874332a25b437da4a5706faea85ff719331dfe7beb255403a1a66477

Hi @LukeWood 
It should be already added at 
https://github.com/keras-team/keras-cv/pull/1433/files#diff-67ea20e78d28e97daedb7b45698d50914d1ffef26cc62311c855606d6d171fc0R98-R105

benchmarks/vectorized_channel_shuffle.py
```diff
@@ -0,0 +1,195 @@
...
+import numpy as np
+import tensorflow as tf
+
+from keras_cv.layers import ChannelShuffle
+from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
+    BaseImageAugmentationLayer,
+)
+
+
+class OldChannelShuffle(BaseImageAugmentationLayer):
```
ianstenbit, line 26: Unless the behavior is changing as part of this PR, we should add a PyTest to this benchmark to verify that the old/new implementations are numerically identical (like [this one](https://github.com/keras-team/keras-cv/blob/master/benchmarks/vectorized_random_translation.py#L259))

ianstenbit: @LukeWood to confirm

ianstenbit (approved):
Thanks for the PR!

james77777778:
Hi @ianstenbit 

Luckily, I got a magic number (`seed=2023` 😂) that has the same randomness for old/new implementation. I have added the unit test to the benchmark.

Also I ran the script `examples/layers/preprocessing/classification/channel_shuffle_demo.py` to visualize the output of new ChannelShuffle

output

benchmarks/vectorized_channel_shuffle.py
james77777778: Please see https://github.com/keras-team/keras-cv/pull/1433#issuecomment-1472992645

ianstenbit:
GCB CI tests failing due to unrelated changed, should be green post-merge.

PR #1435, opened 2023-02-22
Vectorize RandomZoom

### LINKED ISSUE #1416: Vectorize RandomZoom
This PR should be base on:

https://github.com/keras-team/keras-cv/pull/1394

james77777778:
# What does this PR do?

Fixes #1416 

Benchmark results:

comparison

OldRandomZoom is so slow that I need to set `num_images = [100, 200, 500, 1000]` for benchmarks

Both fail to run with XLA due to lacking `ImageProjectiveTransformV3` support.
https://github.com/tensorflow/tensorflow/issues/55194

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood @ianstenbit

### DIFF
--- a/keras_cv/layers/preprocessing/random_zoom.py
+++ b/keras_cv/layers/preprocessing/random_zoom.py
@@ -184,14 +184,6 @@ class RandomZoom(VectorizedBaseImageAugmentationLayer):
     def augment_labels(self, labels, transformations, **kwargs):
         return labels
 
-    def augment_segmentation_masks(
-        self, segmentation_masks, transformations, **kwargs
-    ):
-        return segmentation_masks
-
-    def augment_bounding_boxes(self, bounding_boxes, transformations, **kwargs):
-        return bounding_boxes
-
     def get_zoom_matrix(self, zooms, image_height, image_width, name=None):
         """Returns projective transform(s) for the given zoom(s).
 
--- a/benchmarks/vectorized_random_zoom.py
+++ b/benchmarks/vectorized_random_zoom.py
@@ -11,13 +11,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import time
 
-
+import matplotlib.pyplot as plt
+import numpy as np
 import tensorflow as tf
 from keras import backend
 
-from keras_cv.layers.preprocessing.vectorized_base_image_augmentation_layer import (
-    VectorizedBaseImageAugmentationLayer,
+from keras_cv.layers import RandomZoom
+from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
+    BaseImageAugmentationLayer,
 )
 from keras_cv.utils import preprocessing as preprocessing_utils
 
@@ -27,8 +30,7 @@ H_AXIS = -3
 W_AXIS = -2
 
 
-@tf.keras.utils.register_keras_serializable(package="keras_cv")
-class RandomZoom(VectorizedBaseImageAugmentationLayer):
+class OldRandomZoom(BaseImageAugmentationLayer):
     """A preprocessing layer which randomly zooms images during training.
 
     This layer will randomly zoom in or out on each axis of an image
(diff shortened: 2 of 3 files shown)


keras_cv/layers/preprocessing/random_zoom.py
```diff
@@ -28,7 +28,7 @@
 
 
 @tf.keras.utils.register_keras_serializable(package="keras_cv")
-class RandomZoom(BaseImageAugmentationLayer):
+class RandomZoom(VectorizedBaseImageAugmentationLayer):
```
LukeWood, line 31: can we add a unit test to ensure that each image is zoomed independently?

This can be achieved by:

- generate an image using random.normal((1, 100, 100, 3))
- generate a batch with tf.stack([image, image], axis=0)
- call a random zoom layer
- assertNotAllClose(result[0], result[1])

This ensures we still have element-wise augmentation

LukeWood (changes requested):
Thanks for the swift work on this PR.

I'd say that this looks good to me, just want one more unit test to be sure we don't have a regression on the element-wise operation.  The test in the benchmark does not actually cover this as we have a constant sample - so let's add an explicit unit test for it.

Thanks again for the hard work.

keras_cv/layers/preprocessing/random_zoom.py
james77777778: Done.

I added a unit test:
```python
    def test_random_zoom_on_batched_images_independently(self):
        image = tf.random.uniform(shape=(100, 100, 3))
        input_images = tf.stack([image, image], axis=0)

        layer = RandomZoom(
            height_factor=(-0.4, -0.5), width_factor=(-0.2, -0.3)
        )

        results = layer(input_images)
        self.assertNotAllClose(results[0], results[1])
```
and it passed as expected.

LukeWood (approved):
Awesome!  Thanks for the quick update.  This looks good to me!

sebastian-sz:
@james77777778 good work with this vectorization.

I have one question:
This seems like geometrical transform, are functions `augment_bounding_boxes` and `augment_segmentation_masks` correct as passthrough?

I can see that they are not present in the non vectorized implementation.

james77777778:
> @james77777778 good work with this vectorization.
> 
> I have one question: This seems like geometrical transform, are functions `augment_bounding_boxes` and `augment_segmentation_masks` correct as passthrough?
> 
> I can see that they are not present in the non vectorized implementation.

Yeah, they are just passthrough. I think the support for bbox and segmentation masks should be another PR?

sebastian-sz:
@james77777778 
>Yeah, they are just passthrough. I think the support for bbox and segmentation masks should be another PR?

Maybe they could be left unimplemented? That way if someone uses them for Object Detection / Image Segmentation they will explicitly get an error. 
Otherwise, this might introduce silent error - if someone doesn't know that this is not yet implemented, they will use it and get incorrect boxes/masks.

james77777778:
> @james77777778
> 
> > Yeah, they are just passthrough. I think the support for bbox and segmentation masks should be another PR?
> 
> Maybe they could be left unimplemented? That way if someone uses them for Object Detection / Image Segmentation they will explicitly get an error. Otherwise, this might introduce silent error - if someone doesn't know that this is not yet implemented, they will use it and get incorrect boxes/masks.

Thank you for pointing that out. I just copied these functions from another vectorized layer.
I'm going to remove `augment_bounding_boxes`, `augment_segmentation_masks` ASAP

sebastian-sz:
@james77777778 Thanks! And great job with this layer!

LukeWood:
> @james77777778
> 
> > Yeah, they are just passthrough. I think the support for bbox and segmentation masks should be another PR?
> 
> Maybe they could be left unimplemented? That way if someone uses them for Object Detection / Image Segmentation they will explicitly get an error. Otherwise, this might introduce silent error - if someone doesn't know that this is not yet implemented, they will use it and get incorrect boxes/masks.

ah good catch @sebastian-sz -- we should definitely be raising an error on these!  Thank you for the review.

keras_cv/layers/preprocessing/random_zoom.py
```diff
@@ -138,60 +138,51 @@ def __init__(
...
             fill_value=self.fill_value,
             interpolation=self.interpolation,
         )
-        output = tf.squeeze(output, 0)
-        output.set_shape(original_shape)
-        return output
+        outputs.set_shape(original_shape)
+        return outputs
 
-    def augment_label(self, label, transformation, **kwargs):
```
LukeWood, line 181: label can be a passthrough

LukeWood (approved):

LukeWood:
Thank for the great contribution @james77777778 !!!

PR #1438, opened 2023-02-22
Use `from_preset` to load architecture and weights

jbischof:
This PR pilots the biggest single step to unifying the KerasCV and KerasNLP APIs. It sits downstream of the pilot for functional subclasses #1401 and will be followed by a PR introducing `Task` models for classification.

## Highlights of this PR:
* Use `from_preset` constructor to load weights instead of `weights` arg
* Reimplement config-in-code classes (e.g., `ResNet50V2Backbone`) with `from_preset` constructor
* Remove `pooling` and `include_top` args to be handled by `Task` models
* Add docstring examples to show basic usage
* Introduce `Backbone` class to hold generic methods and properties
* Rename `as_backbone` to `get_feature_extractor`
* Decouple testing per model to improve readability
* Test weight loading with `pytest`
* Introduce `conftest.py` to allow control of weight RCP testing

## API preview
See [gist](https://colab.sandbox.google.com/gist/jbischof/a8b15976911fa04c7472cfc5dfa6c342/kerascv-new-api-demo.ipynb) for a preview of the new API.

### Quick summary:
We will rely on the docstring and keras.io to communicate preset usage.

Example docstring:
```
from_preset(*args, **kwargs) method of builtins.type instance
    Instantiate ResNetV2Backbone model from preset architecture and weights.
    Args:
        preset: string. Must be one of "resnet18_v2", "resnet34_v2", "resnet50_v2", "resnet101_v2", "resnet152_v2", "resnet50_v2_imagenet".
            If looking for a preset with pretrained weights, choose one of
            "resnet50_v2_imagenet".
        load_weights: Whether to load pre-trained weights into model.
            Defaults to `None`, which follows whether the preset has
            pretrained weights available.
    
    Examples:
    ```python
    # Load architecture and weights from preset
    model = keras_cv.models.ResNetV2Backbone.from_preset(
        "resnet50_v2_imagenet",
    )
    
    # Load randomly initialized model from preset architecture with weights
    model = keras_cv.models.ResNetV2Backbone.from_preset(
        "resnet50_v2_imagenet",
        load_weights=False,
    ```
```
Example usage:
```python
# Generic constructor
model = ResNetV2Backbone(
    stackwise_filters=[64, 128, 256, 512],
    stackwise_blocks=[2, 2, 2, 2],
    stackwise_strides=[1, 2, 2, 2],
    include_rescaling=False,
    input_shape=[256, 256, 3],
)

# Load preset architecture without weights
# Can also include overrides
model = ResNetV2Backbone.from_preset(
    "resnet18_v2",
    include_rescaling=False,
)

# Load preset architecture with weights
model = ResNetV2Backbone.from_preset("resnet50_v2_imagenet")

# Use as backbone
model = RetinaNet(
    classes=20,
    bounding_box_format="xywh",
    backbone=ResNetV2Backbone.from_preset(
        "resnet50_v2_imagenet").get_feature_extractor(),

)
```

# keras.applications aliases

We've also reintroduced versions of the `keras.applications` config-in-code classes now powered by `from_preset`.

```python
# Load a preset without weights
model = ResNet50V2Backbone(
    include_rescaling=False,
)

# Load a preset with weights
model = ResNet50V2Backbone.from_preset("resnet50_v2_imagenet")

# Use as backbone
model = RetinaNet(
    classes=20,
    bounding_box_format="xywh",
    backbone=ResNet50V2Backbone().get_feature_extractor(),
)
```


jbischof:
\gcbrun

keras_cv/callbacks/pycoco_callback_test.py
```diff
@@ -34,9 +34,11 @@ def test_model_fit_retinanet(self):
...
             classes=10,
             bounding_box_format="xywh",
-            backbone=keras_cv.models.ResNet50V2(
-                include_top=False, include_rescaling=True, weights=None
-            ).as_backbone(),
+            backbone=keras_cv.models.ResNetV2Backbone.from_preset(
+                "resnet50_v2_imagenet",
+                load_weights=False,
+                include_rescaling=True,
+            ).extract_features(),
```
LukeWood, line 41: can be a discussion for outside of this PR but do we want to support the passing of arbitrary levels/layers here?

keras_cv/callbacks/pycoco_callback_test.py
```diff
@@ -65,9 +67,11 @@ def test_model_fit_rcnn(self):
         model = keras_cv.models.FasterRCNN(
             classes=10,
             bounding_box_format="xywh",
-            backbone=keras_cv.models.ResNet50V2(
-                include_top=False, include_rescaling=True, weights=None
-            ).as_backbone(),
+            backbone=keras_cv.models.ResNetV2Backbone.from_preset(
+                "resnet50_v2_imagenet",
+                load_weights=False,
+                include_rescaling=True,
```
LukeWood, line 73: what's your intuition on `include_rescaling` at this time?  Are we still intending to include it in every from_preset call?

keras_cv/models/__init__.py
```diff
@@ -91,9 +91,7 @@
 from keras_cv.models.resnet_v1 import ResNet50
 from keras_cv.models.resnet_v1 import ResNet101
 from keras_cv.models.resnet_v1 import ResNet152
-from keras_cv.models.resnet_v2 import ResNet50V2
-from keras_cv.models.resnet_v2 import ResNet101V2
-from keras_cv.models.resnet_v2 import ResNet152V2
+from keras_cv.models.resnet_v2_backbone import ResNetV2Backbone
```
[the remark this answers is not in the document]
LukeWood: or just `backbones/resnet50v2.py`

keras_cv/models/backbone.py
```diff
@@ -0,0 +1,197 @@
...
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_cv.utils.python_utils import classproperty
+from keras_cv.utils.python_utils import format_docstring
+
+
+class Backbone(keras.Model):
+    """Base class for Backbone models."""
```
LukeWood, line 26: perhaps a little description of what we mean by "backbone" would be good - especially for those who stumble on it by browsing the code base.

keras_cv/models/backbone.py
```diff
@@ -0,0 +1,197 @@
...
+                f""" "{'", "'.join(cls.presets_with_weights)}"."""
+            )
+
+        config = metadata["config"]
+        model = cls.from_config({**config, **kwargs})
+
+        if not load_weights:
+            return model
+
+        weights = keras.utils.get_file(
```
LukeWood, line 97: so, in the case where we have `anchors` (probably akin to vocabulary?) do we have to special case this in the subclass ?

keras_cv/models/resnet_v2_backbone.py
```diff
@@ -440,40 +350,25 @@ def __init__(
         )(x)
         x = layers.Activation("relu", name="post_relu")(x)
 
-        if include_top:
+        if pooling == "avg":
```
LukeWood, line 353: here's an open question: should pooling belong to backbones or task models.

keras_cv/models/resnet_v2_backbone.py
```diff
@@ -369,35 +301,13 @@ def __init__(
         stackwise_blocks,
         stackwise_strides,
         include_rescaling,
-        include_top,
```
LukeWood, line 372: one thing I don't like is that `stackwise_blocks`, `stackwise_strides`, are now required parameters --- most users don't really care about the details of these arguments but rather want something like ResNet50V2?   Are we basically committing to users only constructing ResNets from pretrained in this change?  I see we've removed ResNet50/etc now

LukeWood, line 372: for boolean parameters such as include_rescaling I think a * to make keyword required helps readability

keras_cv/models/resnet_v2_backbone.py
```diff
@@ -482,189 +377,20 @@ def get_config(self):
...
-        input_tensor=input_tensor,
-        pooling=pooling,
-        classes=classes,
-        classifier_activation=classifier_activation,
-        block_type="block",
-        **kwargs,
-    )
-
-
-def ResNet101V2(
```
LukeWood, line 602: a meta comment: but I do think that `ResNet50v2(...)` does read a bit nicer than `ResNetv2Backbone.from_pretrained(...)`

keras_cv/models/resnet_v2_backbone.py
```diff
@@ -482,189 +377,20 @@ def get_config(self):
...
+    def presets(cls):
+        return copy.deepcopy(backbone_presets)
 
-setattr(ResNet18V2, "__doc__", BASE_DOCSTRING.format(name="ResNet18V2"))
-setattr(ResNet34V2, "__doc__", BASE_DOCSTRING.format(name="ResNet34V2"))
-setattr(ResNet50V2, "__doc__", BASE_DOCSTRING.format(name="ResNet50V2"))
-setattr(ResNet101V2, "__doc__", BASE_DOCSTRING.format(name="ResNet101V2"))
-setattr(ResNet152V2, "__doc__", BASE_DOCSTRING.format(name="ResNet152V2"))
+    @classproperty
+    def presets_with_weights(cls):
```
LukeWood, line 395: I like that theres an abstraction here to make this easier/more browsable.

LukeWood (commented):
I did a first pass --- overall I really like the abstractions around presets/weight loadings.

My only real concern revolves around the readability of `ResNet50v2(...)` vs `ResNetV2Backbone.from_preset(...)1

keras_cv/models/__init__.py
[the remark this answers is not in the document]
jbischof: Where would we put the classifier `Task` models in this directory structure? I could put them in a `classifiers/` folder to mimic other tasks like `object_detection/`, but would it be awkward to separate `ResNetV2Backbone` and `ResNetV2Classifier`?

keras_cv/callbacks/pycoco_callback_test.py
jbischof: The body of this function is identical to `as_backbone`. If we want to enhance that method it should probably be a separate PR.

keras_cv/callbacks/pycoco_callback_test.py
jbischof: Either we add `include_rescaling` as a positional arg in `from_preset` or we add a default in all the presets. From talking to the team I sense preference would be toward the former.

keras_cv/models/backbone.py
jbischof: No, there will be a separate `Task` class to handle e2e models like classification or object detection. It will have its own `from_preset` method.

keras_cv/models/resnet_v2_backbone.py
jbischof: When looking at the existing code it seems like pooling is only an option when `include_top=False`. Therefore I was a little confused when this option is actually used---maybe for another task like objection detection?

Either way, these layers have no weights so their presence does not impact preset loading.

keras_cv/models/resnet_v2_backbone.py
jbischof: Yes, you either use the generic constructor or you use `from_preset`.

keras_cv/models/resnet_v2_backbone.py
jbischof: @fchollet had an idea to improve this further by diverging from the KerasNLP API. Instead of a `load_weights` argument we could have some presets come without weights (`"resnetv2_50"`) and other with weights (`"resnetv2_50_imagenet"`). Our documentation / docstrings would have to clarify which was which.

keras_cv/models/__init__.py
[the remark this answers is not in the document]
LukeWood: (Discussed in a meeting: we will have classifier and backbone directories)

keras_cv/callbacks/pycoco_callback_test.py
LukeWood: +1, sounds good

keras_cv/callbacks/pycoco_callback_test.py
LukeWood: Sgtm

fchollet: IMO we should default it to True and not require it anywhere.

keras_cv/models/backbone.py
```diff
@@ -0,0 +1,197 @@
...
+            "model.h5",
+            metadata["weights_url"],
```
fchollet, line 99: This bit of logic should be made optional, conditional on the presence of the `weights_url` entry.

keras_cv/models/backbone.py
```diff
@@ -0,0 +1,197 @@
...
+        if cls.from_preset.__doc__ is None:
+            cls.from_preset.__func__.__doc__ = Backbone.from_preset.__doc__
+            format_docstring(
+                model_name=cls.__name__,
+                example_preset_name=next(iter(cls.presets_with_weights), ""),
+                preset_names='", "'.join(cls.presets),
+            )(cls.from_preset.__func__)
+
+    @property
+    def backbone_level_outputs(self):
```
fchollet, line 130: This is unclear to me

keras_cv/models/backbone.py
```diff
@@ -0,0 +1,197 @@
...
+        """Backbone outputs at each resolution level for transfer learning."""
+        return None
+
+    @backbone_level_outputs.setter
+    def backbone_level_outputs(self, value):
```
fchollet, line 135: Likewise

keras_cv/models/backbone.py
```diff
@@ -0,0 +1,197 @@
...
+        self._backbone_level_outputs = value
+
+    def extract_features(self, min_level=None, max_level=None):
```
fchollet, line 138: To me, intuitively, "extract features" means that the method call returns some features (maybe KerasTensor instances, since this is a Functional API model). But this method actually returns a Model, which in turn extracts the features. This should be reflect in the name, e.g. `get_feature_extractor()`.

keras_cv/models/backbone.py
```diff
@@ -0,0 +1,197 @@
...
+        level. Those outputs can be feed to network downstream, like FPN and RPN.
+
+        The output of the backbone model will be a dict with int as key and
+        tensor as value. The int key represent the level of the feature output.
+        A typical feature pyramid has five levels corresponding to scales P3,
+        P4, P5, P6, P7 in the backbone. Scale Pn represents a feature map 2n
+        times smaller in width and height than the input image.
+
+        Args:
+            min_level: optional int, the lowest level of feature to be included
```
fchollet, line 152: This min_level / max_level API is very mysterious. Doubt anyone not already familiar with the implementation will figure out what it means. Can we find better argument names and descriptions?

keras_cv/models/backbone.py
```diff
@@ -0,0 +1,197 @@
...
+            outputs = {}
+            for level in range(min_level, max_level + 1):
+                outputs[level] = backbone_level_outputs[level]
+
+            return tf.keras.Model(inputs=self.inputs, outputs=outputs)
+
+        else:
+            raise ValueError(
+                "The current model doesn't have any feature level "
+                "information so extraction not possible."
```
fchollet, line 196: A helpful error message should give you pointers to steps you can take to address the problem (the purpose of error messages is to minimize debugging time when you encounter the error)

keras_cv/models/resnet_v2_backbone.py
fchollet: It is usually the case in CV that architectures and their weights are entirely orthogonal. However it's nice that we have an API that is general enough to support both situations -- the orthogonal case and the interdependent case. If we run into architectures that are checkpoint-dependent it's easy to add a preset for them. Should we keep the `load_weights` arg or not?

keras_cv/models/backbone.py
jbischof: @fchollet this is just a renamed `as_backbone` method. I haven't changed anything in the body.

If we want to rewrite this method completely within this PR, I might need a lot of help.

keras_cv/models/resnet_v2_backbone.py
fchollet: We discussed this earlier -- I think we can add nice aliases like `class ResNet50V2Backbone(**kwargs)` that just call `ResNetv2Backbone.from_preset("resnet50_v2", **kwargs)`.

`ResNet50V2Backbone(weights="imagenet", **kwargs)` would call `ResNetv2Backbone.from_preset("resnet50_v2_imagenet", **kwargs)`.

This is a pure alias so there is no added maintenance / complexity or even change in the user-facing mental model.

keras_cv/models/backbone.py
jbischof: This is some logic from KerasNLP that copies the templated docstring here to the subclass. It's not something we expect users to interact with.

keras_cv/models/backbone.py
jbischof: This formats the templated docstring with the subclass-specific args (like preset names)

keras_cv/models/backbone.py
jbischof: Good point! Will file an Issue.

keras_cv/callbacks/pycoco_callback_test.py
jbischof: OK I will set it to `True` in all the presets and we can see how we feel.

keras_cv/models/backbone.py
jbischof: Thanks will add this to the Issue as well!

keras_cv/models/resnet_v2_backbone.py
jbischof: I used the `None` default trick to keep this arg in the signature.

keras_cv/models/resnet_v2_backbone.py
jbischof: This isn't quite true because we also have to keep the `weights` strings in sync with the `presets`.

However, my major reservation is allowing multiple ways to do the same thing often adds more confusion it removes. Why did we create a new library only to be tied to 2016 conventions from `keras.applications`? We have a "backward compatible" library already---the original one!

jbischof: What about `include_top`? Would we be grafting this back on?

jbischof: Also: what design pattern would we be highlighting in the documentation?

keras_cv/models/backbone.py
LukeWood: We can literally add a link to open a github issue perhaps?

keras_cv/models/backbone.py
LukeWood: yeah I really do not like this API.  realistically it should be a list - either of level names or layer names.

keras_cv/models/resnet_v2_backbone.py
LukeWood: > What about include_top? Would we be grafting this back on?

For this I'd say not worth it.  In Docs I think it depends how the model is being used.

keras_cv/models/backbones/backbone.py
```diff
@@ -0,0 +1,208 @@
...
+
+    @classmethod
+    def from_preset(
+        cls,
+        preset,
+        load_weights=None,
+        **kwargs,
+    ):
+        """Instantiate {{model_name}} model from preset architecture and weights.
+        Args:
```
mattdangerw, line 59: empty space before Args: (github eats empty newlines when you copy from there)

keras_cv/utils/python_utils.py
```diff
@@ -0,0 +1,44 @@
...
+class classproperty(property):
+    """Define a class level property."""
+
+    def __get__(self, _, owner_cls):
+        return self.fget(owner_cls)
+
+
+def format_docstring(**replacements):
+    """Format a python docstring using a dictionary of replacements.
+    This decorator can be placed on a function, class or method to format it's
```
mattdangerw, line 26: space before this line and after the next (more missing newlines eaten)

keras_cv/models/backbones/resnet_v2/resnet_v2_backbone_presets.py
```diff
@@ -0,0 +1,132 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""ResNetV2 model preset configurations."""
+
+backbone_presets_no_weights = {
+    "resnet18_v2": {
+        "metadata": {
+            "description": (
+                "ResNet model with 18 layers where the batch normalization "
+                "and ReLU activation precede the convolution layers."
```
mattdangerw, line 21: Is there another option than batch norm and relu precede convolution? Seems weird to comment on this but then all models have this description.

mattdangerw (commented):
Thanks! Mainly took a pass to educate myself, but thoughts on this.

Overall I think this is great at brining the two libraries into a similar style and make it so to an outside observer they look like they are written by the same folks. Which I think is the big win to have here.

I think the biggest thing I noticed was that

```
model = keras_cv.models.RetinaNet(
    classes=10,
    bounding_box_format="xywh",
    backbone=keras_cv.models.ResNetV2Backbone.from_preset(
        "resnet50_v2",
    ).get_feature_extractor(),
)
```

feels a bit different than the approach we took on KerasNLP. In KerasNLP our high-level models do some surgery on our backbones often, but that logic stay in the class that need to extract certain features, rather than the backbone having an "export" function. The more similar move to KerasNLP seems like it would be

```
backbone = keras_cv.models.ResNetV2Backbone.from_preset("resnet50_v2")
model = keras_cv.models.RetinaNet(
    classes=10,
    bounding_box_format="xywh",
    backbone=backbone,
    min_backbone_level=xx,  # None can still be the deafult here.
    max_backbone_level=yy,  # None can still be the deafult here.
)
```

Or even just

```
    "some_id",
    classes=10,
    bounding_box_format="xywh",
)
```

I suspect this is more something to think about (or it already has been), then to change on this PR, but what's the big reasons for the discrepancy here?

keras_cv/models/backbones/resnet_v2/resnet_v2_backbone_presets.py
jbischof: Yes, the v1 model has the opposite ordering. Added clarification.

keras_cv/models/resnet_v2_backbone.py
jbischof: I tried to add a readable, minimalist version of this without a lot of automation. Let me know what you think.

keras_cv/models/resnet_v2_backbone.py
jbischof: See below discussion about reintroducing the keras.applications style aliases.

fchollet:
> I suspect this is more something to think about (or it already has been), then to change on this PR, but what's the big reasons for the discrepancy here?

The approach you describe is cleaner, in fact. I believe the reason for the current discrepancy is that each backbone might require custom handling. If there is no universal way to do feature extraction, then it's more practical for each backbone implementation to specify how to do it for that particular backbone.

But perhaps this is not a correct assumption.

fchollet (commented):
Thanks for the PR! Looking great.

keras_cv/models/backbones/resnet_v2/resnet_v2_backbone.py
```diff
@@ -482,189 +383,150 @@ def get_config(self):
...
+            path, or name of pretrained checkpoint if available.
+            Available checkpoints: None
+        kwargs: Args to be passed to `keras_cv.models.ResNetV2Backbone`
+    """
+
+    if weights and not tf.io.gfile.exists(weights):
+        raise ValueError(
+            "The `weights` argument should be either `None` or the path to the "
+            "weights file to be loaded. Weights file not found at location: "
+            f"{weights}."
```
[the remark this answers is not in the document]
jbischof: There are no presets with weights available for this architecture. See the `ResNet50V2Backbone` for the only model with weights ATM.

keras_cv/models/backbones/resnet_v2/resnet_v2_backbone.py
```diff
@@ -482,189 +383,150 @@ def get_config(self):
...
-        pooling=pooling,
-        classes=classes,
-        classifier_activation=classifier_activation,
-        block_type="block",
-        **kwargs,
-    )
+    return model
+
+
+def ResNet34V2Backbone(weights=None, **kwargs):
```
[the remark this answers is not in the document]
jbischof: @fchollet you recommended the style I used above: https://github.com/keras-team/keras-cv/pull/1438#discussion_r1116356420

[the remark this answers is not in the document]
jbischof: Confirmed offline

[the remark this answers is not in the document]
jbischof: Confirmed offline

keras_cv/models/backbones/backbone.py
```diff
@@ -0,0 +1,209 @@
...
+    @property
+    def backbone_level_outputs(self):
+        """Backbone outputs at each resolution level for transfer learning."""
+        return None
+
+    @backbone_level_outputs.setter
+    def backbone_level_outputs(self, value):
+        self._backbone_level_outputs = value
+
+    def get_feature_extractor(self, min_level=None, max_level=None):
```
LukeWood, line 150: No need to address this in this PR - but I wonder if we should make a FeatureExtractor (or FeaturePyramid) a class instead of a class method:

featureExtractor(ResNet50V2Backbone())

keras_cv/models/backbones/resnet_v2/resnet_v2_backbone.py
```diff
@@ -482,189 +388,240 @@ def get_config(self):
...
-        include_top=include_top,
-        name=name,
-        weights=weights,
-        input_shape=input_shape,
-        input_tensor=input_tensor,
-        pooling=pooling,
-        classes=classes,
-        classifier_activation=classifier_activation,
-        block_type="basic_block",
+    return _backbone_alias(
```
LukeWood, line 501: just a comment:

now that we don't have to specify include_rescaling or include_top it is REALLY cool that you can write

ResNet18V2() and have a model!

LukeWood (approved):
I'm comfortable merging this --- but it might make more sense to wait until the Classifier task model is done.  Up to you --- 

another note: when we do merge this we should ctrl-f ResNet50V2 in keras-io and any other r

Also: we should probably get a universal agreement

LukeWood:
(apologies for any typos in my comments - my GitHub UI is bugged and my comments don't display as I type.  The box simply remains empty - so no way to edit typos out.)

jbischof:
Thanks @LukeWood am working on a cleaner implementation. Will update Monday.

LukeWood (approved):
I think this looks good - I also am glad to see that the `Sequential([ResNet50V2Backbone(), layers.GlobalAveragePooling2D()])` API works as expected (per the Simclr training test)

keras_cv/models/backbones/backbone.py
fchollet: This is a good idea IMO

keras_cv/models/backbones/backbone.py
```diff
@@ -0,0 +1,215 @@
...
+                max_level = model_levels[-1]
+
+            outputs = {}
+            for level in range(min_level, max_level + 1):
+                outputs[level] = backbone_level_outputs[level]
+
+            return tf.keras.Model(inputs=self.inputs, outputs=outputs)
+
+        else:
+            raise ValueError(
```
fchollet, line 212: No need to address in this PR, but this error message is not helpful. It should tell you how to fix the issue.

keras_cv/models/backbones/resnet_v2/resnet_v2_backbone.py
```diff
@@ -320,9 +254,21 @@ def apply_stack(
...
+    Reference:
+        - [Identity Mappings in Deep Residual Networks](https://arxiv.org/abs/1603.05027) (ECCV 2016)
+
+    The difference in Resnet and ResNetV2 rests in the structure of their
+    individual building blocks. In ResNetV2, the batch normalization and
+    ReLU activation precede the convolution layers, as opposed to ResNetV1 where
+    the batch normalization and ReLU activation are applied after the
+    convolution layers.
+
+    For transfer learning use cases, make sure to read the [guide to transfer
```
fchollet, line 269: Break the markdown link after `(` (otherwise it won't render)

keras_cv/models/backbones/resnet_v2/resnet_v2_backbone.py
```diff
@@ -482,189 +364,223 @@ def get_config(self):
...
+                "input_shape": input_shape,
+                "input_tensor": input_tensor,
+            }
+        )
+        return ResNetV2Backbone.from_preset("resnet50_v2", **kwargs)
+
+    @classproperty
+    def presets(cls):
+        """Dictionary of preset names and configurations."""
+        return copy.deepcopy(
```
fchollet, line 490: The copy call should be moved to right before `backbone_presets` (the rest of the dict gets generated on the fly) to avoid redundant work

fchollet (commented):
Thanks for the updates. Looking good, I think we can ship it.

keras_cv/models/backbone.py
jbischof: Filed #1447

keras_cv/models/backbone.py
jbischof: Filed #1447

keras_cv/models/backbones/backbone.py
jbischof: Filed #1447

keras_cv/models/backbones/backbone.py
jbischof: Yes, filed #1447

keras_cv/models/backbones/resnet_v2/resnet_v2_backbone.py
jbischof: Moved to one line

keras_cv/models/resnet_v2_backbone.py
jbischof: Talked offline. Removed the `pooling` arg to reduce conflicts with `Task` model behavior.

jbischof:
Thanks @LukeWood, I also replicated our quickstart using the following model:
```python
resnet_classifier = keras.Sequential(
    [
        keras_cv.models.ResNet18V2Backbone(),
        keras.layers.GlobalAveragePooling2D(name="avg_pool"),
        keras.layers.Dense(3, activation="softmax", name="predictions"),
    ],
)
```

The model accuracy improved from 0.33 -> 0.51 over 10 epochs. See the attached colab for details.

LukeWood:
> Thanks! Mainly took a pass to educate myself, but thoughts on this.
> 
> Overall I think this is great at brining the two libraries into a similar style and make it so to an outside observer they look like they are written by the same folks. Which I think is the big win to have here.
> 
> I think the biggest thing I noticed was that
> 
> ```
> model = keras_cv.models.RetinaNet(
>     classes=10,
>     bounding_box_format="xywh",
>     backbone=keras_cv.models.ResNetV2Backbone.from_preset(
>         "resnet50_v2",
>     ).get_feature_extractor(),
> )
> ```
> 
> feels a bit different than the approach we took on KerasNLP. In KerasNLP our high-level models do some surgery on our backbones often, but that logic stay in the class that need to extract certain features, rather than the backbone having an "export" function. The more similar move to KerasNLP seems like it would be
> 
> ```
> backbone = keras_cv.models.ResNetV2Backbone.from_preset("resnet50_v2")
> model = keras_cv.models.RetinaNet(
>     classes=10,
>     bounding_box_format="xywh",
>     backbone=backbone,
>     min_backbone_level=xx,  # None can still be the deafult here.
>     max_backbone_level=yy,  # None can still be the deafult here.
> )
> ```
> 
> Or even just
> 
> ```
> model = keras_cv.models.RetinaNet.from_preset(
>     "some_id",
>     classes=10,
>     bounding_box_format="xywh",
> )
> ```
> 
> I suspect this is more something to think about (or it already has been), then to change on this PR, but what's the big reasons for the discrepancy here?

I do quite like the idea of having the `get_feature_extractor()` logic live in the task models -- but I also think that can be a follow up PR.

thoughts?

IMvision12:
@jbischof Are these modifications required for resentv1?

jbischof:
Yes @IMvision12 but not quite yet! I'm planning some followup PRs to refine the design somewhat and will start filing Issues later this week. Thank you for all your amazing contributions 🚀

PR #1459, opened 2023-03-03
Add exclude_class param to some 3D preprocessing layers. credit: lengzq

ianstenbit:
This change has already been made downstream and I'm syncing it here (since we don't have piper -> copybara set up).

### DIFF
--- a/keras_cv/layers/preprocessing_3d/frustum_random_dropping_points.py
+++ b/keras_cv/layers/preprocessing_3d/frustum_random_dropping_points.py
@@ -50,7 +50,7 @@ class FrustumRandomDroppingPoints(
       theta_width: A float scalar sets the theta width of a frustum.
       phi_width: A float scalar sets the phi width of a frustum.
       drop_rate: A float scalar sets the probability threshold for dropping the points.
-      exclude_class: An optional int scalar or a list of ints. Points with the specified class(es) will not be dropped.
+      exclude_classes: An optional int scalar or a list of ints. Points with the specified class(es) will not be dropped.
 
     """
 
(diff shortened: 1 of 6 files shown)


keras_cv/layers/preprocessing_3d/frustum_random_dropping_points.py
```diff
@@ -50,10 +50,18 @@ class FrustumRandomDroppingPoints(
...
     """
 
     def __init__(
-        self, r_distance, theta_width, phi_width, drop_rate=None, **kwargs
+        self,
+        r_distance,
+        theta_width,
+        phi_width,
+        drop_rate=None,
+        exclude_class=-1,
```
ianstenbit, line 63: This should probably use `None` instead of -1.

ianstenbit: Updated -- this now defaults to None.

keras_cv/layers/preprocessing_3d/frustum_random_dropping_points.py
```diff
@@ -50,10 +50,18 @@ class FrustumRandomDroppingPoints(
       theta_width: A float scalar sets the theta width of a frustum.
       phi_width: A float scalar sets the phi width of a frustum.
       drop_rate: A float scalar sets the probability threshold for dropping the points.
+      exclude_class: A int scalar. When the last feature of a point == exclude_class, the point is not dropped.
```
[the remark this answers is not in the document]
ianstenbit: Good call -- updated to support a list (or `None`) throughout.

keras_cv/layers/preprocessing_3d/frustum_random_dropping_points.py
```diff
@@ -60,11 +60,14 @@ def __init__(
         theta_width,
         phi_width,
         drop_rate=None,
-        exclude_class=-1,
```
LukeWood: nit: classes

LukeWood (approved):
Approvod with a nit, happy to delegate on this

soma2000-lang:
@ianstenbit @LukeWood just want to educate myself on this part ,why we are doing this?

ianstenbit:
> @ianstenbit @LukeWood just want to educate myself on this part ,why we are excluding the class params to some 3d processing layers?

The purpose here is to allow specifying classes that shouldn't be affected by these layers. There will be an upcoming blog post in the next few months about the usage of these layers which should help demonstrate the use case!

PR #1465, opened 2023-03-07
Remove lambda layer from ViT model

ianstenbit:
Serialization/deserialization of lambdas is not considered safe, so we should avoid using one here.

I've verified that weight loading still works as expected.


jbischof (approved):
Wow, that is waaaay cleaner! LGTM

PR #1477, opened 2023-03-09
EfficientNet V1 subclassing 

soma2000-lang:
https://colab.research.google.com/drive/1JyeD1Fp3Dcb3C8JstcPGv-s2w3VsMyxy?usp=sharing
#1424

@LukeWood @ianstenbit


ianstenbit:
Hey @soma2000-lang it looks like @srikesh-07 opened up #1482 as well, which does the same thing. Maybe you two can collaborate on that PR?

soma2000-lang:
@ianstenbit Does the colab looks good.

ianstenbit:
@soma2000-lang I don't have access to your Colab. Fine by me if you'd prefer to pick up a different model. Let's just close one of these two PRs since they do the same thing. You two can choose how to handle that

soma2000-lang:
@ianstenbit Sorry I have given the access to the colab now

ianstenbit:
> @ianstenbit Sorry I have given the access to the colab now

Colab looks good. Both here and in #1482, we should make the Block a function for now instead of a layer.

Maybe you can commit that on the branch for #1482 so you and  @srikesh-07 will have co-authorship.

Issue #1420, opened 2023-02-21
Make CSPDarknet a Functional subclass model and remove closures


ianstenbit:
This should follow the example of #1412 and #1401, including

- Docstring updates
- Colab demonstrating that weight loading works (if pre-trained weights exist) and numerics are identical

soma2000-lang:
working on this.

soma2000-lang:
@YimjuneLiu I am already  working on this,will open a pr soon .Can you do some other task?

YimjuneLiu:
Sure.

PR #1478, opened 2023-03-09
CSPDARKNET Subclassing 

soma2000-lang:
[colab](https://colab.research.google.com/drive/12XKas34I6eBmAMBjckFzwLDtQubnxskG?usp=sharing)
#1420
@ianstenbit

### DIFF
--- a/keras_cv/models/csp_darknet.py
+++ b/keras_cv/models/csp_darknet.py
@@ -23,6 +23,7 @@ import types
 
 import tensorflow as tf
 from tensorflow import keras
+from tensorflow.keras import backend
 from tensorflow.keras import layers
 
 from keras_cv.models import utils
@@ -36,10 +37,9 @@ from keras_cv.models.__internal__.darknet_utils import (
 from keras_cv.models.weights import parse_weights
 
 
-@keras.utils.register_keras_serializable(package="keras_cv.models")
-class CspDarknet(keras.Model):
-
-    """Instantiates the CSPDarkNet architecture.
+@keras.utils.regester_keras_serializable(package="keras_cv.models")
+class CSPDarkNet(keras.Model):
+    """This class Instantiates the CSPDarkNet architecture.
 
     Although the DarkNet architecture is commonly used for detection tasks, it is
     possible to extract the intermediate dark2 to dark5 layers from the model for
@@ -58,13 +58,13 @@ class CspDarknet(keras.Model):
             this changes based the detection model being used.
         width_multiplier: A float value used to calculate the base width of the model
             this changes based the detection model being used.
-        include_rescaling: whether or not to Rescale the inputs.If set to True,
+        include_rescaling: bool ,whether or not to Rescale the inputs.If set to True,
             inputs will be passed through a `Rescaling(1/255.0)` layer.
-        include_top: whether to include the fully-connected layer at the top of
+        include_top: bool, whether to include the fully-connected layer at the top of
             the network.  If provided, `num_classes` must be provided.
         use_depthwise: a boolean value used to decide whether a depthwise conv block
             should be used over a regular darknet block. Defaults to False
-        num_classes: optional number of classes to classify images into, only to be
+        num_classes: optional int,optional number of classes to classify images into, only to be
             specified if `include_top` is True.
         weights: one of `None` (random initialization), a pretrained weight file
             path, or a reference to pre-trained weights (e.g. 'imagenet/classification')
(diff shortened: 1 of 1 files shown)


keras_cv/models/csp_darknet.py
```diff
@@ -36,22 +36,9 @@
...
-    input_tensor=None,
-    pooling=None,
-    classifier_activation="softmax",
-    name="CSPDarkNet",
-    **kwargs,
-):
-    """Instantiates the CSPDarkNet architecture.
+@keras.utils.regester_keras_serializable(package="keras_cv.models")
+class CSPDarkNet(keras.Model):
+    """This class Instantiates the CSPDarkNet architecture.
```
ianstenbit, line 41: nit: s/Instantiates/represents

keras_cv/models/csp_darknet.py
```diff
@@ -266,8 +276,8 @@ def CSPDarkNet(
 
 def CSPDarkNetTiny(
     *,
-    include_rescaling,
-    include_top,
+    include_rescaling=None,
```
ianstenbit, line 279: let's avoid including defaults here

ianstenbit (changes requested):
Thanks for the PR!

Let's also remove the closures from `darknet_utils`, and use the `apply_xyz` naming technique (this will also need to be done in #1486)

ID6109:
Hey @soma2000-lang , this is in reference to #1486 . I have already started working on `darknet_utils.py`, so how about I get back to you once I am through with the changes that can be incorporated into your PR? Thanks.

soma2000-lang:
Ok @ID6109 seems good

ID6109:
@soma2000-lang I have made the necessary changes in `darknet_utils.py` but the tests are failing due to function name mismatches with `csp_darknet.py`. How about I go ahead with making the naming changes there as well? This might cause merge conflicts with your PR.

soma2000-lang:
@ID6109 let me make the name changes.
So that it does not cause the merge conflicts.Please tell me what name changes do you want to do

soma2000-lang:
@ID6109 I will do that

ianstenbit:
> Let's also remove the closures from `darknet_utils`, and use the `apply_xyz` naming technique (this will also need to be done in #1486)

Upon further inspection, let's _not_ update the layers in `darknet_utils` and instead leave them as-is.

So this PR is very close -- just the few small comments I left.

Additionally, let's test pre-trained weights in the Colab. `"imagenet"` weights are available for `CSPDarkNetL`

keras_cv/models/csp_darknet.py
```diff
@@ -267,7 +277,7 @@ def CSPDarkNet(
 def CSPDarkNetTiny(
     *,
     include_rescaling,
-    include_top,
+    include_top=None,
```
ianstenbit, line 280: nit: please remove this default value of `include_top`

ianstenbit (approved):
Looks good -- just one more small fix and then we can merge :)

Your Colab no longer runs, but I was able to adapt it to make sure weight loading still works:
https://colab.research.google.com/gist/ianstenbit/b2ef30c13329624e45200008ce4acb46/cspdarknetm.ipynb

With respect to KerasIO guides -- right now, we're still a bit short on model offerings and we're still in the process of updating our API to unify with KerasNLP, so we're holding off on guides for`KerasCV` models. However, if there are models that we don't include in KerasCV (yet!), feel free to create guides on those, and we can use that as a starting point for introducing those models here.

ianstenbit:
Also @soma2000-lang please take a look at the lint and test errors

keras_cv/models/csp_darknet.py
```diff
@@ -36,22 +36,9 @@
...
-    weights=None,
-    input_shape=(None, None, 3),
-    input_tensor=None,
-    pooling=None,
-    classifier_activation="softmax",
-    name="CSPDarkNet",
-    **kwargs,
-):
-    """Instantiates the CSPDarkNet architecture.
+@keras.utils.regester_keras_serializable(package="keras_cv.models")
```
ianstenbit, line 39: nit: s/regester/register

keras_cv/models/csp_darknet.py
```diff
@@ -257,12 +136,134 @@ def CSPDarkNet(
...
+            x = layers.GlobalMaxPooling2D(name="max_pool")(x)
+
+        model = keras.Model(inputs, x, name=name, **kwargs)
+        model._backbone_level_outputs = _backbone_level_outputs
+        # Bind the `to_backbone_model` method to the application model.
+        model.as_backbone = types.MethodType(utils.as_backbone, model)
+
+        if weights is not None:
+            model.load_weights(weights)
+        return model
```
ianstenbit, line 265: Sorry I missed this in the first review.

We shouldn't be `return`ing anything, as this is now a constructor.

We do need to set model-level properties, though, and add a `get_config` and `from_config`

See e.g. https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/regnet.py#L791 for an example of this

ianstenbit (changes requested):

PR #1482, opened 2023-03-10
EfficientNetV1 as a Functional subclass model & remove closures

### LINKED ISSUE #1424: Make EfficientNetV1 a Functional subclass model & remove closures
This should follow the example of #1412 and #1401, including

- Docstring updates
- Colab demonstrating that weight loading works (if pre-trained weights exist) and numerics are identical

srikesh-07:
# What does this PR do?

Fixes #1424  

This [Colab](https://colab.research.google.com/drive/1lkxDaixqDuZtyiisK6fSV3wLrPQp9zzn?usp=sharing) helps to reproduce the same results from existing and new EfficientNet-V1. 

In `EfficientNetBlock`, `__call__()` has been used instead of `call()` in order to reproduce the old results and verify the new implementation. Once after the approval `__call__` could be replaced with `call()` to  make the architecture modular.

Kindly please check the PR and let me know if any changes to be made.

Thank You

## Who can review?
@ianstenbit  @LukeWood @tanzhenyu

### DIFF
--- a/keras_cv/models/efficientnet_v1.py
+++ b/keras_cv/models/efficientnet_v1.py
@@ -401,7 +401,7 @@ class EfficientNetBlock(layers.Layer):
         return x
 
     def get_config(self):
-        return {
+        config = {
             'filters_in': self.filters_in,
             'filters_out': self.filters_out,
             'kernel_size': self.kernel_size,
@@ -414,6 +414,9 @@ class EfficientNetBlock(layers.Layer):
             'name': self.name,
             'trainable': self.trainable
         }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
+
 
     @classmethod
     def from_config(cls, config):
--- a/keras_cv/models/efficientnet_v1.py
+++ b/keras_cv/models/efficientnet_v1.py
@@ -419,7 +419,7 @@ class EfficientNetBlock(layers.Layer):
     def from_config(cls, config):
         return cls(**config)
 
-
+@keras.utils.register_keras_serializable(package="keras_cv.models")
 class EfficientNet(keras.Model):
     """This class represents a Keras VGG16 model.
     Args:
--- a/keras_cv/models/efficientnet_v1.py
+++ b/keras_cv/models/efficientnet_v1.py
@@ -508,9 +508,6 @@ class EfficientNet(keras.Model):
                 f"Received pooling={pooling} and include_top={include_top}. "
             )
 
-        if input_shape[0] is None or input_shape[1] is None:
-            input_shape = (default_size, default_size, 3)
-
         img_input = utils.parse_model_inputs(input_shape, input_tensor)
 
         # Build stem
--- a/keras_cv/models/efficientnet_v1.py
+++ b/keras_cv/models/efficientnet_v1.py
@@ -618,7 +618,6 @@ class EfficientNet(keras.Model):
         self.depth_divisor = depth_divisor
         self.activation = activation
         self.block_args = blocks_args
-        self.weights = weights
         self.input_tensor = input_tensor
         self.pooling = pooling
         self.num_classes = num_classes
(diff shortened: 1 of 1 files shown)


keras_cv/models/efficientnet_v1.py
```diff
@@ -26,7 +26,8 @@
 
 import tensorflow as tf
 from keras import backend
```
ianstenbit, line 28: For consistency, let's also import `keras.backend` using `from tensorflow.keras import backend`

keras_cv/models/efficientnet_v1.py
```diff
@@ -201,151 +201,252 @@ def correct_pad(inputs, kernel_size):
...
-    name="",
-    filters_in=32,
-    filters_out=16,
-    kernel_size=3,
-    strides=1,
-    expand_ratio=1,
-    se_ratio=0.0,
-    id_skip=True,
-):
+class EfficientNetBlock(layers.Layer):
```
ianstenbit, line 204: What's the reasoning behind making this a `Layer` instead of a function like `apply_efficientnet_block`, like we have for other models?

If it's practical to do so, we should keep this consistent with other models by using the `apply_xyz_block` function approach.

ianstenbit (changes requested):
Thanks for the PR!

keras_cv/models/efficientnet_v1.py
srikesh-07: The main purpose to use of `layer`  is to develop a architecture as a separate blocks as `EfficientNetV2` has been developed. 

In `EfficientNetV2`, the model is constructed as a different blocks connected together (`FusedMBConvBlock`, `MBConvBlock`). So I have thought `layer` implementation of `efficientnet_block` would make the model modular.  

Can you please have a suggestion on this ? If the implementations wants to be in functional type, I am ready to make changes.

ianstenbit: Gotcha -- thanks! 
I think for now let's make this a function, not a layer.

If we find that this is re-usable for other models, we can move it to `keras_cv/layers/` and use it from here

srikesh-07: Thank you for direction.

Successfully made the changes.

keras_cv/models/efficientnet_v1.py
srikesh-07: Fixed

srikesh-07 (commented):
I have done the changes as suggested previously. Kindly please check the PR and kindly let me know if any changes to be done.

Thank you!

keras_cv/models/efficientnet_v1.py
```diff
@@ -201,151 +201,217 @@ def correct_pad(inputs, kernel_size):
     )
 
 
-def EfficientNetBlock(
+def conv_bn(
```
ianstenbit, line 204: nit: apply_conv_bn

ianstenbit (changes requested):
Thank you @srikesh-07!

Very close -- just two very small fixes.

keras_cv/models/efficientnet_v1.py
```diff
@@ -423,92 +627,34 @@ def round_filters(filters, divisor=depth_divisor):
             new_filters += divisor
         return int(new_filters)
 
-    def round_repeats(repeats):
+    def round_repeats(self, repeats, depth_coefficient):
```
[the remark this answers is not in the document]
srikesh-07: Fixed.

keras_cv/models/efficientnet_v1.py
srikesh-07: Fixed.

srikesh-07 (commented):
I have done the final changes. Kindly please check the PR and kindly let me know if any further changes to be made.

Thank you!

ianstenbit (approved):
Thank you for the PR!

PR #1485, opened 2023-03-10
Reorganize RetinaNet model structure

LukeWood:
Removing the first __internal__ namespace in preparation for the generated API.

We are promptly moving towards a generated API - and these __internal__ structures are very user-hostile.  

At this time we don't have backwards compat guarantees, so in my opinion this is a move in the right direction.  We can just exclude these from the API when using the API generator @fchollet wrote.


LukeWood:
Looks like this breaks some tests/namespacing, will take a look

ianstenbit (approved):
This LGTM once merge conflicts are resolved. Thanks Luke!

jbischof (approved):
Love the approachability here, thanks @LukeWood ! Machine learning for humans indeed

LukeWood:
I realized I never merged this, I'll rebase tomorrow and redo this change on the updated RetinaNet.

LukeWood:
Doing in a new PR

PR #1488, opened 2023-03-11
Regnet as a Functional subclass model & removal of closures

### LINKED ISSUE #1428: Make RegNet a Functional subclass model & remove closures
This should follow the example of #1412 and #1401, including

- Docstring updates
- Colab demonstrating that weight loading works (if pre-trained weights exist) and numerics are identical

srikesh-07:
# What does this PR do?

Fixes #1428 

This [Colab](https://colab.research.google.com/drive/1jHcTW2JupMdnUH6xCEyAHrXdr78aYOa8?usp=sharing) helps to reproduce the same results from existing and new Regnet.

1. During the development, I have found a bug that `SqueezeAndExcite2D` block hasn't been included in `ZBlock` which could be seen at [Line-559](https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/regnet.py#L559) in `regnet.py`.  I haven't fixed the error to verify the new implementation with the old implementation. On approval the bug could be fixed.

2. Currently the three different blocks `XBlock`, `YBlock`, `ZBlock` are defined as three different functions. So,
    Do we have to define a single function where these three blocks could be created ?
                                                       (or)
    Does the current implementation seems to be good ?

Kindly please check the PR and let me know  the suggestions on above points and also if there any changes that is to be made on PR.

Thank You !

## Who can review?
@ianstenbit  @LukeWood @tanzhenyu


keras_cv/models/regnet.py
```diff
@@ -255,11 +256,47 @@
 """
 
 
-def Stem(name=None):
+def conv2d_bn(
```
ianstenbit, line 259: nit: apply_conv2d_bn

keras_cv/models/regnet.py
```diff
@@ -268,119 +305,97 @@ def Stem(name=None):
...
+        padding="same",
+        name=name + "_stem_conv",
+    )
 
-    return apply
+    return x
 
 
-def XBlock(filters_in, filters_out, group_width, stride=1, name=None):
+def apply_XBlock(
```
ianstenbit, line 320: how about `apply_x_block` here? (same for y and z)

srikesh-07:
> 

Thank you for the suggestion, I have created a separate issue for bug in `Z block`. Once it is approved and merged, then I can make the final changes as you commented above.

PR #1487, opened 2023-03-11
custom_ops: fix formatting

justinledford:
Fix formatting of some cc files to follow
https://www.tensorflow.org/community/contribute/code_style#c_coding_style

## Who can review?
@ianstenbit  @LukeWood

PR #1490, opened 2023-03-12
add custom op to compute "is within any box"

justinledford:
# What does this PR do?
add custom op to compute "is within any box"

This op directly checks if each point belongs to any box, and shows a
~6x speedup over using within_box to determine the same (which computes
the entire matrix of whether each point is within each box).

Ran a benchmark script using timeit to compare performance against existing implementations with
200K random points and 1K random boxes:

```
----------------------------------------
benchmark_within_any_box3d
1 loop, best of 5: 763 msec per loop
----------------------------------------
benchmark_within_any_box3d_v2
1 loop, best of 5: 442 msec per loop
----------------------------------------
benchmark_within_any_box3d_v3
5 loops, best of 5: 69.5 msec per loop
```

## Who can review?

@ianstenbit @LukeWood


keras_cv/custom_ops/ops/within_any_box_op.cc
```diff
@@ -0,0 +1,28 @@
+/* Copyright 2022 The KerasCV Authors. All Rights Reserved.
```
ianstenbit, line 1: nit: 2023

keras_cv/custom_ops/kernels/within_any_box_op.cc
```diff
@@ -0,0 +1,71 @@
+/* Copyright 2022 The KerasCV Authors. All Rights Reserved.
```
ianstenbit, line 1: nit: 2023

keras_cv/custom_ops/kernels/within_any_box_op.cc
```diff
@@ -0,0 +1,71 @@
...
+
+    std::vector<box::Upright3DBox> boxes_vec = box::ParseBoxesFromTensor(boxes);
+    std::vector<box::Vertex> points_vec = box::ParseVerticesFromTensor(points);
+
+    auto within_fn = [&boxes_vec, &points_vec, &within_any_box_t](int64_t begin,
+                                                                  int64_t end) {
+      for (int64_t idx = begin; idx < end; ++idx) {
+        box::Upright3DBox& box = boxes_vec[idx];
+        for (uint64_t p_idx = 0; p_idx < points_vec.size(); ++p_idx) {
+          auto point = points_vec[p_idx];
```
ianstenbit, line 54: Might be worth `continue`ing from this iteration if `within_any_box_t(p_idx) == true`

(Since that op is probably much faster than `box.WithinBox3D`)

I'm not sure how this would behave in parallel, though, since it would require synchronized reads.

keras_cv/custom_ops/box_util.cc
```diff
@@ -269,7 +269,8 @@ double RotatedBox2D::IoU(const RotatedBox2D& other) const {
   return intersection_area / union_area;
 }
 
-bool RotatedBox2D::left_hand_side(const Vertex& point, const Vertex& v1, const Vertex& v2) const {
+bool RotatedBox2D::left_hand_side(const Vertex& point, const Vertex& v1,
```
ianstenbit, line 272: I assume these format changes will be gone from the diff after merging past #1487 -- right?

ianstenbit (commented):
Thank you so much!

Could you add your benchmark script under a new folder `keras-cv/benchmarks/custom_ops`?

keras_cv/custom_ops/kernels/within_any_box_op.cc
justinledford: good idea, this cut back ~4 ms. this should be fine in parallel since `within_any_box_t(p_idx)` will never go back to false once we set it to true, so it should never return any incorrect results

justinledford:
> Could you add your benchmark script under a new folder keras-cv/benchmarks/custom_ops?
done

ianstenbit:
> > Could you add your benchmark script under a new folder keras-cv/benchmarks/custom_ops?
> 
> done

Sorry for the delay on my review -- I don't see the benchmark script. Otherwise this LGTM. Thank you @justinledford!

benchmarks/custom_ops/within_any_box.sh
```diff
@@ -0,0 +1,65 @@
...
+    boxes_y = tf.random.uniform(
+        shape=[num_boxes, 1],
+        minval=0,
+        maxval=box_dimension - 1.0,
+        dtype=tf.float32,
+    )
+    boxes_dx = tf.random.uniform(
+        shape=[num_boxes, 1], minval=0, maxval=5.0, dtype=tf.float32
+    )
+    boxes_dx = tf.math.minimum(10 - boxes_x, boxes_dx)
```
ianstenbit, line 35: shouldn't this be `box_dimension - boxes_x`?

ianstenbit (approved):
LGTM -- thank you!

Just two quick questions. (Want to make sure our benchmark is solid for whenever we try to make a faster op)

ianstenbit (approved):
Thank you Justin!

If you'd like, we can also switch all callsites of `is_within_any_box3d` to your `v3` version.
(In another PR is probably best)

LukeWood:
🔥 🚒 🧯 🧨 🧑‍🚒  @justinledford

PR #1498, opened 2023-03-13
Switch to dictionary format for inputs and targets of CenterPillar model

ianstenbit:
This is a better UX, and is more consistent with the 2D object detection API, where we use dictionaries instead of tuples of Tensors.

### DIFF
--- a/keras_cv/models/object_detection_3d/center_pillar.py
+++ b/keras_cv/models/object_detection_3d/center_pillar.py
@@ -158,7 +158,12 @@ class MultiHeadCenterPillar(tf.keras.Model):
         self._prediction_decoder = prediction_decoder
         self._head_names = self._multiclass_head._head_names
 
-    def call(self, point_xyz, point_feature, point_mask, training=None):
+    def call(self, input_dict, training=None):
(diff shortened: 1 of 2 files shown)


LukeWood (approved):

keras_cv/models/object_detection_3d/center_pillar.py
```diff
@@ -158,7 +158,12 @@ def __init__(
         self._prediction_decoder = prediction_decoder
         self._head_names = self._multiclass_head._head_names
 
-    def call(self, point_xyz, point_feature, point_mask, training=None):
+    def call(self, input_dict, training=None):
+        point_xyz, point_feature, point_mask = (
+            input_dict["point_xyz"],
+            input_dict["point_feature"],
```
LukeWood, line 164: feel free to target in a follow up --- but should we revisit any of these dictionary keys?  I lack context on the problem domain unfortunately so no insight from me here.

ianstenbit: Yes, more API updates to come for these models + layers to clean this up. For now, I just wanted to get standardized on a dictionary.

PR #1505, opened 2023-03-14
Use SUM_OVER_BATCH_SIZE as default OD loss reduction

ianstenbit:
Currently, this results in unusually small losses.

I'm training on 2 GPUs with reduce `SUM`, I'm seeing a first epoch loss of ~3.

The initial loss I'm seeing with reduce `SUM_OVER_BATCH_SIZE` is `0.0000052`, which seems ... incorrect.

@LukeWood any ideas?

### DIFF
--- a/keras_cv/models/object_detection/faster_rcnn.py
+++ b/keras_cv/models/object_detection/faster_rcnn.py
@@ -431,7 +431,8 @@ class FasterRCNN(tf.keras.Model):
         rpn_box_loss = _validate_and_get_loss(rpn_box_loss, "rpn_box_loss")
         if rpn_classification_loss == "BinaryCrossentropy":
             rpn_classification_loss = tf.keras.losses.BinaryCrossentropy(
-                from_logits=True, reduction=tf.keras.losses.Reduction.SUM
+                from_logits=True,
+                reduction=tf.keras.losses.Reduction.SUM_OVER_BATCH_SIZE,
             )
         rpn_classification_loss = _validate_and_get_loss(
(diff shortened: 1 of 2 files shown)


keras_cv/models/object_detection/faster_rcnn.py
```diff
@@ -431,7 +431,8 @@ def compile(
         rpn_box_loss = _validate_and_get_loss(rpn_box_loss, "rpn_box_loss")
         if rpn_classification_loss == "BinaryCrossentropy":
             rpn_classification_loss = tf.keras.losses.BinaryCrossentropy(
-                from_logits=True, reduction=tf.keras.losses.Reduction.SUM
+                from_logits=True,
+                reduction=tf.keras.losses.Reduction.SUM_OVER_BATCH_SIZE,
```
LukeWood, line 435: This actually changes the scaling needed for learning rate to rely on batch size.  This is a good reason to use sum instead of `SUM_OVER_BATCH_SIZE`.  

If you use SUM and you use learning rate 0.01, it doesn't matter if batch size is 256 or 8 or 1, you'll move by 0.01*sum(dL_i/dW) for all i in the training set.  The same is not the case for sum_over_batch size, which actually reduces the loss scale and therefore requires adjustment of learning rate.

LukeWood:
> 0.0000052

In any case, good change or bad, we should wait until I have integration testing.

ianstenbit:
> > 0.0000052
> 
> In any case, good change or bad, we should wait until I have integration testing.

sg -- I'm going to close this for now then. Was mostly just for conversation.

PR #1509, opened 2023-03-15
resolve inconsistency between lint.sh and setup.cfg

haifeng-jin:
`setup.cfg` says 80 chars per line, but `lint.sh` says 200 chars per line. This PR resolves this inconsistency.

The inconsistency between `lint.sh` and `setup.cfg` is causing the IDEs to print false lint warnings, because IDEs use `setup.cfg` only.

Now 80 char line length is only ensured by `black`.
There are a lot of over length strings, docstrings, and comments.

Ideally, we should still ensure 80 char line length and use `# noqa: E501` after the docstrings like the following.
```py
class SomeClass:
    """My docstring.

    This is a line with over length.
    """  # noqa: E501
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,8 +12,8 @@ filterwarnings =
     ignore::PendingDeprecationWarning
     ignore::FutureWarning
 [flake8]
-max-line-length = 80
-max-doc-length = 200
+# Allow --max-line-length=200 to support long links in docstrings
+max-line-length = 200
 per-file-ignores =
     ./keras_cv/__init__.py:E402, F401
     ./examples/**/*:E402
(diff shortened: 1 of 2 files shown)


LukeWood:
Thanks for the PR!  The reason we have this as of now is for links, i.e.

```
class MyLayer:
    """
    References:
         - https://arxiv.org/my/super/long/link/that/is/10233243232/characters/long/and/definitely/over/80
    """
```

Open to alternatives, but thats the history of this configuration!

haifeng-jin:
@LukeWood Thanks for the context!
To be clear, this PR does not change this configuration, but only move the configuration into `setup.cfg` to make it universal.

We can change it later when we have the resource to do it.

ianstenbit (approved):
LGTM, thanks Haifeng! 

I've opened #1532 for the long-term solution of enforcing 80 universally and exempting docstrings using `noqa`

PR #1511, opened 2023-03-15
Improve the devcontainer settings

haifeng-jin:
This PR improves the devcontainer setup.
* Use the image specially prepared for devcontainers and install tensorflow on post commands.
* Extracted `isort` and `black` settings to `pyproject.toml` to make consistent behavior no matter how the commands are run. (Note that `pyproject.toml` is just there for the formatting setup not for package releases).
* Enable the GitHub CLI `gh` inside devcontainers.
* Fix the bug that `settings` in `devcontainer.json` did not work.
* Add a verticle ruler in the editor for line length limit.

### DIFF
--- a/shell/lint.sh
+++ b/shell/lint.sh
@@ -31,7 +31,7 @@ then
   exit 1
 fi
 [ $# -eq 0 ] && echo "no issues with flake8"
-black --line-length 80 --check $files
+black --check $files
 if ! [ $? -eq 0 ]
 then
   echo "Please run \"sh shell/format.sh\" to format the code."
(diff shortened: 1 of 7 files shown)


shell/lint.sh
```diff
@@ -31,7 +31,7 @@ then
   exit 1
 fi
 [ $# -eq 0 ] && echo "no issues with flake8"
-black --line-length 80 --check $files
```
ianstenbit, line 34: Just to verify -- if using `lint.sh` outside of devcontainer, this setting will still be taken from `pyproject.toml` -- correct?

haifeng-jin: Yes, that's correct.

haifeng-jin:
> This all looks reasonable to me, but I have to admit that I don't use the Devcontainer or VSCode -- perhaps I should be using them?

You can try. I use chromebook so not many IDEs are available. I switched from vim and VS code has all the features I need.

ianstenbit (approved):
Thanks Haifeng!

PR #1515, opened 2023-03-15
Create a Recall metric computation benchmark with 3 versions of recall

LukeWood:
Begins implementing a benchmark to allow us to gauge the performance of various Recall implementations.  The goal of this effort is to ship a performant, easy to use box recall metric.  As of now, a py coco tools based implementation gets the top speeds.  We should strive to be as fast as this implementation.

@ianstenbit 

Figure_1

---

Plan of action:

- [ ] try to reduce retracing in the in-graph version
- [ ] try to jit-compile subgraphs of the in-graph version
- [ ] use tensorflow profiler to identify critical sub-graphs that are costing lots of compute time
- [ ] move forward with PyMetric recall.

---

Further plans:

This is a possible area for community contribution.  I'd like to document the path forward, then open this up if possible.


LukeWood:
Removed in favor of the final PyCOCOTools OD metrics

PR #1534, opened 2023-03-17
Fixes RandomShearDemo

LukeWood:
FYI @sebastian-sz - it looks like the demo utils are just not robust to arbitrary format w/ ragged images.  The conversion function was not accounting for the fact that ragged images have a different width/height than the originals, so calling `images.to_tensor()` will actually modify the locations of boxes within a batch.  After this change boxes are more in line with what you'd expect, but they still aren't perfect.

Before this change:
225731150-05270c4c-7d77-4e9d-86a2-a798207d72bf

After this change:

Still actually not perfect and there is a rounding error/minor bug somewhere, but an improvement.

### DIFF
--- a/examples/layers/preprocessing/bounding_box/demo_utils.py
+++ b/examples/layers/preprocessing/bounding_box/demo_utils.py
@@ -64,14 +64,14 @@ def visualize_data(data, bounding_box_format):
 def visualize_bounding_boxes(image, bounding_boxes, bounding_box_format):
     color = np.array([[255.0, 0.0, 0.0]])
     bounding_boxes = bounding_box.to_dense(bounding_boxes)
-    if isinstance(image, tf.RaggedTensor):
-        image = image.to_tensor(0)
     bounding_boxes = bounding_box.convert_format(
         bounding_boxes,
         source=bounding_box_format,
         target="rel_yxyx",
         images=image,
     )
+    if isinstance(image, tf.RaggedTensor):
+        image = image.to_tensor(0)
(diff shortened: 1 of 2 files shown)


sebastian-sz:
@LukeWood yes that looks better, thanks!

examples/layers/preprocessing/bounding_box/demo_utils.py
```diff
@@ -64,11 +64,17 @@ def visualize_data(data, bounding_box_format):
 def visualize_bounding_boxes(image, bounding_boxes, bounding_box_format):
     color = np.array([[255.0, 0.0, 0.0]])
     bounding_boxes = bounding_box.to_dense(bounding_boxes)
+    bounding_boxes = bounding_box.convert_format(
```
ianstenbit, line 67: Why convert to yxyx at all? Can't we just convert to rel_yxyx directly?

LukeWood: what you just commented is turning out to be one of the most difficult issues to catch in KerasCV code bases.

calling to_dense() actually changes image sizes, so if you are in `rel_xywh` this causes nonsense boxes.

Wonder if we can catch this somehow

ianstenbit: Ack -- please open an Issue about this

ianstenbit (approved):

LukeWood:
Failure is unrelated

PR #1536, opened 2023-03-17
Added autoformatting clang script

### LINKED ISSUE #1496: Add auto-formatting for CPP code
Perhaps https://github.com/marketplace/actions/clang-format-lint is a good approach, per @justinledford in #1487

cosmo3769:
# What does this PR do?

Fixes #1496 

Added auto-formatting for cpp code. I used this [clang-format Wrapper Script Examples](https://embeddedartistry.com/blog/2017/10/30/clang-format-wrapper-script-examples/) and also [clang-format-lint](https://github.com/marketplace/actions/clang-format-lint) as suggested by @ianstenbit in #1496. 

With this we can auto format .h, .c, .cpp, .hpp, .cc file extensions. We also can verify it at CI run as the code added in actions.yml file helps to do so. 

@LukeWood @ianstenbit Please check. Thank you.


shell/lint.sh
```diff
@@ -13,6 +13,7 @@ if [ $# -ne 0  ]
     files=$@
 fi
 
+#verify isort
```
ianstenbit, line 16: nit: I think we can get rid of these comments as the code is pretty self-descriptive

ianstenbit (approved):
Thank you! Just one small nit and then I think we can merge this.

I checked the [lint action](https://github.com/keras-team/keras-cv/actions/runs/4446877913/jobs/7807728329?pr=1536) and it looks like it did in fact check the clang format.

Also -- (optional) if you'd be interested in adding clang-format to the devcontainer as a vscode extension in this PR, that would be nice too!

shell/lint.sh
cosmo3769: Okay. I will remove these comments.

cosmo3769:
> Also -- (optional) if you'd be interested in adding clang-format to the devcontainer as a vscode extension in this PR, that would be nice too!

Okay. I will add this too.

cosmo3769:
Added. Please check @ianstenbit. Thank you.

PR #1530, opened 2023-03-16
Replace tf.keras with keras throughout KerasCV 

soma2000-lang:
Resolves #1527 @ianstenbit 
I have reformatted my files using shell/format.sh  but dont know why the code format is still failing

PR #1538, opened 2023-03-18
tf.keras repalced with keras

vijaybirju:
tf.keras repalced with keras from a file


vijaybirju:
I have signed CLA. Now what's next

ianstenbit:
Hi @vijaybirju -- this was fixed in #1530 which was just merged.

Sorry for the overlap here, and thank you for your PR!

PR #1540, opened 2023-03-18
Update efficientnet_lite.py 

vijaybirju:
File cleaned remove tf.keras from keras

### DIFF
--- a/keras_cv/models/efficientnet_lite.py
+++ b/keras_cv/models/efficientnet_lite.py
@@ -24,6 +24,7 @@ import copy
 import math
 
 import tensorflow as tf
+from tf import Keras
 from keras import backend
 from keras import layers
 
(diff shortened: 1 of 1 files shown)


keras_cv/models/efficientnet_lite.py
```diff
@@ -24,6 +24,7 @@
 import math
 
 import tensorflow as tf
+from tf import Keras
```
innat, line 27: from tensorflow and not tf

keras_cv/models/efficientnet_lite.py
```diff
@@ -24,6 +24,7 @@
...
 from keras import backend
 from keras import layers
```
innat, line 29: import from tensorflow

keras_cv/models/efficientnet_lite.py
vijaybirju: ok

keras_cv/models/efficientnet_lite.py
vijaybirju: thank you made it

ianstenbit:
Hi @vijaybirju -- this was fixed in #1530 which was just merged.

Sorry for the overlap here, and thank you for your PR!

PR #1541, opened 2023-03-18
Expand definition of 'value-range' in equalization docs

Anirban0011:
# What this PR does:

* Updates the definition of value_range argument of Equalization class in preprocessing/layers/equalization.py

### DIFF
--- a/keras_cv/layers/preprocessing/equalization.py
+++ b/keras_cv/layers/preprocessing/equalization.py
@@ -24,9 +24,9 @@ class Equalization(BaseImageAugmentationLayer):
     """Equalization performs histogram equalization on a channel-wise basis.
 
     Args:
-        value_range: a tuple or a list of two elements. The first value represents
-            the lower bound for values in passed images, the second represents the
-            upper bound. Images passed to the layer should have values within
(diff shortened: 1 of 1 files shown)


Anirban0011:
#1544

Anirban0011:
#1545

Anirban0011:
#1547

Anirban0011:
@ianstenbit @LukeWood , please have a look at the doc fix.Thanks

keras_cv/layers/preprocessing/equalization.py
```diff
@@ -24,9 +24,9 @@ class Equalization(BaseImageAugmentationLayer):
     """Equalization performs histogram equalization on a channel-wise basis.
 
     Args:
-        value_range: a tuple or a list of two elements. The first value represents
-            the lower bound for values in passed images, the second represents the
-            upper bound. Images passed to the layer should have values within
+        value_range: a tuple, list, numpy array, tf.tensor or any object that can return integer on indexing of two elements.
```
ianstenbit, line 27: While it is true that a numpy array or any int-indexible type can be passed here, we expect users to use a tuple or list, and this makes the docstring a bit harder to parse, so I think we should leave the docs as-is.

PR #1544, opened 2023-03-19
Update definitions of tensor like args in preprocessing and auto_contrast docs

Anirban0011:
# What does this PR do?

## Updates definition of :

*  'value_range' in  layers/preprocessing/auto_contrast.py

* original_range' & 'target_range' in utils/preprocessing.py

### Fixes some minor docstring errors.

### DIFF
--- a/keras_cv/layers/preprocessing/auto_contrast.py
+++ b/keras_cv/layers/preprocessing/auto_contrast.py
@@ -31,7 +31,7 @@ class AutoContrast(VectorizedBaseImageAugmentationLayer):
 
     Args:
         value_range: the range of values the incoming images will have.
-            Represented as a two number tuple written [low, high]. 
+            Represented as a two number tuple written [low, high].
             This is typically either `[0, 1]` or `[0, 255]` depending
             on how your preprocessing pipeline is set up.
             (Note: It can be passed as any datatype that returns an integer on indexing like a list, tuple, numpy array or tf.tensor)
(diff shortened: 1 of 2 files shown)


Anirban0011:
#1547

Anirban0011:
@ianstenbit @LukeWood kindly review. Thanks

keras_cv/layers/preprocessing/auto_contrast.py
```diff
@@ -31,7 +31,8 @@ class AutoContrast(VectorizedBaseImageAugmentationLayer):
 
     Args:
         value_range: the range of values the incoming images will have.
-            Represented as a two number tuple written [low, high].
+            Represented as a two number tuple written [low, high]. 
+            It can be passed as any datatype that returns an integer on indexing like a list, tuple, numpy array or tf.tensor
```
LukeWood, line 35: I'm not sure this change is actually helpful - I think it just adds noise to end users.

LukeWood: A more valuable update would be an example:

"""
i.e. `value_range=(0, 255)`
"""

Anirban0011: I thought in case users pass the input as a different datatype than a tuple, they must be aware that it actually works. 
About the example update, I will try to add it as soon as this conversation is resolved.

Anirban0011: @LukeWood Kindly reply if you still feel there is an issue.

PR #1553, opened 2023-03-21
Change __new__() args from self to cls

haifeng-jin:
Minor fixes for ResNetV2Backbone.
Changing the `self` to `cls` in `__new__()` signature.


ianstenbit:
I re-merged from Master to verify that the GCB tests are happy, because as-is they were broken by an existing error.

ianstenbit:
Agh sorry -- it turns out this CI test is broken at head. @LukeWood can you PTAL?

ianstenbit:
Looks like Luke fixed the CI issue in #1558, and it's unrelated.

jbischof:
Thank you @haifeng-jin !

PR #1555, opened 2023-03-21
Vectorize Mosaic

james77777778:
# What does this PR do?

Related to #291 

||Mode|n=1000|n=2000|n=5000|n=10000|
|-|-|-|-|-|-|
|OldMosaic|Eager|1.097s|1.591s|2.980s|6.105s|
|OldMosaic|Graph|0.347s|0.560s|1.576s|3.733s|
|Mosaic|Eager|0.014s|0.020s|0.077s|0.143s|
|Mosaic|Graph|0.010s|0.016s|0.070s|0.138s|

comparison

XLA failed to run due to lacking `CropAndResize` kernel.

To visualize the outputs, I fixed `examples/layers/preprocessing/bounding_box/mosaic_demo.py` with the help from #1534:

```python
import tensorflow as tf
from examples.layers.preprocessing.bounding_box import demo_utils
from keras_cv.layers import preprocessing
from keras_cv.layers.preprocessing.jittered_resize import JitteredResize

def main():
    dataset = demo_utils.load_voc_dataset(bounding_box_format="xyxy")
    resize = JitteredResize(
        target_size=(256, 256),
        scale_factor=(0.5, 1.0),
        bounding_box_format="xyxy",
    )
    mosaic = preprocessing.Mosaic(bounding_box_format="xyxy")
    dataset = dataset.map(resize, num_parallel_calls=tf.data.AUTOTUNE)
    dataset = dataset.map(mosaic, num_parallel_calls=tf.data.AUTOTUNE)
    demo_utils.visualize_data(dataset, bounding_box_format="xyxy")

if __name__ == "__main__":
    main()
```

This layer (and the old implementation as well) cannot run with ragged images directly. But it works with dense images.

output

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@LukeWood @ianstenbit

### DIFF
--- a/keras_cv/layers/preprocessing/mosaic.py
+++ b/keras_cv/layers/preprocessing/mosaic.py
@@ -1,4 +1,4 @@
-# Copyright 2022 The KerasCV Authors
+# Copyright 2023 The KerasCV Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


james77777778:
Some additional notes for reviewing:

This PR decouples the old `_batch_augment` into `get_random_transformation_batch`, `augment_images`, `augment_labels` and `augment_bounding_boxes`

Now, Mosaic raises ValueError with message when input is ragged image.
Old Mosaic raised ValueError at https://github.com/keras-team/keras-cv/blob/8c34dbb2c34d1e4aa19fd8bbfbb14a0f2f78cc86/keras_cv/layers/preprocessing/mosaic.py#L103-L107
because `input_width` is None. I think it was too implicit for end-user.


There should be a bug at
https://github.com/keras-team/keras-cv/blob/8c34dbb2c34d1e4aa19fd8bbfbb14a0f2f78cc86/keras_cv/layers/preprocessing/vectorized_base_image_augmentation_layer.py#L166
This function call drops **kwargs so we cannot access `bounding_boxes` and `images` in `self.augment_labels`.

LukeWood:
@james77777778 thanks a lot for this contribution.  Can you open:
```
augment_targets drops **kwargs so we cannot access bounding_boxes and images in self.augment_labels.
This PR overwrites augment_targets in Mosaic as a workaround.
```

In a but?  This is not the expected behavior and is a bug.  Feel free to fix it in the base layer.

keras_cv/layers/preprocessing/mosaic.py
```diff
@@ -72,159 +86,98 @@ def __init__(
...
 
-    def _augment(self, inputs):
+    def augment_ragged_image(self, image, transformation, **kwargs):
         raise ValueError(
-            "Mosaic received a single image to `call`.  The layer relies on "
-            "combining multiple examples, and as such will not behave as "
-            "expected.  Please call the layer with 4 or more samples."
+            "Mosaic received ragged images to `call`. The layer relies on "
+            "combining multiple examples with same size, and as such will not "
+            "behave as expected. Please call the layer with dense images."
```
LukeWood, line 126: maybe add a note that if you would find this method useful open an issue on KerasCV.

Its actually possible to support this - just an engineering investment :)

LukeWood: (basically lets make it clear that this is an implementation constraint - not an algorithm constraint)

LukeWood (approved):
Thank you @james77777778 !

LukeWood:
this is super excited - we can now train `YOLOx` end to end with the correct KPL pipeline 🔥

james77777778:
> In a but? This is not the expected behavior and is a bug. Feel free to fix it in the base layer.

I will open another PR to fix it!

> maybe add a note that if you would find this method useful open an issue on KerasCV.
> Its actually possible to support this - just an engineering investment :)

Since this PR is merged, I will add a note at another PR.

> this is super excited - we can now train YOLOx end to end with the correct KPL pipeline 🔥

🚀

keras_cv/layers/preprocessing/mosaic.py
james77777778: I will open another PR to add this note!

PR #1558, opened 2023-03-21
Disable COCO fit test

LukeWood:
removes the COCO fit test until we have forward pass diff testing.

This is currently causing so many CI issues and the API is under review still for COCO metrics.

We should disable this until we have the forward pass diff testing we are working on done.


keras_cv/models/object_detection/retina_net/retina_net_coco_test.py
```diff
@@ -32,7 +32,8 @@ def cleanup_global_session(self):
         # Reset soft device placement to not interfere with other unit test files
         tf.config.set_soft_device_placement(True)
         keras.backend.clear_session()
-
+    
+    @pytest.skip(reason="Disabled until forward-pass numerical testing is implemented")
```
ianstenbit, line 36: We'll need to make sure this is also skipped internally.

ianstenbit (approved):

PR #1562, opened 2023-03-22
Fix the serialization of RetinaNet

haifeng-jin:
removed the `anchor_generator` from `get_config()` because it cannot be provided together with `label_encoder` and `prediction_decoder` while initializing the class.

I suggest we start to add these serialization tests when implementing new models.
Otherwise, many models may not work in saving.

### DIFF
--- a/keras_cv/models/object_detection/retina_net/__internal__/layers/prediction_head.py
+++ b/keras_cv/models/object_detection/retina_net/__internal__/layers/prediction_head.py
@@ -64,9 +64,22 @@ class PredictionHead(layers.Layer):
 
     def get_config(self):
         config = {
-            "bias_initializer": self.bias_initializer,
+            "bias_initializer": keras.initializers.serialize(
+                self.bias_initializer
+            ),
             "output_filters": self.output_filters,
             "num_conv_layers": self.num_conv_layers,
         }
         base_config = super().get_config()
         return dict(list(base_config.items()) + list(config.items()))
+
+    @classmethod
+    def from_config(cls, config):
+        config.update(
+            {
+                "bias_initializer": keras.initializers.deserialize(
+                    config["bias_initializer"]
+                )
+            }
+        )
+        return super().from_config(config)
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -193,7 +193,9 @@ class RetinaNet(keras.Model):
             self.backbone, extractor_layer_names, extractor_levels
         )
         self.feature_pyramid = layers_lib.FeaturePyramid()
-        prior_probability = tf.constant_initializer(-np.log((1 - 0.01) / 0.01))
+        prior_probability = keras.initializers.Constant(
+            -np.log((1 - 0.01) / 0.01)
+        )
 
         self.classification_head = (
             classification_head
@@ -204,7 +206,7 @@ class RetinaNet(keras.Model):
         )
 
         self.box_head = box_head or layers_lib.PredictionHead(
-            output_filters=9 * 4, bias_initializer="zeros"
+            output_filters=9 * 4, bias_initializer=keras.initializers.Zeros()
         )
 
     def make_predict_function(self, force=False):
@@ -516,12 +518,9 @@ class RetinaNet(keras.Model):
         return {
             "num_classes": self.num_classes,
             "bounding_box_format": self.bounding_box_format,
-            # TODO(bischof): actually serialize the backbone
             "backbone": self.backbone,
-            "anchor_generator": self.anchor_generator,
             "label_encoder": self.label_encoder,
             "prediction_decoder": self._prediction_decoder,
-            "feature_pyramid": self.feature_pyramid,
             "classification_head": self.classification_head,
             "box_head": self.box_head,
         }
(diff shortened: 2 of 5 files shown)


keras_cv/models/object_detection/retina_net/__internal__/layers/prediction_head.py
```diff
@@ -64,9 +64,22 @@ def call(self, x, training=False):
...
+    @classmethod
+    def from_config(cls, config):
+        config.update(
+            {
+                "bias_initializer": keras.initializers.deserialize(
+                    config["bias_initializer"]
+                )
+            }
+        )
+        return super().from_config(config)
```
haifeng-jin, line 85: @nkovela1 Would you please take a look if I did something weird here? Is there a way to implicitly serialize and deserialize the initializer? Thanks.

nkovela1: Hi Haifeng,

I believe the explicit serialization and deserialization of the bias initializer is unnecessary here. It should already be taken care of unless you are using a custom (non-Keras) initializer. Please let me know if removing serialization there results in any errors.

keras_cv/models/object_detection/retina_net/__internal__/layers/prediction_head.py
```diff
@@ -64,9 +64,22 @@ def call(self, x, training=False):
 
     def get_config(self):
         config = {
-            "bias_initializer": self.bias_initializer,
+            "bias_initializer": keras.initializers.serialize(
```
ianstenbit, line 67: Does this mean that we won't be able to serialize the string "zeros", and the actual keras initializer will have to be passed?

keras_cv/models/object_detection/retina_net/__internal__/layers/prediction_head.py
haifeng-jin: We will stick to the explicit serialization for initializers for now : )
Thanks, @nkovela1!

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -516,12 +518,9 @@ def get_config(self):
         return {
             "num_classes": self.num_classes,
             "bounding_box_format": self.bounding_box_format,
-            # TODO(bischof): actually serialize the backbone
             "backbone": self.backbone,
-            "anchor_generator": self.anchor_generator,
```
ianstenbit, line 521: Won't removing this break serialization when a custom anchor_generator is used?

ianstenbit (commented):
A few questions -- sorry I'm a bit of a rookie when it comes to this serialization stuff!

keras_cv/models/object_detection/retina_net/__internal__/layers/prediction_head.py
haifeng-jin: Yes. I see `PredictionHead` as a internal only class, so we may enforce it. If we still want the "zeros", we  can do a special case here for the string types.

keras_cv/models/object_detection/retina_net/retina_net.py
haifeng-jin: Yes, this is a problem. I am not so sure about the logics between `anchor_generator`, `label_encoder`, `prediction_decoder`. When `get_config()` is called, they all exist. However, they cannot be all passed to `__init__()`. What would be a good way to judge which one to omit here?

ianstenbit: Okay actually I think what you've done is exactly right -- since the encoder/decoder are dependent on this, just serializing the encoder/decoder should be good.

This makes me wonder though -- @LukeWood why can't we pass a custom `anchor_generator` in addition to a custom `prediction_decoder`? It seems like those are now decoupled, so I wonder if this is an artifact of how this used to be structured?

ianstenbit (approved):
Thanks! This LGTM -- I left one question for @LukeWood that we can probably handle in a follow-up

haifeng-jin:
\gcbrun

ianstenbit:
> \gcbrun

Gotta use a forward-slash 😄

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: > Yes, this is a problem. I am not so sure about the logics between anchor_generator, label_encoder, prediction_decoder. When get_config() is called, they all exist. However, they cannot be all passed to __init__(). What would be a good way to judge which one to omit here?

We definitely should not  prevent the use of custom anchor generators.  One thing we could do is exclude anchor generators, and encourage only the use of `save_weights()` and `load_weights()`?  I'm not sure the right approach, but anchor generators and weights are tightly coupled unfortunately.

LukeWood:
Realistically, this needs a deeper dive (probably after I finish forward pass testing). 

I think this is a step in the right direction, so lets go for it with the note to revisit once we have strong numberics.

keras_cv/models/object_detection/retina_net/retina_net.py
ianstenbit: @LukeWood -- since the label encoder contains the anchor generator, by serializing the label encoder aren't we also effectively serializing the anchor generator?

I feel like I'm missing something. Sorry if it's a stupid question!

haifeng-jin:
@LukeWood Was this PR somehow pending on https://github.com/keras-team/keras-cv/pull/1458? Just add this note for future reference.

Issue #1548, opened 2023-03-20
Remove `Augmenter` and use `keras.Sequential` instead


haifeng-jin:
Because `Augmenter` is similar to `keras.Sequential` in functionality, we would like to remove the `Augmenter` from KerasCV and recommend to use `keras.Sequential` instead.

The change require the following steps.
* Change the keras.io notebooks to use `keras.Sequential` instead of `Augmenter` based on KerasCV v0.4.1. Starting from [this notebook](https://keras.io/guides/keras_cv/retina_net_overview/).
* Remove the `Augmenter` and its usages from the codebase. Remember to update the release note.

soma2000-lang:
@haifeng-jin It would be great to work on this issue.Can I send a PR for this.

PR #1568, opened 2023-03-22
Remove Augmenter usage in README

haifeng-jin:
Related to #1548

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ from tensorflow import keras
 import tensorflow_datasets as tfds
 
 # Create a preprocessing pipeline
-augmenter = keras_cv.layers.Augmenter(
+augmenter = keras.Sequential(
     layers=[
         keras_cv.layers.RandomFlip(),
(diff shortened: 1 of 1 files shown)


LukeWood:
@haifeng-jin do you know what TensorFlow this bumps our minimal dependency to?

LukeWood:
> @LukeWood I don't know. I don't think it would change the minimal TF version. The sequantial model should support KPLs all the time, right?

Actually this is why the augmenter class exists.  Previously the Sequential model would error out when using KPLs with dictionary formatted inputs.  I believe it might be 2.12 minimum to support, perhaps 2.11 - no matter the case lets make sure to bump the minimum version so we don't confuse users.

ianstenbit (approved):
Thanks!

I'm honestly surprised this doesn't show up in any of our other examples / codebase, but I suppose that's a good thing!

@LukeWood do you want to verify the TF version dependency before merging?

PR #1576, opened 2023-03-23
Add layer name in convmixer

### LINKED ISSUE #1571: Add layer names [ConvMixer]

Anirban0011:
# What does this PR do?

* Attempts to potentially fix  #1571 

#### Special thanks to @IMvision12 for intellectually co-authoring the PR


IMvision12:
@Anirban0011 for formatting you can use black : `black --line-length 80 file_path`

Anirban0011:
@LukeWood , @ianstenbit , kindly review whenever possible, thanks!

keras_cv/models/convmixer.py
```diff
@@ -106,20 +107,24 @@ def apply_conv_mixer_layer(x, dim, kernel_size):
     Returns:
         the updated input tensor.
     """
+    if name is None:
+        name = str(backend.get_uid("apply_conv_mixer_layer"))
```
LukeWood, line 111: This breaks our pretrained weights, and `apply_` should not end up in layer names.

Anirban0011: removed apply_ from names

Anirban0011: > This breaks our pretrained weights, and `apply_` should not end up in layer names.

@LukeWood , then name string should not be generated this way if name is `None` ?

PR #1582, opened 2023-03-24
Fix padding issue in StableDiffusion encoder

### LINKED ISSUE #1172: BUG in stable_diffusion.image_encoder.ImageEncoder creates edge padding on encoded image.
This was found when debugging: https://github.com/keras-team/keras-cv/issues/1158 and is likely the root cause.

The pretrained image encoder adds zero padding to the top and left edges of the image when encoding. This can be reproduced and seen as follows:
Screenshot 2022-12-20 at 12 24 43 PM
The original image is transformed via `out = decoder(encoder(img))` resulting in an output image that is visually identical other than the gray border at the top and left edges. Furthermore, we can likely assume that there's no problem with the decoder since that is being used in the standard `StableDiffusion.text_to_image` function which does not have this issue visually.

ianstenbit:
Fixes #1172

In the [testing colab](https://colab.research.google.com/gist/ianstenbit/4c20db0d98dde03b0a68e7afaf713084/sd-encoder-padding.ipynb), you can see that the gray bar on the top and left sides of the auto-encoded image are now gone.

Additionally, the mean error of the reconstructed image is down about 10x (probably just from getting rid of the bars).

This is not the most elegant fix the world has ever seen, but it does seem to fix the numerics of the model, and the qualitative results from the encoder are much better.

I've manually run the SD tests, and I am currently also re-running the Textual Inversion notebook (as it depends on the encoder heavily). I'll post results once it's done, but it's looking fine so far.

### DIFF
--- a/keras_cv/models/stable_diffusion/__internal__/layers/padded_conv2d.py
+++ b/keras_cv/models/stable_diffusion/__internal__/layers/padded_conv2d.py
@@ -16,10 +16,20 @@ from tensorflow import keras
 
 
 class PaddedConv2D(keras.layers.Layer):
-    def __init__(self, filters, kernel_size, padding=0, strides=1, **kwargs):
+    def __init__(
+        self,
+        filters,
+        kernel_size,
+        padding=0,
+        strides=1,
+        conv_padding="valid",
+        **kwargs
+    ):
         super().__init__(**kwargs)
         self.padding2d = keras.layers.ZeroPadding2D(padding)
-        self.conv2d = keras.layers.Conv2D(filters, kernel_size, strides=strides)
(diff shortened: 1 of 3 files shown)


ianstenbit:
Update: textual inversion is working
https://colab.research.google.com/gist/ianstenbit/cf9af08d2de90b5080ed7447c55778c1/fine_tune_via_textual_inversion.ipynb

LukeWood:
> This is not the most elegant fix the world has ever seen, but it does seem to fix the numerics of the model, and the qualitative results from the encoder are much better.

Thanks for running these tests @ianstenbit - I'm glad we were able to fix this issue.

LukeWood (approved):

PR #1587, opened 2023-03-26
Update if include_top condition in resnet class

Anirban0011:
## What this PR does:

* Attempts to fix the if include_top condition by letting it evaluate only once as opposed to twice before to raise value error.


Anirban0011:
@ianstenbit @LukeWood , kindly review whenever possible. Thanks!

LukeWood:
> access

Check out this article: https://deviq.com/design-patterns/guard-clause

I think following a guard clause style here is more readable, lets leave it for now!

PR #1593, opened 2023-03-27
Update the definition of IoULoss in iou_loss docs

Anirban0011:
## What this PR does:

* Attempts to expand upon the existing definition of IoULoss

### DIFF
--- a/keras_cv/losses/iou_loss.py
+++ b/keras_cv/losses/iou_loss.py
@@ -23,11 +23,13 @@ class IoULoss(keras.losses.Loss):
     """Implements the IoU Loss
 
     IoU loss is commonly used for object detection. This loss aims to directly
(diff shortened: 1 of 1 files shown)


Anirban0011:
@ianstenbit @LukeWood , kindly review whenever possible. Thanks!

keras_cv/losses/iou_loss.py
```diff
@@ -23,11 +23,13 @@ class IoULoss(keras.losses.Loss):
     """Implements the IoU Loss
 
     IoU loss is commonly used for object detection. This loss aims to directly
```
ianstenbit, line 25: How about instead we just say "IoU (Intersection over Union) loss ..." and scrap the other changes?

ianstenbit (changes requested):

PR #1597, opened 2023-03-28
Expand GIoU loss definition in GIoULoss docs

Anirban0011:
## What this PR does:

* Attempts to expand upon the existing definition of GIoU Loss


ianstenbit:
Thanks for the PR. I don't think this update really improves the docstring since we already point to the GIoU paper where folks can go read about the details + theory if they are interested.

PR #1603, opened 2023-03-28
Trivial fix to WaymoEvaluationCallback tests.

ianstenbit:
When running this test against TF compiled with strict type checking, it fails due to casting a negative number to an unsigned int.

This change resolves that issue.


haifeng-jin (approved):
LGTM. as long as the classes should never be negative.

ianstenbit:
> LGTM. as long as the classes should never be negative.

Yep, the classes should never be negative. They are negative in the test model because it's just a randomly-initialized dense layer.

Thanks Haifeng!

PR #1604, opened 2023-03-28
Remove PYCOCOCallback

LukeWood:
Now we have the PyMetric version of this which is superior.  You can wrap the PyMetric version in a callback in 5-10 lines of code if you'd like TPU support (I have this in my updated guide), so this is a better approach.


ianstenbit (commented):
IMO we should stop recommending this in guides and wait a while to make sure the PyMetric approach lands before we axe this.

Marking it as deprecated for now might be a better step?
wdyt?

LukeWood:
> IMO we should stop recommending this in guides and wait a while to make sure the PyMetric approach lands before we axe this.
> 
> Marking it as deprecated for now might be a better step? wdyt?

Personally I think we should just remove it and focus on getting to an OD flow we're proud of.  I think the marginal value of keeping it around a few releases is not worth the effort.

PR #1605, opened 2023-03-28
Explicitly serialize backbone in RetinaNet

haifeng-jin:
This is to address the future serialization & deserialization mechanism changes in TF 2.13.
It would no longer implicitly serialize & deserialize any attribute of the object (e.g. self.backbone).
So we do it explicitly.

### DIFF
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -514,11 +514,21 @@ class RetinaNet(keras.Model):
                 metrics[metric.name] = result
         return metrics
 
+    @classmethod
+    def from_config(cls, config):
+        config["backbone"] = keras.utils.deserialize_keras_object(
+            config["backbone"]
+        )
+        return super().from_config(config)
+
     def get_config(self):
         return {
             "num_classes": self.num_classes,
             "bounding_box_format": self.bounding_box_format,
-            "backbone": self.backbone,
+            "backbone": keras.utils.serialize_keras_object(self.backbone),
+            # TODO(haifengj): handle custom anchor_generator. we now rely on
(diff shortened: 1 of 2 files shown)


keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -514,11 +514,21 @@ def compute_metrics(self, x, y, y_pred, sample_weight):
...
+            config["backbone"]
+        )
+        return super().from_config(config)
+
     def get_config(self):
         return {
             "num_classes": self.num_classes,
             "bounding_box_format": self.bounding_box_format,
-            "backbone": self.backbone,
+            "backbone": keras.utils.serialize_keras_object(self.backbone),
```
ianstenbit, line 528: I didn't realize that serialization also needs to be explicit, I thought it was just deserialization?

But either way this LGTM, as this seems like the most defensive approach.

ianstenbit (approved):
Thanks Haifeng!

keras_cv/models/object_detection/retina_net/retina_net.py
nkovela1: Only deserialization should be explicitly necessary here. But this seems like a good defensive approach either way! Approved.

nkovela1 (approved):

LukeWood (approved):

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: This actually breaks old TF versions, we should bump to 2.12

haifeng-jin: That would affect the experience of Colab users. We need to make sure Colab is updated to TF 2.12 when keras-cv 0.5 is released.

PR #1606, opened 2023-03-29
Remove the Augmenter class

### LINKED ISSUE #1548: Remove `Augmenter` and use `keras.Sequential` instead

haifeng-jin:
Remove the `Augmenter` class completely.
Fixes #1548


ianstenbit (approved):
LGTM -- thank you!

PR #1610, opened 2023-03-29
Update focal loss definition in focal_loss docs

Anirban0011:
## What this PR does:

* Attempts to expand upon the existing definition of focal loss in the docstring.

### DIFF
--- a/keras_cv/losses/focal.py
+++ b/keras_cv/losses/focal.py
@@ -22,7 +22,11 @@ class FocalLoss(keras.losses.Loss):
     """Implements Focal loss
 
     Focal loss is a modified cross-entropy designed to perform better with
-    class imbalance. For this reason, it's commonly used with object detectors.
+    class imbalance. It minimizes the weight of easy -ves on the loss function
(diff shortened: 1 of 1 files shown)


Anirban0011:
@ianstenbit @LukeWood , kindly review whenever possible . Thanks!

keras_cv/losses/focal.py
```diff
@@ -22,7 +22,11 @@ class FocalLoss(keras.losses.Loss):
     """Implements Focal loss
 
     Focal loss is a modified cross-entropy designed to perform better with
-    class imbalance. For this reason, it's commonly used with object detectors.
+    class imbalance. It minimizes the weight of easy -ves on the loss function
```
ianstenbit, line 25: Since we already reference the FocalLoss paper below, I think this detail is a bit extraneous (it's not critical information to a user of this API, and if they want to learn more they're already referred to the paper)

ianstenbit:
Thanks for the PR! See my comment above -- I think we should leave this doc as-is for now.

PR #1611, opened 2023-03-29
Mark a pair of test cases as XL

ianstenbit:
This is better than using our current `INTEGRATION` tag, and is one (final?) step toward automatic syncing to G3


LukeWood:
@ianstenbit are these skipped by default?

ianstenbit:
> @ianstenbit are these skipped by default?

Yeah, the DeepLab test is skipped because our CI doesn't set `REGRESSION=true`. The numerical comparison test for the `within_box3d` versions will become skipped by default, which is fine because it's a super expensive test that's irrelevant to most PRs, and should be run with the XL flag.

PR #1612, opened 2023-03-29
Polish the RetinaNet experience

LukeWood:
This PR implements a few distinct fixes across the RetinaNet.  These were found while ironing out the reasons for the performance regressions I'd been seeing in the RetinaNet.

## Consistent delta box format handling

There was a bug in the box format handling that was introduced while unifying the _encode_to_box_delta and _decode_to_box_deltas utility functions.  There was always a bit of ambiguity and tradeoff in this space: either your deltas have to be the same format, or your numerics will vary depending on your chosen box format.  The reason for this is that box deltas are fed to your loss function, and if you are in `xywh` for example you will get radically different numerics than if you are in `rel_xywh` or `center_yxwh`.  `center_` formats DO have the strongest numerical performance, so we should standardize on this internally.  The only place this poses issues is specific box-targeted loss functions (IoU loss, GIoU loss), so these losses will require some special handling.

To workaround these, I've proposed we just standardize on the `center_yxwh` INTERNALLY for loss computation.  In this process I've also supported ANY bounding box format in the RetinaNet - a huge upgrade from only supporting `xywh`.

## Backbone changes

The `backbone` parameter is now required.  Before I had a default of ResNet50V2, which we accidentally changed - and this is just not a good pattern.  This is extremely unintuitive, but ResNet50V2 is  HORRIBLE backbone - while ResNet50 works fine.  I've worked around this now in this PR.  Following Haifeng's PR we can merge this.

Unfortunately, this is the sort of bug that is very hard to catch until you go train the model.  Luckily, integration testings hould catch this.

## Other Suspected issues

I'm fairly sure the BoxCOCOMetrics() suite is broken - I get higher scores with the `PyCOCOCallback()`.  I'll do some experimentation and see if I can fix this.  For now, I am adding back in PYCOCOCallback until I can fix this.

##  Next Steps + Integration Testing

- [x] verify 0.33 MaP with the old ResNet50.as_backbone()

After this PR I score 0.38.

- [ ] rebase on Haifeng's change to ResNet50, verify 0.33 MaP
- [ ] verify PyCOCOCallback and custom COCO callback get identical scores

## Follow ups

- Following this PR I will be training a fully pretrained model, and adding a preset to the RetinaNet to support this.
- Add integration testing


LukeWood:
https://github.com/keras-team/keras-cv/pull/1612/files Fixes the RetinaNet, across all bounding_box formats, with identical numerics (`bounding_box_format="xywh"` is now numerically the same as `bounding_box_format="yxyx"`). This also fixes a double-conversion bug that snuck in when we were unifying logic to MaskRCNN.  Lastly, it catches a numerical regression that was in the backbone (ResNet50->ResNet50V2 gets substantially worse Numerics for RetinaNet).

There are some open steps I need to do (i.e. use @haifeng-jin new Task level ResNet50, verify PyCOCOCallback and my custom callback are the same, save the weights) but after that we should be good to go on launching a pretrained RetinaNet w/ decent numerics🔥

PTAL, lets do a pass on this - please note there is also an API change.  we can now accept `feature_extractor` instead of just backbone.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -80,9 +80,15 @@ class RetinaNet(keras.Model):
         bounding_box_format: The format of bounding boxes of input dataset. Refer
             [to the keras.io docs](https://keras.io/api/keras_cv/bounding_box/formats/)
             for more details on supported bounding box formats.
-        backbone: optional `keras.Model`. Must implement the `pyramid_level_inputs`
+        backbone: (Optional) `keras.Model`. Must implement the `pyramid_level_inputs`
             property with keys 3, 4, and 5 and layer names as values. If
-            `None`, defaults to `keras_cv.models.ResNet50V2Backbone()`.
+            `None`, defaults to
+            `keras_cv.models.ResNet50Backbone.from_preset('imagenet')`.
+        feature_extractor: (Optional) `keras.Model`.  A feature extractor takes
```
jbischof, line 87: We just cut this in #1454 to simplify the code. Why are we reintroducing?

My best summary of the discussion is that no one could provide a compelling example of why a custom feature extractor would be needed.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -171,11 +178,28 @@ def __init__(
...
+        if backbone is not None and feature_extractor is not None:
+            raise ValueError(
+                "Expected one of `backbone` or `feature_extractor`, but got "
+                f"`backbone={backbone}`"
+                f"`feature_extractor={feature_extractor}`"
+            )
+
+        # Only create a backbone if feature_extractor is None
+        if backbone is None and feature_extractor is None:
+            backbone = keras_cv.models.ResNet50.from_preset('resnet50_imagenet')
```
jbischof, line 191: Is an untrained model actually a good backbone? If not I would prefer to change `backbone` to a positional arg and forget the defaults. We can recommend a `backbone` in the docstring.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -516,19 +528,13 @@ def compute_metrics(self, x, y, y_pred, sample_weight):
...
 
     def get_config(self):
         return {
             "num_classes": self.num_classes,
             "bounding_box_format": self.bounding_box_format,
-            "backbone": keras.utils.serialize_keras_object(self.backbone),
-            # TODO(haifengj): handle custom anchor_generator. we now rely on
-            # label_encoder and prediction_decoder to reconstruct
-            # anchor_gnerator.
+            "feature_extractor": self.feature_extractor,
```
jbischof, line 537: Why is superior to serializing the `backbone`? We serialize the backbone without issue in `ImageClassifier`

jbischof (commented):
Thanks @LukeWood! Starting with a couple questions since we're undoing several recent changes in this PR.

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: untrained is BAD (0.08~-0.12 MAP).  We can swap it to a positional argument.

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: the problem is that some `trainable_variables` on `backbone` are unused.  So you get all these annoying warnings telling you that you didn't pass a `loss` to compile(), when in fact the variables are just not used.

LukeWood: I'll fix this in a separate PR - this way I can fix one thing at a time and make it easier to review.

LukeWood: I've reverted, we can address in a separate PR.

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: I've made this required.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -80,9 +83,10 @@ class RetinaNet(keras.Model):
         bounding_box_format: The format of bounding boxes of input dataset. Refer
             [to the keras.io docs](https://keras.io/api/keras_cv/bounding_box/formats/)
             for more details on supported bounding box formats.
-        backbone: optional `keras.Model`. Must implement the `pyramid_level_inputs`
-            property with keys 3, 4, and 5 and layer names as values. If
-            `None`, defaults to `keras_cv.models.ResNet50V2Backbone()`.
+        backbone: `keras.Model`. Must implement the `pyramid_level_inputs`
+            property with keys 3, 4, and 5 and layer names as values. A somewhat
+            sensible backbone to use in many cases is the:
+            `keras_cv.models.ResNet50Backbone.from_preset('imagenet')`.
```
jbischof, line 89: This isn't a real preset. I'd recommend using the `keras_cv.models.ResNetBackbone.from_preset("resnet50_imagenet")` syntax.

LukeWood:
Ok, this no longer has an API change - so I've reduced the reviewer list a bit.

jbischof (approved):
Thanks!

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: Fixed, thanks

PR #1613, opened 2023-03-29
Fixes result(force=True) in PyMetric

LukeWood:
Added tests as well.

### DIFF
--- a/keras_cv/metrics/object_detection/box_coco_metrics.py
+++ b/keras_cv/metrics/object_detection/box_coco_metrics.py
@@ -192,17 +192,17 @@ class BoxCOCOMetrics(keras.metrics.Metric):
         # Wrap the result function in a py_function and scope it to /cpu:0
         obj_result = obj.result
 
-        def result_on_host_cpu():
+        def result_on_host_cpu(force):
             with tf.device("/cpu:0"):
                 # Without the call to `constant` `tf.py_function` selects the
                 # first index automatically and just returns obj_result()[0]
-                return tf.constant(obj_result(), obj.dtype)
+                return tf.constant(obj_result(force), obj.dtype)
 
         obj.result_on_host_cpu = result_on_host_cpu
 
-        def result_fn(self):
+        def result_fn(self, force=False):
             py_func_result = tf.py_function(
-                self.result_on_host_cpu, inp=[], Tout=obj.dtype
+                self.result_on_host_cpu, inp=[force], Tout=obj.dtype
             )
             result = {}
             for i, key in enumerate(METRIC_NAMES):
--- a/keras_cv/metrics/object_detection/box_coco_metrics_test.py
+++ b/keras_cv/metrics/object_detection/box_coco_metrics_test.py
@@ -90,6 +90,21 @@ class BoxCOCOMetricsTest(tf.test.TestCase):
             # passed which actually modifies the final area under curve value.
             self.assertNotEqual(metrics[metric], 0.0)
 
+    def test_coco_metric_suite_force_eval(self):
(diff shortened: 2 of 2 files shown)


ianstenbit:
It seems like `force` doesn't exists as a concept in upstream Keras metrics -- what's the use case?

keras_cv/metrics/object_detection/box_coco_metrics_test.py
```diff
@@ -90,6 +90,21 @@ def test_coco_metric_suite_evaluate_freq(self):
...
+        suite.update_state(y_true, y_pred)
+        metrics = suite.result()
+        self.assertAllEqual(metrics, {key: 0 for key in golden_metrics})
+
+        suite.update_state(y_true, y_pred)
+        metrics = suite.result(force=True)
+        for metric in metrics:
+            # The metrics do not match golden metrics because two batches were
+            # passed which actually modifies the final area under curve value.
+            self.assertNotEqual(metrics[metric], 0.0)
```
ianstenbit, line 106: Nit (optional): if we expect these to be >0, we could use assertAllGreater and avoid the loop

keras_cv/metrics/object_detection/box_coco_metrics.py
```diff
@@ -192,17 +192,17 @@ def update_state_fn(self, y_true, y_pred, sample_weight=None):
...
-                return tf.constant(obj_result(), obj.dtype)
+                return tf.constant(obj_result(force), obj.dtype)
 
         obj.result_on_host_cpu = result_on_host_cpu
 
-        def result_fn(self):
+        def result_fn(self, force=False):
             py_func_result = tf.py_function(
-                self.result_on_host_cpu, inp=[], Tout=obj.dtype
+                self.result_on_host_cpu, inp=[force], Tout=obj.dtype
```
ianstenbit, line 205: Have you verified that this works in Graph mode on TPU?
Since force is a Python boolean, I would expect this to break, but perhaps I am misunderstanding

keras_cv/metrics/object_detection/box_coco_metrics_test.py
ianstenbit: I assumed so, since you're using `assertAllEqual` with a dictionary above. But maybe not 🤷

keras_cv/metrics/object_detection/box_coco_metrics.py
LukeWood: This is invoked from python, so whatever value is passed will be converted to a Tensor.

LukeWood: Also, this already doesn't work on TPU, right?

keras_cv/metrics/object_detection/box_coco_metrics_test.py
LukeWood: Equal makes sense, because it will compare the keys via iteration.

For _ in dictionary pulls out the keys, so it will run "string" > 0.0

ianstenbit: Ahh yes, that's goofy. Nevermind this then :)

LukeWood:
https://colab.research.google.com/drive/1BOYJcjV_2G4gXQKp178AVedmb3J2pUls?usp=sharing

ianstenbit (approved):
LGTM, mod:

- ~Linter is unhappy~ edit: this is unrelated and should be fixed at head
- Need either a test or a TODO to add a training test with these metrics (not related to this PR necessarily, but would have made it easier to test this)
- ~Please link to your manual test colab/gist in the description~ nvm you already did this :)
Thanks!

PR #1618, opened 2023-03-30
Allow `from_preset()` for RetinaNet

LukeWood:
At long last, we can now load pretrained RetinaNets!

Create a RetinaNet with an untrained FPN:

```
import keras_cv

retina_net = keras_cv.models.RetinaNet.from_preset(
    "resnet50_imagenet", bounding_box_format="xywh"
)
```

Or a fully trained RetinaNet:

```
import keras_cv

retina_net = keras_cv.models.RetinaNet.from_preset(
    "retinanet_resnet50_pascalvoc", bounding_box_format="xywh"
)
```

## Basic Inference

```
import keras
import tensorflow as tf
from tensorflow import keras
from tensorflow.keras import optimizers
import keras_cv
import numpy as np
import os
import resource
from keras_cv import visualization
import tqdm

retina_net = keras_cv.models.RetinaNet.from_preset(
    'retinanet_resnet50_pascalvoc',
    bounding_box_format='xywh',
)

filepath = tf.keras.utils.get_file(origin="https://i.imgur.com/9i63gLN.jpg")
image = keras.utils.load_img(filepath)
image = np.array(image)
inference_resizing = keras_cv.layers.Resizing(640, 640, pad_to_aspect_ratio=True, bounding_box_format='xywh')
image_batch = image[None, ...]
image_batch = inference_resizing(image_batch)

y_true = {
    'classes': [[7]],
    'boxes': [[[63, 235, 400, 300]]]
}

retina_net.prediction_decoder = keras_cv.layers.MultiClassNonMaxSuppression(
    bounding_box_format="xywh",
    from_logits=True,
    # Decrease the required threshold to make predictions get pruned out
    iou_threshold=0.35,
    # Tune confidence threshold for predictions to pass NMS
    confidence_threshold=0.6,
)
y_pred = retina_net.predict(image_batch)

class_ids = [
    "Aeroplane",
    "Bicycle",
    "Bird",
    "Boat",
    "Bottle",
    "Bus",
    "Car",
    "Cat",
    "Chair",
    "Cow",
    "Dining Table",
    "Dog",
    "Horse",
    "Motorbike",
    "Person",
    "Potted Plant",
    "Sheep",
    "Sofa",
    "Train",
    "Tvmonitor",
    "Total",
]
class_mapping = dict(zip(range(len(class_ids)), class_ids))

visualization.plot_bounding_box_gallery(
    image_batch,
    value_range=(0, 255),
    rows=1, cols=1,
    y_pred=y_pred,
    y_true=y_true,
    scale=5,
    bounding_box_format='xywh',
    class_mapping=class_mapping,
    legend=True
)
```
Result: 

ODModel

Pretty cool!!!

---

Tasks:
- [x] fix weight loading for fully pretrained
- [x] verify inference behavior
- [x] Consider renaming the fully pretrained presets perhaps...

Other testing:
---

```
keras_cv.models.ImageClassifier.from_preset("resnet50_imagenet", num_classes)
```

Still works.


keras_cv/models/object_detection/retina_net/retina_net_presets.py
```diff
@@ -0,0 +1,35 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""RetinaNet Task presets."""
+from keras_cv.models.backbones.resnet_v1 import resnet_v1_backbone_presets
+
+retina_net_presets = {
+    "resnet50_pascal_voc": {
```
jbischof, line 18: Since every preset name is unique, call this `"retinanet_resnet50_pascal_voc"` or something similar.

keras_cv/models/object_detection/retina_net/retina_net_presets.py
```diff
@@ -0,0 +1,35 @@
...
+                "layers (v2 style). Trained on Imagenet 2012 classification "
+                "task."
+            ),
+        },
+        "config": {
+            "backbone": resnet_v1_backbone_presets.backbone_presets["resnet50"],
+            # Should be 21
+            "num_classes": 20,
+        },
+        "weights_url": "https://storage.googleapis.com/keras-cv/models/retinanet/pascal_voc/resnet50.weights.h5",
```
jbischof, line 32: Why `.weights`?

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -98,26 +107,28 @@ class RetinaNet(keras.Model):
...
 
     def __init__(
         self,
+        backbone,
         num_classes,
         bounding_box_format,
-        backbone,
         anchor_generator=None,
         label_encoder=None,
         prediction_decoder=None,
```
jbischof, line 134: The `name` arg is redundant here.

keras_cv/models/task.py
```diff
@@ -136,7 +136,7 @@ def from_preset(
             return model
 
         weights = keras.utils.get_file(
-            "model.h5",
+            "model.weights.h5",
```
jbischof, line 139: Why this change? This will break out `ImageClassifier` presets for sure.

jbischof (changes requested):
This is great! Love how clean everything looks

keras_cv/models/object_detection/retina_net/retina_net_presets.py
LukeWood: there is a pretty long thread about this in the main chat;
TLDR: This is the new Keras model SavedFormat, and when you're using a sub-class model including any sort of floating layers/floating weights this will behave much more consistently.

This is also meant to be the new Keras model format going forwards

keras_cv/models/task.py
LukeWood: Oops, I had a fix for this but must not have pushed

keras_cv/models/object_detection/retina_net/retina_net_presets.py
LukeWood: i can't omit it for now.  I couldn't get the model to load without the new format.

fchollet: URLs are arbitrary strings, so it doesn't make any user-facing difference. Long-term all files should standardize on the new format. This is low priority though since it doesn't really change anything.

fchollet (approved):
LGTM

keras_cv/models/object_detection/retina_net/retina_net_presets.py
jbischof: Sorry I actually misunderstood this change. This only affects the name of the file in the local cache.

keras_cv/models/object_detection/retina_net/retina_net_presets.py
```diff
@@ -0,0 +1,34 @@
...
+"""RetinaNet Task presets."""
+from keras_cv.models.backbones.resnet_v1 import resnet_v1_backbone_presets
+
+retina_net_presets = {
+    "retinanet_resnet50_pascalvoc": {
+        "metadata": {
+            "description": (
+                "ResNet classifier with 50 layers where the batch "
+                "normalization and ReLU activation precede the convolution "
+                "layers (v2 style). Trained on Imagenet 2012 classification "
```
jbischof, line 23: You're using the v1 resnet and the task was trained on Pascal VOC. Please update the description

keras_cv/models/object_detection/retina_net/retina_net_presets.py
```diff
@@ -0,0 +1,34 @@
...
+                "task."
+            ),
+        },
+        "config": {
```
jbischof, line 27: Why isn't `bounding_box_format` part of the config? I guess a user can pass a different one than used in training b/c it's all converted to the same internal representation?

LukeWood:
Thanks @fchollet, @jbischof let me know if anything else needs review or updating!

After the last update, I did check to make sure this doesn't break the classifier by running a classifier locally.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -514,13 +518,6 @@ def compute_metrics(self, x, y, y_pred, sample_weight):
                 metrics[metric.name] = result
         return metrics
 
-    @classmethod
-    def from_config(cls, config):
-        config["backbone"] = keras.utils.deserialize_keras_object(
-            config["backbone"]
-        )
-        return super().from_config(config)
-
```
jbischof, line 523: In `get_config` should the other objects after `backbone` be serialized by default as well? We can file an issue to clean this up later if so; custom objects don't seem to be tested.

jbischof (approved):
Super exciting, thank you!

Some minor nits to address and then ship away. I updated the description to match the final API/preset choices

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: the short answer is I don't know, but this was breaking the tests and I dont think this is the proper approach. 

I think we should explicitly address from_config() and get_config() in a separate PR.

LukeWood: the short answer is I don't know, but this was breaking the tests and I dont think this is the proper approach. 

I think we should explicitly address from_config() and get_config() in a separate PR.

keras_cv/models/object_detection/retina_net/retina_net_presets.py
LukeWood: > Why isn't bounding_box_format part of the config? I guess a user can pass a different one than used in training b/c it's all converted to the same internal representation?

Correct, and they can perform prediction in any format, train on any format, etc.

There is NO "best" choice for this, so we have chosen to require an argument.  This is also a common bug surface in OD, so we are trying to mitigate against that.  I have some documentation on this in my upcoming OD guide.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -20,27 +22,34 @@
...
-class RetinaNet(keras.Model):
-    """A Keras model implementing the RetinaNet architecture.
+class RetinaNet(Task):
+    """A Keras model implementing the RetinaNet meta-architecture.
 
     Implements the RetinaNet architecture for object detection.  The constructor
-    requires `num_classes`, and `bounding_box_format`.  Optionally, a backbone,
-    custom label encoder, and prediction decoder may all be provided.
+    requires `num_classes`, `bounding_box_format`, and a backbone.  Optionally,
+    a custom label encoder, and prediction decoder may all be provided.
```
ianstenbit, line 52: nit: s/may all/may

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -50,7 +59,7 @@ class RetinaNet(keras.Model):
             [
                 [0, 0, 100, 100],
                 [100, 100, 200, 200],
-                [300, 300, 400, 400],
+                [300, 300, 100, 100],
```
ianstenbit, line 62: Nice catch!

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -98,32 +107,33 @@ class RetinaNet(keras.Model):
...
-            transforming RetinaNet predictions into usable bounding box Tensors.  If
-            not provided, a default is provided.  The default `prediction_decoder`
-            layer uses a `MultiClassNonMaxSuppression` operation for box pruning.
+            `RetinaNetLabelEncoder` is created and used.  Results of this object's
+            `call()` method are passed to the `loss` object for `box_loss` and
+            `classification_loss` the `y_true` argument.
+        prediction_decoder: (Optional)  A `keras.layers.Layer` that is responsible
+            for transforming RetinaNet predictions into usable bounding box
+            Tensors.  If not provided, a default is provided.  The default
+            `prediction_decoder` layer is a
```
ianstenbit, line 116: Maybe not in this PR, but I think it's worth considering whether this should be renamed to something other than `decoder`.

Since decoding the deltas to boxes happens in-model, the name is now a bit confusing to me.

ianstenbit (approved):
LGTM -- should we add a SmokeTest which verifies numerics for a given input against the pre-trained preset with weights?

keras_cv/models/object_detection/retina_net/retina_net_presets.py
```diff
@@ -0,0 +1,36 @@
...
+            "description": (
+                "RetinaNet with a ResNet50 v1 backbone. "
+                "Trained on PascalVOC 2012 object detection task, which "
+                "consists of 20 classes.  This model achieves a final MaP of "
+                "0.33 on the evaluation set."
+            ),
+        },
+        "config": {
+            "backbone": resnet_v1_backbone_presets.backbone_presets["resnet50"],
+            # 21 used as an implicit background class marginally improves
```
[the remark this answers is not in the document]
LukeWood: I'm not sure they need to know this, its kind of an implementation detail in a way.  They'll pretty much never see this class, it exists just to provide a floor for softmax basically.

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -20,27 +22,34 @@
...
+class RetinaNet(Task):
+    """A Keras model implementing the RetinaNet meta-architecture.
 
     Implements the RetinaNet architecture for object detection.  The constructor
-    requires `num_classes`, and `bounding_box_format`.  Optionally, a backbone,
-    custom label encoder, and prediction decoder may all be provided.
+    requires `num_classes`, `bounding_box_format`, and a backbone.  Optionally,
+    a custom label encoder, and prediction decoder may be provided.
 
     Usage:
```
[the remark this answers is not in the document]
LukeWood: I feel `Usage:` is a pretty standard wording in Keras:

https://github.com/keras-team/keras/search?q=%22Usage%3A%22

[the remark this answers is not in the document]
jbischof: Apologies I meant standardizing with KerasNLP. But just having a simple example in each docstring is already 90% of the way there

LukeWood:
@ianstenbit can you take a look at the new failures?  Bumping to 2.12 fixed the issue, but it is now complaining about custom OPs.  Can we build a custom-op release for 2.12?

ianstenbit:
> @ianstenbit can you take a look at the new failures? Bumping to 2.12 fixed the issue, but it is now complaining about custom OPs. Can we build a custom-op release for 2.12?

Should just need to bump the version to 2.12 here: https://github.com/keras-team/keras-cv/blob/master/.github/workflows/actions.yml#L35 (add this to #1658, and then give CI another go)

I take it you ran the CloudBuild script to update the Docker container to 2.12? Fine by me, but also I'd like to make sure that we're not unnecessarily adding a dependency on TF 2.12 if we can avoid it.

LukeWood:
> > @ianstenbit can you take a look at the new failures? Bumping to 2.12 fixed the issue, but it is now complaining about custom OPs. Can we build a custom-op release for 2.12?
> 
> Should just need to bump the version to 2.12 here: https://github.com/keras-team/keras-cv/blob/master/.github/workflows/actions.yml#L35 (add this to #1658, and then give CI another go)
> 
> I take it you ran the CloudBuild script to update the Docker container to 2.12? Fine by me, but also I'd like to make sure that we're not unnecessarily adding a dependency on TF 2.12 if we can avoid it.
> 
> Edit: what I said above actually should only affect normal CI, the error in the GCBRun CI here is caused by this:
> 
> https://github.com/keras-team/keras-cv/blob/15056344d08d234537fdf5b61d77c66c1fae691b/keras_cv/utils/resource_loader.py#L25
> 
> which should be updated to 2.12 if we're building against 2.12.
> There are a handful of other spots in the library we'll want to update to 2.12 as well (just searching 2.11 will make them all show up)

Ok so we're not actually dependent on 2.12 and we're not failing on it.  I feel we should configure these custom op tests to not outright fail on these version mismatches, any thoughts on this?  I did not realize bumping the cloudbuild version would break our CI

ianstenbit:
> Ok so we're not actually dependent on 2.12 and we're not failing on it. I feel we should configure these custom op tests to not outright fail on these version mismatches, any thoughts on this? I did not realize bumping the cloudbuild version would break our CI

I think it's a good failure. ABI incompatibility is an absolute nightmare for users and being defensive about it seems prudent. It's just a good reminder to update our entire buildchain when we have a new dep.

PR #1508, opened 2023-03-15
Migrate ResNet V1 to subclass Backbone

haifeng-jin:
Fixes #1504 

This PR migrate ResNet V1 from `keras.Model` subclass to `Backbone` subclass.
The migration steps are as follows:
1. Add the resnet_v1 files.
2. Add the license header.
3. Move the code from the old file to the new files under models/backbones
4. Add the presets
    * The presets information are summarized from the model configurations in `resnet_v1_backbone.py` and `weights.py`. The format of `resnet_v1_backbone_presets.py` follows `resnet_v2_backbone_presets.py`.
    * Add the presets to `backbone_presets.py`.
5. Change the base class of `ResNetBackbone`.
    * Rename the symbols.
    * Change the args and attributes to follow the `Backbone` style.
    * Removed `include_top`, `name`, `weights`, `pooling`, `num_classes`, `classifier_activation` from the args, attributes, and `get_config()`.
    * Change `get_config()` to call `super().get_config()`.
    * Added a `*` in the args.
    * Removed `self.as_backbone`.
    * Use `self.pyramid_level_inputs` instead of `self._backbone_level_outputs`.
6. Make all different ResNet architectures extend the `ResNetBackbone` class.
    * Change the usages of the `ResNet` classes.
7. Update the docstrings.
8. Add unit tests. Limit the batch size to 2 for all generated data.
    * Add tests for all pretrained weights for producing the same outputs as before. The notebook used to produce old outputs. ([link](https://github.com/haifeng-jin/Colabs/blob/main/KerasCV_old_ResNet50_imagenet_weights_golden_number.ipynb)). The newly added test: ([link](https://github.com/keras-team/keras-cv/pull/1508/files#diff-623290d6b13340c267c8724520d0f9f62cc4c22d491a67ed474a33961da75a05R48-R81))

PR #1651, opened 2023-03-31
Migrate VGG19 to subclass Backbone

### LINKED ISSUE #1629: Migrate VGG19 to the new `Backbone` API
To align KerasCV with the KerasNLP API, we're migrating our backbone models to subclass keras_cv.models.Backbone.
This allows us to have a more structured, modular API for high-level task models, including the new ImageClassifier model (see https://github.com/keras-team/keras-cv/pull/1458).

A great step-by-step guide to this migration can be found in the description of https://github.com/keras-team/keras-cv/pull/1508, which performed this migration for ResNetV1.

If you'd like to take this issue up, please leave a comment and the KerasCV team can assign the issue to you. When opening a PR, please tag this issue as fixed by your PR, and tag our tracking bug https://github.com/keras-team/keras-cv/issues/1627 for bookkeeping.

sabribarac:
## What does this PR do?

Fixes #1629 

## Who can review?

@LukeWood @ianstenbit


keras_cv/models/backbones/vgg19/vgg19_backbone.py
sabribarac, line 1: In vgg19_backbone,
``` python
from keras_cv.models.backbones.vgg16.vgg16_backbone import apply_vgg_block
```
depends on #1630

keras_cv/models/__init__.py
```diff
@@ -45,6 +45,7 @@
 from keras_cv.models.backbones.resnet_v2.resnet_v2_backbone import (
     ResNetV2Backbone,
 )
+from keras_cv.models.backbones.vgg19.vgg19_backbone import VGG19Backbone
```
ianstenbit, line 48: We'll also want to drop `from keras_cv.models.vgg19 import VGG19` below

keras_cv/models/backbones/vgg19/vgg19_backbone.py
```diff
@@ -0,0 +1,147 @@
...
+            kernel_size=(3, 3),
+            activation="relu",
+            padding="same",
+            max_pool=True,
+            name="block5",
+        )
+
+        super().__init__(inputs=inputs, outputs=x, **kwargs)
+
+        self.include_rescaling = include_rescaling
```
ianstenbit, line 130: In order to use this as a Backbone for Object detection, we'll also need to setup self.pyramid_level_outputs.

The pyramid_level_outputs should be a dict with keys 1...N where the values are the last intermediate outputs of the model where the size of the Tensor has been reduced by 2^N.

(So, for example, if the input is 512x512, the key 1 would point to the output of the last intermediate res block whose output is 256x256, and the key 3 would point to the output of the last intermediate res block whose output is 64x64)

I would guess that this will be somewhat straightforward for VGG, because we can create `pyramid_level_outputs = {}` before all of the `apply_vgg_block`s, and then after each block we can just say `pyramid_level_outputs[block_index] = x`. But I'm not certain.

keras_cv/models/backbones/vgg19/vgg19_backbone_presets.py
```diff
@@ -0,0 +1,36 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""VGG19 model preset configurations."""
+
+backbone_presets_no_weights = {
```
ianstenbit, line 17: This is an interesting corner case that @jbischof perhaps can weigh in on -- what do we do when we have a preset that is basically just a default for `include_rescaling` (VGG has a simpler API so presets are less meaningful).

Perhaps this should be a `VGGBackbone` and `vgg19` and `vgg16` should be presets?

keras_cv/models/backbones/vgg19/vgg19_backbone_test.py
```diff
@@ -0,0 +1,55 @@
...
+        model_output = model(self.input_batch)
+        save_path = os.path.join(self.get_temp_dir(), filename)
+        model.save(save_path, save_format=save_format)
+        restored_model = keras.models.load_model(save_path)
+
+        self.assertIsInstance(restored_model, VGG19Backbone)
+
+        restored_output = restored_model(self.input_batch)
+        self.assertAllClose(model_output, restored_output)
+######
```
ianstenbit, line 53: nit: rm

keras_cv/models/backbones/vgg19/vgg19_backbone_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def test_valid_call_applications_model(self):
+        model = VGG19Backbone()
+        model(self.input_batch)
+    
+    @parameterized.named_parameters(
+        ("tf_format", "tf", "model"),
+        ("keras_format", "keras_v3", "model.keras")
+    )
+    def test_saved_alias_model(self, save_format, filename):
```
ianstenbit, line 41: Let's also add test cases like `test_create_backbone_model_from_alias_model` and `test_model_backbone_layer_names_stability` like the ones in #1508

keras_cv/models/backbones/vgg19/vgg19_backbone_presets.py
```diff
@@ -0,0 +1,36 @@
...
+    "vgg19": {
+        "metadata": {
+            "description": (
+                "VGG19 model with 19 layers where the ReLU activation is "
```
ianstenbit, line 21: Oddly enough, the VGG19 backbone actually has 16 layers. VGG19 has 16 conv layers and 3 dense layers, and the backbone omits those dense layers.

ianstenbit (changes requested):
Thanks for the PR! We'll have to wait for the VGG16 PR to merge this one, but I've left some comments to take a look at while we wait on that one.

ianstenbit:
Closing as this PR is now stale (never had another review requested), and we've decided to drop the VGG models.

PR #1656, opened 2023-04-03
Add a 3D Box regression loss for CenterPillar

ianstenbit:
This loss is used for box regression in the CenterPillar model, and is the last component missing in KerasCV to have an end-to-end trainable 3D OD model

### DIFF
--- a/keras_cv/losses/box_3d_regression_loss.py
+++ b/keras_cv/losses/box_3d_regression_loss.py
@@ -1,0 +1,129 @@
+# Copyright 2023 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import math
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_cv.bounding_box_3d import CENTER_XYZ_DXDYDZ_PHI
+
+
+def l1(y_true, y_pred, sigma=9.0):
+    """Computes element-wise l1 loss."""
+
+    absolute_difference = tf.abs(y_pred - y_true)
+    loss = tf.where(
+        absolute_difference < 1.0 / sigma,
+        0.5 * sigma * absolute_difference**2,
+        absolute_difference - 0.5 / sigma,
+    )
+
+    return tf.reduce_sum(loss, axis=-1)
+
+
+class Box3DRegressionLoss(keras.losses.Loss):
+    """Implements a bin-based box regression loss for 3D bounding boxes.
+
+    Reference: https://arxiv.org/abs/1812.04244
+
+    3DBoxRegressionLoss uses L1 loss on the individual components of boxes, with
+    the exception of the bin-based heading component of each box, where the bin
+    indicator outputs use cross entropy loss, and the heading residual uses L1
+    loss.
+
+    Ground truth boxes are expected to follow the CENTER_XYZ_DXDYDZ_PHI format.
+    Refer to https://github.com/keras-team/keras-cv/blob/master/keras_cv/bounding_box_3d/formats.py
+    for more details on supported bounding box formats.
+
+    Box predictions are expected to be in CenterPillar heatmap-encoded format.
+
+    Args:
+        num_heading_bins: int, number of bins used for predicting box heading.
+        anchor_size: list of 3 ints, anchor sizes for the x, y, and z axes.
+    """
+
+    def __init__(self, num_heading_bins, anchor_size, **kwargs):
+        super().__init__(**kwargs)
+        self.num_heading_bins = num_heading_bins
+        self.anchor_size = anchor_size
+
+    def heading_regression_loss(self, heading_true, heading_pred):
+        # Set the heading to within 0 -> 2pi
+        heading_true = tf.math.floormod(heading_true, 2 * math.pi)
+
+        # Divide 2pi into bins. shifted by 0.5 * angle_per_class.
+        angle_per_class = (2 * math.pi) / self.num_heading_bins
+        shift_angle = tf.math.floormod(
+            heading_true + angle_per_class / 2, 2 * math.pi
+        )
+
+        heading_bin_label_float = tf.math.floordiv(shift_angle, angle_per_class)
+        heading_bin_label = tf.cast(heading_bin_label_float, dtype=tf.int32)
+        heading_res_label = shift_angle - (
+            heading_bin_label_float * angle_per_class + angle_per_class / 2.0
+        )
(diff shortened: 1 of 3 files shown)


keras_cv/losses/box_3d_regression_loss.py
```diff
@@ -0,0 +1,129 @@
...
+
+import math
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_cv.bounding_box_3d import CENTER_XYZ_DXDYDZ_PHI
+
+
+def l1(y_true, y_pred, sigma=9.0):
```
LukeWood, line 23: can we use a keras L1 loss component to do this, just for standardization? 

Consider this a nit - not a blocker.

keras_cv/losses/box_3d_regression_loss.py
```diff
@@ -0,0 +1,129 @@
...
+    loss = tf.where(
+        absolute_difference < 1.0 / sigma,
+        0.5 * sigma * absolute_difference**2,
+        absolute_difference - 0.5 / sigma,
+    )
+
+    return tf.reduce_sum(loss, axis=-1)
+
+
+class Box3DRegressionLoss(keras.losses.Loss):
```
LukeWood, line 36: do you like the name Box3D?  I feel like theres a bit of an inversion with ordering here (3D box is my expected ordering), but I understand python probably prevents the name 3DBox.  

Wonder if there's anything else we can call these components.

No strong opinion, and honestly no better name in mind, just thought I'd mention it in case you have a better anme.

keras_cv/losses/box_3d_regression_loss.py
```diff
@@ -0,0 +1,129 @@
...
+    """Implements a bin-based box regression loss for 3D bounding boxes.
+
+    Reference: https://arxiv.org/abs/1812.04244
+
+    Box3DRegressionLoss uses L1 loss on the individual components of boxes, with
```
LukeWood, line 41: are these delta encoded?  if so, good to clarify this.

keras_cv/losses/box_3d_regression_loss.py
```diff
@@ -0,0 +1,129 @@
...
+                * heading_bin_one_hot,
+                axis=-1,
+                keepdims=True,
+            ),
+            heading_res_norm_label[..., tf.newaxis],
+        )
+
+        return loss_heading_bin + loss_heading_res
+
+    def regression_loss(self, y_true, y_pred):
```
LukeWood, line 98: I feel like this class is doing 2~ things at once:

- unpackaging formats
- computing loss on these formats.

I feel something analogous to this is in the RetinaNet's compile method.  Instead of having a generic `BoxRegressionLoss`, we take 2 losses to compile(), `box_loss`, `classification_loss`.  Is the reason that this is packaged together due to inter-dependencies?

Perhaps it can be refactored to something like:

```
Box3DLoss(heading_loss='...', box_loss='...',)
```

Or we can attach some of the logic to the CenterPillar model instead.

Thoughts?  Am I missing some context as to why this wouldn't work, or why it is optimal to keep it all in one place?

keras_cv/losses/box_3d_regression_loss.py
```diff
@@ -0,0 +1,129 @@
...
+    """Computes element-wise l1 loss."""
+
+    absolute_difference = tf.abs(y_pred - y_true)
+    loss = tf.where(
+        absolute_difference < 1.0 / sigma,
+        0.5 * sigma * absolute_difference**2,
+        absolute_difference - 0.5 / sigma,
+    )
+
+    return tf.reduce_sum(loss, axis=-1)
```
LukeWood, line 33: can you just remove this and then remove the `tf.new_axis` used below (along with `reduce_sum(keep_axis=True)`?

LukeWood (changes requested):
Did a first pass, my first instinct is that this is handling 2-3 distinct responsibilities in one component.  Perhaps we can break it down into more modular building blocks.

(I mentioned this in a comment too, just to echo I've copy pasted that comment below):

For clarity/to echo the comment:

I feel like this class is doing 2~ things at once:

unpackaging formats
computing loss on these formats.
I feel something analogous to this is in the RetinaNet's compile method. Instead of having a generic BoxRegressionLoss, we take 2 losses to compile(), box_loss, classification_loss. Is the reason that this is packaged together due to inter-dependencies?

Perhaps it can be refactored to something like:

Box3DLoss(heading_loss='...', box_loss='...',)
Or we can attach some of the logic to the CenterPillar model instead.

Thoughts? Am I missing some context as to why this wouldn't work, or why it is optimal to keep it all in one place?

keras_cv/losses/box_3d_regression_loss.py
ianstenbit: Yeah, I wanted to call it `3DBoxRegressionLoss`, but Python doesn't like starting with the `3`

keras_cv/losses/box_3d_regression_loss.py
ianstenbit: This is a bit of a complex issue -- I agree we should think about it.

CenterNet will already take two losses:
- Heatmap loss
- Box regression loss

IMO it makes sense for this to be one loss, because fundamentally the whole thing represents a regression loss for 3D boxes. The reason that the headings are handled separately is simply that the headings are predicted using bins instead of a single regression target, which makes the code more complex.

Happy to chat about this more. There's also an additional concept that is sometimes added to the box loss (which we'll want to support) which is a 3D IoU component. In that case, it may make sense to include this as a separate loss function.

keras_cv/losses/box_3d_regression_loss.py
ianstenbit: Keras doesn't offer an L1 loss component, only a regularizer (which is a different function)

keras_cv/losses/box_3d_regression_loss.py
ianstenbit: Updated to clarify the encoding a bit

keras_cv/losses/box_3d_regression_loss.py
ianstenbit: The `l1` function is called from a few different spots, and this would only work for the heading residual, not the location and size components

ianstenbit:
With respect to doing 2 things, I'm not sure it makes any sense to split this up.

The only purpose of decoding the predictions (which we only do a little bit of here) is to correctly compute loss, and it's not relevant to other parts of the model. This loss is all conceptually 1 thing, which is regressing against ground truth boxes, whereas e.g. Heatmap loss in CenterNet is a separate concept (and is thus handled separately in the model)

ianstenbit:
@LukeWood -- renamed this to make it more over that it's really only useful for CenterPillar, and I also updated the docstring to reflect this.

keras_cv/losses/box_3d_regression_loss.py
LukeWood: (As discussed in 1:1 let’s call it CenterPillarLoss)

LukeWood:
> @LukeWood -- renamed this to make it more over that it's really only useful for CenterPillar, and I also updated the docstring to reflect this.

Cool, will do another pass

LukeWood (approved):

LukeWood:
Thanks @ianstenbit - with the rename and all now it is clear how it is used and it all looks logical to me.

PR #1658, opened 2023-04-04
Bump TF version in GCB image

LukeWood:
Justification: Seeing some extremely strange behavior in our GCBRun environment with the `keras.Model` class.

See failures in https://github.com/keras-team/keras-cv/pull/1618

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,9 +3,9 @@
 	"build": {
 		"dockerfile": "Dockerfile",
 		"args": {
-			"VERSION": "2.11.0"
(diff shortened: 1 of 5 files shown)


jbischof (approved):

ianstenbit (approved):
Bumping our CI + custom op buildchain to 2.12 LGTM, even if we don't have a 2.12 dependency I think it makes sense to use the latest version for testing.

LGTM as long as the Docker container is also running 2.12.

LukeWood:
(Closing as there are issues w/ the Windows tests and Jon has been doing some refactoring)

PR #1663, opened 2023-04-04
Task model take 2

LukeWood:
Same as https://github.com/keras-team/keras-cv/pull/1618 but with functional subclassing.


LukeWood:
Ok so it looks like functional subclassing works.  Let me finish training a model, then we should merge.

keras_cv/models/object_detection/retina_net/retina_net_test.py
```diff
@@ -199,3 +195,21 @@ def test_serialization(self):
         )
         serialized_2 = keras.utils.serialize_keras_object(restored)
         self.assertEqual(serialized_1, serialized_2)
+
+
+@pytest.mark.large
+class RetinaNetSmokeTest(tf.test.TestCase):
+    def test_backbone_preset_weight_loading(self):
+        # Check that backbone preset weights loaded correctly
+        _ = keras_cv.models.RetinaNet.from_preset(
```
jbischof, line 204: Don't assign the input at all if unused

jbischof (commented):
Thanks just some small comments!

keras_cv/models/object_detection/retina_net/retina_net.py
```diff
@@ -196,22 +223,43 @@ def __init__(
...
+        self.box_head = box_head
+        self.build(backbone.input_shape)
 
-        self.classification_head = (
-            classification_head
-            or layers_lib.PredictionHead(
-                output_filters=9 * num_classes,
-                bias_initializer=prior_probability,
+    @staticmethod
+    def build_forward(
```
ianstenbit, line 232: This feels a bit hacky to me, but maybe it's unavoidable?
If we do decide we want to go with this approach, I think this at least warrants a comment linking to a bug in Keras core describing the issue we're working around by doing this.

wdyt @LukeWood ?

ianstenbit (changes requested):

keras_cv/models/object_detection/retina_net/retina_net.py
jbischof: Why is it unavoidable? Sounds like I'm missing something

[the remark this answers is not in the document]
LukeWood: Interesting, I personally really dislike doing it in the constructor.  To me I feel it mixes up too many responsibilities/makes the constructor super chaotic.  Thoughts?

keras_cv/models/object_detection/retina_net/retina_net_test.py
```diff
@@ -199,3 +195,21 @@ def test_serialization(self):
...
+            "resnet50_v2_imagenet",
+            num_classes=20,
+            bounding_box_format="xywh",
+        )
+
+    def test_full_preset_weight_loading(self):
```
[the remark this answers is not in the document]
LukeWood: Let’s wait until I have some weights I’m happy with locked in.

keras_cv/models/object_detection/retina_net/retina_net.py
LukeWood: Which part do you find hacky?  The functional subclassing?  The build_forwards method?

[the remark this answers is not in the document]
jbischof: I think @fchollet's take would be that separating this out adds indirection without any payoff. We have inlined the functional model declaration across KerasNLP ([example](https://github.com/keras-team/keras-nlp/blob/0d7e4a91b94ce211085884f78fadaaeb6a8ef817/keras_nlp/models/bert/bert_backbone.py#L108)).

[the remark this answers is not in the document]
LukeWood: I see.  Not personally sold, but happy to compromise on this (at least in the short term to preserve consistency).  Will update this!

ianstenbit: The part that I find hacky is that we have this separate method for building forward. The part that I meant was unavoidable was defining the forward pass in a method for construction time, but Jonathan points out that we can just do this in the constructor a la Functional model style.

That option SGTM (and bonus points that apparently NLP does that too!)

LukeWood: Seems we have consensus for doing it in line in the constructor.  Will update and make sure weights still load.

LukeWood: Fixed

LukeWood:
Ok @jbischof and @ianstenbit, all comments addressed!

keras_cv/models/object_detection/retina_net/retina_net_test.py
```diff
@@ -199,3 +195,25 @@ def test_serialization(self):
...
+        # Check that backbone preset weights loaded correctly
+        # TODO(lukewood): need to forward pass test once proper weights are
+        # implemented
+        _ = keras_cv.models.RetinaNet.from_preset(
```
jbischof, line 216: Remove output assignment if unused

jbischof (approved):
Looks great! Some nits then ship away.

ianstenbit (approved):
thanks Luke!

LukeWood:
😀

Issue #1659, opened 2023-04-04
Remove the `training` argument from `.call()` function of all preprocessing layers


haifeng-jin:
According to the discussion with @LukeWood and @fchollet , we will remove the `training` argument from all `.call()` functions of all preprocessing layers in KerasCV.

The reason is to easily build the preprocessing pipeline into a `Sequential` model and run it with `tf.data.Dataset.map()`.
This preprocessing pipeline is separate from the main neural network model, thus do not receive the `training` argument when calling `model.fit()` in the main neural network model. Therefore, managing this `training` argument in the preprocessing pipeline is an extra burden on the users.

After removing the `training` argument, all preprocessing layers are expected only to perform training time behaviors all the time. The user should construct separate pipelines for training and inferencing.

Following are the preprocessing layers in KerasCV that overrides the `call()` function, which needs to be changed in addressing this issue.

`keras_cv/layers/preprocessing/`
- [ ] randomly_zoomed_crop.py
- [ ] mosaic.py
- [ ] base_image_augmentation_layer.py
- [ ] random_crop_and_resize.py
- [ ] vectorized_base_image_augmentation_layer.py
- [ ] resizing.py
- [x] jittered_resize.py

`keras_cv/layers/preprocessing_3d/`
- [ ] group_points_by_bounding_boxes.py
- [ ] base_augmentation_layer_3d.py
- [ ] random_copy_paste.py

haifeng-jin:
I only included the layers in `keras_cv/layers/preprocessing/` and `keras_cv/layers/preprocessing_3d/`.

@LukeWood Would you help confirm the following?
1. The description above is accurate.
2. There is no preprocessing layers in other directories.
3. These preprocessing layers are extended from `BaseRandomLayer`, which is not a subclass of KPL. That doesn't matter.

Thanks.

LukeWood:
Hey @haifeng-jin !  This looks correct to me - I think everything here is accurate.

> The reason is to easily build the preprocessing pipeline into a Sequential model and run it with tf.data.Dataset.map().
This preprocessing pipeline is separate from the main neural network model, thus do not receive the training argument when calling model.fit() in the main neural network model. Therefore, managing this training argument in the preprocessing pipeline is an extra burden on the users.

[...]

PR #1672, opened 2023-04-05
Remove trinaing from Resize in the KPL

soma2000-lang:
#1659

 @LukeWood and @ianstenbit @haifeng-jin


keras_cv/layers/preprocessing/resizing.py
```diff
@@ -273,22 +273,6 @@ def resize_with_crop_to_aspect(x):
         inputs["images"] = images
         return inputs
 
-    def call(self, inputs, training=True):
```
LukeWood, line 276: I don't see any references to `training` here!  What behavior are we removing here?

PR #1681, opened 2023-04-06
Export ResNet backbone

jbischof:
Let's make sure to export the generic backbone class in all PRs following #1508


ianstenbit (approved):
Thanks!

Issue #1683, opened 2023-04-06
Avoid NaNs in RetinaNet training by adding an epsilon to `normalizer`


ianstenbit:
If a training row has no boxes, the `normalizer` is 0, causing division by 0 and thus NaNs.

For YOLOv8 I'm just using an epsilon factor for now.

@LukeWood

LukeWood:
Surprised this wasn't problematic in the RetinaNet, thanks for the fix.  I will add it to RetinaNet too

PR #1684, opened 2023-04-07
Add epsilon for NaN prevention + test case to RetinaNet

LukeWood:
fixes https://github.com/keras-team/keras-cv/issues/1683

I actually took a `tf.cond()` based approach here to zero out sample weights instead.  I do this because I found that when we divide by normalizer, and normalizer is zero, the background classes all get absolutely massive values for class weight - causing massive gradients.

An alternative is perhaps to use a small value, i.e. 0.3~ for the class weights?  I think the ideal mask is something like: `tf.where(positive_mask, 0.1, 0.0)` where 0.1 is any tiny value.   Though, using all zeros is probably safer and easier.

### DIFF
--- a/keras_cv/models/object_detection/retinanet/retinanet.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet.py
@@ -416,7 +416,7 @@ class RetinaNet(Task):
         )
 
         positive_mask = tf.cast(tf.greater(classes, -1.0), dtype=tf.float32)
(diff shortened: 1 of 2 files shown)


ianstenbit (approved):
Awesome -- thank you Luke!

I'll try this same approach for YOLOV8

LukeWood:
Looks like we need to upgrade tensorflow

PR #1686, opened 2023-04-07
Fix bounding box visualization tool to draw boxes inside of the box when OOB

LukeWood:
Also adds confidence scores to the plotting utility

Some edge case testing:

Normal font size:
Figure_2

Massive font size:

Figure_1

### DIFF
--- a/examples/visualization/plot_bounding_box_gallery.py
+++ b/examples/visualization/plot_bounding_box_gallery.py
@@ -36,12 +36,16 @@ def unpackage_tfds_inputs(inputs):
         target="xywh",
     )
     classes = tf.cast(inputs["objects"]["label"], tf.float32)
-    bounding_boxes = {"classes": classes, "boxes": boxes}
+    bounding_boxes = {
+        "classes": classes,
+        "confidence": tf.random.uniform(tf.shape(classes), minval=0, maxval=1),
+        "boxes": boxes,
+    }
     return image, bounding_boxes
 
 
 train_ds = train_ds.map(unpackage_tfds_inputs)
-train_ds = train_ds.apply(tf.data.experimental.dense_to_ragged_batch(16))
+train_ds = train_ds.ragged_batch(16)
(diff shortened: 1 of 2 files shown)


examples/visualization/plot_bounding_box_gallery.py
```diff
@@ -80,30 +84,29 @@ def unpackage_tfds_inputs(inputs):
     images,
     value_range=(0, 255),
     bounding_box_format="xywh",
-    y_true=boxes,
-    scale=3,
+    y_pred=boxes,
```
ianstenbit, line 87: why not `y_true` here anymore?

LukeWood: I was just trying to show off the confidence scores in a way that would actually be used/logical

LukeWood:
@ianstenbit reverted!

keras_cv/visualization/draw_bounding_boxes.py
```diff
@@ -117,11 +128,26 @@ def draw_bounding_boxes(
...
                     cv2.FONT_HERSHEY_SIMPLEX,
                     font_scale,
                     color,
                     text_thickness,
                 )
         result.append(image)
     return np.array(result).astype(int)
+
+
+def _find_text_location(x, y, font_scale, line_thickness, outline_factor):
```
ianstenbit, line 141: Just to sanity check -- have you tested this on a few potential edge cases? I can't totally parse the logic, but the demos look good.

ianstenbit (approved):
lgtm -- one question below

keras_cv/visualization/draw_bounding_boxes.py
LukeWood: I have, I tried fiddling with thicknesses and font scales and outline factors quite a bit.  I've failed to find a failure case, but if there is one we can submit new cases to handle it.

jbischof:
Nice, thank you @LukeWood!!!

PR #1690, opened 2023-04-09
Updates weights for RetinaNet

LukeWood:
Much better scores- 0.36 MaP.

This was achieved by NOT freezing batchnorm, and using a slower LR schedule with more epochs.

From https://github.com/keras-team/keras-cv/pull/1691


jbischof (approved):
Still not 100% on the `.weights` story. Otherwise, great thanks!

ianstenbit (approved):

PR #1691, opened 2023-04-09
Update RetinaNet training script

LukeWood:
Scores 0.36~ MaP now.

### DIFF
--- a/examples/training/object_detection/pascal_voc/retina_net.py
+++ b/examples/training/object_detection/pascal_voc/retina_net.py
@@ -38,7 +38,7 @@ resource.setrlimit(resource.RLIMIT_NOFILE, (high, high))
 
 flags.DEFINE_integer(
     "epochs",
-    35,
+    100,
     "Number of epochs to run for.",
 )
 
@@ -68,7 +68,7 @@ except ValueError:
 
 BATCH_SIZE = 4
 GLOBAL_BATCH_SIZE = BATCH_SIZE * strategy.num_replicas_in_sync
-BASE_LR = 0.01 * GLOBAL_BATCH_SIZE / 16
+BASE_LR = 0.005 * GLOBAL_BATCH_SIZE / 16
 print("Number of accelerators: ", strategy.num_replicas_in_sync)
 print("Global Batch Size: ", GLOBAL_BATCH_SIZE)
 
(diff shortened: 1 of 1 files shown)


examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -187,10 +183,6 @@ def pad_fn(inputs):
     bounding_box_format="xywh", confidence_threshold=0.5, from_logits=True
 )
 
-for layer in model.feature_extractor.layers:
```
jbischof, line 190: Another option here would be to add a `--freeze_batch_norm` flag

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -29,22 +29,18 @@
...
-# a 1:1 comparison with the PyMetrics version.
-from keras_cv.callbacks import PyCOCOCallback
-
 low, high = resource.getrlimit(resource.RLIMIT_NOFILE)
 resource.setrlimit(resource.RLIMIT_NOFILE, (high, high))
 
 flags.DEFINE_integer(
     "epochs",
-    35,
+    100,
```
jbischof, line 38: I'm shocked going over the same dataset so many times could be productive without an early stopping callback

examples/training/object_detection/pascal_voc/retina_net.py
```diff
@@ -127,6 +124,24 @@ def unpackage_tfds_inputs(inputs, bounding_box_format):
         ),
     ]
 )
+
```
jbischof, line 127: Not crucial, but it seems like you could easily package all these transforms into one augmenter instead of doing many maps

jbischof (approved):
Very cool, thank you!

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: I do have early stopping :)

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: This is verifiably worse, I've ruin it like 10-15 times with and without randaugment, etc

examples/training/object_detection/pascal_voc/retina_net.py
LukeWood: you would hope so, but that is not the case.  RandAugment doesn't work on dictionary inputs at this point.

PR #1197, opened 2022-12-23
[Refactor RetinaNet] Step 9 unify NMSDecoder.

tanzhenyu:
1. This PR significantly reduces the logic for NMS decode.
2. It allows user to more easily fork and customize their own NMS decoder.
3. It unifies FasterRCNN and RetinaNet decoding logic.

PR #1697, opened 2023-04-10
Revert accidental change to logits activation

LukeWood:
At some point while we were trying to create a single decoder to use for MaskRCNN + RetinaNet the activation for the prediction decoder was changed to be softmax instead of sigmoid.  This has a few undesirable properties, but the most notable is that at times the model will make a prediction for a box with a logit of 0.000001 if the rest of the logits are all 0 or negative

This also removes the need for the extra class in RetinaNet.  We can just use len(class_ids) now.


jbischof (approved):
Could this be related to the high confidence false positives in Martin's bug dataset?

ianstenbit (approved):
Looks like we missed this in #1197

Nice catch Luke!

PR #1698, opened 2023-04-10
Support multiple box losses for CenterPillar (one-per-class)

ianstenbit:
For CenterPillar, we may need separate box losses for separate classes.

This updates the API to support this.


LukeWood (changes requested):
can we add a docstring?

LukeWood (approved):
LGTM thank you Ian

PR #1699, opened 2023-04-10
Fix RetinaNet serialization for Keras 2.13+

ianstenbit:
#1663 accidentally removed `from_config` from the RetinaNet, which didn't break our CI but does break deserialization in TF/Keras 2.13, which broke our sync to G3.

I re-introduced it here.

### DIFF
--- a/keras_cv/models/object_detection/retina_net/retina_net.py
+++ b/keras_cv/models/object_detection/retina_net/retina_net.py
@@ -534,9 +534,6 @@ class RetinaNet(Task):
         config["backbone"] = keras.utils.deserialize_keras_object(
             config["backbone"]
         )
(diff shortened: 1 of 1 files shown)


LukeWood (approved):

LukeWood:
> > Looks great, but shouldn't this be handled by the `Task` base class? Also, you probably want to match the logic in the base class, unless that is somehow causing the error.
> > https://github.com/keras-team/keras-cv/blob/c142683e55b05e21b6f5b302105e45f133c77190/keras_cv/models/task.py#L50
> 
> I didn't realize that this existed in the `Task` class. This means explicit deserialization here should be unnecessary, so I'll close this PR.
> 
> @LukeWood were you ever able to track down the root cause of the serialization issue in G3?

not yet :(

PR #1700, opened 2023-04-10
Update the preset to now work with the new sigmoid activation in the decoder

LukeWood:


ianstenbit:
Nice! I verified that these weights are public on GCS.

Does the SmokeTest work with this change? Or is that breakage totally unrelated?

LukeWood:
It should work!

ianstenbit (approved):

PR #1701, opened 2023-04-10
Update 0.4.3 after RetinaNet changes + finalization of OD API

LukeWood:


jbischof:
I'm nervous about this for two reasons:
* We don't have much documentation for the new API
* Many of our models have not been migrated to `Task` or `Backbone` but they are still exported

Wouldn't a release cause a lot of confusion at this time?

PR #1296, opened 2023-01-20
[YOLOX] Step 2/? : Setting up YoloX structure and add internal layers

quantumalaviya:
# What does this PR do?
This PR is one of many upcoming PRs that will attempt to break down https://github.com/keras-team/keras-cv/pull/982 so that it's easier to review. This one deals with the internal layers.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/keras-team/keras-cv/blob/master/.github/CONTRIBUTING.md),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue? Please add a link
      to it if that's the case.
- [ ] Did you write any new necessary tests?
- [ ] If this adds a new model, can you run a few training steps on TPU in Colab to ensure that no XLA incompatible OP are used?

## Who can review?
@LukeWood 
Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #1707, opened 2023-04-13
Remove a redundant check from IoULoss

ianstenbit:
This was introduced in #1296 and is a duplicate of the check below which is more exhaustive.


haifeng-jin (approved):
Thanks for the PR! LGTM

PR #1708, opened 2023-04-13
Allow RetinaNet to use exactly 1 class

ianstenbit:
This is no longer a relevant check after #1697


jbischof:
Should we add testing for the 1 class case?

ianstenbit:
> Should we add testing for the 1 class case?

I'm not really sure what the value would be. The 1 class case isn't really a special case, it's just that softmax doesn't behave nicely with only 1 input. I'm happy to add one though if there's a reason I'm missing.

jbischof (approved):
OK, just curious if there was some test that would have caught the bug. If not LGTM

LukeWood:
Cool w/ me - should be no issue now

PR #1714, opened 2023-04-17
fix failed tests in resnet_v1 presets

haifeng-jin:
The old tests breaks in TF 2.12. This PR fixes it in both 2.11 and 2.12.

### DIFF
--- a/keras_cv/models/backbones/resnet_v1/resnet_v1_backbone_presets_test.py
+++ b/keras_cv/models/backbones/resnet_v1/resnet_v1_backbone_presets_test.py
@@ -37,10 +37,6 @@ class ResNetPresetSmokeTest(tf.test.TestCase, parameterized.TestCase):
     def setUp(self):
         self.input_batch = tf.ones(shape=(2, 224, 224, 3))
 
-    @parameterized.named_parameters(
-        ("preset_with_weights", "resnet50_imagenet"),
-        # ("preset_no_weights", "resnet50"),
-    )
     def test_backbone_output(self):
         model = ResNetBackbone.from_preset("resnet50")
         model(self.input_batch)
@@ -63,10 +59,9 @@ class ResNetPresetSmokeTest(tf.test.TestCase, parameterized.TestCase):
         # We should only update these numbers if we are updating a weights
         # file, or have found a discrepancy with the upstream source.
 
-        outputs = feature_extractor(tf.ones(1, 512, 512, 3))
+        outputs = feature_extractor(tf.ones((1, 512, 512, 3)))
         outputs = [outputs[i][0, 0, 0, :5] for i in extractor_levels]
         expected = [
-            [0.50249904, 0.35751498, 0.9474212, 1.0659311, 1.1105202],
             [0.65718395, 0.7209194, 0.39707005, 0.5164382, 0.73338735],
             [
                 1.0615100e00,
--- a/keras_cv/models/backbones/resnet_v1/resnet_v1_backbone_presets_test.py
+++ b/keras_cv/models/backbones/resnet_v1/resnet_v1_backbone_presets_test.py
@@ -15,7 +15,6 @@
 
 import pytest
 import tensorflow as tf
-from absl.testing import parameterized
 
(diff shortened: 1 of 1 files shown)


keras_cv/models/backbones/resnet_v1/resnet_v1_backbone_presets_test.py
```diff
@@ -63,10 +59,9 @@ def test_backbone_output_with_weights(self):
         # We should only update these numbers if we are updating a weights
         # file, or have found a discrepancy with the upstream source.
 
-        outputs = feature_extractor(tf.ones(1, 512, 512, 3))
+        outputs = feature_extractor(tf.ones((1, 512, 512, 3)))
         outputs = [outputs[i][0, 0, 0, :5] for i in extractor_levels]
```
jbischof, line 63: I think this test deviated from the resnet_v2 version by mistake. That is doing what we want ([link](https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/backbones/resnet_v2/resnet_v2_backbone_presets_test.py#L29))

haifeng-jin: This test is more comprehensive since it tests all feature pyramid outputs. There is no other difference from the resnet v2 tests.

jbischof: I'm not sure that's necessary honestly. What would it catch that a forward pass would not?

Just don't want to add unnecessary overhead for our model development.

haifeng-jin: @LukeWood may provide some comments here. It is requested in [this comment](https://github.com/keras-team/keras-cv/pull/1508#discussion_r1152394503). I am fine with either way.

jbischof: This test would be better as a RetinaNet forward pass test rather than requiring every backbone test to write out 15 floats.

keras_cv/models/backbones/resnet_v1/resnet_v1_backbone_presets_test.py
```diff
@@ -37,10 +37,6 @@ class ResNetPresetSmokeTest(tf.test.TestCase, parameterized.TestCase):
     def setUp(self):
         self.input_batch = tf.ones(shape=(2, 224, 224, 3))
 
-    @parameterized.named_parameters(
```
ianstenbit, line 40: It looks like after this change, there's no meaningful usage of `parameterized`, so we can probably remove it as a superclass of these tests and drop the import

keras_cv/models/backbones/resnet_v1/resnet_v1_backbone_presets_test.py
```diff
@@ -63,10 +59,9 @@ def test_backbone_output_with_weights(self):
         # We should only update these numbers if we are updating a weights
         # file, or have found a discrepancy with the upstream source.
 
-        outputs = feature_extractor(tf.ones(1, 512, 512, 3))
```
ianstenbit, line 66: This fix seems correct to me -- as is, this test should fail because it's an invalid call to `tf.ones`

keras_cv/models/backbones/resnet_v1/resnet_v1_backbone_presets_test.py
ianstenbit: I'm fine either way here -- but as-is this test is broken due to this call to `tf.ones`, so it's probably worth fixing.

ianstenbit (approved):
Thanks Haifeng!

keras_cv/models/backbones/resnet_v1/resnet_v1_backbone_presets_test.py
haifeng-jin: Done. Changed to only testing the final output without feature extractor.

LukeWood (approved):
this looks correct to me as well!

PR #1671, opened 2023-04-05
Updated Docstring into smaller chunks, waiting for review and feedback In center_pillar.py

NiharJani2002:
# What does this PR do?
Update of The Docstring In center_pillar.py 
Divided The Docstring into small chunks.

## Who can review?
@ianstenbit 
@LukeWood 
@tanzhenyu 

## This PR template is modified from here:
Updated Docstring of center_pillar.py #1535

PR #1717, opened 2023-04-18
Docstring update of center_pillar.py, now it contains class descripti…

NiharJani2002:
# What does this PR do?
Docstring Update of center_pillar.py now contains class information, parameters, an explanation of __init__ and call functions for each class, and along with that, it includes a description of computing loss.

Focus was given to following a universal rule of 80, but some required more explanation; to provide more proof, the rule has been unfollowed for those particular cases.

Fixes # (issue)
#1671 

# Who can review this Pr.
Who can review:
@ianstenbit 
@jbischof


NiharJani2002:
@ianstenbit 

Any reviews further ?

ianstenbit:
thanks for the PR!

These updated docs add a lot of redundant information that is somewhat implied by method names / the Keras ecosystem (e.g. explaining what `compute_loss` does). If there are isolated smaller errors or improvements to the docstrings that you'd like to contribute, feel free to send a smaller PR, but we'd like to keep the high-level docs structure as-is.

PR #1723, opened 2023-04-19
Migrate EfficientNetv2 to Backbone + migrate to ImageClassifier

LukeWood:
co-authored-by [soma2000-lang](https://github.com/soma2000-lang)


keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,551 @@
...
+                        activation=activation,
+                        bn_momentum=bn_momentum,
+                        survival_probability=drop_connect_rate * b / blocks,
+                        name="block{}{}_".format(i + 1, chr(j + 97)),
+                    ),
+                }[args["conv_type"]]
+
+                x = block(x)
+                b += 1
+
```
ianstenbit, line 247: `pyramid_level_inputs[i+2] = x.node.layer.name`

LukeWood: doesn't work because some blocks don't downsample so the i index doesn't match up with pyramid level idnices

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,569 @@
...
+def round_repeats(repeats, depth_coefficient):
+    """Round number of repeats based on depth multiplier."""
+    return int(math.ceil(depth_coefficient * repeats))
+
+
+@keras.utils.register_keras_serializable(package="keras_cv.models")
+class EfficientNetV2Backbone(Backbone):
+    """Instantiates the EfficientNetV2 architecture using given scaling
+    coefficients.
```
jbischof, line 108: newline below

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,569 @@
...
+
+
+def round_repeats(repeats, depth_coefficient):
+    """Round number of repeats based on depth multiplier."""
+    return int(math.ceil(depth_coefficient * repeats))
+
+
+@keras.utils.register_keras_serializable(package="keras_cv.models")
+class EfficientNetV2Backbone(Backbone):
+    """Instantiates the EfficientNetV2 architecture using given scaling
```
jbischof, line 107: Description should fit on a single line

keras_cv/models/backbones/efficientnet_v2/block_args.py
```diff
@@ -0,0 +1,474 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The EfficientNet authors carefully tuned the block parameters with a
+# hyperparameter search for each sub-architecture.
+# This file contains constants to reproduce these architectures.
+
+DEFAULT_BLOCKS_ARGS = {
```
jbischof, line 19: These are preset args and should be moved to the presets file

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,569 @@
...
+)
+from keras_cv.models.backbones.efficientnet_v2.efficientnet_v2_backbone_presets import (  # noqa: E501
+    backbone_presets,
+)
+from keras_cv.models.backbones.efficientnet_v2.efficientnet_v2_backbone_presets import (  # noqa: E501
+    backbone_presets_with_weights,
+)
+from keras_cv.utils.python_utils import classproperty
+
+CONV_KERNEL_INITIALIZER = {
```
jbischof, line 43: Let's make these inlined classes rather than serialized objects. See [example](https://github.com/keras-team/keras-nlp/blob/edd6b6ba58afde819e7b07d4a0aa47f15a8a996b/keras_nlp/models/bert/bert_backbone.py#L30)

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,569 @@
...
+            to `True`, inputs will be passed through a `Rescaling(1/255.0)`
+            layer.
+        width_coefficient: float, scaling coefficient for network width.
+        depth_coefficient: float, scaling coefficient for network depth.
+        default_size: integer, default input image size.
+        dropout_rate: float, dropout rate before final classifier layer.
+        drop_connect_rate: float, dropout rate at skip connections.
+        depth_divisor: integer, a unit of network width.
+        min_depth: integer, minimum number of filters.
+        bn_momentum: float. Momentum parameter for Batch Normalization layers.
```
jbischof, line 120: We haven't been exposing this for other models. Drop if there's not a good reason to expose here

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets.py
```diff
@@ -0,0 +1,284 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""EfficientNetV2 model preset configurations."""
+
+DESCRIPTION = "One of the many EfficientNetV2 variants.  Each variant is built "
```
jbischof, line 17: This description is not helpful for a preset. Let's drop

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets.py
```diff
@@ -0,0 +1,284 @@
...
+"publication.  To learn more about the parameterizations and their tradeoffs, "
+"please check keras.io.  As a starting point, we recommend starting with the "
+'"efficientnetv2-s" architecture, and increasing in size to "efficientnetv2-m" '
+'or "efficientnetv2-l" if resources permit.'
+
+backbone_presets_no_weights = {
+    "efficientnetv2-s": {
+        "model_name": "efficientnetv2-s",
+        "metadata": {
+            "description": DESCRIPTION,
```
jbischof, line 28: Describe what makes this preset unique vs the others.

keras_cv/models/classification/image_classifier_test.py
```diff
@@ -109,6 +109,43 @@ class ImageClassifierPresetSmokeTest(tf.test.TestCase, parameterized.TestCase):
     def setUp(self):
         self.input_batch = tf.ones(shape=(2, 224, 224, 3))
 
+    @parameterized.named_parameters(
```
jbischof, line 112: Unless there's a good reason, we don't need to test the output of each possible preset. Just one should be enough to catch changes to the forward pass. I would drop

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,569 @@
...
+                    args["input_filters"] = args["output_filters"]
+
+                if args["strides"] != 1:
+                    pyramid_level_inputs[
+                        i+1
+                    ] = x.node.layer.name
+                    pyramid_level_tracker += 1
+
+                # Determine which conv type to use:
+                block = None
```
jbischof, line 227: Why can't this be checked before the loop if it only relies on the config?

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,569 @@
...
+    "config": {
+        "scale": 1.0 / 3.0,
+        "mode": "fan_out",
+        "distribution": "uniform",
+    },
+}
+
+BN_AXIS = 3
+
+BASE_DOCSTRING = """Instantiates the {name} architecture.
```
jbischof, line 63: Move down to the start of the alias class definitions

jbischof (changes requested):
Thanks for the PR @LukeWood! 

My first reaction is that this doesn't look much like the other backbones we have migrated like ResNet, CSPDarknet and the (almost finished) MobileNet. Let's try to unify the style as much as possible unless there's a good reason.

keras_cv/models/classification/image_classifier_test.py
LukeWood: Keeping until we LGTM, just to confirm that we don't accidentally shift architecture in code review.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: Done

LukeWood:
Ok @jbischof - updated and rebased!

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,556 @@
...
+        width_coefficient,
+        depth_coefficient,
+        default_size,
+        dropout_rate=0.2,
+        drop_connect_rate=0.2,
+        depth_divisor=8,
+        min_depth=8,
+        activation="swish",
+        blocks_args="default",
+        model_name="efficientnet",
```
jbischof, line 137: This arg appears unused. Let's remove.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,556 @@
...
+        include_rescaling,
+        width_coefficient,
+        depth_coefficient,
+        default_size,
+        dropout_rate=0.2,
+        drop_connect_rate=0.2,
+        depth_divisor=8,
+        min_depth=8,
+        activation="swish",
+        blocks_args="default",
```
jbischof, line 136: A couple weird things here:
* Passing configs is an antipattern in Keras. We need to find a way around this
* This is named `blocks_args` here and `block_args` above
* The default value is `"default"`, which actually throws in the helper method.
* There is no description of legal values in the docstring.

jbischof: I would recommend the pattern we developed in MobileNet instead where the block args are passed as parallel lists. See #1641

jbischof (commented):
Much better! We need a different strategy for the block configurations however

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,556 @@
...
+        drop_connect_rate: float, dropout rate at skip connections.
+        depth_divisor: integer, a unit of network width.
+        min_depth: integer, minimum number of filters.
+        activation: activation function.
+        blocks_args: list of dicts, parameters to construct block modules.
+        model_name: string, model name.
+        input_shape: optional shape tuple, defaults to (None, None, 3).
+        input_tensor: optional Keras tensor (i.e. output of `layers.Input()`)
+            to use as image input for the model.
+    """
```
jbischof, line 122: An example call at the end would be very valuable here.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: > This is named blocks_args here and block_args above
> the default value is "default", which actually throws in the helper method.
> There is no description of legal values in the docstring.

Good points; these are definitely not correct - I'll update.

> Passing configs is an antipattern in Keras. We need to find a way around this... I would recommend the pattern we developed in MobileNet instead where the block args are passed as parallel lists. See https://github.com/keras-team/keras-cv/pull/1641

I'm not super sold on that - especially given that block types here may not even match!  In this case the configs are identical, but you can certainly imagine a scenario where theres an arg that  I.e. MbConvBlock vs FusedMBConvBlock require different arguments.

I feel we're still passing a config if its parallel lists... its just split up and harder to pick out which values go with which layers.  

Thoughts?  Personally I definitely prefer a list of dictionaries, but I suppose there is added noise comes in keys with this approach.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: Added

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: the configs also render very ugly for long lists of ints: the presets in https://github.com/keras-team/keras-cv/pull/1641

Perhaps fixable with some more `black` configuration.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,562 @@
...
+            kernel_size=1,
+            strides=1,
+            kernel_initializer=conv_kernel_initializer(),
+            padding="same",
+            data_format="channels_last",
+            use_bias=False,
+            name="top_conv",
+        )(x)
+        x = layers.BatchNormalization(
+            axis=BN_AXIS,
```
ianstenbit, line 253: I think in both cases in this file, `BN_AXIS` (3) is equivalent to the default axis of `-1`.
It's worth verifying but I think we can just get rid of this.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,562 @@
...
+
+def round_repeats(repeats, depth_coefficient):
+    """Round number of repeats based on depth multiplier."""
+    return int(math.ceil(depth_coefficient * repeats))
+
+
+def get_block_conv(args, activation, survival_probability, name):
+    # Determine which conv type to use:
+    if args["conv_type"] == "mb_conv":
+        return MBConvBlock(
```
ianstenbit, line 69: It seems like here we could say `conv_block = MBConvBlock`, and in the `elif` we could say `conv_block = FusedMBConvBlock` and then we can just call it once, since the calls are identical for the two.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,562 @@
...
+            name="stem_bn",
+        )(x)
+        x = layers.Activation(activation, name="stem_activation")(x)
+
+        # Build blocks
+        block_args = copy.deepcopy(block_args)
+        b = 0
+        blocks = float(sum(args["num_repeat"] for args in block_args))
+
+        pyramid_level_tracker = 1
```
ianstenbit, line 190: Instead of keeping this tracker, why don't we just make pyramid_level_outputs a list and then convert it to a dict at the end?

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,562 @@
...
+        return copy.deepcopy(backbone_presets_with_weights)
+
+
+BASE_DOCSTRING = """Instantiates the {name} architecture.
+
+    Reference:
+    - [EfficientNetV2: Smaller Models and Faster Training](https://arxiv.org/abs/2104.00298)
+      (ICML 2021)
+
+    This function returns a Keras image classification model.
```
ianstenbit, line 318: This is no longer true (same for the line(s) below)

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets.py
```diff
@@ -0,0 +1,778 @@
...
+        },
+    },
+    "efficientnetv2-b0": {
+        "metadata": {
+            "description": "The EfficientNet B0 architecture.  In this "
+            "variant of the EfficientNet architecture, there are "
+            f"{len(B0_BLOCK_ARGS)} convolutional blocks. As with all of the B "
+            "style EfficientNet variants, the number of filters in each "
+            "convolutional block is scaled by a `width_coefficient` and a "
+            "`depth_coefficient`.  Please see the GitHub source to find the "
```
ianstenbit, line 552: Why not just write the coefficients here? (same throughout the file)

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets.py
```diff
@@ -0,0 +1,778 @@
...
+}
+
+backbone_presets_with_weights = {
+    "efficientnetv2-s_imagenet": {
+        "metadata": {
+            "description": {
+                "description": "The EfficientNet small architecture.  In this "
+                "variant of the EfficientNet architecture, there are "
+                f"{len(SMALL_BLOCK_ARGS)} convolutional blocks. Weights are "
+                "initialized to pretrained imagenet classification weights."
```
ianstenbit, line 662: Do we want to state the scores of these pre-trained weights?

ianstenbit: (We do for OD models, probably should here too I think)

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets_test.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2023 The KerasCV Authors
```
ianstenbit, line 1: We should also have a separate `efficientnet_v2_backbone_test`, as well as one test case here which asserts the numerical results of a forward pass with a preset.

ianstenbit (changes requested):

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets_test.py
LukeWood: thats done in the ImageClassifierTest.  Do you think thats sufficient?

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: I feel this is a case where you reduce the lines of code, but actually reduce readability.

I really don't like assigning class constructors to variables then invoking them.

Open to further opinions, just my own and dont feel super strongly.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: Good idea -updated

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: Works!

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: Good point - done.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
jbischof: Here's why I don't think this is a productive direction @LukeWood:
* You have a `block_args` param that is 100% undocumented. What fields need to be populated? What type of values are expected? This is a bad user experience.
* The most obvious way to document the params is to create an official "config" using `dataclass` or a proto
* You have just created a nested config system, which will be vetoed by @fchollet 

I do agree that the parallel lists in #1641 aren't perfect but each block param is well documented and it's easy to trace how each param evolves across the block. What we have here doesn't look like Keras code or other backbones. If we encounter a more complex case we can have a bigger design discussion, but I haven't seen one so far.

jbischof (commented):
Left note on https://github.com/keras-team/keras-cv/pull/1723#discussion_r1180873094.

I don't think nested configs are the right design choice for our library.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: Yep yep - sorry I mean't to comment back on this after our 1:1 discussion but I  couldnt find the thread.  Definitely updating to match the others

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets.py
LukeWood: Great point - updated

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets.py
LukeWood: Done!  Thanks

LukeWood:
> Left note on [#1723 (comment)](https://github.com/keras-team/keras-cv/pull/1723#discussion_r1180873094).
> 
> I don't think nested configs are the right design choice for our library.

Should officially be good to go.  Let me know if you have any more comments on the approach here.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,610 @@
...
+    raise ValueError(
+        "Expected `block_args['conv_type']` to be "
+        "one of 'mb_conv', 'fused_mb_conv', but got "
+        f"`block_args['conv_type']={args['conv_type']}`"
+    )
+
+
+@keras.utils.register_keras_serializable(package="keras_cv.models")
+class EfficientNetV2Backbone(Backbone):
+    """Instantiates the EfficientNetV2 architecture.
```
jbischof, line 100: This docstring does not appear updated

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,610 @@
...
+    )
+    images = tf.ones((1, 256, 256, 3))
+    outputs = efficientnet.predict(images)
+    ```
+    """
+
+    def __init__(
+        self,
+        *,
+        include_rescaling,
```
jbischof, line 136: Can we rename all list parameters `stackwise_{name}` to match our current backbones? E.g., `stackwise_se_ratios`, `stackwise_strides` etc

jbischof (changes requested):
I think we can remove the `block_args` entirely without issue to match the other backbone migrations.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: Great point; will update this.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,610 @@
...
+        dropout_rate=0.2,
+        drop_connect_rate=0.2,
+        depth_divisor=8,
+        min_depth=8,
+        activation="swish",
+        input_shape=(None, None, 3),
+        input_tensor=None,
+        **kwargs,
+    ):
+        block_args = [
```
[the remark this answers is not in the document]
LukeWood: I'm using these for input validation only.  We can just drop it I suppose.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: I went ahead and refactored the names quite a bit across the board.  They should be much more readable now - and make it more obvious what is singular and what is plural.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: Updated the arguments section!

LukeWood:
> I think we can remove the `block_args` entirely without issue to match the other backbone migrations.

Good point - Removed all of them.  We lose little bit of input argument validation, but I think it should be fine overall.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_test.py
```diff
@@ -0,0 +1,230 @@
...
+)
+from keras_cv.utils.train import get_feature_extractor
+
+
+class EfficientNetV2BackboneTest(tf.test.TestCase, parameterized.TestCase):
+    def setUp(self):
+        self.input_batch = tf.ones(shape=(8, 224, 224, 3))
+
+    def test_valid_call(self):
+        model = EfficientNetV2Backbone(
```
ID6109, line 35: Can we just load from a Preset here instead?

Referring to @ianstenbit 's [comment](https://github.com/keras-team/keras-cv/pull/1641/commits/699ff06fd421642e56e885c7eee0fcf7c84f3f2f#r1171871160) on #1641

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_test.py
```diff
@@ -0,0 +1,230 @@
...
+        self.assertEquals(list(outputs.keys()), [3, 4])
+        self.assertEquals(outputs[3].shape, [None, 32, 32, 64])
+        self.assertEquals(outputs[4].shape, [None, 16, 16, 160])
+
+    @parameterized.named_parameters(
+        ("one_channel", 1),
+        ("four_channels", 4),
+    )
+    def test_application_variable_input_channels(self, num_channels):
+        # ResNet50 model
```
ID6109, line 203: Needs to be addressed

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_test.py
LukeWood: we can; but this way we get lots of coverage

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,587 @@
...
+        stackwise_squeeze_and_excite_ratios: list of ints, the squeeze and
+            excite ratios passed to the squeeze and excitation blocks.
+        stackwise_strides: list of ints, stackwise_strides for each conv block.
+        stackwise_conv_types: list of strings.  Each value is either 'mb_conv'
+            or 'fused_mb_conv' depending on the desired blocks.
+        dropout_rate: float, dropout rate before final classifier layer.
+        drop_connect_rate: float, dropout rate at skip connections.
+        depth_divisor: integer, a unit of network width.
+        min_depth: integer, minimum number of filters.
+        activation: activation function.
```
jbischof, line 126: Activation for what? Let's add some color.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,587 @@
...
+                "input_shape": input_shape,
+                "input_tensor": input_tensor,
+            }
+        )
+        return EfficientNetV2Backbone.from_preset("efficientnetv2-s", **kwargs)
+
+    @classproperty
+    def presets(cls):
+        """Dictionary of preset names and configurations."""
+        return {}
```
jbischof, line 384: Remember to add all relevant presets. We have been adding the presets with weights applicable to this architecture (e.g., `"efficientnetv2-s_imagenet"` here)

keras_cv/models/classification/image_classifier_presets.py
```diff
@@ -13,8 +13,13 @@
 # limitations under the License.
 """ImageClassifier Task presets."""
 
+from keras_cv.models.backbones.efficientnet_v2 import (
+    efficientnet_v2_backbone_presets,
+)
 from keras_cv.models.backbones.resnet_v2 import resnet_v2_backbone_presets
 
+EFFICIENTNET_DESCRIPTION = "TODO"
```
jbischof, line 21: Please do!

keras_cv/models/classification/image_classifier_test.py
jbischof: OK, remember to remove.

jbischof (changes requested):
@LukeWood we're missing a lot of attention to detail here, I tried to catch what I could but please give this a careful once-over before trying to push.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: Yep yep - added

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
```diff
@@ -0,0 +1,587 @@
...
+                    args["input_filters"] = args["output_filters"]
+
+                if args["strides"] != 1:
+                    pyramid_level_inputs.append(x.node.layer.name)
+
+                block = get_block_conv(
+                    **args,
+                    activation=activation,
+                    survival_probability=drop_connect_rate * b / blocks,
+                    name="block{}{}_".format(i + 1, chr(j + 97)),
```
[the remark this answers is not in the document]
LukeWood: This came from the original implementation.  block_{number}{letter} basically to have dual indexing.

1a, 1b, 1c, 1d... etc

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone.py
LukeWood: I was under the impression these were intentionally empty due to https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/backbones/resnet_v2/resnet_v2_backbone.py#L441

I see the others are populated; will populate.

LukeWood: Done

LukeWood:
I reorganized the files and I think they should be much easier to review now.  Primarily following https://github.com/keras-team/keras-cv/pull/1745

LukeWood:
ok @ianstenbit @jbischof  I think that the comments should be addressed (excluding chr(j+97), not sure what is the optimal replacement for that...).  I did another pass, reorganized the code, and think it looks pretty reasonable.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets.py
```diff
@@ -0,0 +1,478 @@
...
+                "unfused",
+                "unfused",
+                "unfused",
+                "unfused",
+            ],
+        },
+    },
+    "efficientnetv2-b0": {
+        "metadata": {
+            "description": "The EfficientNet B0 architecture.  In this "
```
jbischof, line 150: Nit: Shorten to 
```
"EfficientNet B-style architecture with six convolutional blocks. This B-style model has `width_coefficient=1.0` and `depth_coefficient=1.0`."
``` 

This has to fit in a small table on keras.io. The scaling is clearly different for each B-style preset for there's no "all" that I can see.

keras_cv/models/classification/image_classifier_test.py
jbischof: It's still here tho?

jbischof (approved):
Thanks @LukeWood, down to nits now.

The biggest one is that our model descriptions need to be short to fit in a table of presets on keras.io. See https://keras.io/api/keras_nlp/models/ for example.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_aliases.py
ianstenbit, line 1: (Maybe I missed this in other comments)
Why the separate file for this?

keras_cv/models/classification/image_classifier_presets.py
```diff
@@ -39,4 +42,117 @@
         "weights_url": "https://storage.googleapis.com/keras-cv/models/resnet50v2/imagenet-classifier-v0.h5",  # noqa: E501
         "weights_hash": "77fa9f1cd1de0e202309e51d4e598e441d1111dacb6c41a182b6c63f76ff26cd",  # noqa: E501
     },
+    "efficientnetv2-s_imagenet_classifier": {
```
ianstenbit, line 45: It seems like in the above resnet preset, we use all underscores, but here we have a dash.

How about we do all underscores? (Throughout all the presets in this PR)

ianstenbit (approved):
LGTM, just a few little things

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_aliases.py
LukeWood: I made this change because I couldn't realistically review the entire efficientnet_v2_backbone file in one sitting.  I think putting them all into one file is a bit of an anti-pattern - as they're pretty independent pieces and it makes the backbone files needlessly long.

keras_cv/models/classification/image_classifier_test.py
LukeWood: I kept one `b` variant and one normal variant.  They're slightly logically distinct so I intentionally kept one of each.  Before there were 4.

keras_cv/models/classification/image_classifier_presets.py
LukeWood: done

keras_cv/models/classification/image_classifier_test.py
LukeWood: actually - just the B variant should cover everything

PR #1731, opened 2023-04-24
Update the KerasCV bounding box output experience

LukeWood:
After some piloting and user feedback it is clear that the following are true:
- ragged outputs are less confusing than receiving a tensor of all -1s
- NMS wasn't filtering confidence, this is just a bad/inconsistent UX

I propose we update the behavior to output ragged boxes by default, as well as filter confidence scores by class ID.  This will overall improve consistency and UX.

### DIFF
--- a/keras_cv/bounding_box/mask_invalid_detections.py
+++ b/keras_cv/bounding_box/mask_invalid_detections.py
@@ -61,9 +61,10 @@ def mask_invalid_detections(bounding_boxes, output_ragged=False):
 
     boxes = bounding_boxes.get("boxes")
     classes = bounding_boxes.get("classes")
+    confidence = bounding_boxes.get("confidence", None)
     num_detections = bounding_boxes.get("num_detections")
 
-    # Create a mask to select only the first N boxes from each batch
+    # Create a mask to selxect only the first N boxes from each batch
     mask = tf.repeat(
         tf.expand_dims(tf.range(tf.shape(boxes)[1]), axis=0),
         repeats=tf.shape(boxes)[0],
@@ -81,6 +82,10 @@ def mask_invalid_detections(bounding_boxes, output_ragged=False):
     result = bounding_boxes.copy()
     result["boxes"] = boxes
     result["classes"] = classes
+
+    if confidence:
+        confidence = tf.where(mask, confidence, -tf.ones_like(confidence))
+        result["confidence"] = confidence
 
     if output_ragged:
         return to_ragged(result)
--- a/keras_cv/bounding_box/mask_invalid_detections.py
+++ b/keras_cv/bounding_box/mask_invalid_detections.py
@@ -74,17 +74,19 @@ def mask_invalid_detections(bounding_boxes, output_ragged=False):
 
     classes = tf.where(mask, classes, -tf.ones_like(classes))
 
+    if confidence is not None:
+        confidence = tf.where(mask, confidence, -tf.ones_like(confidence))
+
     # reuse mask for boxes
     mask = tf.expand_dims(mask, axis=-1)
     mask = tf.repeat(mask, repeats=boxes.shape[-1], axis=-1)
     boxes = tf.where(mask, boxes, -tf.ones_like(boxes))
 
     result = bounding_boxes.copy()
+
     result["boxes"] = boxes
     result["classes"] = classes
-
-    if confidence:
-        confidence = tf.where(mask, confidence, -tf.ones_like(confidence))
+    if confidence is not None:
         result["confidence"] = confidence
 
     if output_ragged:
--- a/keras_cv/bounding_box/mask_invalid_detections.py
+++ b/keras_cv/bounding_box/mask_invalid_detections.py
@@ -64,7 +64,7 @@ def mask_invalid_detections(bounding_boxes, output_ragged=False):
     confidence = bounding_boxes.get("confidence", None)
     num_detections = bounding_boxes.get("num_detections")
 
-    # Create a mask to selxect only the first N boxes from each batch
+    # Create a mask to select only the first N boxes from each batch
     mask = tf.repeat(
         tf.expand_dims(tf.range(tf.shape(boxes)[1]), axis=0),
         repeats=tf.shape(boxes)[0],
--- a/keras_cv/bounding_box/to_ragged_test.py
+++ b/keras_cv/bounding_box/to_ragged_test.py
@@ -23,13 +23,27 @@ class ToRaggedTest(tf.test.TestCase):
                 [[[0, 0, 0, 0], [0, 0, 0, 0]], [[2, 3, 4, 5], [0, 1, 2, 3]]]
             ),
             "classes": tf.constant([[-1, -1], [-1, 1]]),
+            "confidence": tf.constant([[0.5, 0.7], [0.23, 0.12]]),
         }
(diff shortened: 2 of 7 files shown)


jbischof:
I might lack sufficient context without a demo here. Under which circumstances are `-1`s returned?

LukeWood:
> I might lack sufficient context without a demo here. Under which circumstances are `-1`s returned?

In order to convert a ragged dimension to Dense (for TPU compat, etc) you pad -1s on to all values that are to be ignored.  Consider two images:

image 1 has:
boxes: [[0, 0, 0.5 0.5], [0.25, 0.25, .1, .1]]
classes: [0, 1]

image 2 has:
boxes: [[0, 0, 0.5 0.35]]
classes: [0]

To make these a TPU compatible batch, you will need to make these Dense tensors before passing them to `train_step()`.  This can be done by calling:

`keras_cv.bounding_box.to_dense(max_boxes=2)`, which would pad the following boxes on to make it a dense tensor:

boxes: [[[0, 0, 0.5 0.5], [0.25, 0.25, .1, .1]], [[0, 0, 0.5 0.35], [-1, -1, -1, -1]]]
classes: [[0, 1], [0, -1]]

Notice the final box is all -1s.   Our components simply know to ignore these values, and you can re-filter them out using `keras_cv.bounding_box.to_ragged(boxes)`.  In this PR, I'm simply filtering padding boxes.

keras_cv/layers/object_detection/multi_class_non_max_suppression.py
```diff
@@ -48,6 +50,7 @@ def __init__(
         confidence_threshold=0.5,
         max_detections=100,
         max_detections_per_class=100,
+        output_ragged=True,
```
ianstenbit, line 53: IIRC we had discussed this API in the past and decided it wasn't worthwhile -- is there a reason this can't always be true?

Conversely, can it always be false? It seems like since we provide a `to_ragged` construct, we can direct users to that if they want to prune results from a model?

keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
```diff
@@ -35,6 +35,7 @@ def decode_predictions_output_shapes():
         bounding_box_format="xyxy",
         from_logits=True,
         max_detections=100,
+        output_ragged=False,
```
ianstenbit, line 38: If we do decide to add this API flexibility, we should test both cases.

keras_cv/bounding_box/mask_invalid_detections.py
```diff
@@ -61,6 +61,7 @@ def mask_invalid_detections(bounding_boxes, output_ragged=False):
 
     boxes = bounding_boxes.get("boxes")
     classes = bounding_boxes.get("classes")
+    confidence = bounding_boxes.get("confidence", None)
```
ianstenbit, line 64: Was it just a bug that this was missing before? Or is it needed as a consequence of the change to the multi-class NMS?

I can't quite compute how it's related to the rest of the PR

ianstenbit (changes requested):

keras_cv/layers/object_detection/multi_class_non_max_suppression.py
LukeWood: We can just make it true.

LukeWood: > Conversely, can it always be false? It seems like since we provide a to_ragged construct, we can direct users to that if they want to prune results from a model?

This is more just confusing to users - we have seen this in the issues already.  "Why does my model output all -1s". Its a lot clearer if it outputs no boxes that it just didn't make any detections.

keras_cv/layers/object_detection/multi_class_non_max_suppression_test.py
LukeWood: Removed dense case - I feel its better to only have ragged, and we can add dense if a use case arises.

keras_cv/bounding_box/mask_invalid_detections.py
LukeWood: Its a bug, but  I had to fix it to get the PYCOCOCallback to operate correctly (it needs to match the other two)

keras_cv/bounding_box/mask_invalid_detections.py
```diff
@@ -73,14 +74,20 @@ def mask_invalid_detections(bounding_boxes, output_ragged=False):
 
     classes = tf.where(mask, classes, -tf.ones_like(classes))
 
+    if confidence is not None:
```
jbischof, line 77: Wait, so confidence filtering was just broken?

keras_cv/layers/object_detection/multi_class_non_max_suppression.py
```diff
@@ -107,7 +107,9 @@ def call(self, box_prediction, class_prediction):
             "num_detections": valid_det,
         }
         # this is required to comply with KerasCV bounding box format.
-        return bounding_box.mask_invalid_detections(bounding_boxes)
+        return bounding_box.mask_invalid_detections(
+            bounding_boxes, output_ragged=True
```
jbischof, line 111: Could we expose `output_ragged` as a param?

ianstenbit: Luke had previously done this, but I left a comment about it. (Sorry for the churn!)

Basically my take is that it's sorta a weird config option to have. In general, parameters that look like `output_xyz` should be instead handled as post-processing by the caller IMO.

Feel free to disagree!

jbischof: OK, fine by me

keras_cv/bounding_box/mask_invalid_detections.py
LukeWood: Not the filtering but converting confidence tensors you ragged.  Basically confidence score still had -1s padded on - even in ragged mode

keras_cv/bounding_box/to_ragged_test.py
```diff
@@ -23,13 +23,27 @@ def test_converts_to_ragged(self):
                 [[[0, 0, 0, 0], [0, 0, 0, 0]], [[2, 3, 4, 5], [0, 1, 2, 3]]]
             ),
             "classes": tf.constant([[-1, -1], [-1, 1]]),
+            "confidence": tf.constant([[0.5, 0.7], [0.23, 0.12]]),
```
jbischof, line 26: How is there a confidence if no class is detected? Confused

Also, why is the one positive class label have the lowest confidence.

jbischof (approved):
Idea is good! I'm a bit less clear on the details but maybe it's not important to dive in here

keras_cv/bounding_box/to_ragged_test.py
LukeWood: Any box with a -1 class is considered padding.  These values could be anything - they’re irrelevant.

LukeWood: (One case this could happen is if there is a high IOU between a box with a high confidence, and a box with EVEN higher confidence.  The high confidence box will be pruned, yielding a class ID of -1, but a high confidence.  Filtering out using to_ragged() removes these values which is best for UX

PR #1737, opened 2023-04-28
Some cleanup of the 3D OD API

ianstenbit:
While writing the guide, I found these three issues which are fixed in this PR:
- `label_encoder` is a parameter to the CenterPillar model even though it's not used (encoding should happen outside the model)
- Loss scaling was incorrect -- (we need to divide by num_boxes)
- The metrics callback uses a bad default configuration that causes lots of unnecessary results


LukeWood (approved):

LukeWood:
Looks good; thanks!

PR #1740, opened 2023-05-01
Remove unneeded layer stability tests

jbischof:
These tests were carried over from the old API, but no longer make sense now that we've removed closure-style submodels.


ianstenbit (approved):
LGTM -- we can also drop these from the `RetinaNet` tests I think

LukeWood (approved):

LukeWood:
Thanks!

PR #1745, opened 2023-05-03
Reorganize MobileNetV3 Directory to be more browsable.

LukeWood:
# What does this PR do?

I have a few minor complaints with our current backbone structure:

- the backbone files are massive
- finding the actual backbone class can be a bit annoying
- the aliases clog up the file but add very little information to code browsers.

In this PR I propose a solution to this.  This is two fold:

1.) move aliases to their own file:

self explanatory - seems strictly better from my perspective unless I'm missing something.

2.) move helpers below model declarations

This makes it super easy to find the actual modeling logic; which is what most of us reading this code care about!  If you want to answer "What is a MobileNetV3" its a lot easier to start reading the model than to start from the building blocks.  All tests pass; and there should be no logic delta due to this PR.  Only code browse-ability improvements.

---

If this PR looks good lets do this to the other models too.

### DIFF
--- a/keras_cv/models/__init__.py
+++ b/keras_cv/models/__init__.py
@@ -30,14 +30,14 @@ from keras_cv.models.backbones.csp_darknet.csp_darknet_backbone import (
 from keras_cv.models.backbones.csp_darknet.csp_darknet_backbone import (
     CSPDarkNetXLBackbone,
 )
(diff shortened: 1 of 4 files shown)


jbischof (approved):
@LukeWood I don't mind these changes, but could you file issues on the other migrated backbones to get everything in a consistent format?

jbischof:
Closing since now has a gross merge conflict. Will refile as a community contrib

PR #1746, opened 2023-05-03
Update quickstart for v0.5

jbischof:
* Showcase unified KerasCV/KerasNLP API
* Remove `Augmenter`
* Add `jbischof` as author

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,7 @@ pip install git+https://github.com/keras-team/keras-cv.git tensorflow --upgrade
 ```python
 import tensorflow as tf
 from tensorflow import keras
-from keras_cv.layers import RandomFlip, RandAugment, CutMix, Mixup
-from keras_cv.models import ResNet50V2Backbone, ImageClassifier
+import keras_cv
 import tensorflow_datasets as tfds
 
 # Create a preprocessing pipeline with augmentations
@@ -61,10 +60,9 @@ BATCH_SIZE = 16
 NUM_CLASSES = 3
 augmenter = keras.Sequential(
     [
-        RandomFlip(),
-        RandAugment(value_range=(0, 255)),
-        CutMix(),
-        MixUp()
+        keras_cv.layers.RandomFlip(),
+        keras_cv.layers.RandAugment(value_range=(0, 255)),
+        keras_cv.layers.CutMix(),
     ]
 )
 
@@ -88,10 +86,12 @@ test_dataset = test_dataset.batch(BATCH_SIZE).map(
         tf.data.AUTOTUNE)
 
 # Create a model using a pretrained backbone
-backbone = ResNet50V2Backbone.from_preset("resnet50_v2_imagenet")
-model = ImageClassifier(
+backbone = keras_cv.models.ResNetV2Backbone.from_preset(
+    "resnet50_v2_imagenet"
+)
+model = keras_cv.models.ImageClassifier(
     backbone=backbone,
-    num_classes=3,
+    num_classes=NUM_CLASSES,
     activation="softmax",
 )
 model.compile(
@@ -99,9 +99,6 @@ model.compile(
     optimizer=keras.optimizers.Adam(learning_rate=1e-5),
     metrics=['accuracy']
 )
-
-# Freeze backbone to avoid overfitting to small dataset
-model.backbone.trainable = False
 
 # Train your model
 model.fit(
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ To learn more about the future project direction, please check the [roadmap](.gi
 - [Call for Contributions](https://github.com/keras-team/keras-cv/issues?q=is%3Aopen+is%3Aissue+label%3Acontribution-welcome)
 - [Roadmap](.github/ROADMAP.md)
 - [API Design Guidelines](.github/API_DESIGN.md)
+- [Developer Guides](https://keras.io/guides/keras_cv/)
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,55 +50,65 @@ pip install git+https://github.com/keras-team/keras-cv.git tensorflow --upgrade
 ## Quickstart
 
 ```python
-import keras_cv
 import tensorflow as tf
 from tensorflow import keras
+from keras_cv.layers import RandomFlip, RandAugment, CutMix, Mixup
```
LukeWood, line 55: maybe from keras_cv import layers as cv_layers?  Typically we don't really like to prescribe Symbol imports as it hurts trace-ability a bit.

README.md
```diff
@@ -49,55 +50,65 @@ pip install git+https://github.com/keras-team/keras-cv.git tensorflow --upgrade
...
     preprocess_data, num_parallel_calls=tf.data.AUTOTUNE).prefetch(
         tf.data.AUTOTUNE)
 
-# Create a model
-densenet = keras_cv.models.DenseNet121(
-    include_rescaling=True,
-    include_top=True,
-    num_classes=3
+# Create a model using a pretrained backbone
+backbone = ResNet50V2Backbone.from_preset("resnet50_v2_imagenet")
```
LukeWood, line 91: Maybe EfficientNetV2B0Backbone is the most impressive.   Many fewer parameters; super efficient while maintaining strong power.

LukeWood (commented):
Minor comments but LGTM overall

README.md
jbischof: I don't think that's merged yet.

README.md
```diff
@@ -49,55 +50,65 @@ pip install git+https://github.com/keras-team/keras-cv.git tensorflow --upgrade
...
-    layers=[
-        keras_cv.layers.RandomFlip(),
-        keras_cv.layers.RandAugment(value_range=(0, 255)),
-        keras_cv.layers.CutMix(),
-        keras_cv.layers.MixUp()
+    [
+        RandomFlip(),
+        RandAugment(value_range=(0, 255)),
+        CutMix(),
+        MixUp()
```
innat, line 67: Feels a bit sceptical to use these two heavy transformation sequentially. I wonder if there's any literature explore this area. Otherwise what if showcase the random_choice layer on this two, no?

```python
layers.RandomChoice(
   layers=[CutMix(), MixUp(), FourierMix()]
)
```

ianstenbit (approved):
LGTM -- thanks Jonathan!

README.md
LukeWood: Figure we may as well put our best foot forward and show people we're not just fiddling with outdated components; WRT rock paper scissors EfficientNetV2B0 is much lighter weight than ResNet50V2: 

EfficientNetV2B0: 7_200_312 params
ResNet50V2: 25_613_800 params

Just a thought; we can just change it in a follow up PR too as its a one liner - no need to block these updates.

LukeWood (approved):

README.md
```diff
@@ -49,55 +50,65 @@ pip install git+https://github.com/keras-team/keras-cv.git tensorflow --upgrade
...
         tf.data.AUTOTUNE)
 
-# Create a model
-densenet = keras_cv.models.DenseNet121(
-    include_rescaling=True,
-    include_top=True,
-    num_classes=3
+# Create a model using a pretrained backbone
+backbone = ResNet50V2Backbone.from_preset("resnet50_v2_imagenet")
+model = ImageClassifier(
```
innat, line 92: As this read-me file is the first thing a visitor will make impression on keras-cv, it might be needed to consider adding advanced modeling, i.e. object detection (2d, 3d), segmentation. Maybe adding individual readme to corresponding [modeling directories](https://github.com/keras-team/keras-cv/tree/master/keras_cv/models); just a thoughts.

README.md
```diff
@@ -49,55 +50,65 @@ pip install git+https://github.com/keras-team/keras-cv.git tensorflow --upgrade
...
-densenet.compile(
+model.compile(
     loss='categorical_crossentropy',
-    optimizer='adam',
+    optimizer=keras.optimizers.Adam(learning_rate=1e-5),
     metrics=['accuracy']
 )
 
+# Freeze backbone to avoid overfitting to small dataset
+model.backbone.trainable = False
```
innat, line 104: Doesn't it require to set `model.compile` right after changing the trainable attributes?

README.md
jbischof: Eventually we will have a "getting started" guide with a tour of the library, but for now I added a link to our developer guides above

README.md
jbischof: Worked just fine in my colab (looking at `model.summary()`). Will add a gist once I respond to some comments

README.md
jbischof: Great we're just waiting for #1749 @LukeWood

README.md
innat: From the [doc](https://keras.io/guides/transfer_learning/)

image

The `model.summary` might have possible bugs, relevant [ticket](https://github.com/keras-team/keras/issues/18092). Mentioning the author of the above doc to give comment on it, cc. @fchollet

Frightera: @innat I guess bug happens when you try to freeze the whole model, when there are nested models they behave like a layer so `model.backbone.trainable` may not be an issue (not tested).

innat: @Frightera 
Yes, I think so. 

Now, if we don't recompile the model after changing its trainable attributes, what would happen in real? How it would affect the rest of the code? Does the model summary report something useful? 

[similar [feedback](https://github.com/keras-team/keras-io/pull/1314#discussion_r1164088972) cc. @martin-gorner]

Frightera: @innat Model summary should show something useful if you freeze and unfreeze layers like this:
```python
for layer in model.layers[0:10]:
    layer.trainable = True
```
But I am not sure if they are tracked properly without calling `compile()`. Needs further testing I guess.

jbischof: Thanks this is a great catch @innat! The model does not train very well with a frozen backbone actually 🥶 🥶 🥶

innat: > The model does not train very well with a frozen backbone actually 

I think it's a different thing from what I was concerned about. I wondered about 

```python
# using!
model = keras.Sequentail([backbone, neck, head])
model.compile( ... )
model.backbone.trainable = Fasle

# or should it be this way!
model = keras.Sequentail([backbone, neck, head])
model.backbone.trainable = Fasle
model.compile( ... )
```

jbischof: No, we are agreeing @innat! My previous runs had your first setup and the backbone was actually training. When I switched to the second the model behaved totally differently and I verified that the backbone was actually frozen. So
* only the second approach works
* a frozen backbone doesn't seem to work well (or would need to some fiddling to work well) in my quickstart

PR #1749, opened 2023-05-04
Add batchwise parameter to RandomChoice

LukeWood:
# What does this PR do?

Adds a batchwise parameter to RandomChoice for use with `CutMix`, `MixUp`, etc

Usage:

```
pipeline = layers.RandomChoice(
    layers=[layers.CutMix(), layers.MixUp()], batchwise=True
)
xs = {
    "images": tf.random.uniform((4, 5, 5, 3), 0, 100, dtype=tf.float32),
    "labels": tf.random.uniform((4, 10), 0, 1, dtype=tf.float32),
}
pipeline(xs)
```

### DIFF
--- a/keras_cv/layers/preprocessing/random_choice.py
+++ b/keras_cv/layers/preprocessing/random_choice.py
@@ -53,10 +53,10 @@ class RandomChoice(BaseImageAugmentationLayer):
             boost, but can only be used if all the layers provided to the
             `layers` argument support auto vectorization.
         batchwise: (Optional) bool, whether to pass entire batches to the
-            underlying layer. When set to true, only a single random sample is
-            drawn to determine if the batch should be passed to the underlying
-            layer. This is useful when using `MixUp()`, `CutMix()`, `Mosaic()`,
-            etc.
+            underlying layer. When set to `True`, each batch is passed to a
+            single layer, instead of each sample to an independent layer. This
+            is useful when using `MixUp()`, `CutMix()`, `Mosaic()`, etc.
+            Defaults to `False`.
         seed: Integer. Used to create a random seed.
     """
 
--- a/keras_cv/layers/preprocessing/random_choice.py
+++ b/keras_cv/layers/preprocessing/random_choice.py
@@ -52,7 +52,7 @@ class RandomChoice(BaseImageAugmentationLayer):
             apply the augmentations. This offers a significant performance
             boost, but can only be used if all the layers provided to the
             `layers` argument support auto vectorization.
-        batchwise: (Optional) bool, whether to pass entire batches to the
+        batchwise: Boolean, whether to pass entire batches to the
             underlying layer. When set to `True`, each batch is passed to a
             single layer, instead of each sample to an independent layer. This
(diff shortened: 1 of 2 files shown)


keras_cv/layers/preprocessing/random_choice_test.py
```diff
@@ -53,6 +53,14 @@ def call_pipeline(xs):
 
         self.assertAllClose(xs + 1, os)
 
+    def test_batchwise(self):
+        layer = AddOneToInputs()
+        pipeline = layers.RandomChoice(layers=[layer], batchwise=True)
+        xs = tf.random.uniform((4, 5, 5, 3), 0, 100, dtype=tf.float32)
+        os = pipeline(xs)
+
+        self.assertAllClose(xs + 1, os)
```
ianstenbit, line 62: Does this meaningfully test the batchwise behavior? It seems like it doesn't but maybe I'm missing somethiing

keras_cv/layers/preprocessing/random_choice.py
```diff
@@ -52,19 +52,26 @@ class RandomChoice(BaseImageAugmentationLayer):
             apply the augmentations. This offers a significant performance
             boost, but can only be used if all the layers provided to the
             `layers` argument support auto vectorization.
+        batchwise: (Optional) bool, whether to pass entire batches to the
```
ianstenbit, line 55: Let's mention that the default is `False`.

IMO the docstring would be a bit easier to follow as something like:
"Whether to chose a layer on a batchwise basis instead of an example-wise basis." wdyt?

ianstenbit (changes requested):

keras_cv/layers/preprocessing/random_choice.py
LukeWood: Improved the docstring a bit!

keras_cv/layers/preprocessing/random_choice_test.py
LukeWood: Good point; I updated it to reference `call_counter` in the `AddOne` layer to ensure we are testing this behavior.

keras_cv/layers/preprocessing/random_choice.py
```diff
@@ -73,11 +80,16 @@ def call_layer():
 
         return call_layer
 
+    def _batch_augment(self, inputs):
+        if self.batchwise:
```
jbischof, line 84: Is this correct? If `batchwise` we don't call `_batch_augment`?

LukeWood: yeah it is correct.  if you use batchwise you should pass everything to the same layer.

keras_cv/layers/preprocessing/random_choice.py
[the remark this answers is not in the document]
LukeWood: Done

jbischof (approved):
Thanks! If any other layers should have this flag, please file issues.

LukeWood:
> Thanks! If any other layers should have this flag, please file issues.

I think it should just be RandomChoice and MaybeApply.

PR #1750, opened 2023-05-04
add params count to mobilenet and yolo

haifeng-jin:


ianstenbit (approved):
lgtm -- what is the "path" metadata used for?

keras_cv/models/backbones/mobilenet_v3/mobilenet_v3_backbone_presets.py
```diff
@@ -21,6 +21,9 @@
                 "normalization and hard-swish activation are applied after the "
                 "convolution layers."
             ),
+            "params": 933502,
```
LukeWood, line 24: lets add underscores:

933_502

LukeWood (approved):
minor change in adding underscores but LGTM.

keras_cv/models/backbones/mobilenet_v3/mobilenet_v3_backbone_presets.py
jbischof: I don't think this will make any difference to the python interpreter run by keras.io

jbischof (approved):
@ianstenbit if we added a segmentation head on YOLOv8, would these backbone presets still make sense to use? Would there be another set of backbone presets? If so, what would we name them?

PR #1751, opened 2023-05-05
Use the new ImageClassifier API in our ImageNet training script

ianstenbit:
I've got a training run for MobileNetV3Large running now using these changes.

### DIFF
--- a/examples/training/classification/imagenet/basic_training.py
+++ b/examples/training/classification/imagenet/basic_training.py
@@ -250,7 +250,8 @@ with strategy.scope():
     model = models.ImageClassifier(
         backbone=backbone(input_shape=IMAGE_SIZE + (3,)),
         num_classes=NUM_CLASSES,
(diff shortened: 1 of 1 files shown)


LukeWood (approved):

examples/training/classification/imagenet/basic_training.py
```diff
@@ -246,12 +246,11 @@ def augment(img, label):
 """
 
 with strategy.scope():
-    model = models.__dict__[FLAGS.model_name]
-    model = model(
-        include_rescaling=True,
-        include_top=True,
+    backbone = models.__dict__[FLAGS.model_name]
+    model = models.ImageClassifier(
```
LukeWood, line 250: nit: maybe we should just explicitly construct using `from_preset`?

ianstenbit: I like this idea. I can switch to that paradigm for the next time I run this

PR #1756, opened 2023-05-05
Fix numerics of MobileNet

ianstenbit:
After this change, the forward pass of the backbone is identical to the forward pass of the `keras.applications` backbone (verified by porting weights and running the forward pass)


keras_cv/layers/regularization/squeeze_excite.py
```diff
@@ -37,8 +37,8 @@ class SqueezeAndExcite2D(layers.Layer):
     Args:
         filters: Number of input and output filters. The number of input and
             output filters is same.
-        ratio: Ratio for bottleneck filters. Number of bottleneck filters =
-            filters * ratio, defaults to 0.25.
+        bottleneck_filters: (Optional) Number of bottleneck filters. Defaults
```
jbischof, line 40: Why is this better UI than leaving as a ratio?

jbischof (approved):
Some small questions, but looks good overall!

keras_cv/layers/regularization/squeeze_excite.py
ianstenbit: Well the `ratio` param is a bit weird for this use case, because we want to apply a ratio and then subsequently round to a multiple of 8. Doing this rounding inside the SE block is awkward because it's a caller detail, not an SE block detail, and using the ratio argument would be very weird as a caller because you'd have to specify some odd ratio that you'd calculate by multiplying by your intended approximate ratio, rounding to a multiple of 8, and then computing your actual ratio.

jbischof: Interesting, I wonder why they just don't require the `bottleneck_filters` to be divisible by 8?

ianstenbit: Well in the MobileNet paper this is expressed as a squeeze excite ratio, not a number of filters.

The original SqueezeAndExcite paper (https://arxiv.org/pdf/1709.01507.pdf) refers to a "ratio" for SE blocks, but defines it as the inverse of how it's used elsewhere, so it's somewhat jumbled in the literature.

LukeWood (approved):
Looks reasonable to me - nice to have more fine grained control

PR #1764, opened 2023-05-08
Pre-trained MobileNetV3LargeBackbone weights

ianstenbit:


keras_cv/models/backbones/mobilenet_v3/mobilenet_v3_backbone.py
```diff
@@ -441,7 +454,24 @@ def __new__(
     @classproperty
     def presets(cls):
         """Dictionary of preset names and configurations."""
-        return {}
+        return {
+            "mobilenetv3large": copy.deepcopy(
```
jbischof, line 458: We only put in presets with weights into the alias `presets` method because the no weights version is the same as init!

jbischof: See ResNetV2 for an example ([link](https://github.com/keras-team/keras-cv/blob/6c59877637a49a76f9af0e0cefc2573b106c4d19/keras_cv/models/backbones/resnet_v2/resnet_v2_backbone.py#L520)).

keras_cv/models/backbones/mobilenet_v3/mobilenet_v3_backbone.py
```diff
@@ -441,7 +454,24 @@ def __new__(
...
+            "mobilenetv3large_imagenet": copy.deepcopy(
+                backbone_presets["mobilenetv3large_imagenet"]
+            ),
+        }
+
+    @classproperty
+    def presets_with_weights(cls):
+        """Dictionary of preset names and configurations that include
+        weights."""
+        return {
```
jbischof, line 470: `return cls.presets`

jbischof (approved):
Looks good! Are we going to train a checkpoint for `small` down the line? No rush but curious

ianstenbit:
> Looks good! Are we going to train a checkpoint for `small` down the line? No rush but curious

Yes, I'll start one right now.

PR #1769, opened 2023-05-09
Remove randomly zoomed crop

LukeWood:


ianstenbit:
@LukeWood could you please verify that this doesn't show up in any docs / examples anywhere?

Also could you please add some info in the description about why we're getting rid of this and what users should use instead

ianstenbit (approved):

PR #1785, opened 2023-05-14
Fix MobileNetBackbone.feature_pyramid_inputs

### LINKED ISSUE #1779: Using MobileNetV3Backbone for object detection
As discussed here: https://github.com/keras-team/keras-cv/discussions/1772

I've run into some issues using the MobileNetV3Backbone with RetinaNet:

```
backbone = keras_cv.models.MobileNetV3Backbone.from_preset("mobilenet_v3_large_imagenet")
model = keras_cv.models.RetinaNet(
    backbone,
    num_classes=20,
    bounding_box_format="xywh",
)

img = tf.ones(shape=(1, 512, 512, 3))
output = model.predict(img, verbose=0)
```

Results in an error:
```
in user code:
    
        File "C:\Users\barry\work\keras_object_detection\src\main\python\venv_idea\lib\site-packages\keras_cv\models\object_detection\retinanet\feature_pyramid.py", line 46, in call  *
            p3_output = p3_output + self.upsample_2x(p4_output, training=training)
    
        ValueError: Dimensions must be equal, but are 64 and 128 for '{{node retina_net/feature_pyramid/add_1}} = AddV2[T=DT_FLOAT](retina_net/feature_pyramid/conv2d_18/BiasAdd, retina_net/feature_pyramid/up_sampling2d/resize_1/ResizeNearestNeighbor)' with input shapes: [?,64,64,256], [?,128,128,256].
    
    
    Call arguments received by layer 'feature_pyramid' (type FeaturePyramid):
      • inputs={3: 'tf.Tensor(shape=(None, 64, 64, 24), dtype=float32)', 4: 'tf.Tensor(shape=(None, 64, 64, 24), dtype=float32)', 5: 'tf.Tensor(shape=(None, 32, 32, 40), dtype=float32)'}
      • training=False
```

I suspect its related to [extractor_levels in retinanet.py](https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/object_detection/retinanet/retinanet.py#L166) which are hard-coded as [3, 4, 5] but I presume they need to change based on the backbone used?

barrypitman:
As described in #1779, the feature_pyramid_inputs in MobileNetBackbone are wrong. Each level should be half the size of the next, but MobileNetBackbone was including multiple layers with the same input size (where stride=1).

With this change, the MobileNetBackbone will only include levels with stride=2, similar to how it is implemented in `EfficientNetBackbone`. 

With this change, for mobilenet_v3_large with an input size of 512x512, the backbone will have the following feature_pyramid_inputs:

1: expanded_conv/project/BatchNorm(None, 256, 256, 16)
2: expanded_conv_1/project/BatchNorm(None, 128, 128, 24)
3: expanded_conv_3/project/BatchNorm(None, 64, 64, 40)
4: expanded_conv_6/project/BatchNorm(None, 32, 32, 80)
5: expanded_conv_12/project/BatchNorm(None, 16, 16, 160)

With this change I'm able to successfully use YOLOV8Detector and RetinaNet with MobileNetV3Backbone.

Discussion here:
https://github.com/keras-team/keras-cv/discussions/1772#discussioncomment-5888306

Fixes #1779

@jbischof @LukeWood


LukeWood (approved):
Looks good to me - @ianstenbit most recently touched this so lets get a sign off from him as well.

ianstenbit (approved):
Thanks for the PR!

jbischof (approved):
Thank you @barrypitman! Love the initiative!

barrypitman:
Pleasure - I've fixed the failing test. For the large backbone on a 224x224 input, these are the pyramid_level_inputs:
```
1: multiply(None, 112, 112, 16)
2: expanded_conv_1/project/BatchNorm(None, 56, 56, 24)
3: expanded_conv_3/project/BatchNorm(None, 28, 28, 40)
4: expanded_conv_6/project/BatchNorm(None, 14, 14, 80)
5: expanded_conv_12/project/BatchNorm(None, 7, 7, 160)
``` 
And for the large backbone on the same size image, these are the pyramid_level_inputs:
```
1: multiply(None, 112, 112, 16)
2: expanded_conv/project/BatchNorm(None, 56, 56, 16)
3: expanded_conv_1/project/BatchNorm(None, 28, 28, 24)
4: expanded_conv_3/project/BatchNorm(None, 14, 14, 40)
5: expanded_conv_8/project/BatchNorm(None, 7, 7, 96)
```
If someone can maybe confirm that those are correct relative to [the paper](https://arxiv.org/pdf/1905.02244.pdf), because I'm not sure.

image

ianstenbit (changes requested):
Thanks for updating the test! The tests seems to show that we have a small mistake.

keras_cv/models/backbones/mobilenet_v3/mobilenet_v3_backbone.py
```diff
@@ -289,7 +289,8 @@ def __init__(
                 activation=stackwise_activation[stack_index],
                 expansion_index=stack_index,
             )
-            pyramid_level_inputs[stack_index + 2] = x.node.layer.name
+            if stackwise_stride[stack_index] != 1:
```
ianstenbit, line 292: Okay I think this is actually incorrect, and this should be _before_ the inverted_res_block.

Basically, this results in us taking the _first_ output block for a given resolution instead of the last.

As a result, we're seeing `14x14x40` for `p4` on MobileNetv3Small, but we should see `14x14x48`

barrypitman: Thanks for the feedback @ianstenbit. I've pushed an update to resolve this and updated the test

PR #1786, opened 2023-05-15
Add a numeric preset test for our MobileNet backbone

ianstenbit:
This uses our new pre-trained MobileNetV3Large weights


jbischof (approved):
Very nice, thank you!

PR #1787, opened 2023-05-15
Refactor correct_pad_downsample as a util

ianstenbit:
I realized after #1756 that I introduced a duplicate of the same logic for a second model, so we should just make it a util.


jbischof (approved):
Nice, thank you!

PR #1789, opened 2023-05-16
Support the updated 3D data format in most 3D preprocessing KPLs

ianstenbit:
There are a few that still need to be updated to use the new format.

This resolves a big ugly TODO in https://github.com/keras-team/keras-io/pull/1369, where we have to convert between formats after augmentation and before training the model. (This is the same API friction we previously had in 2D OD)

I added a new test to verify that the numerics are identical with both formats

### DIFF
--- a/keras_cv/layers/preprocessing_3d/base_augmentation_layer_3d.py
+++ b/keras_cv/layers/preprocessing_3d/base_augmentation_layer_3d.py
@@ -261,7 +261,12 @@ def convert_from_model_format(inputs):
 
     box_tensors = [
         inputs["3d_boxes"]["boxes"],
-        tf.expand_dims(inputs["3d_boxes"]["classes"], axis=-1),
+        tf.expand_dims(
+            tf.cast(
+                inputs["3d_boxes"]["classes"], inputs["3d_boxes"]["boxes"].dtype
+            ),
+            axis=-1,
+        ),
         tf.expand_dims(
             tf.cast(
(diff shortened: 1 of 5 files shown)


keras_cv/layers/preprocessing_3d/base_augmentation_layer_3d.py
```diff
@@ -165,12 +165,23 @@ def get_random_transformation(self, point_clouds=None, bounding_boxes=None):
         return None
 
     def call(self, inputs):
-        point_clouds = inputs[POINT_CLOUDS]
-        bounding_boxes = inputs[BOUNDING_BOXES]
+        if "3d_boxes" in inputs.keys():
+            # TODO(ianstenbit): Consider using the better format internally
```
jbischof, line 169: s/internally internally/very internally

jbischof (commented):
Clarification: why not update the augmentations to support the new format organically?

ianstenbit:
> Clarification: why not update the augmentations to support the new format organically?

This is what I'd like to do eventually, but we have users relying on the old format so I need to start by supporting both formats, migrate the users to the new format, and then we can swap to using the new format internally.

keras_cv/layers/preprocessing_3d/base_augmentation_layer_3d.py
ianstenbit: Done (sorta -- I reworded it)

jbischof (approved):
Thanks!

PR #1790, opened 2023-05-16
bug fix for csp_darknet presets

haifeng-jin:
Make sure the `.presets_with_weights()` for specific architectures are overridden.
Otherwise, it would return all the presets with weights from the base class.


jbischof:
Thank you @haifeng-jin!

I think the test is less important. Once this kinda thing is fixed it's fixed.

ianstenbit (approved):
lgtm -- thanks!

PR #1791, opened 2023-05-16
YOLOV8Detector fpn_depth defaults to 2

haifeng-jin:


jbischof:
Thanks @haifeng-jin! This is indeed a thorny issue.

After chatting with @ianstenbit, the problem is that the depth should really be a function of the backbone size. Therefore I wonder if we should instead attach a preference to each YOLO-specific backbone and otherwise force the user to choose? The FPN would pull the preference, if available, the same time it pulls the `pyramid_level_inputs`. WDYT?

ianstenbit:
> Thanks @haifeng-jin! This is indeed a thorny issue.
> 
> After chatting with @ianstenbit, the problem is that the depth should really be a function of the backbone size. Therefore I wonder if we should instead attach a preference to each YOLO-specific backbone and otherwise force the user to choose? The FPN would pull the preference, if available, the same time it pulls the `pyramid_level_inputs`. WDYT?

I had considered this, but it violated our principle of "no task level information on a backbone", so I decided not to use this approach. We can technically figure this out based on looking at the depth of the last block in the backbone, but that's a bit hacky. I thought about it a little more after we chatted about it and I think that a default is probably okay. If a user uses a big backbone with the default smaller FPN, they're still getting a working model, so that's probably reasonable. (Plus, the presets for the `YOLOV8Detector` give us a reasonable value for this depth anyway

ianstenbit (approved):
I'm happy with this approach as long as you two agree that it's okay.

This is a sensible default, and it won't silently break anyone if they don't specify this value.

Sorry for the churn @jbischof, I just changed my mind after we first chatted about this.

haifeng-jin:
I agree. The user can always put a good value by themselves. We are not forcing anything at least.

jbischof (approved):
Sounds like a good first step! Could we added a comment in the arg description about which presets would benefit from depth 3? For example, 

```python
"""
We recommend using `3` for "yolo_v8_l_backbone" and "yolo_v8_xl_backbone".
"""
```

haifeng-jin:
Added.

jbischof:
@ianstenbit I was just guessing which backbones need "3". Can you verify?

ianstenbit:
> @ianstenbit I was just guessing which backbones need "3". Can you verify?

Can confirm it's `L` and `XL`.
the logic in the original implementation is that they take the final `stackwise_depth` from the backbone and use that as an FPN depth

Issue #1725, opened 2023-04-20
ValueError: Layer count mismatch


innat:
When loading `retinanet_resnet50_pascalvoc` preset in TF 2.11, it gives error but works fine in TF 2.12.

```python
import keras_cv
import tensorflow as tf

tf.__version__, keras_cv.__version__
('2.11.0', '0.4.2')

pretrained_model = keras_cv.models.RetinaNet.from_preset(
    "resnet50_imagenet",
    bounding_box_format="xywh",
    num_classes=1
)
# OK
```
```python
pretrained_model = keras_cv.models.RetinaNet.from_preset(
    "retinanet_resnet50_pascalvoc", bounding_box_format="xywh"
)
```
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_27/3432473294.py in <module>
      1 pretrained_model = keras_cv.models.RetinaNet.from_preset(
----> 2     "retinanet_resnet50_pascalvoc", bounding_box_format="xywh"
      3 )

/opt/conda/lib/python3.7/site-packages/keras_cv/models/task.py in from_preset(calling_cls, *args, **kwargs)
    162 
    163             def from_preset(calling_cls, *args, **kwargs):
--> 164                 return super(cls, calling_cls).from_preset(*args, **kwargs)
    165 
    166             cls.from_preset = classmethod(from_preset)

/opt/conda/lib/python3.7/site-packages/keras_cv/models/task.py in from_preset(cls, preset, load_weights, **kwargs)
    150         )
    151 
--> 152         model.load_weights(weights)
    153         return model
    154 

/opt/conda/lib/python3.7/site-packages/keras/utils/traceback_utils.py in error_handler(*args, **kwargs)
     68             # To get the full stack trace, call:
     69             # `tf.debugging.disable_traceback_filtering()`
---> 70             raise e.with_traceback(filtered_tb) from None
     71         finally:
     72             del filtered_tb

/opt/conda/lib/python3.7/site-packages/keras/saving/legacy/hdf5_format.py in load_weights_from_hdf5_group(f, model)
    811     if len(layer_names) != len(filtered_layers):
    812         raise ValueError(
--> 813             "Layer count mismatch when loading weights from file. "
    814             f"Model expected {len(filtered_layers)} layers, found "
    815             f"{len(layer_names)} saved layers."

ValueError: Layer count mismatch when loading weights from file. Model expected 5 layers, found 0 saved layers.
```

[...]

PR #1797, opened 2023-05-17
Mark RetinaNetSmokeTest "extra_large"

jbischof:
Our RetinaNet task preset throws on TF 2.11, which we officially support and use for GCB testing. Marking "extra_large" (manual runs only) until #1725 resolved.


ianstenbit (approved):
Thanks for keeping our CI happy!

jbischof:
\gcbrun

jbischof:
> Thanks for keeping our CI happy!

@ianstenbit it will have no effect but would unblock #1770

Issue #1738, opened 2023-04-28
Make automated testing useful


jbischof:
# Problem

Our current Github automated testing does not catch most errors in our PRs. The reasons are that:

* our standard windows/ubuntu tests exclude ALL code in the `models/` directory where most of the delta is generated,
* our GCB testing does check the `models/` folder but can only be run by maintainers,
* GCB testing has a small budget and should not be run often and
* our current GCB testing still skips all `large` and `extra_large` tests. Unless run manually they are never executed.

For productive code review the only solution is for contributors to run
```
pytest keras_cv/models --run_large
```
before pinging for review. However, this produces no artifacts and relies too much on the contributor.

# Proposal
The motivation for disabling so many tests is concerns about timeouts/OOM for Github actions or GCB. We should aggressively employ pytest flags like `large` and `extra_large` on any tests that take longer than some threshold time to run. This way we can run basic forward pass and other quick checks in `keras_cv/models` on every commit. 

On top of this we need some procedure for nightly/weekly/manual runs of `large` and `extra_large` tests or they will never be executed! At a minimum this should happen before every release, but ideally these could be run before merging a PR.

jbischof:
Looking at the slowest pytest fixtures (see [gist](https://gist.github.com/jbischof/cc27b2fedda53b68a64851fa2c9d7578)) with 

```
pytest --ignore keras_cv/models/legacy/ --run_large --durations=0
```

it seems that many but not the majority of the slowest tests are related to saving. We should audit these to decide which are useless, which should be run manually only, and which are worth running all the time.

ianstenbit:
> @ianstenbit almost all of our longest tests not about saving are in `keras_cv/training/contrastive/`. Since we're not actively developing in that area, should we mark them all as `"extra_large"` (manual run only)?

That SGTM. As far as I know, nobody is really using that API at this point and it will need polish + a guide eventually

jbischof:
This is now fixed!
* Regular tests run on CI
* "large" tests run on GCB
* "extra_large" tests are manual run only
* the old "INTEGRATION" flag is now "large"

PR #1802, opened 2023-05-18
Mark contrastive test suit as "extra_large"

jbischof:
Part of #1738 

These are the slowest tests in our repo currently but this module is not in active development.

"extra_large" tests will be run manually when we push a new release

### DIFF
--- a/cloudbuild/unit_test_jobs.jsonnet
+++ b/cloudbuild/unit_test_jobs.jsonnet
@@ -34,7 +34,7 @@ local unittest = base.BaseTest {
     |||
   ],
   command: [
-    'pytest --ignore keras_cv/models/legacy/',
+    'pytest --ignore keras_cv/models/legacy/ --durations 0',
     'keras_cv',
   ],
 };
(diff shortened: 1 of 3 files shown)


keras_cv/training/contrastive/contrastive_trainer_test.py
```diff
@@ -24,6 +25,7 @@
 from keras_cv.training import ContrastiveTrainer
```
ianstenbit, line 27: Should we leave a comment here about why these are XL?
Maybe something like
"These tests are expensive and our contrastive training API is not yet released"?

cloudbuild/unit_test_jobs.jsonnet
```diff
@@ -34,7 +34,7 @@ local unittest = base.BaseTest {
     |||
   ],
   command: [
-    'pytest --ignore keras_cv/models/legacy/',
+    'pytest --ignore keras_cv/models/legacy/ --durations 0',
```
ianstenbit, line 37: This is a nice feature -- didn't know this existed

ianstenbit (approved):
Thanks Jonathan!

cloudbuild/unit_test_jobs.jsonnet
jbischof: Yes, and this way we can easily check in on slow tests from the GCB logs

PR #1803, opened 2023-05-18
Run large tests on GCB

jbischof:
Part for #1738.

Checking if GCB times out.

Our "large" tests include a lot of important functionality that needs to be tested on each PR. Any tests that will be only be run manually should be marked "extra_large"


ianstenbit:
Very nice! 23 mins should be well within the bounds of not getting timeouts.

Also, adding the time-per-test-case logs is awesome. Easy to see now that `efficientnet_v2_backbone_test.py` has the 2 most expensive test cases at 90s each.

PR #1804, opened 2023-05-19
Mark saving & fitting tests "large" and test `models/` on CI

jbischof:
Part of #1738

Saving and fit tests are in aggreggate the most of our testing time. Mark these as `"large"` and only run on GCB.

We can then run the rest of the tests in `keras_cv/models` as part of our CI.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -43,7 +43,7 @@ jobs:
       env:
         TEST_CUSTOM_OPS: true
       run: |
-        pytest keras_cv/ --ignore keras_cv/models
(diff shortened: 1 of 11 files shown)


jbischof:
\gcbrun

ianstenbit:
I'm curious if the huge delta between the Ubuntu test time and the Windows test time is consistent, so I'm going to restart them. The current runs show 6 mins on Ubuntu (which is great!) and 18 mins on Windows (which is not so great)

ianstenbit (approved):
LGTM. I'm not sure how well this is going to scale down the stretch as we offer more models, as this does seem to add meaningful time to the CI runs, but it seems reasonable for now.

PR #1805, opened 2023-05-19
Test legacy on GCB

jbischof:
Part of #1738 

We should keep monitoring legacy tests on GCB at least. These tests are small and since we plan to migrate the code soon we don't actually want it to be broken.

We can also remove the `"INTEGRATION"` pytest tag, which is now replaced by `"large"`. That's a lot less boilerplate!


ianstenbit (approved):
LGTM -- now that the tests are much more sane I'm going to try to restart Copybara next week since it's been blocked again for a while

Thank you!

PR #1810, opened 2023-05-19
Fix __init__.py behavior for `legacy` folder

jbischof:
Currently get error

```
AttributeError: module 'keras_cv.models' has no attribute 'legacy'
```

when trying to use any of the models in the `keras_cv.models.legacy` namespace.


ianstenbit (approved):
Thank you!

PR #1815, opened 2023-05-22
Add saving tests for object detection

### LINKED ISSUE #1812: Bug with model.save()
Hi !

I am working with retinaNet model and I want to save and reload the model. However when I use model.save(file_path), I got the following error : 
```
WARNING:absl:Found untraced functions such as retina_net_label_encoder_layer_call_fn, retina_net_label_encoder_layer_call_and_return_conditional_losses, anchor_generator_layer_call_fn, anchor_generator_layer_call_and_return_conditional_losses, multi_class_non_max_suppression_3_layer_call_fn while saving (showing 5 of 109). These functions will not be directly callable after loading.
  File "/snap/pycharm-community/332/plugins/python-ce/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
  File "/home/hugo/miniconda3/envs/sinbad/lib/python3.10/site-packages/keras/utils/traceback_utils.py", line 70, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/home/hugo/miniconda3/envs/sinbad/lib/python3.10/site-packages/tensorflow/python/saved_model/nested_structure_coder.py", line 206, in do_encode
    encoded_dict.dict_value.fields[key].CopyFrom(encode_fn(value))
TypeError: 3 has type int, but expected one of: bytes, unicode

```

Is model.save() working on kerasCV models ? 

Thank you in advance 

Hugo PIRES

jbischof:
Fixes #1812.

Currently we only have serialization tests for OD models. This is a much lower threshold and it seems that saving does not work.

The fix is to properly serialize all subclasses and to use string keys for `pyramid_level_inputs`. We do this post hoc here but the correct long term fix is to change the format of this method (#1820).

Also move to `ResNet18Backbone` for `RetinaNet` tests to improve speed.

Ran `pytest keras_cv/models/object_detection/retinanet/ --run_extra_large` before merging.

### DIFF
--- a/keras_cv/models/object_detection/retinanet/feature_pyramid.py
+++ b/keras_cv/models/object_detection/retinanet/feature_pyramid.py
@@ -34,9 +34,9 @@ class FeaturePyramid(keras.layers.Layer):
 
     def call(self, inputs, training=False):
         if isinstance(inputs, dict):
-            c3_output = inputs[3]
-            c4_output = inputs[4]
-            c5_output = inputs[5]
+            c3_output = inputs["3"]
+            c4_output = inputs["4"]
+            c5_output = inputs["5"]
         else:
             c3_output, c4_output, c5_output = inputs
         p3_output = self.conv_c3_1x1(c3_output, training=training)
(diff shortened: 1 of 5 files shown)


ianstenbit (approved):
LGTM, although the linter is unhappy

PR #1817, opened 2023-05-22
Fix DeepLabV3 tests

ianstenbit:
Verified tests are happy with 
```
pytest --run_extra_large keras_cv/models/legacy/segmentation/deeplab_test.py
```


jbischof (approved):
Thanks! I guess these tests never ran because of the `extra_large` flag

PR #1825, opened 2023-05-24
Mark the RetinaNetSmoke test as large, since #1725 is fixed

ianstenbit:
This is passing on tf 2.11 as of #1824 so we can add it to our CI again


jbischof (approved):
Thanks! Did you check our other two `Task` models as well?
* `ImageClassifier`
* `YOLOV8Detector`

ianstenbit:
> Thanks! Did you check our other two `Task` models as well?
> 
> * `ImageClassifier`
> * `YOLOV8Detector`

yes, both of their smoke tests are also `large` and therefore running on gcbrun

PR #1826, opened 2023-05-24
Change pyramid_level_inputs to have string keys

### LINKED ISSUE #1820: Change `pyramid_level_inputs` to have string keys
Currently `pyramid_level_inputs` uses `int` keys, but this breaks serialization (#1812). We patched this up in #1815 at the Task level, but a better fix is to change this attribute to use string keys in all the backbones ([example](https://github.com/keras-team/keras-cv/blob/ef7707f87362eb340190c71baa5b0b5d7f344a40/keras_cv/models/backbones/resnet_v2/resnet_v2_backbone.py#L359)) and in its docstring ([link](https://github.com/keras-team/keras-cv/blob/ef7707f87362eb340190c71baa5b0b5d7f344a40/keras_cv/models/backbones/backbone.py#L159)).

After that we can remove the ad hoc fix in the OD tasks themselves.

avinair:
This PR changes the data type of the keys in the pyramid_level_inputs dictionary from int to string since int keys do not seem to work nicely with TensorFlow model saving flow.

Fixes #1820 

## Who can review?
@jbischof @ianstenbit

### DIFF
--- a/keras_cv/layers/object_detection/roi_align.py
+++ b/keras_cv/layers/object_detection/roi_align.py
@@ -178,6 +178,8 @@ def multilevel_crop_and_resize(
 
     Args:
       features: A dictionary with key as pyramid level and value as features.
+        The pyramid level keys need to be represented by strings like so:
+        "P2", "P3", "P4", and so on.
         The features are in shape of [batch_size, height_l, width_l,
         num_filters].
       boxes: A 3-D Tensor of shape [batch_size, num_boxes, 4]. Each row
@@ -192,10 +194,14 @@ def multilevel_crop_and_resize(
     """
 
     with tf.name_scope("multilevel_crop_and_resize"):
-        levels = list(features.keys())
-        min_level = int(min(levels))
-        max_level = int(max(levels))
-        features_shape = tf.shape(features[min_level])
+        levels_str = list(features.keys())
+        # Levels are represented by strings with a prefix "P" to represent
+        # pyramid levels. The integer level can be obtained by looking at
+        # the value that follows the "P".
+        levels = [int(level_str[1:]) for level_str in levels_str]
+        min_level = min(levels)
+        max_level = max(levels)
+        features_shape = tf.shape(features[f"P{min_level}"])
         batch_size, max_feature_height, max_feature_width, num_filters = (
             features_shape[0],
             features_shape[1],
@@ -211,13 +217,13 @@ def multilevel_crop_and_resize(
         feature_heights = []
         feature_widths = []
         for level in range(min_level, max_level + 1):
-            shape = features[level].get_shape().as_list()
+            shape = features[f"P{level}"].get_shape().as_list()
             feature_heights.append(shape[1])
             feature_widths.append(shape[2])
             # Concat tensor of [batch_size, height_l * width_l, num_filters] for
             # each level.
             features_all.append(
-                tf.reshape(features[level], [batch_size, -1, num_filters])
+                tf.reshape(features[f"P{level}"], [batch_size, -1, num_filters])
             )
         features_r2 = tf.reshape(tf.concat(features_all, 1), [-1, num_filters])
 
@@ -377,7 +383,7 @@ class _ROIAligner(keras.layers.Layer):
         bounding_box_format,
         target_size=7,
         sample_offset: float = 0.5,
-        **kwargs
+        **kwargs,
     ):
         """
         Generates ROI Aligner.
--- a/keras_cv/models/backbones/csp_darknet/csp_darknet_backbone.py
+++ b/keras_cv/models/backbones/csp_darknet/csp_darknet_backbone.py
@@ -144,7 +144,7 @@ class CSPDarkNetBackbone(Backbone):
                 residual=(index != len(stackwise_depth) - 1),
                 name=f"dark{index + 2}_csp",
             )(x)
-            pyramid_level_inputs["P" + str(index + 2)] = x.node.layer.name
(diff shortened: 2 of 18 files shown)


keras_cv/models/legacy/object_detection/faster_rcnn/faster_rcnn.py
```diff
@@ -53,10 +53,10 @@ def __init__(self, **kwargs):
...
-        c3_output = inputs[3]
-        c4_output = inputs[4]
-        c5_output = inputs[5]
+        c2_output = inputs["P2"]
+        c3_output = inputs["P3"]
+        c4_output = inputs["P4"]
+        c5_output = inputs["P5"]
 
         c6_output = self.conv_c6_pool(c5_output)
         p6_output = c6_output
```
avinair, line 62: [Leaving a comment here since Github isn't allowing me to make one at the actual lines but I'm referring to lines 78-84]
I could not change the output of this function to use the new string naming convention since it requires some changes to the following:
- Anchor Generator in lines 293 and 296
- RoI Aligner - https://github.com/keras-team/keras-cv/blob/9d62238f7b99f18eaa1d83727e4dd8d5dd5a920a/keras_cv/layers/object_detection/roi_align.py#L195-L204

keras_cv/models/backbones/csp_darknet/csp_darknet_backbone.py
```diff
@@ -144,7 +144,7 @@ def __init__(
                 residual=(index != len(stackwise_depth) - 1),
                 name=f"dark{index + 2}_csp",
             )(x)
-            pyramid_level_inputs[index + 2] = x.node.layer.name
+            pyramid_level_inputs["P" + str(index + 2)] = x.node.layer.name
```
ianstenbit, line 147: let's use f-strings throughout for these (e.g `f"P{str(index+2)}"`)

keras_cv/models/legacy/object_detection/faster_rcnn/faster_rcnn.py
ianstenbit: I think it would be best to update this + the ROIAligner so that we have consistency across the package.

If you'd like to also update the RetinaNet FPN, feel free to update that one as well (which currently outputs a tuple instead of a dict)

keras_cv/models/object_detection/retinanet/retinanet.py
```diff
@@ -193,7 +192,7 @@ def __init__(
         batch_size = tf.shape(images)[0]
         cls_pred = []
         box_pred = []
-        for feature in features:
+        for feature in features.values():
```
avinair, line 195: @ianstenbit Let me know if you want me to handle the output of the FeaturePyramid differently. I wasn't sure what would be the best way to handle changing the output to a dict and decided to go with this option because of how it is handled in the [yolo_v8_detector](https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py#L409). I believe dicts are ordered in Python3.6 and beyond, so this should be deterministic.

keras_cv/models/backbones/efficientnet_v2/efficientnet_v2_backbone_presets_test.py
```diff
@@ -48,12 +48,14 @@ def test_efficientnet_feature_extractor(self):
             include_rescaling=False,
             input_shape=[256, 256, 3],
         )
-        levels = [3, 4]
-        layer_names = [model.pyramid_level_inputs[level] for level in [3, 4]]
+        levels = ["P3", "P4"]
```
jbischof, line 51: This list should be used everywhere or the levels should be inlined everywhere.

keras_cv/models/object_detection/retinanet/retinanet.py
ianstenbit: Yes, as of Python3.7 I think, dicts are guaranteed to be insertion-ordered.

That said, it may be nice to be a bit more overt where we can, e.g. by saying something like:
```
for pyramid_level in ["P1", "P2", ...]:
    feature = features[pyramid_level]
    ...
```

avinair: Done

keras_cv/layers/object_detection/roi_align.py
```diff
@@ -192,10 +194,14 @@ def multilevel_crop_and_resize(
...
     with tf.name_scope("multilevel_crop_and_resize"):
-        levels = list(features.keys())
-        min_level = int(min(levels))
-        max_level = int(max(levels))
-        features_shape = tf.shape(features[min_level])
+        levels_str = list(features.keys())
+        # Levels are represented by strings with a prefix "P" to represent
+        # pyramid levels. The integer level can be obtained by looking at
+        # the value that follows the "P".
+        levels = [int(level_str[1:]) for level_str in levels_str]
```
ianstenbit, line 201: Is this strictly required? If we take the max using the strings, shouldn't we get the same result due to the String sort order matching the int order in this case?

ianstenbit (approved):
Thanks! Looks good to me.

I'll wait for an approval from Jonathan too

keras_cv/layers/object_detection/roi_align.py
avinair: You are right, if you use max using strings you would get the same result and that's what I initially did. However, it seemed easier to convert it to an int list upfront to avoid messing with the operations on Lines 219, 260, 263 & 280 (the line numbers are the numbers after the edits).

jbischof (approved):
Looks great, thank you!

PR #1834, opened 2023-05-29
Segmentation map in Resizing layer

### LINKED ISSUE #1830: Support Segmentation Maps in Resizing Layer
It would be really helpful is the `Resizing` layer supports segmentation maps.

cosmo3769:
# What does this PR do?

Fixes #1830 (issue)

Demo added by @ariG23498  

## Before submitting
      Pull Request section?
      to it if that's the case.

@ianstenbit @LukeWood @jbischof

### DIFF
--- a/keras_cv/layers/preprocessing/resizing.py
+++ b/keras_cv/layers/preprocessing/resizing.py
@@ -308,7 +308,7 @@ class Resizing(BaseImageAugmentationLayer):
         else:
             input_dtype = tf.float32
 
-        def resize_with_crop_to_aspect_images(x, interpolation_method):
+        def resize_with_crop_to_aspect(x, interpolation_method):
             if isinstance(x, tf.RaggedTensor):
                 x = x.to_tensor()
             return keras.preprocessing.image.smart_resize(
@@ -317,24 +317,25 @@ class Resizing(BaseImageAugmentationLayer):
                 interpolation=interpolation_method,
             )
 
+        def resize_with_crop_to_aspect_images(x):
+            return resize_with_crop_to_aspect(
+                x, interpolation_method=self._interpolation_method
+            )
+
+        def resize_with_crop_to_aspect_masks(x):
+            return resize_with_crop_to_aspect(x, interpolation_method="nearest")
+
         if isinstance(images, tf.RaggedTensor):
             size_as_shape = tf.TensorShape(size)
             shape = size_as_shape + images.shape[-1:]
             spec = tf.TensorSpec(shape, input_dtype)
-            map_fn = partial(
-                resize_with_crop_to_aspect_images,
-                interpolation_method=self._interpolation_method,
-            )
             images = tf.map_fn(
-                map_fn,
+                resize_with_crop_to_aspect_images,
                 images,
                 fn_output_signature=spec,
             )
         else:
-            images = resize_with_crop_to_aspect_images(
-                images,
-                interpolation_method=self._interpolation_method,
-            )
+            images = resize_with_crop_to_aspect_images(images)
 
         inputs["images"] = images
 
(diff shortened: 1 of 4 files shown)


keras_cv/layers/preprocessing/resizing.py
```diff
@@ -143,24 +154,37 @@ def _augment(self, inputs):
...
         size = [self.height, self.width]
         images = tf.image.resize(
             images, size=size, method=self._interpolation_method
         )
         images = tf.cast(images, self.compute_dtype)
 
+        if segmentation_maps is not None:
+            segmentation_maps = tf.image.resize(
+                segmentation_maps, size=size, method=self._interpolation_method
+            )
```
ariG23498, line 176: I think we should have a way to input different resizing methods of the image and the segmentation masks.

WDYT? @ianstenbit

ayulockin: Usually segmentation masks are resized with nearest neighbor interpolation in practice. Using say bilinear interpolation ruin the masks.

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -24,7 +24,13 @@
 H_AXIS = -3
 W_AXIS = -2
 
-supported_keys = ["images", "labels", "targets", "bounding_boxes"]
+supported_keys = [
+    "images",
+    "labels",
+    "targets",
+    "bounding_boxes",
+    "segmentation_maps",
```
ianstenbit, line 32: This should be `"segmentation_masks"` for consistency with the other KPLs

ianstenbit: (Same throughout -- s/maps/masks)

keras_cv/layers/preprocessing/resizing.py
ianstenbit: Yes, this should be resized with nearest neighbor. Numerical interpolation is not safe for masks unless they're one-hot encoded. (See RandomRotation for an example of where we had to deal with this).

For resizing I think it's probably reasonable to just always use nearest neighbor, regardless of whether masks are one-hot encoded.

ianstenbit (changes requested):
Thanks for the PR!

Please also update the resizing test to include a mask resizing, and also please update `with_segmentation_masks_test.py`

Thank you!

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -255,24 +302,53 @@ def _resize_with_crop(self, inputs):
         else:
             input_dtype = tf.float32
 
-        def resize_with_crop_to_aspect(x):
+        def resize_with_crop_to_aspect_images(x):
```
ianstenbit, line 305: Instead of making two versions of this method, let's just add `interpolation_method` as a parameter, and call the same method for both masks and images, just with different parameters.

ianstenbit (commented):
Thanks! 
Looking closer -- just need to add tests, and let's also add an example in 
`keras-cv/examples/layers/preprocessing/segmentation/resize_demo.py` (and please include the output of the demo in the PR description.

cosmo3769:
Apologies for the delayed work. I am busy with my exams now. I will resolve the issues and changes requested on 8 June.

keras_cv/layers/preprocessing/resizing.py
ariG23498: I have used the `functool.partial` to take care of this. Let me know if you want it to be implemented in any other way.

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -255,24 +308,58 @@ def _resize_with_crop(self, inputs):
         else:
             input_dtype = tf.float32
 
-        def resize_with_crop_to_aspect(x):
+        def resize_with_crop_to_aspect_images(x, interpolation_method):
```
ianstenbit, line 311: nit: let's keep the name the way it was (without "_images")

keras_cv/layers/preprocessing/resizing.py
```diff
@@ -255,24 +308,58 @@ def _resize_with_crop(self, inputs):
...
+                x,
+                size=size,
+                interpolation=interpolation_method,
             )
 
         if isinstance(images, tf.RaggedTensor):
             size_as_shape = tf.TensorShape(size)
             shape = size_as_shape + images.shape[-1:]
             spec = tf.TensorSpec(shape, input_dtype)
+            map_fn = partial(
```
ianstenbit, line 324: Ahh I didn't realize we'd end up having to use a `partial` here -- my bad.

How about we keep a base method `resize_with_crop_to_aspect` and make two wrapper methods with the same name and `_images` and `_masks` suffixes which pass through the interpolation method (just to avoid using a partial)

ianstenbit (approved):
Thanks for the PR! Just a few small comments left

PR #1837, opened 2023-05-30
Support for `tf.data.Dataset` added in `plot_image_gallery`, removed hard dependency on `rows` and `cols` params

suvadityamuk:
# What does this PR do?

This PR is meant to add support for using `tf.data.Dataset` instances directly with `keras_cv.visualization.plot_image_gallery` instead of the user having to manually convert the dataset and then perform processing. It still supports NumPy arrays or Tensors as it used to.

Also, handled a possible future deprecation of removing `plt.tight_layout()` by making use of `plt.subplot(...,layout='tight')` as per [this warning in Matplotlib](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.5.0.html#discouraged-figure-parameters-tight-layout-and-constrained-layout)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

cc: @ianstenbit (as discussed previously) @jbischof @tanzhenyu

### DIFF
--- a/examples/visualization/plot_image_gallery.py
+++ b/examples/visualization/plot_image_gallery.py
@@ -1,8 +1,8 @@
 """
 Title: Plot an image gallery
-Author: [lukewood](https://lukewood.xyz)
+Author: [lukewood](https://lukewood.xyz), updated by [Suvaditya Mukherjee](https://twitter.com/halcyonrayes)
 Date created: 2022/10/16
-Last modified: 2022/05/31
+Last modified: 2022/06/24
 Description: Visualize ground truth and predicted bounding boxes for a given
              dataset.
 """
@@ -22,8 +22,6 @@ train_ds = tfds.load(
     shuffle_files=True,
 )
 
-train_ds = train_ds.ragged_batch(16)
-
 keras_cv.visualization.plot_image_gallery(
     train_ds,
     value_range=(0, 255),
@@ -34,13 +32,9 @@ keras_cv.visualization.plot_image_gallery(
 If you want to use plain NumPy arrays, you can do that too:
 """
 
-# Prepare some NumPy arrays from random noise
+# Prepare some sample NumPy arrays from random noise
 
-samples = []
-for sample in train_ds.take(20):
-    samples.append(sample["image"].numpy())
-
-samples = np.array(samples, dtype="object")
+samples = np.random.randint(0, 255, (20, 224, 224, 3))
 
 keras_cv.visualization.plot_image_gallery(
     samples, value_range=(0, 255), scale=3, rows=4, cols=5
--- a/examples/visualization/plot_image_gallery.py
+++ b/examples/visualization/plot_image_gallery.py
@@ -1,6 +1,7 @@
 """
 Title: Plot an image gallery
-Author: [lukewood](https://lukewood.xyz), updated by [Suvaditya Mukherjee](https://twitter.com/halcyonrayes)
+Author: [lukewood](https://lukewood.xyz), updated by
+[Suvaditya Mukherjee](https://twitter.com/halcyonrayes)
 Date created: 2022/10/16
 Last modified: 2022/06/24
 Description: Visualize ground truth and predicted bounding boxes for a given
--- a/examples/visualization/plot_image_gallery.py
+++ b/examples/visualization/plot_image_gallery.py
@@ -12,7 +12,7 @@ Plotting images from a TensorFlow dataset is easy with KerasCV. Behold:
 """
 
 import tensorflow_datasets as tfds
-
+import numpy as np
 import keras_cv
 
 train_ds = tfds.load(
@@ -28,4 +28,20 @@ keras_cv.visualization.plot_image_gallery(
     train_ds,
     value_range=(0, 255),
     scale=3,
+)
+
+"""
+If you want to use plain NumPy arrays, you can do that too:
+"""
+
+# Prepare some NumPy arrays from random noise
+
+samples = []
+for sample in train_ds.take(20):
+    samples.append(sample["image"].numpy())
+
+samples = np.array(samples, dtype="object")
+
+keras_cv.visualization.plot_image_gallery(
+    samples, value_range=(0, 255), scale=3, rows=4, cols=5
 )
--- a/examples/visualization/plot_image_gallery.py
+++ b/examples/visualization/plot_image_gallery.py
@@ -2,7 +2,7 @@
 Title: Plot an image gallery
 Author: [lukewood](https://lukewood.xyz)
 Date created: 2022/10/16
-Last modified: 2022/10/16
+Last modified: 2022/05/31
 Description: Visualize ground truth and predicted bounding boxes for a given
              dataset.
 """
@@ -11,7 +11,6 @@ Description: Visualize ground truth and predicted bounding boxes for a given
 Plotting images from a TensorFlow dataset is easy with KerasCV. Behold:
 """
 
-import tensorflow as tf
 import tensorflow_datasets as tfds
 
 import keras_cv
@@ -23,18 +22,10 @@ train_ds = tfds.load(
     shuffle_files=True,
 )
 
-
-def unpackage_tfds_inputs(inputs):
-    return inputs["image"]
-
-
-train_ds = train_ds.map(unpackage_tfds_inputs)
-train_ds = train_ds.apply(tf.data.experimental.dense_to_ragged_batch(16))
+train_ds = train_ds.ragged_batch(16)
 
 keras_cv.visualization.plot_image_gallery(
-    next(iter(train_ds.take(1))),
+    train_ds,
     value_range=(0, 255),
     scale=3,
-    rows=2,
-    cols=2,
 )
--- a/examples/visualization/plot_image_gallery.py
+++ b/examples/visualization/plot_image_gallery.py
@@ -11,8 +11,9 @@ Description: Visualize ground truth and predicted bounding boxes for a given
 Plotting images from a TensorFlow dataset is easy with KerasCV. Behold:
 """
 
-import tensorflow_datasets as tfds
 import numpy as np
+import tensorflow_datasets as tfds
+
 import keras_cv
 
 train_ds = tfds.load(
--- a/keras_cv/visualization/plot_image_gallery.py
+++ b/keras_cv/visualization/plot_image_gallery.py
@@ -47,7 +47,8 @@ def _extract_image_batch(images, num_images, batch_size):
     else:
         if len(images.shape) != 4:
             raise ValueError(
-                "`plot_images_gallery()` requires you to batch your `np.array` samples together."
+                "`plot_images_gallery()` requires you to "
+                "batch your `np.array` samples together."
             )
         else:
             num_samples = (
@@ -90,14 +91,17 @@ def plot_image_gallery(
     ![example gallery](https://i.imgur.com/r0ndse0.png)
 
     Args:
-        images: a Tensor, `tf.data.Dataset` or NumPy array containing images to show in the
-            gallery. Note: If using a `tf.data.Dataset`, images should be
-            present in the `FeaturesDict` under the key `image`.
+        images: a Tensor, `tf.data.Dataset` or NumPy array containing images
+            to show in the gallery. Note: If using a `tf.data.Dataset`,
+            images should be present in the `FeaturesDict` under
+            the key `image`.
         value_range: value range of the images. Common examples include
             `(0, 255)` and `(0, 1)`.
         scale: how large to scale the images in the gallery
-        rows: (Optional) number of rows in the gallery to show. Required if inputs are unbatched.
-        cols: (Optional) number of columns in the gallery to show. Required if inputs are unbatched.
+        rows: (Optional) number of rows in the gallery to show.
+            Required if inputs are unbatched.
+        cols: (Optional) number of columns in the gallery to show.
+            Required if inputs are unbatched.
         path: (Optional) path to save the resulting gallery to.
         show: (Optional) whether to show the gallery of images.
         transparent: (Optional) whether to give the image a transparent
--- a/keras_cv/visualization/plot_image_gallery.py
+++ b/keras_cv/visualization/plot_image_gallery.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import numpy as np
 import keras_cv
+import tensorflow as tf
 from keras_cv import utils
 from keras_cv.utils import assert_matplotlib_installed
 
@@ -25,14 +27,16 @@ except:
 def plot_image_gallery(
     images,
     value_range,
-    rows=3,
-    cols=3,
+    rows=None,
+    cols=None,
+    batch_size=8,
     scale=2,
     path=None,
     show=None,
     transparent=True,
     dpi=60,
     legend_handles=None,
+    image_key="image",
 ):
     """Displays a gallery of images.
 
@@ -45,19 +49,10 @@ def plot_image_gallery(
         shuffle_files=True,
     )
 
-
-    def unpackage_tfds_inputs(inputs):
-        return inputs["image"]
-
-    train_ds = train_ds.map(unpackage_tfds_inputs)
-    train_ds = train_ds.apply(tf.data.experimental.dense_to_ragged_batch(16))
-
     keras_cv.visualization.plot_image_gallery(
-        next(iter(train_ds.take(1))),
+        train_ds,
         value_range=(0, 255),
         scale=3,
-        rows=2,
-        cols=2,
     )
     ```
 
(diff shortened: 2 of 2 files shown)


keras_cv/visualization/plot_image_gallery.py
```diff
@@ -25,14 +26,16 @@
...
     scale=2,
+    rows=None,
+    cols=None,
+    batch_size=8,
     path=None,
     show=None,
     transparent=True,
     dpi=60,
     legend_handles=None,
+    image_key="image",
```
ianstenbit, line 38: Let's keep this out of the visualization API, as it's not super relevant to the visualization component.

We have a pattern in these visualization tools of expecting only images in the `images` parameter, and we should keep it that way by assuming that the contents of the tf dataset are just image tensors (and documenting this assumption)

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -92,27 +91,69 @@ def unpackage_tfds_inputs(inputs):
...
-    fig = plt.figure(figsize=(cols * scale, rows * scale))
-    fig.tight_layout()  # Or equivalently,  "plt.tight_layout()"
-    plt.subplots_adjust(wspace=0, hspace=0)
-    plt.margins(x=0, y=0)
-    plt.axis("off")
+    if isinstance(images, tf.data.Dataset):
+
+        # Find final dataset batch size
+        sample = next(iter(images.take(1)))
+        if len(sample[image_key].shape) == 3:
```
ianstenbit, line 98: If `rows` and `cols` are `None`, let's assume images are batched and throw otherwise.

If they are specified, we should assume that the dataset is unbatched and take the first `rows * cols` images.

(This way we can also drop the `batch_size` parameter)

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -92,27 +91,69 @@ def unpackage_tfds_inputs(inputs):
...
+
+        def unpack_images(inputs):
+            return inputs[image_key]
+
+        images = images.map(unpack_images)
+        images = images.take(batches)
+        images = next(iter(images))
+
+    # Calculate appropriate number of rows and columns
+    if rows is None and cols is None:
```
ianstenbit, line 118: To infer rows/cols based on batch size, we should probably strive for a square instead of a 2xN, which is what we have here (I think?)

What about 
```
rows = math.ceil(batch_size ** 0.5)
cols = math.ceil(batch_size / cols)
```

ianstenbit (changes requested):
Thanks for the PR!

Let's also update `examples/visualization/plot_image_gallery.py` to show the usage of both paths for this API

keras_cv/visualization/plot_image_gallery.py
suvadityamuk: Got it, removed

keras_cv/visualization/plot_image_gallery.py
suvadityamuk: Got it. Changed

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -80,6 +73,10 @@ def unpackage_tfds_inputs(inputs):
         legend_handles: (Optional) matplotlib.patches List of legend handles.
             I.e. passing: `[patches.Patch(color='red', label='mylabel')]` will
             produce a legend with a single red patch and the label 'mylabel'.
+
+    Note:
```
ianstenbit, line 77: Instead of assuming the dataset has a dict with the key "image", how about the elements are just the images?

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -92,27 +89,87 @@ def unpackage_tfds_inputs(inputs):
...
+            images = images.map(unpack_images)
+            images = images.take(batch_size)
+            images = next(iter(images))
+        else:
+            sample_shape = images.shape
+            if len(sample_shape) == 4:
+                batch_size = sample_shape[0]
+            else:
+                raise ValueError(
+                    f"Passed '`{type(images)}`' does not appear to be batched. Please batch using the `.batch()."
```
ianstenbit, line 116: Let's say

"plot_image_gallery received unbatched images and `cols` and `rows` were both `None`. Either images should be batched, or `cols` and `rows` should be specified.

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -92,27 +89,87 @@ def unpackage_tfds_inputs(inputs):
...
-    fig = plt.figure(figsize=(cols * scale, rows * scale))
-    fig.tight_layout()  # Or equivalently,  "plt.tight_layout()"
-    plt.subplots_adjust(wspace=0, hspace=0)
-    plt.margins(x=0, y=0)
-    plt.axis("off")
+    def unpack_images(inputs):
+        return inputs["image"]
+
+    # Calculate appropriate number of rows and columns
+    if rows is None and cols is None:
```
ianstenbit, line 96: Probably `or`?

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -92,27 +89,87 @@ def unpackage_tfds_inputs(inputs):
...
+            images = images.ragged_batch(batch_size=batch_size)
+
+            images = images.map(unpack_images)
+            images = images.take(batch_size)
+            images = next(iter(images))
+        else:
+            batch_size = rows * cols
+            images = images[:batch_size, ...]
+    else:
+        raise ValueError(
```
ianstenbit, line 138: This error seems like it will be thrown iff one of `rows` or `cols` is set, but not both. Let's update the error message to reflect that.

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -45,32 +47,23 @@ def plot_image_gallery(
...
     Args:
-        images: a Tensor or NumPy array containing images to show in the
+        images: a Tensor, `tf.data.Dataset` or NumPy array containing images to show in the
             gallery.
         value_range: value range of the images. Common examples include
             `(0, 255)` and `(0, 1)`.
-        rows: number of rows in the gallery to show.
-        cols: number of columns in the gallery to show.
         scale: how large to scale the images in the gallery
+        rows: (Optional) number of rows in the gallery to show.
```
ianstenbit, line 65: Required if inputs are unbatched (for this and `cols`)

examples/visualization/plot_image_gallery.py
```diff
@@ -23,18 +22,10 @@
...
-    return inputs["image"]
-
-
-train_ds = train_ds.map(unpackage_tfds_inputs)
-train_ds = train_ds.apply(tf.data.experimental.dense_to_ragged_batch(16))
+train_ds = train_ds.ragged_batch(16)
 
 keras_cv.visualization.plot_image_gallery(
-    next(iter(train_ds.take(1))),
+    train_ds,
```
ianstenbit, line 28: For the example, let's show both ways of calling this (maybe one with a Numpy array of images, and the other with a tf.data.Dataset

ianstenbit (changes requested):
Thanks for the updates!

A few more comments for you

keras_cv/visualization/plot_image_gallery.py
suvadityamuk: Normally, TFDS and most other datasets which are pre-loaded have a dictionary. Do you think it is unnecessary?

keras_cv/visualization/plot_image_gallery.py
suvadityamuk: Done!

keras_cv/visualization/plot_image_gallery.py
suvadityamuk: Ah yes, my bad :sweat_smile:

suvadityamuk: Solved!

keras_cv/visualization/plot_image_gallery.py
suvadityamuk: Done!

keras_cv/visualization/plot_image_gallery.py
suvadityamuk: Done

examples/visualization/plot_image_gallery.py
suvadityamuk: Done

examples/visualization/plot_image_gallery.py
```diff
@@ -23,18 +22,26 @@
...
+    value_range=(0, 255),
+    scale=3,
+)
 
+"""
+If you want to use plain NumPy arrays, you can do that too:
+"""
+
+# Prepare some NumPy arrays from random noise
```
ianstenbit, line 37: This isn't random noise -- these are images from the tf.ds, right?

keras_cv/visualization/plot_image_gallery.py
ianstenbit: okay, for compat with tfds that seems reasonable. Let's move this note to the `images` docstring.

At the end of the docstring we can say "If a `tf.data.Dataset` is used, images should be present the `FeaturesDict` under the key `"image"`

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -92,27 +89,88 @@ def unpackage_tfds_inputs(inputs):
...
+            images = images.ragged_batch(batch_size=batch_size)
+
+            images = images.map(unpack_images)
+            images = images.take(batch_size)
+            images = next(iter(images))
+        else:
+            batch_size = rows * cols
+            images = images[:batch_size, ...]
+    else:
+        raise ValueError(
```
ianstenbit, line 139: This branch seems unreachable -- the `elif` above. is the inverse of the original `if`

So we can just make it an `else` and drop this branch.

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -92,27 +89,88 @@ def unpackage_tfds_inputs(inputs):
...
-    fig = plt.figure(figsize=(cols * scale, rows * scale))
-    fig.tight_layout()  # Or equivalently,  "plt.tight_layout()"
-    plt.subplots_adjust(wspace=0, hspace=0)
-    plt.margins(x=0, y=0)
-    plt.axis("off")
+    def unpack_images(inputs):
+        return inputs["image"]
+
+    # Calculate appropriate number of rows and columns
+    if rows is None or cols is None:
```
ianstenbit, line 96: If one of them is set and the other isn't, this behaves a bit weirdly. 

If we set `rows=2` and `cols=None`, and our batch size is `9`, we'll end up with a 3x3, which seems counterintuitive.

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -92,27 +89,88 @@ def unpackage_tfds_inputs(inputs):
...
+            if len(sample_shape) == 4:
+                batch_size = sample_shape[0]
+            else:
+                raise ValueError(
+                    f"`plot_image_gallery` received unbatched images and `cols` and `rows` "
+                    "were both `None`. Either images should be batched, or `cols` and `rows` should be specified."
+                )
+
+    elif rows is not None and cols is not None:
+        if isinstance(images, tf.data.Dataset):
```
ianstenbit, line 121: If rows and cols are both specified, why do we need to compute batch size? I'm not sure what the purpose of this branch is

examples/visualization/plot_image_gallery.py
suvadityamuk: Aha, yes. My bad, corrected it

keras_cv/visualization/plot_image_gallery.py
suvadityamuk: So in this case, we can do something like divide the not-`None` argument such that the batch will create something which at least fixes the user's args. 
```python
if rows is None and cols is not None:
    rows = ceil(batch_size / cols)
elif rows is not None and cols is None:
    cols = ceil(batch_size / rows)
```
This should work in theory.
For your example, it would mean we do `rows=2`, `columns=None`, `batch_size=9` and get `rows=2`, `cols=5`

The final grid would have 5 images in the first row and 4 images in the second row. Is that fine?

keras_cv/visualization/plot_image_gallery.py
suvadityamuk: This is done to take out the required number of images from the dataset

keras_cv/visualization/plot_image_gallery.py
ianstenbit: yeah this sgtm

keras_cv/visualization/plot_image_gallery.py
ianstenbit: I see. It seems like it would be a lot easier to just use `tfds.as_numpy(images.take(rows*cols)` 

(Probably have to optionally unbatch first)

ianstenbit: Overall, I think the logic would be much simpler if we separated the logic for converting a tf.data.Dataset into a numpy array from the logic for computing batch size / rows / cols. As-is there's some unnecessary indirection so the resulting code is more complex than it needs to be.

One simple approach would be to convert all possible input types to an unbatched iterator at the top of the method, and record the batch size if there was one.

suvadityamuk: `tfds` is a special case. Most folks may also be invoking and creating `tf.data.Dataset` instances from a simple `image_dataset_from_directory()` call or `from_tensor_slices()` too. Using `tfds.as_numpy` might not be the best choice for that reason.

keras_cv/visualization/plot_image_gallery.py
```diff
@@ -45,32 +47,23 @@ def plot_image_gallery(
...
     Args:
-        images: a Tensor or NumPy array containing images to show in the
+        images: a Tensor, `tf.data.Dataset` or NumPy array containing images to show in the
             gallery.
         value_range: value range of the images. Common examples include
             `(0, 255)` and `(0, 1)`.
-        rows: number of rows in the gallery to show.
-        cols: number of columns in the gallery to show.
         scale: how large to scale the images in the gallery
+        rows: (Optional) number of rows in the gallery to show. Required if inputs are unbatched.
```
ianstenbit, line 65: After chatting with the team a bit, and thinking about this code some more, I think we should keep rows/cols required and get rid of the logic to try to infer how many images should be shown.

It seems reasonable to ask users for a number of rows/cols, and it makes the code nice and overt on the caller's side.

It also dramatically simplifies how this can be implemented, because we don't need to try to infer batch sizes for different data types, and to add support for tf.data.Datasets, we can just use `tfds.as_numpy(dataset.take(rows*cols))` (after optionally unbatching the dataset 😄 

Sorry for the churn on this, and thanks for the awesome work!

martin-gorner: I'd prefer an automatic rows and cols, mostly to remove many useless lines of code from our examples.
This can be implemented in just a few lines, even with a fancy aspect ratio adaptation:

```
N = len(images)
sqrtN = int(math.ceil(math.sqrt(N)))
aspect = sum([im.shape[1]/im.shape[0] for im in images]) / N # mean aspect ratio of images
fig = plt.figure(figsize=(15,15/aspect), frameon=False)
```

(I've been using this code in all my samples from the past couple of years)
And it's fine to require either both rows and cols or neither and not handle the fancy cases in between.

ianstenbit (commented):
See Martin's comment about rows/cols specification.

I think the code here would be a lot cleaner if we structured it like this:

```
def visualize(....):
    image_batch = _extract_image_batch(images, rows, cols) # Throws if input is invalid
    rows = rows or ...
    cols = cols or ...
    # actually do the plotting
```

Basically the current logic of intermixing the conditionals of input data type and whether rows/cols are specified gets a bit hairy and we should break out the logic to extract an image batch.

Using some guard clauses would make the code a lot easier to understand 🙂

suvadityamuk:
I have performed a ton of refactoring around the code to try and reduce the complexity of the code :sweat_smile:.
Let me know if this is a better method to work with. This code handles more or less every case I could try, except for passing `np.array` in an unbatched format. Most other possibilities seem to be accounted for.

ianstenbit (approved):
Beautiful -- I love the new structure.

Thank you for the PR, and for sticking with it despite the churn on my end 😄

suvadityamuk:
Aha, no worries at all, happy to help! :sweat_smile:

suvadityamuk:
Just committed a linted fix!

suvadityamuk:
Aha, the example was not linted, it seems. This new commit should do the trick, hopefully! :crossed_fingers:

suvadityamuk:
Seems like it needed that one more final "push", haha. Please check it out now, have run `bash shell/format.sh` and `bash shell/lint.sh` and fixed all the problems

PR #1841, opened 2023-05-30
Support updating the NMS decoder in YOLOv8

### LINKED ISSUE #1840: YOLOv8 does not support updating the `prediction_decoder`
If you build a YOLOV8Detector model `yolov8` and call

`yolov8.prediction_decoder = some_new_prediction_decoder`, the predict function isn't being updated correctly.

ianstenbit:
Fixes #1840.

cc @innat

### DIFF
--- a/keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
+++ b/keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
@@ -427,7 +427,7 @@ class YOLOV8Detector(Task):
         super().__init__(inputs=images, outputs=outputs, **kwargs)
 
         self.bounding_box_format = bounding_box_format
-        self.prediction_decoder = (
+        self._prediction_decoder = (
             prediction_decoder
             or keras_cv.layers.MultiClassNonMaxSuppression(
                 bounding_box_format=bounding_box_format,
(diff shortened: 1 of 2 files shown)


keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
```diff
@@ -608,14 +608,34 @@ def decode_predictions(
...
+                "use the same `bounding_box_format`, but got "
+                "`prediction_decoder.bounding_box_format="
+                f"{prediction_decoder.bounding_box_format}`, and "
+                "`self.bounding_box_format="
+                f"{self.bounding_box_format}`."
+            )
+        self._prediction_decoder = prediction_decoder
+        self.make_predict_function(force=True)
+        self.make_train_function(force=True)
+        self.make_test_function(force=True)
```
haifeng-jin, line 629: Don't we need the return values of these functions?

keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
```diff
@@ -608,14 +608,34 @@ def decode_predictions(
...
+
     def get_config(self):
         return {
             "num_classes": self.num_classes,
             "bounding_box_format": self.bounding_box_format,
             "fpn_depth": self.fpn_depth,
             "backbone": keras.utils.serialize_keras_object(self.backbone),
             "label_encoder": self.label_encoder,
-            "prediction_decoder": self.prediction_decoder,
+            "prediction_decoder": self._prediction_decoder,
```
haifeng-jin, line 638: Do we need to update this? Seems directly using the property is safer.

haifeng-jin (commented):
Thanks for the PR!

keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
ianstenbit: I've just copied what we did in RetinaNet -- I think it doesn't particularly matter but if there is a reason I should prefer the property I'm happy to switch it in both models. wdyt?

keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
ianstenbit: No, these just need to get called to update the graph. No need to store the results here, as they'll be stored on the model in the call to `make_xyz_function`

haifeng-jin (approved):

Issue #1832, opened 2023-05-27
YOLOV8 does not work with ragged tensors


eawer:
TF version: 2.12 (nvcr.io/nvidia/tensorflow:23.04-tf2-py3 container)
keras_cv version: 0.5.0
host: Ubuntu 22.04.2 LTS

YoloV8 works as expected with just plain tensors, but when it comes to ragged tensors (which are pretty common in OD task) it starts to fail
code:
```python
images = tf.ones(shape=(1, 512, 512, 3))
boxes = tf.RaggedTensor.from_tensor(
    tf.convert_to_tensor(
        [
            [
                [0, 0, 100, 100],
                [100, 100, 200, 200],
                [300, 300, 100, 100],
            ]
        ],
        dtype=tf.float32,
    )
)
classes = tf.RaggedTensor.from_tensor(
    tf.convert_to_tensor([[1, 1, 1]], dtype=tf.float32)
)

labels = {
    "boxes": boxes,
    "classes": classes,
}

model = keras_cv.models.YOLOV8Detector(
    num_classes=1,
    backbone=keras_cv.models.YOLOV8Backbone.from_preset("yolo_v8_m_backbone"),
    fpn_depth=2,
    bounding_box_format="xywh",
)
model.compile(
    optimizer='adam', 
    box_loss="iou", 
    classification_loss="binary_crossentropy"
)
model.fit({
    "images": images,
    "bounding_boxes": labels,
})
```

error:
```bash
TypeError                                 Traceback (most recent call last)
Cell In[47], line 31
     20 model = keras_cv.models.YOLOV8Detector(
     21     num_classes=1,
     22     backbone=keras_cv.models.YOLOV8Backbone.from_preset("yolo_v8_m_backbone"),
     23     fpn_depth=2,
     24     bounding_box_format="xywh",
     25 )
     26 model.compile(
     27     optimizer='adam', 
     28     box_loss="iou", 
     29     classification_loss="binary_crossentropy"
     30 )
---> 31 model.fit({
     32     "images": images,
     33     "bounding_boxes": labels,
     34 })

File /usr/local/lib/python3.8/dist-packages/keras/utils/traceback_utils.py:70, in filter_traceback.<locals>.error_handler(*args, **kwargs)
     67     filtered_tb = _process_traceback_frames(e.__traceback__)
     68     # To get the full stack trace, call:
     69     # `tf.debugging.disable_traceback_filtering()`
[...]

PR #1844, opened 2023-05-30
Throw a better error when ragged boxes are passed to YOLOv8

ianstenbit:
In the long run, we'd like to actually support ragged boxes here, but for now we should at least throw a better error.

Related to #1832.


haifeng-jin (approved):
Thanks for the PR!

PR #1852, opened 2023-06-01
backbone compatibility for objection detection models [RetinaNet]

IMvision12:
# What does this PR do?

Added Test

@ianstenbit @jbischof

### DIFF
--- a/keras_cv/models/object_detection/retinanet/retinanet_test.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet_test.py
@@ -264,7 +264,7 @@ class RetinaNetSmokeTest(tf.test.TestCase):
         expected_box = tf.constant(
             [-1.2427993, 0.05179548, -1.9953268, 0.32456252]
         )
-        self.assertAllClose(output["box"][0, 123, :], expected_box)
+        self.assertAllClose(output["box"][0, 123, :], expected_box, atol=1e-5)
 
         expected_class = tf.constant(
             [
@@ -291,4 +291,6 @@ class RetinaNetSmokeTest(tf.test.TestCase):
             ]
         )
         expected_class = tf.reshape(expected_class, (20,))
-        self.assertAllClose(output["classification"][0, 123], expected_class)
+        self.assertAllClose(
+            output["classification"][0, 123], expected_class, atol=1e-5
+        )
--- a/keras_cv/models/object_detection/retinanet/retinanet_test.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet_test.py
@@ -279,7 +279,36 @@ class RetinaNetSmokeTest(tf.test.TestCase):
         xs, _ = _create_bounding_box_dataset(bounding_box_format="xywh")
         output = model(xs)
 
-        # How to get expected_box as on every run it will be different
-        expected_box = []
+        expected_box = tf.constant(
+            [-0.40364307, 0.15742484, -1.3398054, 0.38104224]
+        )
+        self.assertAllClose(output["box"][0, 123, :], expected_box, atol=0.2)
 
-        self.assertTrue(output["box"][0, 123, :], expected_box)
+        expected_class = tf.constant(
+            [
+                -8.110557,
+                -7.5584545,
+                -7.9096074,
+                -8.163499,
+                -6.721413,
+                -7.8551226,
+                -6.9021387,
+                -8.499062,
+                -6.190583,
+                -7.294302,
+                -6.341216,
+                -7.6847053,
+                -7.8064694,
+                -7.7984314,
+                -6.5489616,
+                -6.524928,
+                -7.1218467,
+                -7.070874,
+                -8.457317,
+                -6.167918,
+            ]
+        )
+        expected_class = tf.reshape(expected_class, (20,))
+        self.assertAllClose(
+            output["classification"][0, 123], expected_class, atol=0.2
+        )
--- a/keras_cv/models/object_detection/retinanet/retinanet_test.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet_test.py
@@ -231,20 +231,36 @@ class RetinaNetTest(tf.test.TestCase, parameterized.TestCase):
         restored_output = restored_model(input_batch)
         self.assertAllClose(model_output, restored_output)
 
+
 @pytest.mark.large
 class RetinaNetSmokeTest(tf.test.TestCase):
     def test_backbone_preset(self):
-        weights = ["csp_darknet_tiny", "csp_darknet_s",
-                   "csp_darknet_m", "csp_darknet_l",
-                   "csp_darknet_xl", "efficientnetv2_s",
-                   "efficientnetv2_m", "efficientnetv2_l",
-                   "efficientnetv2_b0", "efficientnetv2_b1",
-                   "efficientnetv2_b2", "efficientnetv2_b3",
-                   "mobilenet_v3_small", "mobilenet_v3_small",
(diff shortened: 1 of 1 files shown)


keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -232,23 +232,34 @@ def test_saved_model(self, save_format, filename):
         self.assertAllClose(model_output, restored_output)
 
 
-@pytest.mark.large
```
ianstenbit, line 235: This should still be a `large` test

keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -232,23 +232,34 @@ def test_saved_model(self, save_format, filename):
...
     def test_backbone_preset_weight_loading(self):
         # Check that backbone preset weights loaded correctly
         # TODO(lukewood): need to forward pass test once proper weights are
         # implemented
-        keras_cv.models.RetinaNet.from_preset(
-            "resnet50_v2_imagenet",
-            num_classes=20,
-            bounding_box_format="xywh",
-        )
+        weights = ["csp_darknet_tiny_imagenet", "csp_darknet_l_imagenet",
```
ianstenbit, line 240: Let's load presets without weights here so that we don't have to download all of these weights every CI run.

keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -232,23 +232,34 @@ def test_saved_model(self, save_format, filename):
...
         # TODO(lukewood): need to forward pass test once proper weights are
         # implemented
-        keras_cv.models.RetinaNet.from_preset(
+        model = keras_cv.models.RetinaNet.from_preset(
             "retinanet_resnet50_pascalvoc",
             bounding_box_format="xywh",
         )
+        xs, _ = _create_bounding_box_dataset(bounding_box_format="xywh")
+        output = model(xs)
+        self.assertEqual(output['box'].shape, (xs.shape[0], 49104, 4))
```
ianstenbit, line 265: Since we're loading a preset here, instead of making an assertion about the shape let's make an assertion about the numerics (e.g assert that `output["box"][0, 123, :]` is equal to some constant and `output["class"][0, 123]` is equal to some constant)

ianstenbit (changes requested):
Thanks for the PR!

IMvision12:
@ianstenbit is atol = 0.2 accepted?

```python
self.assertAllClose(output["classification"][0, 123], expected_class, atol=0.2)
```

keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -234,21 +234,81 @@ def test_saved_model(self, save_format, filename):
...
             "retinanet_resnet50_pascalvoc",
             bounding_box_format="xywh",
         )
+        xs, _ = _create_bounding_box_dataset(bounding_box_format="xywh")
+        output = model(xs)
+
+        expected_box = tf.constant(
+            [-0.40364307, 0.15742484, -1.3398054, 0.38104224]
+        )
+        self.assertAllClose(output["box"][0, 123, :], expected_box, atol=0.2)
```
ianstenbit, line 285: This tolerance seems way too high -- with a lower tolerance are you seeing failures?

keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -234,21 +234,81 @@ def test_saved_model(self, save_format, filename):
...
-        # TODO(lukewood): need to forward pass test once proper weights are
-        # implemented
-        keras_cv.models.RetinaNet.from_preset(
-            "resnet50_v2_imagenet",
-            num_classes=20,
-            bounding_box_format="xywh",
-        )
+    def test_backbone_preset(self):
+        weights = [
+            "csp_darknet_tiny",
```
ianstenbit, line 239: I don't think we need to load every preset for every backbone, just 1 preset per backbone should be good (preferably the smallest preset for every backbone)

keras_cv/models/object_detection/retinanet/retinanet_test.py
IMvision12: Yes!

ianstenbit: Ahh wait nevermind I see the problem 😅 

`_create_bounding_box_dataset` creates random noise, so we'd expect to see different results.

We should instead use `xs = tf.ones(1, 512, 512, 3)`

ianstenbit (approved):
Awesome, thanks!

IMvision12:
Will create another PR for yolov8! :)

keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -234,21 +234,63 @@ def test_saved_model(self, save_format, filename):
...
-    def test_backbone_preset_weight_loading(self):
-        # Check that backbone preset weights loaded correctly
-        # TODO(lukewood): need to forward pass test once proper weights are
-        # implemented
-        keras_cv.models.RetinaNet.from_preset(
-            "resnet50_v2_imagenet",
-            num_classes=20,
-            bounding_box_format="xywh",
-        )
+    def test_backbone_preset(self):
```
jbischof, line 237: Sorry for the late response, but this should be using the `presets` and `presets_with_weights` properties to grab the presets. We do not want to enumerate these manually.

keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -234,21 +234,63 @@ def test_saved_model(self, save_format, filename):
...
+        ]
+        for weight in weights:
+            model = keras_cv.models.RetinaNet.from_preset(
+                weight,
+                num_classes=20,
+                bounding_box_format="xywh",
+            )
+            xs, _ = _create_bounding_box_dataset(bounding_box_format="xywh")
+            output = model(xs)
+            self.assertEqual(output["box"].shape, (xs.shape[0], 49104, 4))
```
jbischof, line 254: What are these magic constants?

jbischof (commented):
Some late comments, maybe as a followup @IMvision12?

keras_cv/models/object_detection/retinanet/retinanet_test.py
ianstenbit: 4 is the number of parameters in a box, 49104 is the number of anchors for a 512x512 image.
(Probably we should either comment this or make these named constants)

keras_cv/models/object_detection/retinanet/retinanet_test.py
jbischof: Yes, also we want the same test for YOLOv8 @ianstenbit

ianstenbit: Yeah, I think @IMvision12 mentioned that a follow-up PR for YOLOV8 is en route

keras_cv/models/object_detection/retinanet/retinanet_test.py
IMvision12: I will add comments in the follow-up PR!

keras_cv/models/object_detection/retinanet/retinanet_test.py
IMvision12: @jbischof Yeah i will create a follow-up PR for yolov8 and for retinanet once #1858 is approved :)

Issue #1808, opened 2023-05-19
Refactor `ResNetBackbone` for readability


jbischof:
Our backbone modeling files are very large and hard to read. @LukeWood proposed two changes in #1745:

1. move aliases to their own file
2. move helpers below model declarations

#1745 provides a nice template for this change for `MobileNetBackbone`. Replicate this change for `keras_cv/models/backbones/resnet_v1/`.

ID6109:
[Here](https://github.com/keras-team/keras-cv/issues/1806#issuecomment-1555380503)

PR #1839, opened 2023-05-30
Refactor `ResNetBackbone` for readability

ID6109:
# What does this PR do?

Fixes #1808

PR #1856, opened 2023-06-03
Refactor `ResNetBackbone` for readability

ID6109:
# What does this PR do?

Fixes an incorrect merge in #1808

### DIFF
--- a/keras_cv/models/backbones/resnet_v1/resnet_v1_backbone.py
+++ b/keras_cv/models/backbones/resnet_v1/resnet_v1_backbone.py
@@ -38,6 +38,142 @@ BN_AXIS = 3
 BN_EPSILON = 1.001e-5
 
 
+@keras.utils.register_keras_serializable(package="keras_cv.models")
+class ResNetBackbone(Backbone):
+    """Instantiates the ResNet architecture.
+
+    Reference:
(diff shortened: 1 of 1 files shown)


keras_cv/models/__init__.py
```diff
@@ -61,6 +61,9 @@
 from keras_cv.models.backbones.mobilenet_v3.mobilenet_v3_aliases import (
     MobileNetV3SmallBackbone,
 )
+from keras_cv.models.backbones.mobilenet_v3.mobilenet_v3_backbone import (
```
ianstenbit, line 64: Maybe you have a diffbase issue, but I'm seeing a MobileNetV3 change with this PR.

ianstenbit (changes requested):
Thanks for the PR!

keras_cv/models/__init__.py
ID6109: The merge done in #1839 changed this as well. [Reference](https://github.com/keras-team/keras-cv/pull/1839/files#diff-84c6df388e9b887f61da9c0c85ca04bdbc9f53875ce9f9716caf083411c838c5L65)

jbischof: @ianstenbit this line isn't there in the master branch but should be! But there was a bad merge somewhere but fine to add here.

jbischof (approved):
Thanks!

PR #1858, opened 2023-06-05
Update YOLOv8 Loss

IMvision12:
# What does this PR do?

should we also include IoU loss:

```python
if box_loss == "ciou":
        box_loss = CIoULoss(bounding_box_format="xyxy", reduction="sum")
elif box_loss == "iou":
        box_loss = IoULoss(bounding_box_format="xyxy", reduction="sum")
```
@ianstenbit


keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
```diff
@@ -474,12 +472,12 @@ def compile(
             raise ValueError("User metrics not yet supported for YOLOV8")
 
         if isinstance(box_loss, str):
-            if box_loss == "iou":
-                box_loss = YOLOV8IoULoss(reduction="sum")
+            if box_loss == "ciou":
```
ianstenbit, line 475: Let's also allow "iou" here so we don't break backwards compat (maybe with a comment about why it's included)

keras_cv/models/object_detection/yolo_v8/yolo_v8_label_encoder.py
```diff
@@ -297,7 +297,7 @@ def get_box_metrics(
 
         gt_boxes = tf.repeat(tf.expand_dims(gt_bboxes, axis=2), na, axis=2)
 
-        iou = tf.squeeze(bbox_iou(gt_boxes, pd_boxes), axis=-1)
+        iou = CIoULoss(gt_boxes, pd_boxes)
```
ianstenbit, line 300: It feels very strange to me to be calling a loss here -- this should be something like `bounding_box.compute_ciou` to match `bounding_box.compute_iou`.

How about we move the `ciou` computation helper to `bounding_box/iou.py`, export it as `bounding_box.compute_ciou`, and use it both here and in the loss

(Sorry I should have thought of this during your last PR)

ianstenbit (changes requested):
Thanks for the PR!

Once changes are done, I'm going to run our YOLOV8 benchmark training job as well to verify that this doesn't have any hidden numerical / performance delta.

keras_cv/models/object_detection/yolo_v8/yolo_v8_label_encoder.py
IMvision12: No problem I will add compute_ciou in iou.py :)

keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
```diff
@@ -474,12 +473,14 @@ def compile(
...
+                box_loss = CIoULoss(bounding_box_format="xyxy", reduction="sum")
+            elif box_loss == "iou":
+                box_loss = IoULoss(bounding_box_format="xyxy", reduction="sum")
```
ianstenbit, line 479: To make backwards compat happy, really this should be CIoU loss, although maybe it's best to just allow this as a breaking change since it's a lot more intuitive as an API this way.

So I think I'm happy with this. However, let's print a warning in the "iou" case that says 

```
warnings.warn("YOLOV8 recommends using CIoU loss, but was configured to use standard IoU. Consider using `box_loss='ciou'` instead.")
```

keras_cv/bounding_box/iou.py
```diff
@@ -172,3 +174,71 @@ def compute_iou(
     )
     iou_lookup_table = tf.where(background_mask, mask_val_t, res)
     return iou_lookup_table
+
+
+def compute_ciou(box1, box2, bounding_box_format, eps=1e-7):
```
ianstenbit, line 179: Since this is now defined here, let's drop it from `ciou_loss.py` and just import it from here instead

ianstenbit (changes requested):
Thanks for the changes! Almost ready I think 😄

keras_cv/bounding_box/iou.py
IMvision12: Should we leave 'ciou_loss.py' as a loss in the losses folder given that developers can use it for their own purposes/tasks? wdyt @ianstenbit

ianstenbit: Yes! Sorry I was ambiguous with my comment -- we should still offer `CIoULoss`, we should just import the `compute_ciou` method from here instead of having a second copy of it in `ciou_loss.py`

keras_cv/bounding_box/iou.py
```diff
@@ -172,3 +174,74 @@ def compute_iou(
...
+    """
+    Computes the Complete IoU (CIoU) between two bounding boxes.
```
ianstenbit, line 181: This could also be between two _batches_ of bounding boxes.

PR #1860, opened 2023-06-06
Testing bounding boxes in WithMixedPrecisionTest

### LINKED ISSUE #1784: issue: keras-cv MixedPrecision in KPL
**Describe**

TF 2.12
KerasCV: 0.5

I've tried to use mixed precision in object detection pipelines but it gives error,

```python
keras.mixed_precision.set_global_policy('mixed_float16')
augmenter = keras.Sequential(
    layers=[
        preprocessor,

        keras_cv.layers.RandomFlip(mode="horizontal", bounding_box_format="xywh"),
        keras_cv.layers.RandomChannelShift(value_range=(0, 255), factor=0.2),
        keras_cv.layers.MixUp(alpha=0.4),
        keras_cv.layers.Mosaic(bounding_box_format="xywh")
    ]
)
```
```
ValueError: Error in map_fn:
      Expected `fn` to return a:
        RaggedTensorSpec(TensorShape([None, 4]), tf.float16, 1, tf.int64)
      But it returned a:
        RaggedTensorSpec(TensorShape([None, 4]), tf.float32, 1, tf.int64)
```
[Gist.](https://colab.research.google.com/drive/1-GBk2okZBx-9R1_04aGfzr_GGwvVUUsb?usp=sharing)

jaygala223:
# What does this PR do?

Fixes #1784 

## Before submitting
      Pull Request section?
- [x] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.

## Who can review?

Requesting @ianstenbit and @jbischof for review.


jaygala223:
Hello @ianstenbit, I have added bounding boxes as inputs and made some changes to the test configs as well. Need to fix some failing tests as well. Is this a step in the right direction? Here's a screenshot of the tests for your reference:
image

jaygala223:
Thanks @ianstenbit, the output is pretty long. How can I share it with you? Here are a few screenshots:
image
image
image

ianstenbit:
> Here are a few screenshots

Thanks for sharing! In general, just the error message is a good place to start (in this case it's the `ValueError`)

I left a comment describing the issue (See above)
Thanks!

keras_cv/layers/preprocessing/with_mixed_precision_test.py
```diff
@@ -161,12 +164,26 @@ def test_can_run_in_mixed_precision(self, layer_cls, init_args):
         img = tf.random.uniform(
             shape=(3, 512, 512, 3), minval=0, maxval=255, dtype=tf.float32
         )
-        labels = tf.ones((3,), dtype=tf.float32)
-        inputs = {"images": img, "labels": labels}
+        
+        bounding_boxes = {
+            "boxes": tf.convert_to_tensor(
```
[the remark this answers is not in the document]
jaygala223: Got it. Thanks

jaygala223:
Thanks for your help @ianstenbit. I did exactly as you suggested and we now stand at 19F, 10P, and 4 Skipped.

This keeps popping up: ``To fix, update the `fn_output_signature` (or `dtype`) argument to `map_fn`` along with some notImplementedErrors

Short summary of the remaining failures below:

image
image

ianstenbit:
@jaygala223 some of these issues look like they may be because the layer under test doesn't support bounding boxes (like RandomZoom) -- in those cases we'll need to add box support to those layers.

I'll patch your CL and look at the other errors in some more detail

ianstenbit:
Turns out 5 of the issues were due to using `to_ragged` without a dtype specified.

I've fixed those on your dev branch, and I've also fixed one other one to give you a sense of what the changes look like to resolve the remaining issues.

ianstenbit:
There are 13 remaining failures -- the `ValueError` ones should be straightforward to sort out (like I did above for RandomCrop)

The `TypeError` I haven't looked at.

The `NotImplementedError` we can fix by adding a flag to the test (like the existing `NO_CPU_FP16_KERNEL_LAYERS` list) for layers that don't support boxes yet (or we can add box support!)

jaygala223:
Thanks a lot, @ianstenbit. I'll add a flag (for now) to the tests raising a `NotImplementedError` and I'll also take a look at the other failures!

jaygala223:
Also, one of the tests is now showing this. A StackOverflow [thread](https://stackoverflow.com/questions/51248813/any-ideas-on-what-might-be-causing-this-tensorflow-error-object-was-never-used) suggests that it may be an internal bug with tf v1.2 but I'm on 2.12🤔
image

ianstenbit:
> Also, one of the tests is now showing this. A StackOverflow [thread](https://stackoverflow.com/questions/51248813/any-ideas-on-what-might-be-causing-this-tensorflow-error-object-was-never-used) suggests that it may be an internal bug with tf v1.2 but I'm on 2.12🤔

This is indeed a very weird error message. Looking at the error logs, it seems that there's a dtype issue in the same test (RandomCropAndResize):
```
TensorArray dtype is float16 but Op is trying to write dtype float32 
```

I'd suspect this is the root cause and I'd check out what's happening there.

jaygala223:
Hi @ianstenbit, sorry for the delay. I fixed a few of the tests and added flags for some. Still not able to get around the ones we discussed last time... I'll keep looking

Also, can you take a look at my fix for this layer (in pic) in the commit above? It seems hacky but does the job. We now stand at 6F, 21P, and 6S

image

keras_cv/layers/preprocessing/with_mixed_precision_test.py
```diff
@@ -155,16 +185,56 @@ def test_can_run_in_mixed_precision(self, layer_cls, init_args):
...
+
+        # img = tf.random.uniform(
+        #     shape=(3, 512, 512, 3), minval=0, maxval=255, dtype=tf.float32
+        # )
+        # labels = tf.ones((3,), dtype=tf.float32)
+        # inputs = {"images": img, "labels": labels}
+        
         layer = layer_cls(**init_args)
+
+        if layer_cls == layers.CutMix:
```
ianstenbit, line 235: CutMix doesn't support bounding boxes, so this should just be included in `NO_BOUNDING_BOX_TESTS`

jaygala223: Oh right!

jaygala223:
Hi @ianstenbit, thanks for your tip. All tests are passing now!

> > Also, one of the tests is now showing this. A StackOverflow [thread](https://stackoverflow.com/questions/51248813/any-ideas-on-what-might-be-causing-this-tensorflow-error-object-was-never-used) suggests that it may be an internal bug with tf v1.2 but I'm on 2.12🤔
> 
> This is indeed a very weird error message. Looking at the error logs, it seems that there's a dtype issue in the same test (RandomCropAndResize):
> 
> ```
> TensorArray dtype is float16 but Op is trying to write dtype float32 
> ```
> 
> I'd suspect this is the root cause and I'd check out what's happening there.

keras_cv/layers/preprocessing/with_mixed_precision_test.py
```diff
@@ -30,6 +30,8 @@
             "target_size": (224, 224),
             "crop_area_factor": (0.8, 1.0),
             "aspect_ratio_factor": (3 / 4, 4 / 3),
+            "bounding_box_format": "xywh",
+            "dtype": tf.float32,
```
ianstenbit, line 34: Now that we're using Keras Core, we need to switch to string dtypes instead of `tf.float32`

So for the places in this config where we specify `"dtype": tf.float32` we should instead now specify `"dtype": "float32"`.

After that change I think we can merge this

ianstenbit (changes requested):
Looks good to me -- just one small change to make sure CI is happy and then we can merge 😄

keras_cv/layers/preprocessing/with_mixed_precision_test.py
jaygala223: Got it!

keras_cv/layers/preprocessing/with_mixed_precision_test.py
```diff
@@ -156,14 +211,43 @@ def test_can_run_in_mixed_precision(self, layer_cls, init_args):
                     "operations `tf.image.adjust_saturation`, and "
                     "`tf.image.adjust_hue`. Skipping."
                 )
+            if layer_cls in NO_BOUNDING_BOXES_TESTS:
```
ianstenbit, line 214: Our GPU testing caught this little bug for us -- this only gets checked if there's no GPU present because this is over-indented by one level.

But this also makes the NO_BOUNDING_BOX tests no-ops anyway. How about instead, we check if layer_cls is in `NO_BOUNDING_BOX_TESTS` and if it is we exclude the bounding boxes from the inputs to the layer instead? (So basically just an `if` statement down around line 250

jaygala223: Yes, that makes sense!

jaygala223: CI is still unhappy 😢 I made the change at line 250 as you suggested.

jaygala223:
All tests should be happy now 🥳

keras_cv/layers/preprocessing/with_mixed_precision_test.py
```diff
@@ -86,20 +97,54 @@
...
+        "RandomShear",
+        layers.RandomShear,
+        {
+            "x_factor": 0.3,
+            "x_factor": 0.3,
+            "bounding_box_format": "xyxy",
+            "dtype": "float32",
+        },
+    ),
+    # ("RandomShear", layers.RandomShear, {"x_factor": 0.3, "x_factor": 0.3}),
```
ianstenbit, line 141: Let's either un-comment this or re-add it. Not sure when this originally got commented-out

keras_cv/layers/preprocessing/with_mixed_precision_test.py
ianstenbit: This line should not be here I think -- we can/should run these tests regardless of whether we have a GPU.

keras_cv/layers/preprocessing/with_mixed_precision_test.py
jaygala223: This is weird... the test above it is the same one. I think I may have added it when fixing some styling issues. I'll just remove it to avoid duplicates.

keras_cv/layers/preprocessing/with_mixed_precision_test.py
ianstenbit: Unresolving this -- we should get rid of this check on line 214, as these tests should still be run in this case.

jaygala223: Oh okay

jaygala223: Hello, @ianstenbit, I removed the check on line 214 as you suggested. Sorry for the infinite little errors towards the end😢

ianstenbit:
Thanks for fixing the missing tests -- now one of them is failing. That's good because it means we're actually testing it now :-)

This should be the last bit:
```
FAILED keras_cv/layers/preprocessing/with_mixed_precision_test.py::WithMixedPrecisionTest::test_can_run_in_mixed_precision_CutMix - tensorflow.python.framework.errors_impl.InvalidArgumentError: Exception encountered when calling CutMix.call().

cannot compute Mul as input #1(zero-based) was expected to be a half tensor but is a float tensor [Op:Mul] name: 
```
Same issue as before, I expect that this is just because of a missing cast, or an extra one that needs to be added

jaygala223:
I have fixed it. Thanks @ianstenbit, I couldn't have done it without your support and guidance. 🙌

ianstenbit (approved):
Wohooo -- thank you!

PR #1863, opened 2023-06-06
Retinanet Tests

IMvision12:
# What does this PR do?

@jbischof @ianstenbit


keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -236,9 +238,18 @@ def test_saved_model(self, save_format, filename):
 
 @pytest.mark.large
 class RetinaNetSmokeTest(tf.test.TestCase):
-    def test_resnet_backbone_preset(self):
-        for preset in ResNetBackbone.presets:
-            if preset in ResNetBackbone.presets_with_weights:
+    def test_backbone_preset(self):
+        backbones = [
```
ianstenbit, line 242: I think we can/should probably just take all the presets from `keras_cv.models.Backbone`

(Then we don't need to update this when we add new backbones).

I'm still a little hazy on why we need to test this with every variant of every backbone -- @jbischof what's your take there? (Either way, if this is now testing many backbones it should probably be marked extra_large)

ianstenbit (changes requested):
Thanks for the PR!

Left a question for Jonathan :)

keras_cv/models/object_detection/retinanet/retinanet_test.py
jbischof: Sorry, this is not what I meant! I was suggesting to change the `test_backbone_preset` test to use the `presets` field instead of a static list of presets that need to be updated manually. No new tests required.

IMvision12:
@ianstenbit @jbischof can please check if this is correct, I have removed the static list and used the `presets` field and created the model using `.from_preset`

keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -235,17 +235,11 @@ def test_saved_model(self, save_format, filename):
...
-            "efficientnetv2_b0",
-            "mobilenet_v3_small",
-            "resnet50",
-            "resnet50_v2",
-        ]
-        for weight in weights:
+        for preset in keras_cv.models.RetinaNet.presets:
+            if preset in keras_cv.models.RetinaNet.presets_with_weights:
+                continue
             model = keras_cv.models.RetinaNet.from_preset(
```
ianstenbit, line 241: If I understand correctly, this is what Jonathan was suggesting. This is fine by me.

ianstenbit (approved):

keras_cv/models/object_detection/retinanet/retinanet_test.py
jbischof: Yes, thank you!

keras_cv/models/object_detection/retinanet/test,py
jbischof, line 1: I don't think you meant to commit this file?

ianstenbit: From what I can tell of the diff, this file no longer exists as of the latest commit. So I think we're all good here.

jbischof (approved):

PR #1866, opened 2023-06-08
Add: Visualition API for segmentation maps

### LINKED ISSUE #1865: [Visualization] Adding an API to visualize segmentation maps along with images
**Short Description**

It would be great to include an API `keras_cv.visualization.plot_segmentation_map_gallery()` which would help with the visualition of the image along with their segmentation maps.

An exmaple of using this API would be:
Segmentation maps

ariG23498:
# What does this PR do?

Fixes #1865 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ianstenbit

### DIFF
--- a/keras_cv/visualization/plot_segmentation_mask_gallery.py
+++ b/keras_cv/visualization/plot_segmentation_mask_gallery.py
@@ -57,7 +57,7 @@ def plot_segmentation_mask_gallery(
     )
     ```
 
-    ![Example bounding box gallery](https://i.imgur.com/tJpb8hZ.png)
+    ![Example segmentation mask gallery](https://i.imgur.com/YRswGHz.png)
 
     Args:
         images: a Tensor or NumPy array containing images to show in the
--- a/keras_cv/visualization/plot_segmentation_mask_gallery.py
+++ b/keras_cv/visualization/plot_segmentation_mask_gallery.py
@@ -1,0 +1,104 @@
+# Copyright 2023 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+
+from keras_cv import utils
+from keras_cv.utils import assert_matplotlib_installed
+from keras_cv.visualization.plot_image_gallery import plot_image_gallery
+
+
+def plot_segmentation_mask_gallery(
+    images,
+    value_range,
+    num_classes,
+    y_true=None,
+    y_pred=None,
+    rows=3,
+    cols=3,
+    **kwargs
+):
+    """Plots a gallery of images with corresponding segmentation masks.
+
+    Usage:
+    ```python
+    train_ds = tfds.load(
+        "oxford_iiit_pet", split="train", with_info=False, shuffle_files=True
+    )
+
+    def unpackage_tfds_inputs(inputs):
+        image = inputs["image"]
+        segmentation_mask = inputs["segmentation_mask"]
+        return image, segmentation_mask
+
+    train_ds = train_ds.map(unpackage_tfds_inputs).ragged_batch(16)
+    images, segmentation_masks = next(iter(train_ds.take(1)))
+
+    keras_cv.visualization.plot_segmentation_mask_gallery(
+        images,
+        value_range=(0, 255),
+        num_classes=3, # The number of classes for the oxford iiit pet dataset
+        y_true=segmentation_masks,
+        y_true=segmentation_masks,
+        scale=3,
+        rows=2,
+        cols=2,
+    )
+    ```
+
+    ![Example bounding box gallery](https://i.imgur.com/tJpb8hZ.png)
+
+    Args:
+        images: a Tensor or NumPy array containing images to show in the
+            gallery.
+        value_range: value range of the images. Common examples include
+            `(0, 255)` and `(0, 1)`.
+        num_classes: number of segmentation classes.
+        y_true: (Optional) a Tensor or NumPy array representing the ground truth
+            segmentation masks.
+        y_pred: (Optional)  a Tensor or NumPy array representing the predicted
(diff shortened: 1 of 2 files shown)


keras_cv/visualization/plot_segmentation_mask_gallery.py
```diff
@@ -0,0 +1,104 @@
...
+        y_pred: (Optional)  a Tensor or NumPy array representing the predicted
+            segmentation masks.
+        kwargs: keyword arguments to propagate to
+            `keras_cv.visualization.plot_image_gallery()`.
+    """
+    assert_matplotlib_installed("plot_segmentation_mask_gallery")
+
+    plotted_images = utils.to_numpy(images)
+
+    # Segmentation maps are of 1 channel, here we repeat the channel 3 times
```
ianstenbit, line 79: This seems to assume that the masks are not one-hot encoded. A few questions:

- What if the masks are one-hot encoded
- How does this visualization look when num_classes > 1? 

A little demo script demonstrating all of these would be awesome

ianstenbit: (Perhaps if they are one-hot (because the last dimension is > 1), we should just use argmax to convert them before visualizing

ianstenbit (changes requested):
Thanks for the PR!

keras_cv/visualization/plot_segmentation_mask_gallery.py
ariG23498: Do you think we should have a `is_one_hot` flag in the parameters? -- This was a great edge case (thanks for this)

ianstenbit: I'd prefer to infer this, as it should be relatively practical to do so

keras_cv/visualization/plot_segmentation_mask_gallery.py
```diff
@@ -0,0 +1,104 @@
...
+    num_classes,
+    y_true=None,
+    y_pred=None,
+    rows=3,
+    cols=3,
+    **kwargs
+):
+    """Plots a gallery of images with corresponding segmentation masks.
+
+    Usage:
```
jbischof, line 34: Please move examples section to the end of the docstring. Thank you!

keras_cv/visualization/plot_segmentation_mask_gallery.py
```diff
@@ -0,0 +1,123 @@
...
+
+from keras_cv import utils
+from keras_cv.utils import assert_matplotlib_installed
+from keras_cv.visualization.plot_image_gallery import plot_image_gallery
+
+
+def build_compatile_segmentation_masks(segmentation_masks):
+    rank = len(segmentation_masks.shape)
+    if rank == 3:
+        # (B, H, W)
```
ianstenbit, line 25: This could actually be a bit more complicated.

Users could pass un-batched masks with shape `(H, W, num_channels)`

Let's avoid this by stating in the docstring that masks and images must include a batch axis (it's good to include shape info in the docstring anyway)

keras_cv/visualization/plot_segmentation_mask_gallery.py
```diff
@@ -0,0 +1,123 @@
...
+
+    train_ds = train_ds.map(unpackage_tfds_inputs).ragged_batch(16)
+    images, segmentation_masks = next(iter(train_ds.take(1)))
+
+    keras_cv.visualization.plot_segmentation_mask_gallery(
+        images,
+        value_range=(0, 255),
+        num_classes=3, # The number of classes for the oxford iiit pet dataset
+        y_true=segmentation_masks,
+        y_pred=segmentation_masks,
```
ianstenbit, line 81: For the demo we can probably just show `y_true` instead of copying the same thing into `y_pred`

keras_cv/visualization/plot_segmentation_mask_gallery.py
```diff
@@ -0,0 +1,123 @@
...
+# limitations under the License.
+
+import numpy as np
+
+from keras_cv import utils
+from keras_cv.utils import assert_matplotlib_installed
+from keras_cv.visualization.plot_image_gallery import plot_image_gallery
+
+
+def build_compatile_segmentation_masks(segmentation_masks):
```
ianstenbit, line 22: maybe `reshape_masks`?

keras_cv/visualization/plot_segmentation_mask_gallery.py
```diff
@@ -0,0 +1,123 @@
...
+    rank = len(segmentation_masks.shape)
+    if rank == 3:
+        # (B, H, W)
+        return segmentation_masks[..., np.newaxis]
+    elif rank == 4:
+        # (B, H, W, num_channels) OR (B, H, W, 1)
+        if segmentation_masks.shape[-1] == 1:
+            return segmentation_masks.repeat(repeats=3, axis=-1)
```
ianstenbit, line 30: This `3` is for RGB, right?

That's worth a code comment because it's a bit confusing as a magic constant

ianstenbit (changes requested):
Nice! Love the example -- I think this is super close

keras_cv/visualization/plot_segmentation_mask_gallery.py
ariG23498: That does make more sense! Thanks for the suggestion. 🤗

keras_cv/visualization/plot_segmentation_mask_gallery.py
```diff
@@ -0,0 +1,123 @@
...
+    )
+    ```
+
+    ![Example segmentation mask gallery](https://i.imgur.com/aRkmJ1Q.png)
+    """
+    assert_matplotlib_installed("plot_segmentation_mask_gallery")
+
+    plotted_images = utils.to_numpy(images)
+
+    plotted_y_true = utils.to_numpy(y_true)
```
ianstenbit, line 98: This makes me think `y_true` is required -- should we wrap it in an `if y_true is not None:` (like we do for `y_pred`)

ianstenbit (approved):
This looks good to me after one small fix -- thank you!

keras_cv/visualization/plot_segmentation_mask_gallery.py
ariG23498: My rationale here is that if we do not keep `y_true` required this API is as good as `plot_image_gallery`. Does that not make the usage redundant?

ariG23498: That makes sense to me!

ianstenbit (approved):
Great -- thank you!

PR #1867, opened 2023-06-08
Fix RetinaNet anchor generator guard clause

### LINKED ISSUE #1862: RetinaNet anchor_generator parameter cannot be used
ReproCode:

```
backbone = kcv.models.ResNet50Backbone.from_preset("resnet50_imagenet", include_rescaling=False)
    
anchor_generator = kcv.layers.AnchorGenerator(bounding_box_format='xywh',
                                              sizes = [32.0, 64.0, 128.0, 256.0, 512.0],# custom
                                              strides = [2**i for i in range(3, 8)],    # default
                                              scales=[2**x for x in [0, 1 / 3, 2 / 3]], #default
                                              aspect_ratios=[0.5, 1.0, 2.0])            #default

prediction_decoder = kcv.layers.MultiClassNonMaxSuppression(bounding_box_format="xywh", from_logits=True,
                                                            confidence_threshold=0.4,
                                                            iou_threshold=0.5,
                                                            max_detections=100,
                                                            max_detections_per_class=25)

label_encoder = kcv.models.object_detection.retinanet.RetinaNetLabelEncoder(anchor_generator=anchor_generator,
                                                                            bounding_box_format='xywh')

model = kcv.models.RetinaNet(num_classes=len(CLASSES), bounding_box_format='xywh',
                             backbone=backbone,
                             label_encoder=label_encoder,
                             anchor_generator=anchor_generator,
                             prediction_decoder=prediction_decoder)
```

Error.

ValueError: `anchor_generator` is only to be provided when both `label_encoder` and `prediction_decoder` are both `None`. Received `anchor_generator=<keras_cv.layers.object_detection.anchor_generator.AnchorGenerator object at 0x7fd0c87c75b0>` `label_encoder=<keras_cv.models.object_detection.retinanet.retinanet_label_encoder.RetinaNetLabelEncoder object at 0x7fd0482f6b90>`, `prediction_decoder=<keras_cv.layers.object_detection.multi_class_non_max_suppression.MultiClassNonMaxSuppression object at 0x7fd0482f69e0>`. To customize the behavior of the anchor_generator inside of a custom `label_encoder` or custom `prediction_decoder` you should provide both to `RetinaNet`, and ensure that the `anchor_generator` provided to both is identical

Unfortunately, it is not possible to follow the instructions in the error message because MultiClassNonMaxSuppression does not have a anchor_generator parameter.

It looks like this was a bad refactor. The predictions decoder probably used to include both:
* decoding from anchor-relative deltas to boxes
* non-max-suppression

But now, RetinaNet always decodes boxes using the function [decode_predictions](https://github.com/keras-team/keras-cv/blob/efbff2f0a994dd9e22e56296dd9af6a763b78b1f/keras_cv/models/object_detection/retinanet/retinanet.py#L279) which does:
*  _decode_deltas_to_boxes
* applies prediction_decoder (typically non-max suppression)

Bottom line: decoding deltas to boxes is no longer part of the customizable prediction decoder which is why MultiClassNonMaxSuppression does not have an anchor_generator parameter.

Probably, the only fix needed is to the [error message and the test triggering it](https://github.com/keras-team/keras-cv/blob/efbff2f0a994dd9e22e56296dd9af6a763b78b1f/keras_cv/models/object_detection/retinanet/retinanet.py#L142):
The required condition to test is that the same anchor_generator is pass to the custom label encoder AND RetinaNet itself.

ianstenbit:
Fixes #1862

### DIFF
--- a/keras_cv/models/object_detection/retinanet/retinanet.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet.py
@@ -218,7 +218,7 @@ class RetinaNet(Task):
             **kwargs,
         )
         self.label_encoder = label_encoder
-        self.anchor_generator = anchor_generator
+        self.anchor_generator = label_encoder.anchor_generator
         self.bounding_box_format = bounding_box_format
         self.num_classes = num_classes
         self.backbone = backbone
--- a/keras_cv/models/object_detection/retinanet/retinanet.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet.py
@@ -139,29 +139,28 @@ class RetinaNet(Task):
         box_head=None,
         **kwargs,
     ):
-        if anchor_generator is not None and (
-            prediction_decoder or label_encoder
-        ):
+        if anchor_generator is not None and label_encoder is not None:
             raise ValueError(
                 "`anchor_generator` is only to be provided when "
-                "both `label_encoder` and `prediction_decoder` are both "
-                f"`None`. Received `anchor_generator={anchor_generator}` "
-                f"`label_encoder={label_encoder}`, "
-                f"`prediction_decoder={prediction_decoder}`. To customize the "
-                "behavior of the anchor_generator inside of a custom "
-                "`label_encoder` or custom `prediction_decoder` you should "
+                "`label_encoder` is `None`. Received `anchor_generator="
+                f"{anchor_generator}`, label_encoder={label_encoder}`. To "
+                "customize the behavior of the anchor_generator inside of a "
+                "custom `label_encoder` you should provide both to `RetinaNet`"
                 "provide both to `RetinaNet`, and ensure that the "
                 "`anchor_generator` provided to both is identical"
             )
-        anchor_generator = (
-            anchor_generator
-            or RetinaNet.default_anchor_generator(bounding_box_format)
-        )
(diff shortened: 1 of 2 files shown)


keras_cv/models/object_detection/retinanet/retinanet.py
```diff
@@ -139,29 +139,28 @@ def __init__(
...
-                "behavior of the anchor_generator inside of a custom "
-                "`label_encoder` or custom `prediction_decoder` you should "
+                "`label_encoder` is `None`. Received `anchor_generator="
+                f"{anchor_generator}`, label_encoder={label_encoder}`. To "
+                "customize the behavior of the anchor_generator inside of a "
+                "custom `label_encoder` you should provide both to `RetinaNet`"
                 "provide both to `RetinaNet`, and ensure that the "
                 "`anchor_generator` provided to both is identical"
             )
-        anchor_generator = (
```
jbischof, line 156: Doesn't this mean that we have no anchor generator by default?

ianstenbit: Ahh good catch, in this flow the resulting `anchor_generator` is `None` which would break the model. I've fixed the issue and added test coverage.

jbischof (approved):
Thanks!

PR #1872, opened 2023-06-09
Create `presets_without_weights` classproperty

jbischof:
This property is already determined by the `presets` and `presets_with_weights` properties, so create programatically.

Providing this property can save boilerplate for users.

### DIFF
--- a/keras_cv/models/backbones/backbone.py
+++ b/keras_cv/models/backbones/backbone.py
@@ -53,9 +53,16 @@ class Backbone(keras.Model):
 
     @classproperty
     def presets_with_weights(cls):
-        """Dictionary of preset names and configurations that include
-        weights."""
+        """Dictionary of preset names and configs that include weights."""
         return {}
+
+    @classproperty
+    def presets_without_weights(cls):
+        """Dictionary of preset names and configs that don't include weights."""
+        return {
+            preset: cls.presets[preset]
+            for preset in set(cls.presets) - set(cls.presets_with_weights)
+        }
 
     @classmethod
     def from_preset(
@@ -64,8 +71,7 @@ class Backbone(keras.Model):
         load_weights=None,
         **kwargs,
     ):
-        """Instantiate {{model_name}} model from preset architecture and
-        weights.
+        """Instantiate {{model_name}} model from preset config and weights.
 
         Args:
             preset: string. Must be one of "{{preset_names}}".
--- a/keras_cv/models/object_detection/retinanet/retinanet_test.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet_test.py
@@ -252,9 +252,7 @@ class RetinaNetTest(tf.test.TestCase, parameterized.TestCase):
 @pytest.mark.large
 class RetinaNetSmokeTest(tf.test.TestCase):
     def test_backbone_preset(self):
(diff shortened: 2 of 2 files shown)


keras_cv/models/backbones/backbone.py
```diff
@@ -53,19 +53,25 @@ def presets(cls):
 
     @classproperty
     def presets_with_weights(cls):
-        """Dictionary of preset names and configurations that include
-        weights."""
+        """Dictionary of preset names and configs that include weights."""
         return {}
 
+    @classproperty
+    def presets_without_weights(cls):
```
ianstenbit, line 60: Heck yeah -- I dig this 😎 
Thanks!

keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -252,9 +252,7 @@ def test_call_with_custom_label_encoder(self):
 @pytest.mark.large
 class RetinaNetSmokeTest(tf.test.TestCase):
     def test_backbone_preset(self):
-        for preset in keras_cv.models.RetinaNet.presets:
-            if preset in keras_cv.models.RetinaNet.presets_with_weights:
-                continue
+        for preset in keras_cv.models.RetinaNet.presets_without_weights:
```
ianstenbit, line 255: Maybe now that we don't have this extra logic we can just make this a parameterized test? (This will make error messages much nicer)

ianstenbit (approved):
Thanks Jonathan!

keras_cv/models/backbones/backbone.py
jbischof: Actually I forgot to put this on the Task as well!!!

jbischof:
@ianstenbit I'm getting failures in this and the master branch running large tests ([gist](https://gist.github.com/jbischof/5e6e974589846eff431349e4c73b136d)). I wonder how this wasn't caught by gcb?

jbischof:
Running GCB to see if it catches them.

I will say that the local error messages were more helpful after the change!

ianstenbit:
@jbischof I updated this PR to also create some test presets (the smallest preset from each backbone, for use in testing).

This won't pass GCB until #1873 is merged, but otherwise I think it's ready

keras_cv/models/backbones/test_backbone_presets.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Backbone presets for tests. Includes smallest preset for each backbone."""
+
+test_backbone_presets = [
```
jbischof, line 17: Why not add this to `backbone_presets.py`?

jbischof (commented):
Thanks! Feel free to push

Again this will show up as a comment.

keras_cv/models/backbones/test_backbone_presets.py
ianstenbit: Just to keep the test code separate from the non-test code. 🤷

PR #1873, opened 2023-06-10
Rationalize pyramid_level_inputs tests

jbischof:
Our current testing our `pyramid_level_inputs` involves two fairly redundant tests that don't verify what we actually care about:
1. are the expected levels there
2. are they the expected shape

As a result, our current tests failed to catch incorrect `feature_pyramid_inputs` in our new DenseNet models. They current grab the tensor AFTER pooling when it should be the last tensor BEFORE pooling.

### DIFF
--- a/keras_cv/models/backbones/densenet/densenet_backbone_test.py
+++ b/keras_cv/models/backbones/densenet/densenet_backbone_test.py
@@ -106,7 +106,7 @@ class DenseNetBackboneTest(tf.test.TestCase, parameterized.TestCase):
         self.assertEquals(list(outputs.keys()), levels)
         self.assertEquals(
             outputs["P2"].shape,
-            [None, input_size // 2**1, input_size // 2**1, 256],
+            [None, input_size // 2**2, input_size // 2**2, 256],
         )
         self.assertEquals(
             outputs["P3"].shape,
--- a/keras_cv/models/backbones/densenet/densenet_backbone_test.py
+++ b/keras_cv/models/backbones/densenet/densenet_backbone_test.py
@@ -102,18 +102,24 @@ class DenseNetBackboneTest(tf.test.TestCase, parameterized.TestCase):
         input_size = 256
         inputs = tf.keras.Input(shape=[input_size, input_size, 3])
         outputs = backbone_model(inputs)
-        expected_levels = ["P2", "P3", "P4", "P5"]
-        self.assertEquals(list(outputs.keys()), expected_levels)
-        # Size for each feature map at Pn is represents a feature map 2^n
-        # times smaller in width and height than the input image.
-        for level in model.pyramid_level_inputs:
-            level_int = int(level[1:])
-            self.assertEquals(
-                outputs[level].shape[1], input_size / 2**level_int
-            )
-            self.assertEquals(
-                outputs[level].shape[2], input_size / 2**level_int
-            )
+        levels = ["P2", "P3", "P4", "P5"]
+        self.assertEquals(list(outputs.keys()), levels)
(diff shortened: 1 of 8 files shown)


jbischof:
Now we're not failing on the RetinaNet tests but are getting the bizarre error:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

e = _NotOkStatusException(), name = None

    def raise_from_not_ok_status(e, name):
      e.message += (" name: " + name if name is not None else "")
>     raise core._status_to_exception(e) from None  # pylint: disable=protected-access
E     tensorflow.python.framework.errors_impl.ResourceExhaustedError: {{function_node __wrapped__AddV2_device_/job:localhost/replica:0/task:0/device:GPU:0}} failed to allocate memory [Op:AddV2]

/usr/local/lib/python3.8/dist-packages/tensorflow/python/framework/ops.py:7215: ResourceExhaustedError
----------------------------- Captured stderr call -----------------------------
2023-06-10 15:34:37.327957: W tensorflow/tsl/framework/bfc_allocator.cc:479] Allocator (GPU_0_bfc) ran out of memory trying to allocate 392.00MiB (rounded to 411041792)requested by op AddV2
If the cause is memory fragmentation maybe the environment variable 'TF_GPU_ALLOCATOR=cuda_malloc_async' will improve the situation.
```

keras_cv/models/backbones/densenet/densenet_backbone_test.py
```diff
@@ -92,24 +92,28 @@ def test_saved_alias_model(self, save_format, filename):
...
         )
-        inputs = tf.keras.Input(shape=[256, 256, 3])
+        input_size = 256
+        inputs = tf.keras.Input(shape=[input_size, input_size, 3])
         outputs = backbone_model(inputs)
-
-        levels = ["P2", "P3", "P4"]
-        self.assertLen(outputs, 3)
-        self.assertEquals(list(outputs.keys()), levels)
-        self.assertEquals(outputs["P2"].shape, [None, 32, 32, 128])
```
ianstenbit, line 110: good catch -- one question though:

wdyt about keeping these assertions as-is but using `input_size / 2 ** {2/3/4}` as the width/height fields?

I personally find the `for` loop approach a bit harder to parse, so it makes the test assertion a little less explicit to me.

ianstenbit (changes requested):
Thanks!

keras_cv/models/backbones/densenet/densenet_backbone_test.py
jbischof: How would you handle different models supporting different P-levels? I guess you would just use the expected ones?

ianstenbit: Right -- I think the way we had it before made this pretty overt (we just have an assertion for every separate P-level)

ianstenbit:
GCB is now only failing due to the OOM which should be (hopefully) fixed by #1872.

@jbischof if you're happy with this I can merge -- lmk what you think

ianstenbit (approved):

ianstenbit:
> Thanks! Feel free to push.
> 
> Question: do we actually care about asserting the number of filters in the layer? We don't have clear expectations there.
> 
> Note: you cannot seem to approve one's own PR, so this will appear as a comment

I'm not sure how much we care, but it's nice to be explicit about it and detect changes to it (we'd also detect changes with weight loading tests but those error messages would be quite nasty)

PR #1885, opened 2023-06-15
Add "feature_pyramid" as parameter inside RetinaNet __init__

### LINKED ISSUE #1884: RetinaNet Custom Feature Pyramid
Hi! I'd like to request a simple feature: the possibility of initializing the RetinaNet class using a custom "feature_pyramid", much like the "box_head" and "classification_head" parameters. It shouldn't take that much time to add this feature in, and it would make it a lot easier to create a completely custom object detector using the RetinaNet meta-architecture. In my case, I'm trying to build a RetinaNet detector using MobileNetV1's convolutional blocks, which I'd like to try in the feature pyramid too. Thanks for your time and for the incredible work you've been doing on this library!

gianlucasama:
Fixes #1884. 

This PR adds a "feature_pyramid" parameter so that users can use their own custom FPN inside the RetinaNet meta-architecture.

### DIFF
--- a/keras_cv/models/object_detection/retinanet/feature_pyramid.py
+++ b/keras_cv/models/object_detection/retinanet/feature_pyramid.py
@@ -49,10 +49,4 @@ class FeaturePyramid(keras.layers.Layer):
         p5_output = self.conv_c5_3x3(p5_output, training=training)
         p6_output = self.conv_c6_3x3(c5_output, training=training)
         p7_output = self.conv_c7_3x3(tf.nn.relu(p6_output), training=training)
-        return {
-            "P3": p3_output,
-            "P4": p4_output,
-            "P5": p5_output,
-            "P6": p6_output,
-            "P7": p7_output,
-        }
+        return p3_output, p4_output, p5_output, p6_output, p7_output
--- a/keras_cv/models/object_detection/retinanet/retinanet.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet.py
@@ -119,6 +119,10 @@ class RetinaNet(Task):
             default `prediction_decoder` layer is a
             `keras_cv.layers.MultiClassNonMaxSuppression` layer, which uses
             a Non-Max Suppression for box pruning.
+        feature_pyramid: (Optional) A `keras.layers.Layer` that is
+            responsible for extracting features from intermediate backbone
+            layers. If not provided, the reference implementation from
+            the paper will be used.
         classification_head: (Optional) A `keras.Layer` that performs
             classification of the bounding boxes. If not provided, a simple
             ConvNet with 3 layers will be used.
@@ -135,6 +139,7 @@ class RetinaNet(Task):
         anchor_generator=None,
         label_encoder=None,
         prediction_decoder=None,
+        feature_pyramid=None,
         classification_head=None,
         box_head=None,
         **kwargs,
@@ -169,7 +174,7 @@ class RetinaNet(Task):
         feature_extractor = get_feature_extractor(
             backbone, extractor_layer_names, extractor_levels
         )
-        feature_pyramid = FeaturePyramid()
+        feature_pyramid = feature_pyramid or FeaturePyramid()
 
         prior_probability = keras.initializers.Constant(
             -np.log((1 - 0.01) / 0.01)
--- a/keras_cv/models/object_detection/retinanet/retinanet.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet.py
@@ -122,8 +122,8 @@ class RetinaNet(Task):
             a Non-Max Suppression for box pruning.
         feature_pyramid: (Optional) A `keras.layers.Layer` that produces
             a list of 4D feature maps (batch dimension included)
-            when called on the `backbone`. If not provided, the reference
-            implementation from the paper will be used.
+            when called on the pyramid-level outputs of the `backbone`.
+            If not provided, the reference implementation from the paper will be used.
         classification_head: (Optional) A `keras.Layer` that performs
             classification of the bounding boxes. If not provided, a simple
             ConvNet with 3 layers will be used.
--- a/keras_cv/models/object_detection/retinanet/retinanet.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet.py
@@ -93,8 +93,8 @@ class RetinaNet(Task):
             Refer
             [to the keras.io docs](https://keras.io/api/keras_cv/bounding_box/formats/)
             for more details on supported bounding box formats.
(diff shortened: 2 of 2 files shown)


jbischof (changes requested):
Thanks!

keras_cv/models/object_detection/retinanet/retinanet.py
```diff
@@ -169,7 +174,7 @@ def __init__(
         feature_extractor = get_feature_extractor(
             backbone, extractor_layer_names, extractor_levels
         )
-        feature_pyramid = FeaturePyramid()
+        feature_pyramid = feature_pyramid or FeaturePyramid()
```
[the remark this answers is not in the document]
gianlucasama: So we could make the FPN output a tuple and iterate over it like you did. Right?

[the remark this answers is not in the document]
jbischof: Yes, there's no reason to return a dict and hardcode the outputs extracted. The FPN should output whatever one wants to use.

jbischof:
Can you run `./shell/format.sh` to fix the formatting issues @gianlucasama?

jbischof (approved):
Looks great, thank you!

jbischof:
@gianlucasama can you merge with master? We need to pick up some recent infra changes

keras_cv/models/object_detection/retinanet/feature_pyramid.py
```diff
@@ -49,10 +49,4 @@ def call(self, inputs, training=False):
...
         p6_output = self.conv_c6_3x3(c5_output, training=training)
         p7_output = self.conv_c7_3x3(tf.nn.relu(p6_output), training=training)
-        return {
-            "P3": p3_output,
-            "P4": p4_output,
-            "P5": p5_output,
-            "P6": p6_output,
-            "P7": p7_output,
-        }
+        return p3_output, p4_output, p5_output, p6_output, p7_output
```
ianstenbit, line 52: What is the purpose of switching from dictionary to tuple format here?

keras_cv/models/object_detection/retinanet/retinanet.py
```diff
@@ -119,6 +120,10 @@ class RetinaNet(Task):
             default `prediction_decoder` layer is a
             `keras_cv.layers.MultiClassNonMaxSuppression` layer, which uses
             a Non-Max Suppression for box pruning.
+        feature_pyramid: (Optional) A `keras.layers.Layer` that produces
+            a list of 4D feature maps (batch dimension included)
+            when called on the `backbone`. If not provided, the reference
```
ianstenbit, line 125: When called on the pyramid-level outputs of the backbone

ianstenbit (changes requested):
Thanks for the PR!

Adding the new parameter looks good to me, but I'd like to understand the motivation behind changing the API expectation of a feature pyramid

keras_cv/models/object_detection/retinanet/feature_pyramid.py
jbischof: Because otherwise we're hardcoding the levels output by the FPN for no obvious payoff. There's a lot more flexibility here.

keras_cv/models/object_detection/retinanet/retinanet.py
ianstenbit: Not sure I follow -- the code looks like 
```
backbone_outputs = feature_extractor(images)
features = feature_pyramid(backbone_outputs)
```

I don't really know what it means for a layer to be called on another layer -- it's called on the output of the other layer

jbischof: Yes, you are right! I was thinking of the original design where the FPN also handled feature extraction. I think it's OK to standardize on `pyramid_level_inputs`. @gianlucasama can you update the docstring?

keras_cv/models/object_detection/retinanet/feature_pyramid.py
gianlucasama: In the RetinaNet class' "__init__" we were iterating over the entire dictionary created by the FeaturePyramid class to attach the prediction heads to the FPN, so @jbischof suggested to simplify the contract users will have to adhere to, while also "cleaning" the code. I think it makes sense this way, but the last word is on you guys.

ianstenbit (approved):
Looks good to me after the docstring update -- thank you!

gianlucasama:
We forgot to add "feature_pyramid" inside the "get_config" method of the RetinaNet class. Also, why aren't we doing "config = super().get_config()" inside that method?

jbischof:
Thanks @gianlucasama are you able to file a followup PR?

Our testing for serialization and saving doesn't really cover custom objects unfortunately. Maybe you could add a test case (check that it fails before "fixing" it!).

gianlucasama:
I'm taking a look at it. For now, it seems that just by adding a few lines of code inside the "get_config" and "from_config" method of the RetinaNet class, saving and loading in the ".keras" format works if, when loading, one specifies the custom objects. Saving in "tf" format works but there is a weird error popping up when loading:
`
  File "/home/gianlucasama/Github/Deckard/deckard/deckard/pipelines.py", line 1013, in <module>
    model = tf.keras.saving.load_model(
  File "/home/gianlucasama/.local/lib/python3.10/site-packages/keras/saving/saving_api.py", line 212, in load_model
    return legacy_sm_saving_lib.load_model(
  File "/home/gianlucasama/.local/lib/python3.10/site-packages/keras/utils/traceback_utils.py", line 70, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/home/gianlucasama/.local/lib/python3.10/site-packages/tensorflow/python/trackable/base.py", line 205, in _method_wrapper
    result = method(self, *args, **kwargs)
TypeError: Functional.__init__() missing 2 required positional arguments: 'inputs' and 'outputs'
`
If you have any idea of what this is, please share some insight. Once this is figured out I'll write the necessary test cases to ensure custom model saving and loading in the different formats. I'll keep you updated on the relevant progress.

gianlucasama:
Ok, it seems that by fixing the "get_config" and the "from_config" methods of the RetinaNet class, saving and loading with custom objects works in all formats, but the latter only works by specifying the custom objects, including the RetinaNet class in the case of the "keras" and "h5" formats, which I don't really understand since it is registered using:
`@keras.utils.register_keras_serializable(package="keras_cv")`
If this is still fine for you, I'll go ahead and make a PR to fix and test the whole the thing.

jbischof:
Yes let's make a PR. It will be easier to understand the code 😸

PR #1888, opened 2023-06-19
[Contrastive Trainer] Check for duplicate projectors

### LINKED ISSUE #1887: Double gradient calculation for the projectors in Contrastive Trainer
https://github.com/keras-team/keras-cv/blob/b3798d8140605b651ac46f40fe7072ea08175720/keras_cv/training/contrastive/contrastive_trainer.py#L233-L238

https://github.com/keras-team/keras-cv/blob/b3798d8140605b651ac46f40fe7072ea08175720/keras_cv/training/contrastive/contrastive_trainer.py#L240-L247

In the above two code snippets we assume that both the projectors are distinct. What if we provide only one projector which is then replicated to fill a tuple as shown below 👇 

https://github.com/keras-team/keras-cv/blob/b3798d8140605b651ac46f40fe7072ea08175720/keras_cv/training/contrastive/contrastive_trainer.py#L112-L114

In that case the gradients of the projector is calculated twice and also the projector is updated twice.

If I have read this code snippet differently and am mistaken please feel free to close this issue.

CC: @ID6109

ariG23498:
Fixes #1887 

CC: @ianstenbit

### DIFF
--- a/keras_cv/training/contrastive/contrastive_trainer.py
+++ b/keras_cv/training/contrastive/contrastive_trainer.py
@@ -109,6 +109,8 @@ class ContrastiveTrainer(keras.Model):
             augmenter if type(augmenter) is tuple else (augmenter, augmenter)
         )
         self.encoder = encoder
+        # Check to see if the projector is being shared or are distinct.
+        self._is_shared_projector = False if type(projector) is tuple else True
         self.projectors = (
             projector if type(projector) is tuple else (projector, projector)
         )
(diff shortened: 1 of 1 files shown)


keras_cv/training/contrastive/contrastive_trainer.py
```diff
@@ -109,6 +109,8 @@ def __init__(
             augmenter if type(augmenter) is tuple else (augmenter, augmenter)
         )
         self.encoder = encoder
+        # Check to see if the projector is being shared or are distinct.
+        self._is_shared_projector = False if type(projector) is tuple else True
```
ianstenbit, line 113: How about `not isinstance(projector, tuple)`?

keras_cv/training/contrastive/contrastive_trainer.py
```diff
@@ -230,19 +232,23 @@ def train_step(self, data):
                 regularization_losses=self.encoder.losses,
             )
 
+        # If the projector is shared, then take the trainable weights of just
+        # one of the projector in the tuple. If not, use both the projectors.
```
ianstenbit, line 236: nit: "one of the projectors"

ianstenbit (changes requested):
Thanks! Looks good after a few minor changes

Also -- contrastive tests are not run on CI, so please run `pytest --run_extra_large keras_cv/training/contrastive`

ariG23498:
The tests pass locally!

ianstenbit (approved):
Thanks for the fix!

PR #1898, opened 2023-06-26
Mark TF saving tests as "extra_large"

### LINKED ISSUE #1896: Remove tf saving tests
Right now we can't merge any PRs because GCB is timing out after 30m. Let's remove TF saving tests, which are very slow and less important than the new Keras saving format going forward.

jbischof:
Fixes #1896


keras_cv/conftest.py
```diff
@@ -61,6 +61,8 @@ def pytest_collection_modifyitems(config, items):
     for item in items:
         if "keras_format" in item.name:
             item.add_marker(skip_keras_saving_test)
+        if "tf_format" in item.name:
```
ianstenbit, line 64: Yeah this seems a lot cleaner 😅

ianstenbit (approved):

PR #1899, opened 2023-06-27
Assert metric reset works correctly in RetinaNet

### LINKED ISSUE #1794: re-compiling RetinaNet with `metrics=[]` or `metrics=None` does not delete metrics from previous compile calls
Repro snipped:

```
    "resnet50_imagenet",
    num_classes=len(class_mapping),
    bounding_box_format="center_xywh",
)
coco_metrics = keras_cv.metrics.BoxCOCOMetrics(
    bounding_box_format="center_xywh", evaluate_freq=20
)
model.compile(
    optimizer=keras.optimizers.SGD(global_clipnorm=10.),
    classification_loss="focal",
    box_loss="smoothl1",
    metrics=[coco_metrics]
)
model.fit(
  train_ds,
  epochs=10
)
# THEN:
model.compile(
    optimizer=keras.optimizers.SGD(global_clipnorm=10.),
    classification_loss="focal",
    box_loss="smoothl1",
    metrics=[]
)
model.fit(
  train_ds,
  epochs=10
)
```

Should delete the metrics, but it doesnt!

ianstenbit:
Fixes #1794. That issue wasn't reproducible, and metric reset seems to be WAI, but I've added test coverage for it.

### DIFF
--- a/keras_cv/models/object_detection/retinanet/retinanet_test.py
+++ b/keras_cv/models/object_detection/retinanet/retinanet_test.py
@@ -57,6 +57,32 @@ class RetinaNetTest(tf.test.TestCase, parameterized.TestCase):
         # self.assertIsNotNone(retinanet.backbone.get_layer(name="rescaling"))
         # TODO(lukewood): test compile with the FocalLoss class
 
+    def test_retinanet_recompilation_without_metrics(self):
+        retinanet = keras_cv.models.RetinaNet(
+            num_classes=20,
+            bounding_box_format="xywh",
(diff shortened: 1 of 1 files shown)


ianstenbit:
> How expensive is this test? If it's not catching a likely bug we might want to mark `extra_large`

Over 5 runs, the average runtime of this test on my machine is 0.54 seconds. It's not _super_ likely to catch a bug, but it's also a cheap test.

jbischof (approved):
Thanks!

PR #1900, opened 2023-06-27
Add YOLOV8 backbone presets to the global backbone registry

ianstenbit:
We didn't do this initially because we thought the YOLOV8 backbone wouldn't stick around for long, but that is no longer true.


jbischof:
I think the main issue is whether these presets are useful for other `Task`s besides YOLOv8. WDYT?

jbischof (approved):
Thanks!

PR #1901, opened 2023-06-27
Port preprocessing layers and bounding box utils to Keras Core

ianstenbit:
CI is failing due to auth issues loading Keras Core, but I've run tests locally and CI appears to be happy.

If the diff is too overwhelming, I can remove the changes to the preprocessing layers and put them in a separate PR.

The big points for review here are:
- Backend import infrastructure
- The TFDataScope and `tf_data` decorator
- CI setup

I expect that getting /gcbrun working is going to take some doing, but I've sent this for review now to get the conversation started.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -31,7 +31,7 @@ jobs:
       run: |
         pip install tensorflow==2.12.0
         pip install -e ".[tests]" --progress-bar off --upgrade
-        git clone https://ianstenbit:${{ secrets.PAT }}@github.com/keras-team/keras-core.git
+        git clone https://${{ secrets.PAT }}@github.com/keras-team/keras-core.git
         pip install keras-core/
         pip install "jax[cpu]"
     - name: Test with pytest
@@ -67,7 +67,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install tensorflow==2.12.0
-        git clone https://ianstenbit:${{ secrets.PAT }}@github.com/keras-team/keras-core.git
+        git clone https://${{ secrets.PAT }}@github.com/keras-team/keras-core.git
         pip install keras-core/
         pip install -e ".[tests]" --progress-bar off --upgrade
         pip install "jax[cpu]"
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -27,12 +27,11 @@ jobs:
         key: ${{ runner.os }}-pip-${{ hashFiles('setup.py') }}
         restore-keys: |
           ${{ runner.os }}-pip-
+        ssh-key: ${{ secrets.KERAS_CORE }}
     - name: Install dependencies
       run: |
         pip install tensorflow==2.12.0
         pip install -e ".[tests]" --progress-bar off --upgrade
-        eval `ssh-agent -s`
-        ssh-add - <<< '${{ secrets.KERAS_CORE }}'
         git clone git@github.com:keras-team/keras-core.git
         pip install keras-core/
         pip install "jax[cpu]"
@@ -66,11 +65,10 @@ jobs:
         key: ${{ runner.os }}-pip-${{ hashFiles('setup.py') }}
         restore-keys: |
           ${{ runner.os }}-pip-
+        ssh-key: ${{ secrets.KERAS_CORE }}
     - name: Install dependencies
       run: |
         pip install tensorflow==2.12.0
-        eval `ssh-agent -s`
-        ssh-add - <<< '${{ secrets.KERAS_CORE }}'
         git clone git@github.com:keras-team/keras-core.git
         pip install keras-core/
         pip install -e ".[tests]" --progress-bar off --upgrade
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -67,7 +67,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install tensorflow==2.12.0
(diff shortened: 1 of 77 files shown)


keras_cv/bounding_box/converters.py
```diff
@@ -27,23 +27,22 @@ class RequiresImagesException(Exception):
...
 
 
-ALL_AXES = [1, 1, 1, 1]
+ALL_AXES = 4
 
 
 def _encode_box_to_deltas(
-    anchors: tf.Tensor,
-    boxes: tf.Tensor,
+    anchors: any,
```
ianstenbit, line 34: Remove `any` typehints throughout

keras_cv/backend/config.py
```diff
@@ -0,0 +1,66 @@
...
+    }
+    try:
+        with open(_config_path, "w") as f:
+            f.write(json.dumps(_config, indent=4))
+    except IOError:
+        # Except permission denied.
+        pass
+
+# Set Keras backend based on KERAS_CV_MULTI_BACKEND flag, if applicable.
+if "KERAS_CV_MULTI_BACKEND" in os.environ:
```
mattdangerw, line 60: @fchollet

I think both me and @ianstenbit were thinking it would be way less user friction if in `KERAS_BACKEND=xx` just worked, without the need for this mouthful every time. Something like...

Transient flow:
- `KERAS_BACKEND=xx python cv-script.py` just works.
- `KERAS_BACKEND=tensorflow python cv-script.py` is keras-core with tf.
- `python cv-script.py` is `tf.keras`.

Persistent flow:
- Edit `keras.json` to set `backend: "xx"`.
- Edit `keras_cv.json` to set `multi_backend: True`

mattdangerw (commented):
Just a partial review, wanted to post the last comment here so am posting. Will keep looking at the rest.

keras_cv/layers/preprocessing/vectorized_base_image_augmentation_layer_test.py
```diff
@@ -435,31 +435,26 @@ def test_augment_unbatched_all_data(self):
             "segmentation_masks": segmentation_masks,
         }
 
+        print(input)
```
mattdangerw, line 438: some prints escaped :)

keras_cv/bounding_box/converters_test.py
```diff
@@ -107,6 +103,10 @@ def test_converters(self, source, target):
         )
 
     @parameterized.named_parameters(*test_image_ragged)
+    @pytest.mark.skipif(
```
mattdangerw, line 106: I ended up wanting to use something like this a lot, so just with a `@pytest.mark.tf_only`. Then did the reason and skip logic once in [conftest.py](https://github.com/mattdangerw/keras-nlp/blob/60a4f7c99a325a6b63f3444a91ac25e34b1e3fb4/keras_nlp/conftest.py).

Could consider something like that? Or `@pytest.mark.requires_ragged` if you want to be more find grained. Just a bit less code.

keras_cv/layers/preprocessing/vectorized_base_image_augmentation_layer_test.py
ianstenbit: Welp that's embarassing -- removed, thanks 😄

keras_cv/backend/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_cv.backend.config import multi_backend
+
+# Keys are of the form: "module.where.attr.exists->module.where.to.alias"
+# Value are of the form: ["attr1", "attr2", ...] or
+#                        [("attr1_original_name", "attr1_alias_name")]
+_KERAS_CORE_ALIASES = {
```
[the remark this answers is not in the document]
ianstenbit: Yeah we figured either way could work, and the `keras_core` API was a bit more likely to suddenly change a bit than `tf.keras`

keras_cv/backend/config.py
ianstenbit: I've updated the code to reflect this approach.

keras_cv/bounding_box/converters_test.py
ianstenbit: Done -- thanks for the pointer!

ianstenbit:
@mattdangerw I've addressed your initial comments amd I'm going to move over to #1903 now.

.github/workflows/actions.yml
```diff
@@ -27,29 +27,87 @@ jobs:
...
-        TEST_CUSTOM_OPS: true
+        TEST_CUSTOM_OPS: false
       run: |
         pytest keras_cv/ --ignore keras_cv/models/legacy/ --durations 0
+  multibackend:
+    name: Test the code with Keras Core
+    strategy:
+      fail-fast: false
+      matrix:
+        backend: [tensorflow, jax, torch]
```
innat, line 55: Gone over the head. o_0 !

@ianstenbit 
Curious to know.
What is the purpose of this PR? Is Keras going to support Jax and Torch as the backend, just like it previously did (Theano, MXNet )?

ianstenbit: This is an experimental dev branch that I'm trying some things out with -- I'll share an update in a few weeks

PR #1905, opened 2023-06-30
Port losses to Keras Core

ianstenbit:

### DIFF
--- a/keras_cv/losses/centernet_box_loss.py
+++ b/keras_cv/losses/centernet_box_loss.py
@@ -14,23 +14,22 @@
 
 import math
 
-import tensorflow as tf
-from tensorflow import keras
-
+from keras_cv.backend import keras
+from keras_cv.backend import ops
 from keras_cv.bounding_box_3d import CENTER_XYZ_DXDYDZ_PHI
 
 
 def l1(y_true, y_pred, sigma=9.0):
     """Computes element-wise l1 loss."""
 
-    absolute_difference = tf.abs(y_pred - y_true)
-    loss = tf.where(
+    absolute_difference = ops.abs(y_pred - y_true)
+    loss = ops.where(
         absolute_difference < 1.0 / sigma,
         0.5 * sigma * absolute_difference**2,
         absolute_difference - 0.5 / sigma,
     )
 
-    return tf.reduce_sum(loss, axis=-1)
+    return ops.sum(loss, axis=-1)
 
 
 class CenterNetBoxLoss(keras.losses.Loss):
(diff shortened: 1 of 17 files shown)


keras_cv/losses/centernet_box_loss.py
```diff
@@ -61,40 +60,43 @@ class CenterNetBoxLoss(keras.losses.Loss):
     def __init__(self, num_heading_bins, anchor_size, **kwargs):
         super().__init__(**kwargs)
         self.num_heading_bins = num_heading_bins
-        self.anchor_size = anchor_size
+        self.anchor_size = ops.convert_to_tensor(anchor_size)
```
mattdangerw, line 63: this is curious, why the need to convert? I have been trying to avoid passing tensor types as init args.

mattdangerw (approved):
This looks good to me! Just one little question

keras_cv/losses/focal_test.py
```diff
@@ -68,13 +57,26 @@ def test_output_shape_from_logits(self):
...
+        y_true = rng.uniform(size=(2, 8, 10)).astype("float64")
+        y_logits = rng.uniform(low=-1000, high=1000, size=(2, 8, 10)).astype(
+            "float64"
+        )
+        y_pred = ops.cast(ops.sigmoid(y_logits), "float32")
 
         focal_loss_on_logits = FocalLoss(from_logits=True)
         focal_loss = FocalLoss()
 
+        # TODO(ianstenbit): This probably warrants some more investigation.
```
jbischof, line 70: Is it just the loss scale that differs or also the performance metrics?

keras_cv/losses/simclr_loss.py
```diff
@@ -12,12 +12,19 @@
...
 
-import tensorflow as tf
-from tensorflow import keras
+from keras_cv.backend import keras
+from keras_cv.backend import ops
 
 LARGE_NUM = 1e9
 
 
+def l2_normalize(x, axis):
```
jbischof, line 21: Maybe we need this in Keras Core?

jbischof (approved):
Looks great, thank you!

keras_cv/losses/simclr_loss.py
mattdangerw: It's an interesting question. The numpy way would be `numpy.linalg.norm`, but I don't think `linalg` is covered by tfnp at all.

keras_cv/losses/focal_test.py
ianstenbit: The performance metrics match when I train the model with different backends -- I think the root cause here is op fusion of softmax + xent, but I'm not certain yet (hence leaving the TODO)

keras_cv/losses/simclr_loss.py
ianstenbit: Probably a reasonable thing to exist in Keras Core as either a util or a layer.
For now I figured it was a simple enough little thing to leave in-line here.

keras_cv/losses/centernet_box_loss.py
ianstenbit: Good catch -- I needed this earlier due to some PyTorch device weirdness, but it got fixed in Keras Core so this is no longer needed.

PR #1906, opened 2023-06-30
Port backbones to Keras Core

ianstenbit:

### DIFF
--- a/keras_cv/layers/fusedmbconv.py
+++ b/keras_cv/layers/fusedmbconv.py
@@ -13,9 +13,7 @@
 # limitations under the License.
 
 
-from keras import backend
-from tensorflow import keras
-from tensorflow.keras import layers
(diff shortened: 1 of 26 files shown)


keras_cv/layers/fusedmbconv.py
```diff
@@ -13,9 +13,7 @@
 # limitations under the License.
 
 
-from keras import backend
-from tensorflow import keras
-from tensorflow.keras import layers
+from keras_cv.backend import keras
```
jbischof, line 16: Just curious, why change the import style?

jbischof: This does match KerasNLP more closely, however

jbischof (approved):
Very clean! We might want to add an issue to test model export as some point with the loss of tf saving testing.

PR #1907, opened 2023-06-30
Port object detection layers to Keras Core

ianstenbit:

(Not the RCNN layers yet, because that model isn't really done)

### DIFF
--- a/keras_cv/layers/object_detection/box_matcher.py
+++ b/keras_cv/layers/object_detection/box_matcher.py
@@ -13,13 +13,12 @@
 # limitations under the License.
 
 from typing import List
-from typing import Tuple
 
-import tensorflow as tf
-from tensorflow import keras
+from keras_cv.backend import keras
+from keras_cv.backend import ops
 
 
-@keras.utils.register_keras_serializable(package="keras_cv")
+@keras.saving.register_keras_serializable(package="keras_cv")
 class BoxMatcher(keras.layers.Layer):
     """Box matching logic based on argmax of highest value (e.g., IOU).
 
@@ -73,7 +72,7 @@ class BoxMatcher(keras.layers.Layer):
     box_matcher = keras_cv.layers.BoxMatcher([0.3, 0.7], [-1, 0, 1])
     iou_metric = keras_cv.bounding_box.compute_iou(anchors, boxes)
     matched_columns, matched_match_values = box_matcher(iou_metric)
-    cls_mask = tf.less_equal(matched_match_values, 0)
+    cls_mask = ops.less_equal(matched_match_values, 0)
     ```
 
     TODO(tanzhenyu): document when to use which mode.
@@ -102,7 +101,7 @@ class BoxMatcher(keras.layers.Layer):
         self.force_match_for_each_col = force_match_for_each_col
         self.built = True
 
-    def call(self, similarity_matrix: tf.Tensor) -> Tuple[tf.Tensor, tf.Tensor]:
+    def call(self, similarity_matrix):
         """Matches each row to a column based on argmax
 
         TODO(tanzhenyu): consider swapping rows and cols.
@@ -120,10 +119,10 @@ class BoxMatcher(keras.layers.Layer):
         squeeze_result = False
         if len(similarity_matrix.shape) == 2:
             squeeze_result = True
-            similarity_matrix = tf.expand_dims(similarity_matrix, axis=0)
-        static_shape = similarity_matrix.shape.as_list()
-        num_rows = static_shape[1] or tf.shape(similarity_matrix)[1]
-        batch_size = static_shape[0] or tf.shape(similarity_matrix)[0]
+            similarity_matrix = ops.expand_dims(similarity_matrix, axis=0)
+        static_shape = list(similarity_matrix.shape)
+        num_rows = static_shape[1] or ops.shape(similarity_matrix)[1]
+        batch_size = static_shape[0] or ops.shape(similarity_matrix)[0]
 
         def _match_when_cols_are_empty():
             """Performs matching when the rows of similarity matrix are empty.
@@ -138,12 +137,12 @@ class BoxMatcher(keras.layers.Layer):
                     num_rows] storing the match type indicator (e.g. positive or
                     negative or ignored match).
             """
-            with tf.name_scope("empty_boxes"):
-                matched_columns = tf.zeros(
-                    [batch_size, num_rows], dtype=tf.int32
+            with ops.name_scope("empty_boxes"):
+                matched_columns = ops.zeros(
+                    [batch_size, num_rows], dtype="int32"
                 )
-                matched_values = -tf.ones(
-                    [batch_size, num_rows], dtype=tf.int32
+                matched_values = -ops.ones(
+                    [batch_size, num_rows], dtype="int32"
                 )
                 return matched_columns, matched_values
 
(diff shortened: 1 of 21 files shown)


keras_cv/layers/object_detection/roi_generator_test.py
```diff
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import pytest
 import tensorflow as tf
 
 from keras_cv.layers.object_detection.roi_generator import ROIGenerator
 
 
+@pytest.mark.tf_keras_only
```
mattdangerw, line 21: For stuff like this I was also adding a runtime check, so users know what to do.

https://github.com/keras-team/keras-nlp/blob/c1eef6fe272cdf637b8833134f74b1fd815ea493/keras_nlp/utils/tensor_utils.py#L81-L86

Might be worth doing that in CV as well?

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -0,0 +1,521 @@
...
+    idx = ops.minimum(idx, num_boxes - 1)
+
+    index_offsets = ops.cast(ops.arange(batch_size) * num_boxes, "int32")
+    take_along_axis_idx = ops.reshape(
+        idx + ops.expand_dims(index_offsets, 1), [-1]
+    )
+
+    # TODO(ianstenbit): Fix bug in tfnp.take_along_axis that causes this hack.
+    # (This will be removed anyway when we use built-in NMS for TF.)
+    if multi_backend() and keras.backend.backend() != "tensorflow":
```
mattdangerw, line 290: ah, I see, this is your approach to the shape issues. looks reasonable.

will still poke around and see if we can fix this at tf level.

ianstenbit: yeah this is hacky and awful, but it's not going to be around for long because this NMS op won't be used for TF backends so I figured it wasn't worth a lot of investment here.

That said, I think we should invest in fixing the tfnp op and I'm happy to help with that

keras_cv/layers/object_detection/roi_generator_test.py
ianstenbit: Yeah I've added this with the `supports_ragged` flag for most affected cases. But I agree we should probably add a similar check to yet-unported components. I'll do that before I send this for review.

mattdangerw: oh whoops! didn't notice you hadn't sent yet, was just scanning the top of PRs :)

ianstenbit: All good 😄

keras_cv/layers/object_detection/box_matcher.py
```diff
@@ -158,25 +157,37 @@ def _match_when_cols_are_non_empty():
                     num_rows] storing the match type indicator (e.g. positive or
                     negative or ignored match).
             """
-            with tf.name_scope("non_empty_boxes"):
-                matched_columns = tf.argmax(
-                    similarity_matrix, axis=-1, output_type=tf.int32
+            with ops.name_scope("non_empty_boxes"):
+                # Jax traces this function even when running eagerly and the`
```
jbischof, line 161: Awesome comment! Remove backtick at end of this line

keras_cv/layers/object_detection/multi_class_non_max_suppression.py
```diff
@@ -71,6 +72,12 @@ def call(
                 `bounding_box_format` specified in the constructor.
             class_prediction: Dense Tensor of shape [batch, boxes, num_classes].
         """
+        if multi_backend() and keras.backend.backend() != "tensorflow":
```
jbischof, line 75: This is some fancy NMS that takes the predicted class into account?

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -0,0 +1,521 @@
...
+from keras_cv.backend import keras
+from keras_cv.backend import ops
+from keras_cv.backend.config import multi_backend
+
+
+@keras.saving.register_keras_serializable(package="keras_cv")
+class NonMaxSuppression(keras.layers.Layer):
+    """A Keras layer that decodes predictions of an object detection model.
+
+    Arguments:
```
jbischof, line 29: Args:

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -0,0 +1,521 @@
...
+        of boxes to be selected by non max suppression.
+      iou_threshold: a float representing the threshold for deciding whether boxes
+        overlap too much with respect to IoU (intersection over union).
+      score_threshold: a float representing the threshold for box scores. Boxes
+        with a score that is not larger than this threshold will be suppressed.
+      tile_size: an integer representing the number of boxes in a tile, i.e.,
+        the maximum number of boxes per image that can be used to suppress other
+        boxes in parallel; larger tile_size means larger parallelism and
+        potentially more redundant work.
+    Returns:
```
jbischof, line 183: Newline between sections of docstring

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -0,0 +1,521 @@
...
+
+    def _sort_scores_and_boxes(scores, boxes):
+        """Sort boxes based their score from highest to lowest.
+
+        Args:
+          scores: a tensor with a shape of [batch_size, num_boxes] representing
+            the scores of boxes.
+          boxes: a tensor with a shape of [batch_size, num_boxes, 4] representing
+            the boxes.
+        Returns:
```
jbischof, line 202: Newline here too

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -0,0 +1,521 @@
...
+    """Calculates the overlap (iou - intersection over union) between boxes_a and boxes_b.
+
+    Args:
+      boxes_a: a tensor with a shape of [batch_size, N, 4]. N is the number of
+        boxes per image. The last dimension is the pixel coordinates in
+        [ymin, xmin, ymax, xmax] form.
+      boxes_b: a tensor with a shape of [batch_size, M, 4]. M is the number of
+        boxes. The last dimension is the pixel coordinates in
+        [ymin, xmin, ymax, xmax] form.
+    Returns:
```
jbischof, line 321: Newline

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -0,0 +1,521 @@
...
+        # Calculates IoU.
+        intersection_over_union = i_area / u_area
+
+        return intersection_over_union
+
+
+def _self_suppression(iou, _, iou_sum, iou_threshold):
+    """Suppress boxes in the same tile.
+
+       Compute boxes that cannot be suppressed by others (i.e.,
```
jbischof, line 357: After first line, contents aligned with quotes on the left.

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -0,0 +1,521 @@
...
+        i_ymin = ops.maximum(a_y_min, ops.transpose(b_y_min, [0, 2, 1]))
+        i_ymax = ops.minimum(a_y_max, ops.transpose(b_y_max, [0, 2, 1]))
+        i_area = ops.maximum((i_xmax - i_xmin), 0) * ops.maximum(
+            (i_ymax - i_ymin), 0
+        )
+
+        # Calculates the union area.
+        a_area = (a_y_max - a_y_min) * (a_x_max - a_x_min)
+        b_area = (b_y_max - b_y_min) * (b_x_max - b_x_min)
+        EPSILON = 1e-8
```
jbischof, line 344: Either inline the global or move to the top of the file.

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -0,0 +1,521 @@
...
+        # Adds a small epsilon to avoid divide-by-zero.
+        u_area = a_area + ops.transpose(b_area, [0, 2, 1]) - i_area + EPSILON
+
+        # Calculates IoU.
```
jbischof, line 348: comment unnecessary

keras_cv/layers/object_detection/non_max_suppression.py
jbischof, line 1: This is pretty hard to read and not really in the style of the repo. It's probably worth it's own PR.

However, get the sense that the effort is not worth it now. We can file an issue for the cleanup.

jbischof (approved):
I'm fine shipping, but we should return to NMS at some point.

keras_cv/layers/object_detection/multi_class_non_max_suppression.py
ianstenbit: Indeed -- and we'll be able to support it in Torch as well (coming soon)

keras_cv/layers/object_detection/non_max_suppression.py
ianstenbit: Agreed, it's not the most beautiful component. It's mostly going to be for Jax in the long run, and it may just be a stopgap for Jax.

PR #1908, opened 2023-06-30
Port ImageClassifier to Keras Core

ianstenbit:

### DIFF
--- a/keras_cv/models/task.py
+++ b/keras_cv/models/task.py
@@ -26,7 +26,16 @@ class Task(keras.Model):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.backbone = None
+        self._backbone = None
+
+    @property
+    def backbone(self):
+        """A `keras.Model` instance providing the backbone submodel."""
+        return self._backbone
+
+    @backbone.setter
+    def backbone(self, value):
+        self._backbone = value
 
     def get_config(self):
         # Don't chain to super here. The default `get_config()` for functional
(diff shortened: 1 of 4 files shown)


keras_cv/models/task.py
```diff
@@ -15,28 +15,18 @@
...
+@keras.saving.register_keras_serializable(package="keras_cv")
 class Task(keras.Model):
     """Base class for Task models."""
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._backbone = None
-
-    @property
-    def backbone(self):
```
jbischof, line 33: We created this property for keras.io documentation (it only picks up properties). However, we might need to pick it up explicitly since I only see it on KerasNLP docs ([example](https://keras.io/api/keras_nlp/models/bert/bert_classifier/)).

jbischof (approved):
Very clean, but we might still need the backbone property.

keras_cv/models/task.py
ianstenbit: Ahh interesting -- I've reverted this and used an alternative workaround for the PyTorch insanity (setting `self._backbone` directly)

mattdangerw: If you want to keep the property, normal setters, and not annoy torch, this should work.

https://github.com/mattdangerw/keras-nlp/blob/b50f67d05027a8af1b59ba21e109183623cf07ef/keras_nlp/models/task.py#L88-L93

jbischof: Thanks for the point @mattdangerw!!

PR #1909, opened 2023-07-01
Backbone property fix

ianstenbit:
As per the suggestion from @mattdangerw


jbischof (approved):
Nice! Now we have to index these on keras.io

PR #1917, opened 2023-07-07
Fix default `training` value in SqueezeExcite

ianstenbit:


fchollet (approved):
Thank you!

`None` is fine here, but `training=False` would be correct as well.

PR #1918, opened 2023-07-08
Implement native NMS for PyTorch backend

ianstenbit:


keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -90,13 +90,45 @@ def call(
...
+                for batch_idx in range(batch_size):
+                    conf_mask = (
+                        confidence_prediction[batch_idx]
+                        > self.confidence_threshold
+                    )
+                    conf_mask_idx = ops.squeeze(ops.nonzero(conf_mask), axis=0)
+                    conf_i = confidence_prediction[batch_idx][conf_mask]
+                    box_i = box_prediction[batch_idx][conf_mask]
+
+                    idx_i = torchvision.ops.nms(
```
mattdangerw, line 109: Is this mainly for the efficiency of the torchvision op? If so, might be nice to leave a comment at the top of this block to clarify.

mattdangerw (approved):

keras_cv/layers/object_detection/non_max_suppression.py
ianstenbit: indeed -- good call! I added a comment

PR #1922, opened 2023-07-10
Add keras-core as a dependency

ianstenbit:


jbischof (approved):
I was expecting a lot of hacks to be removed with the new import. Were we just this clever with the rollout?

ianstenbit:
> I was expecting a lot of hacks to be removed with the new import. Were we just this clever with the rollout?

We're still importing from GitHub in CI, but hey I think now we can just import from PyPi instead! I'll update the CI in this PR

PR #1931, opened 2023-07-11
Use TF-native NMS when in TF backend

ianstenbit:


tirthasheshpatel (approved):
Looks good! Thanks for jumping in and fixing this!

keras_cv/layers/object_detection/non_max_suppression.py
```diff
@@ -88,8 +88,17 @@ def call(
...
+                max_output_size=self.max_detections,
+                iou_threshold=self.iou_threshold,
+                score_threshold=self.confidence_threshold,
+                pad_to_max_output_size=True,
+                sorted_input=False,
+            )
+        elif keras.backend.backend() == "torch":
             # Since TorchVision has a nice efficient NMS op, we might as well
             # use it!
             if keras.backend.backend() == "torch":
```
jbischof, line 104: We are checking this condition twice now

jbischof (approved):
Thanks!

keras_cv/layers/object_detection/non_max_suppression.py
ianstenbit: Welp that's embarrassing -- fixed

PR #1934, opened 2023-07-11
Switch back to tf 2.12 as minimum version

ianstenbit:
Colab is still on 2.12, and it's foolish to require 2.13 unnecessarily


jbischof (approved):
great!

PR #1935, opened 2023-07-11
Add GPU testing for Jax+Torch

ianstenbit:
In order to make this work, I also had to create a new base `TestCase` which is why the diff got so massive.

But hey, we're now doing GPU testing with CUDA on 3 different backends -- which is pretty dang cool!


keras_cv/tests/test_case.py
```diff
@@ -0,0 +1,52 @@
...
+import tensorflow as tf
+from absl.testing import parameterized
+
+from keras_cv.backend import ops
+
+
+class TestCase(tf.test.TestCase, parameterized.TestCase):
+    """Base test case class for KerasCV.
+
+    For now we just extend tf.TestCase and parameterized.TestCase, but this
```
mattdangerw, line 23: Heh, I have this in KerasNLP too, but it's no longer true. We can probably just delete and keep the one liner.

keras_cv/layers/object_detection/anchor_generator_test.py
```diff
@@ -189,6 +189,8 @@ def test_relative_generation(self):
             clip_boxes=False,
         )
         boxes = anchor_generator(image=image)
-        boxes = np.concatenate(list(boxes.values()), axis=0)
+        boxes = np.concatenate(
```
mattdangerw, line 192: `ops.concatenate` here, or `np.array` below? otherwise you are converting to a backend tensor and then back to np in the same line.

cloudbuild/unit_test_jobs.jsonnet
```diff
@@ -21,20 +22,28 @@ local unittest = base.BaseTest {
...
-    |||
-      # Build custom ops from source
-      python build_deps/configure.py
-      bazel-5.4.0 build keras_cv/custom_ops:all --verbose_failures
-      cp bazel-bin/keras_cv/custom_ops/*.so keras_cv/custom_ops/
-      export TEST_CUSTOM_OPS=true
+    std.format(
+      |||
+        export KERAS_BACKEND=%s
+        export JAX_ENABLE_X64=true
```
mattdangerw, line 28: Why do you end up needed int64/float64? Just curious really.

In KerasNLP we have just been trying to go int32 everywhere by default.

cloudbuild/README.md
```diff
@@ -44,6 +44,29 @@ RUN pip install -r keras-cv/requirements.txt
 ```
 - Run the following command from the directory with your `Dockerfile`:
 ```
-gcloud builds submit --region=us-west1 --tag us-west1-docker.pkg.dev/keras-team-test/keras-cv-test/keras-cv-image:deps --timeout=10m
+gcloud builds submit --region=us-west1 --tag us-west1-docker.pkg.dev/keras-team-test/keras-cv-test/keras-cv-image-tensorflow:deps --timeout=20m
+```
+- Repeat the last two steps for Jax and Torch (replacing "tensorflow" with "jax"
+ or "torch" in the docker image target name). `Dockerfile` for jax:
+```
+FROM nvidia/cuda:11.7.1-base-ubuntu20.04
```
mattdangerw, line 52: nice! is a this a base ubuntu with nvidia drivers docker image?

mattdangerw (approved):
Looks good!

cloudbuild/README.md
ianstenbit: yep 👍

cloudbuild/unit_test_jobs.jsonnet
ianstenbit: I think we probably _should_ just do int32 everywhere. During NMS porting I had some things that were real sticklers about int64 in TF and I never found a way to work around

keras_cv/tests/test_case.py
ianstenbit: yeah this is copied straight from NLP - sorry I didn't give attribution 😓

mattdangerw: oh all good! we should readily copy back and forth i think

I should just fix this for nlp too

keras_cv/layers/object_detection/anchor_generator_test.py
ianstenbit: Discussed offline -- we're not converting to a backend tensor here. All numpy all the way 😀

keras_cv/tests/test_case.py
```diff
@@ -0,0 +1,47 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+from absl.testing import parameterized
+
+from keras_cv.backend import ops
+
+
+class TestCase(tf.test.TestCase, parameterized.TestCase):
```
jbischof, line 20: We ended up doing the same thing for NLP

cloudbuild/unit_test_jobs.jsonnet
jbischof: Could we potentially just change the dtype for those tests? Asking since we're always flirting with OOM issues and int64 can't help

jbischof (approved):
Thanks!

.github/workflows/actions.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         KERAS_BACKEND: ${{ matrix.backend }}
         JAX_ENABLE_X64: true
       run: |
-        pytest --run_large keras_cv/bounding_box \
```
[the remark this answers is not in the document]
ianstenbit: Not a mistake, I was just doing this during development of the rebase so that we covered all the tests. But yes we shouldn't include it anymore

cloudbuild/unit_test_jobs.jsonnet
ianstenbit: Potentially, yes. I did some looking and it seems like the only place we're using int64 internally with KerasCore is in the YOLOV8 label encoder -- I guess I already got rid of them from NMS. I'll check with Tirth if he's planning on getting rid of those.

PR #1938, opened 2023-07-12
Remove unneeded environment variable

jbischof:
Setting `KERAS_BACKEND` should be enough to trigger Keras Core usage

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -75,7 +75,6 @@ jobs:
     - name: Test with pytest
       env:
         TEST_CUSTOM_OPS: false # TODO(ianstenbit): test custom ops, or figure out what our story is here
-        KERAS_CV_MULTI_BACKEND: true
         KERAS_BACKEND: ${{ matrix.backend }}
         JAX_ENABLE_X64: true
       run: |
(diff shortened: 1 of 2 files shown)


ianstenbit (approved):
`toxic_env` 😭

jbischof:
Is there any way to check that the GCB tests ran as intended? I can't see any print statements about the backend.

ianstenbit:
As-is on this branch, the GCB tests ran the normal test suite (just three times)

You can see one run result here: https://pantheon.corp.google.com/cloud-build/builds/dd860e52-ebe6-4b93-b794-96f1d70cd092;step=5?jsmode=o&project=keras-team-test

So this is all set.

jbischof:
Right, but do we need to check that the `KERAS_BACKEND` var was sufficient to trigger different backends instead of testing `tf.keras` over and over?

ianstenbit:
> Right, but do we need to check that the `KERAS_BACKEND` var was sufficient to trigger different backends instead of testing `tf.keras` over and over?

Oh I see -- yes, right now this just triggered `tf.keras` over and over for the GCB tests (that's expected, because #1935 isn't done). For the non-GCB tests, we can tell that `keras_core` is being tested because some tests (e.g. the RCNN tests) are skipped on these CI runs, which is only true for Keras Core

PR #1943, opened 2023-07-12
Port DeepLabV3 to Keras Core

ianstenbit:
I had to make some small adjustments to the model to a) clean up the docs a bit and b) get it in line with the paper
(https://arxiv.org/pdf/1802.02611.pdf)


keras_cv/layers/spatial_pyramid.py
```diff
@@ -142,21 +144,28 @@ def call(self, inputs, training=None):
...
-        result = tf.concat(result, axis=-1)
+        result[-1] = keras.layers.Resizing(
+            inputs.shape[1], inputs.shape[2], interpolation="bilinear"
+        )(result[-1])
+
+        result = ops.concatenate(result, axis=-1)
         result = self.projection(result, training=training)
         return result
 
+    def compute_output_shape(self, input_shape):
```
jbischof, line 166: Why is this method needed in Keras Core?

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus.py
```diff
@@ -102,13 +107,20 @@ def __init__(
 
         inputs = backbone.input
 
-        backbone_outputs = backbone(inputs)
+        high_level_feature_extractor = keras.Model(
```
jbischof, line 110: Can't we use `get_feature_extractor` as in `RetinaNet` ([link](https://github.com/keras-team/keras-cv/blob/b1888aac446e6e49c0bb2bd9ac33586f27eabee5/keras_cv/models/object_detection/retinanet/retinanet.py#L174))?

jbischof (approved):
Thanks!

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus.py
```diff
@@ -102,13 +107,20 @@ def __init__(
...
+        high_level_features = high_level_feature_extractor(inputs)
 
         if spatial_pyramid_pooling is None:
             spatial_pyramid_pooling = SpatialPyramidPooling(
                 dilation_rates=[6, 12, 18]
             )
-        spp_outputs = spatial_pyramid_pooling(backbone_outputs)
+        spp_outputs = spatial_pyramid_pooling(high_level_features)
 
         low_level_feature_extractor = keras.Model(
```
jbischof, line 125: Same here

keras_cv/layers/spatial_pyramid.py
ianstenbit: It's needed for tracing the shape while tracing the forward pass. Without it we get an error like this one:
```
Could not automatically infer the output shape / dtype of 'spatial_pyramid_pooling_9' (of type SpatialPyramidPooling). 
Either the `SpatialPyramidPooling.call()` method is incorrect, or you need to implement the `SpatialPyramidPooling.compute_output_spec() / compute_output_shape()` method. Error encountered:
```

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus.py
ianstenbit: We can, and we should! Done

Issue #1933, opened 2023-07-11
Porting DeepVision into KerasCV


DavidLandup0:
Hey team! Absolutely stellar work on Core Keras, and bringing it to KCV!

Since one of the main tennets of my own lib, [DeepVision](https://github.com/DavidLandup0/deepvision) , was to unify computer vision through a back-end agnostic high-level API with interchangeable modular components - and this is exactly what Core Keras is, I'd like to re-route the layer and models from there to here.

It's got a few flavors of transformer encoders, several flavors of attention, on-the-fly weight porting, as well as SAM and CLIP (almost pushed to main).

Are there plans to add any of these to KCV soon? I'd personally port all of these.

@jbischof @ianstenbit

ianstenbit:
Hey @DavidLandup0 -- I would be stoked to have you bring these components over to KerasCV!

These are all things that we'd love to include in KerasCV, and I would be happy to help with reviewing ports. I think @tirthasheshpatel has also done some work on SAM and would be interested in that piece.

Transformer components, CLIP, and SAM are all things that are in our roadmap but that we don't currently have the eng time to implement, so that would be fantastic. 🚀 

I'm happy to discuss further over a call, or feel free to just start opening some PRs and tagging me -- I will prioritize taking a look at those.

DavidLandup0:
Thanks for the prompt response @ianstenbit!
It may make sense to coordinate this a bit instead of just opening lots of PRs on my end. The APIs are fairly compatible, but there are nuances in layers and implementations that may be worth ironing out and syncing on.

Let's set up a group call to sync up? I can get started on these tomorrow.

ianstenbit:
> Thanks for the prompt response @ianstenbit! It may make sense to coordinate this a bit instead of just opening lots of PRs on my end. The APIs are fairly compatible, but there are nuances in layers and implementations that may be worth ironing out and syncing on.
> 
> Let's set up a group call to sync up? I can get started on these tomorrow.

Sounds good -- I sent you an invite for tomorrow afternoon, but feel free to suggest a different time!

IMvision12:
This is really awesome!! Let me know if you need a helping hand :) @DavidLandup0

sachinprasadhs:
Hi, The linked PR is already merged here https://github.com/keras-team/keras-cv/pull/1946, could you please close the issue if there is no pending action to be taken.

PR #1848, opened 2023-05-31
Move `Defaults to` to end of arg docstring and standardise values

SamuelMarks:
# What does this PR do?

Hey it's that annoying fellow again, the one from https://github.com/keras-team/keras/pull/17748

## Before submitting

- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #1946, opened 2023-07-13
[DeepVision Port] SegFormer and Mix-Transformers

DavidLandup0:
# What does this PR do?

As discussed in #1933 - setting up a draft PR for porting SegFormer and associated layers into KCV. Draft PR for now with placeholder main model dump, layers and tests incoming soon. Will tag once ready for review.

## Demo Notebooks

- Mix-Transformer components, `from_preset()` usage and training: https://colab.research.google.com/drive/1Q3m9-LKICrFzuUhVMIPd7pY2l9Z3BLhg?usp=sharing
- SegFormer head, `from_preset()` usage and training: #TODO

## Questions and API Considerations

- DeepLabV3 takes any backbone, but SegFormer is meant to be used with MiT (Mix Transformers), and depends on the output channels which is a field defined in the model. Should we make it generally usable with other backbones? IMO, no, since the head is really just an MLP head, and the crux of the paper is MiT.
- There's no name for the type of attention they use, but they refer to it as efficient attention in the paper. What name should we use? `SegFormerMultiHeadAttention` sounds like a mouthful.
- How do we expose the API if we don't support a `backbone` argument? Just `SegFormer.from_preset()`?

## Before submitting
      Pull Request section?
      to it if that's the case. #1933

## Who can review?

@ianstenbit just tagging so you can follow the progress as it comes in. Otherwise, no need to spend time until it's un-drafted for review :)


DavidLandup0:
Just to update you @ianstenbit - the port is going well, but it took me a bit longer than anticipated to get used to Keras Core + the new API 😅 

I ran into a small blocker and documented it [here ](https://github.com/keras-team/keras-core/issues/516)since I'm not sure what the intended usage is when returning tensors and non-tensors from a `call()`.

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,138 @@
...
+        input_shape=None,
+        input_tensor=None,
+        classes=None,
+        include_top=None,
+        embed_dims=None,
+        depths=None,
+        pooling=None,
+        **kwargs,
+    ):
+        if include_top and not classes:
```
ianstenbit, line 49: We dropped `include_top`, `classes`, and `pooling` from the backbone API as the classification task is now handled by the `ImageClassifier` model, which takes a backbone as a parameter.

ianstenbit (commented):
Thanks David! I'm taking a look at the non-tensor return issue.

We have a few options for workarounds, including:
- Computing the shape outside of the layer (because from my cursory view the returned shape is just the input shape / stride)
- Making the PatchingEmbeddingLayer offer a new method which computes these values which callers can use.

That said, I think we should be able to make this work. I'm taking a look at your issue on Keras Core to see if I can get a working fix.

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,138 @@
...
+        self.num_stages = num_stages
+        self.output_channels = embed_dims
+        self.classes = classes
+        self.include_top = include_top
+        self.pyramid_level_inputs = pyramid_level_inputs
+        self.pooling = pooling
+
+        self.patch_embedding_layers = []
+        self.transformer_blocks = []
+
```
ianstenbit, line 125: We'll also need presets to be added here (I see you already have the imports above)

```
    @classproperty
    def presets(cls):
        return copy.deepcopy(backbone_presets)

    @classproperty
        """Dictionary of preset names and configurations that include
        weights."""
        return copy.deepcopy(backbone_presets_with_weights)
```

(Need to `import copy` above as well)

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone_presets.py
```diff
@@ -0,0 +1,114 @@
...
+            "description": (
+                "MiT (MixTransformer) model with 8 transformer blocks."
+            ),
+            "params": 3321962,
+            "official_name": "MiT",
+            "path": "mit",
+        },
+        "class_name": "keras_cv.models>MiTBackbone",
+        "config": {
+            "embedding_dims": [32, 64, 160, 256],
```
ianstenbit, line 28: s/embedding_dims/embed_dims (or rename the arg in the backbone)

Probably it's more clear to keep it as embedding_dims and rename the arg.

keras_cv/layers/overlapping_patching_embedding.py
```diff
@@ -0,0 +1,22 @@
...
+            strides=stride,
+            padding="same",
+        )
+        self.norm = keras.layers.LayerNormalization()
+
+    def call(self, x):
+        x = self.proj(x)
+        # B, H, W, C
+        shape = x.shape
+        x = keras.ops.reshape(x, (-1, shape[1] * shape[2], shape[3]))
```
ianstenbit, line 20: For now, since we're supporting both `tf.keras` and `keras_core`, we have to import `ops` with `from keras_cv.backend import ops`, which supports both versions

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,138 @@
...
+    backbone_presets_with_weights,
+)
+from keras_cv.utils.python_utils import classproperty
+
+
+@keras.saving.register_keras_serializable(package="keras_cv.models")
+class MiTBackbone(Backbone):
+    def __init__(
+        self,
+        input_shape=None,
```
ianstenbit, line 40: Default to `(None, None, 3)` so that channel dims can be known at build time for conv layers

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone_presets.py
DavidLandup0: Good point - I'll rename the arg

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
DavidLandup0: This'll have to default to `(224, 224, 3)` actually, since the input shape will have to be known at instantiation time

DavidLandup0:
@ianstenbit looks like MiTs are shaped up. Here's a demo notebook showcasing the components, inputs/output shapes, pyramid levels, `from_preset()` usage and training MiTs on a classification task: https://colab.research.google.com/drive/1Q3m9-LKICrFzuUhVMIPd7pY2l9Z3BLhg?usp=sharing

There are a couple of weird-looking `ops.cast()` calls that aren't very clean, and a custom reshaping layer since `keras.Reshape()` caused errors for some reason. I'd like to clean these up and sync up on whether there's a cleaner alternative for them :)

keras_cv/layers/hierarchical_transformer_encoder.py
```diff
@@ -0,0 +1,67 @@
...
+            project_dim, num_heads, sr_ratio
+        )
+        self.drop_path = DropPath(drop_prob)
+        self.norm2 = keras.layers.LayerNormalization(epsilon=layer_norm_epsilon)
+        self.mlp = self.MixFFN(
+            channels=project_dim,
+            mid_channels=int(project_dim * 4),
+        )
+
+    def build(self, input_shape):
```
ianstenbit, line 34: When overriding build we should also either call `super().build` or just set `self.built = True`

keras_cv/layers/overlapping_patching_embedding.py
```diff
@@ -0,0 +1,23 @@
+from keras_cv.backend import keras
+from keras_cv.backend import ops
+
+
+@keras.saving.register_keras_serializable(package="keras_cv")
+class OverlappingPatchingAndEmbedding(keras.layers.Layer):
+    def __init__(self, out_channels=32, patch_size=7, stride=4, **kwargs):
```
ianstenbit, line 7: Same here for docstring -- I think particularly the shape transformation is worth noting in the doc

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,150 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""MiT backbone model.
+
+References:
```
ianstenbit, line 17: Maybe this is a TODO?

Let's add the reference and move the docstring to live inside the `MiTBackbone` class

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,150 @@
...
+        self,
+        include_rescaling,
+        input_shape=(None, None, 3),
+        input_tensor=None,
+        embedding_dims=None,
+        depths=None,
+        **kwargs,
+    ):
+        drop_path_rate = 0.1
+        dpr = [x for x in np.linspace(0.0, drop_path_rate, sum(depths))]
```
ianstenbit, line 51: Probably not a huge deal since the presets specify `depths`, but this will raise an error because `depths=None` by default

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,150 @@
...
+            new_height, new_width = (
+                int(ops.shape(x)[1] / stride),
+                int(ops.shape(x)[2] / stride),
+            )
+
+            x = patch_embedding_layers[i](x)
+            for blk in transformer_blocks[i]:
+                x = blk(x)
+            x = layer_norms[i](x)
+            C = x.shape[-1]
```
ianstenbit, line 104: This appears to be unused

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,150 @@
...
+            x = CustomReshaping(new_height, new_width)(x)
```
ianstenbit, line 105: +1 that we should be able to make this work with the built-in Reshaping layer. I'll take a look with you later

keras_cv/models/segmentation/segformer/segformer.py
```diff
@@ -0,0 +1,86 @@
...
+        num_classes=None,
+        backbone=None,
+        embed_dim=None,
+        input_shape=None,
+        input_tensor=None,
+        **kwargs,
+    ):
+        inputs = utils.parse_model_inputs(input_shape, input_tensor)
+        x = inputs
+        y = backbone(x)
```
ianstenbit, line 19: Because the backbone is expected to be MiT, we don't want pyramid-level outputs but instead the standard model output, correct?

If so, we should definitely mention in the docstring for SegFormer than only MiT/ViT are valid backbones

ianstenbit (changes requested):
Generally looks good! Left you a few minor comments.

keras_cv/layers/hierarchical_transformer_encoder.py
```diff
@@ -0,0 +1,67 @@
...
+            self.fc2 = keras.layers.Dense(channels)
+
+        def call(self, x):
+            x = self.fc1(x)
+            shape = ops.shape(x)
+            B, C = ops.cast(shape[0], "float32"), ops.cast(shape[-1], "float32")
+            H, W = ops.sqrt(ops.cast(shape[1], "float32")), ops.sqrt(
+                ops.cast(shape[1], "float32")
+            )
+            # print(B, C, H, W)
```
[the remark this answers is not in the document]
DavidLandup0: Whoops, removed

keras_cv/layers/hierarchical_transformer_encoder.py
DavidLandup0: Nice catch!

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
DavidLandup0: Yup, left as a TODO. By inside the `MiTBackbone` class - you mean within the class def? It looks like references are outside of the classes for other backbones?

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
DavidLandup0: Good point. maybe we should specify a default depths equal to the B0 setup? 🤔

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
DavidLandup0: Got it working!

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
ianstenbit: Yeah that sgtm, or we can optionally raise an error and tell the user to either use a preset or specify `depths`.

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
ianstenbit: Yeah, for the base `Backbone` models we've been including the docstring in the class directly (e.g. [here](https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/backbones/densenet/densenet_backbone.py#L41))

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
DavidLandup0: Removed the default value, that should signal clearly enough that depths are required to be specified or need to be built from a preset

keras_cv/models/segmentation/segformer/segformer.py
DavidLandup0: We want the pyramid-level outputs here, which is the reason they modified a ViT-style arch to have hierarchical outputs.
Technically, this should work with a SwinTransformer-style output as well, or _technically_ any output that matches this signature:

```
{'P1': <KerasTensor shape=(None, 56, 56, 32), dtype=float32, name=keras_tensor_16>,
 'P2': <KerasTensor shape=(None, 28, 28, 64), dtype=float32, name=keras_tensor_31>,
 'P3': <KerasTensor shape=(None, 14, 14, 160), dtype=float32, name=keras_tensor_46>,
 'P4': <KerasTensor shape=(None, 7, 7, 256), dtype=float32, name=keras_tensor_61>}
```

This can arguably be any CNN. However, it's meant to be used with MiTs as the backbone.

I'll update this to operate on them instead of the final output (this was a temp line until MiTs are fully done, which can now be replaced with the correct outputs)

DavidLandup0:
Should we expose the `from_preset()` API for segmentation models as well? It makes sense for SegFormers - so something like:
  ```python
 import tensorflow as tf
 import keras_cv

 images = np.ones(shape=(1, 96, 96, 3))
 labels = np.zeros(shape=(1, 96, 96, 1))
 model = keras_cv.models.segmentation.SegFormer.from_preset('segformer_b0')

 # Evaluate model
 model(images)

 # Train model
 model.compile(
        optimizer="adam",
        loss=keras.losses.BinaryCrossentropy(from_logits=False),
        metrics=["accuracy"],
 )
 model.fit(images, labels, epochs=3)
 ```

DavidLandup0:
Here's an end-to-end segmentation pipeline with training on Oxford IIT: https://colab.research.google.com/drive/1xQdbOLzWAn-zeqphIR5Io6aNyEE_ir83

I can't seem to load the pretrained weights for the time being, so this is without a pre-trained backbone. As soon as I get the weights rolling, I'll evaluate this on Pascal VOC :)
@ianstenbit

keras_cv/models/__init__.py
```diff
@@ -67,6 +67,9 @@
 from keras_cv.models.backbones.efficientnet_v2.efficientnet_v2_aliases import (
     EfficientNetV2SBackbone,
 )
+from keras_cv.models.backbones.mix_transformer.mix_transformer_backbone import (
```
ianstenbit, line 70: Let's add all the aliases here as well

ianstenbit:
> https://colab.research.google.com/drive/1xQdbOLzWAn-zeqphIR5Io6aNyEE_ir83

Wow those results look pretty decent, especially given the lack of a pre-trained backbone!

ianstenbit (commented):
I think `SegFormerMultiHeadAttention` is a fine name even though it's a bit long. Let's just mention in the docstring that it's called something different in the paper. (Also the file needs to be renamed)

I think we _should_ use the normal backbone API, but include a comment in the docs that it's meant to be used with MiT as a backbone.

Let's add some tests -- I think the PR is pretty close!

keras_cv/models/__init__.py
DavidLandup0: Deal!

keras_cv/layers/efficient_multihead_attention.py
```diff
@@ -0,0 +1,116 @@
+import math
+
+from keras_cv.backend import keras
+from keras_cv.backend import ops
+
+
+@keras.saving.register_keras_serializable(package="keras_cv")
```
ianstenbit, line 7: Sorry that this changed while this PR is in-flight, but if you sync to master this should now be

```
from keras_cv.api_export import keras_cv_export

@keras_cv_export("keras_cv.layers.SegFormerMultiheadAttention")
```

(Same for all new public API symbols)

keras_cv/layers/efficient_multihead_attention.py
```diff
@@ -0,0 +1,116 @@
...
+class SegFormerMultiheadAttention(keras.layers.Layer):
+    def __init__(self, project_dim, num_heads, sr_ratio):
+        """
+        Efficient MultiHeadAttention implementation as a Keras layer.
+        A huge bottleneck in scaling transformers is the self-attention layer with an O(n^2) complexity.
+
+        SegFormerMultiheadAttention performs a sequence reduction (SR) operation with a given ratio, to reduce
+        the sequence length before performing key and value projections, reducing the O(n^2) complexity to O(n^2/R) where
```
ianstenbit, line 15: Is this not (n/R)^2

keras_cv/models/segmentation/segformer/segformer.py
```diff
@@ -0,0 +1,158 @@
...
+    Args:
+        backbone: `keras.Model`. The backbone network for the model that is
+            used as a feature extractor for the SegFormer encoder. It is *intended*
+            to be used only with the MiT backbone model which was created specifically
+            for SegFormers. It should either be a `keras_cv.models.backbones.backbone.Backbone` or a
+            `tf.keras.Model` that implements the `pyramid_level_inputs`
+            property with keys "P2", "P3", "P4", and "P5" and layer names as
+            values.
+        num_classes: int, the number of classes for the detection model. Note
+            that the `num_classes` doesn't contain the background class, and the
```
ianstenbit, line 35: We should probably change this docstring. The version in DeepLabV3Plus that this comes from is poorly worded IMO. Users _should_ include a background class, so e.g. for cats vs dogs `num_classes=3`.

ianstenbit (commented):
Let's add tests for the backbone, and then I think this is ready to go!

Thank you @DavidLandup0 for this fantastic contribution!

keras_cv/layers/efficient_multihead_attention.py
DavidLandup0: Done

keras_cv/layers/efficient_multihead_attention.py
DavidLandup0: Took this from the paper, so I honestly took it with blind faith

DavidLandup0: image

keras_cv/layers/hierarchical_transformer_encoder.py
```diff
@@ -0,0 +1,121 @@
...
+        config.update(
+            {
+                "project_dim": self.project_dim,
+                "num_heads": self.num_heads,
+                "drop_prop": self.drop_prop,
+            }
+        )
+        return config
+
+    class MixFFN(keras.layers.Layer):
```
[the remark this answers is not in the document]
DavidLandup0: Added it as part of the config as a serialized object. Should we put `@keras_cv_export()` here too? Doesn't seem like something we'd want to export.

keras_cv/layers/hierarchical_transformer_encoder.py
```diff
@@ -0,0 +1,140 @@
...
+    alternative for computational efficiency, and is meant to be used
+    within the SegFormer architecture.
+
+    References:
+        - [SegFormer: Simple and Efficient Design for Semantic Segmentation with Transformers](https://arxiv.org/abs/2105.15203) (CVPR 2021) # noqa: E501
+        - [Official PyTorch implementation](https://github.com/NVlabs/SegFormer/blob/master/mmseg/models/backbones/mix_transformer.py) # noqa: E501
+        - [Ported from the TensorFlow implementation from DeepVision](https://github.com/DavidLandup0/deepvision/blob/main/deepvision/layers/hierarchical_transformer_encoder.py) # noqa: E501
+
+    Args:
+        project_dim: the dimensionality of the projection of the encoder, and
```
ianstenbit, line 38: Let's start with a typehint on all docstrings (e.g. here: "Integer, the dimensionality ...")

keras_cv/layers/hierarchical_transformer_encoder.py
```diff
@@ -0,0 +1,140 @@
...
+            output of the `SegFormerMultiheadAttention` layer. Due to the
+            residual addition the input dimensionality has to be equal to
+            the output dimensionality.
+        num_heads: the number of heads for the `SegFormerMultiheadAttention`
+            layer
+        drop_prob: default 0.0, the probability of dropping a random sample
```
ianstenbit, line 44: We're trying to adopt a new convention for defaults (see #1848) -- let's use that convention here. (Basically just make "Defaults to XYZ" a second sentence at the end of a docstring.

keras_cv/models/segmentation/segformer/segformer_aliases.py
```diff
@@ -0,0 +1,244 @@
...
+    segformer = keras_cv.models.SegFormer(backbone=backbone, num_classes=19)
+    output = model(input_data)
+    ```
+"""  # noqa: E501
+
+
+class SegFormerB0(SegFormer):
+    def __new__(
+        cls,
+        num_classes=19,
```
ianstenbit, line 45: We shouldn't specify a default for `num_classes` as it needs to be user-specified. A silent default could be very confusing.

ianstenbit (changes requested):
Just a few things to clean up -- in the meantime I am seeing if the tests need any fixing.

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,145 @@
...
+    backbone_presets,
+)
+from keras_cv.models.backbones.mix_transformer.mix_transformer_backbone_presets import (  # noqa: E501
+    backbone_presets_with_weights,
+)
+from keras_cv.utils.python_utils import classproperty
+
+
+@keras_cv_export("keras_cv.models.MiTBackbone")
+class MiTBackbone(Backbone):
```
ianstenbit, line 44: Ah I almost missed this -- let's also add backbone tests for this backbone.

ianstenbit:
Looks like tests are passing locally, but on CI (for TF) it will depend on us getting a new release of Keras Core which includes https://github.com/keras-team/keras-core/pull/722

In the meantime, @DavidLandup0 I left a few review comments for you to take a look at -- thanks!

DavidLandup0:
Awesome, thanks! Getting to these soon. Thanks for the review pass! :)

keras_cv/layers/hierarchical_transformer_encoder.py
DavidLandup0: Done

keras_cv/layers/hierarchical_transformer_encoder.py
DavidLandup0: Got it! I think all of the sections are updated now

keras_cv/models/segmentation/segformer/segformer_aliases.py
DavidLandup0: Removed. Should it be requested as a mandatory arg?

ianstenbit: Yes I think this should be required at init time.

keras_cv/layers/overlapping_patching_embedding.py
```diff
@@ -32,14 +32,12 @@ def __init__(self, project_dim=32, patch_size=7, stride=4, **kwargs):
...
-            layer_norm_epsilon: default 1e-06, the epsilon for `LayerNormalization`
-                layers
-            sr_ratio: default 1, the ratio to use within `SegFormerMultiheadAttention`. If set to > 1,
-                a `Conv2D` layer is used to reduce the length of the sequence.
+            project_dim: integer, the dimensionality of the projection.
+                Defaults to `32`.
+            patch_size: integer, the size of the patches to encode.
+                Defaults to `7`.
+            stride: integer, the stride to use for the patching before
+                projection. Defaults to 5`.
```
ianstenbit, line 40: Nit: missing backtick before 5

keras_cv/models/backbones/mix_transformer/mix_transformer_backbone.py
```diff
@@ -51,6 +51,55 @@ def __init__(
...
     ):
+        """A Keras model implementing the MixTransformer architecture to be
+        used as a backbone for the SegFormer architecture.
+
+        References:
+            - [SegFormer: Simple and Efficient Design for Semantic Segmentation with Transformers](https://arxiv.org/abs/2105.15203) # noqa: E501
+            - [Based on the TensorFlow implementation from DeepVision](https://github.com/DavidLandup0/deepvision/tree/main/deepvision/models/classification/mix_transformer) # noqa: E501
+
+        Args:
+            backbone: `keras.Model`. The backbone network for the model that is
```
ianstenbit, line 62: These args aren't correct

DavidLandup0: 🤦 Meant to change these

ianstenbit (approved):
Thank you David! 

I think this PR is basically all set, I just need to merge #2037 to fix CI

DavidLandup0:
Awesome, thank you for the help in the final stretch! @ianstenbit 🎉

ianstenbit:
It looks like the GCB failure is because I need to update the Docker image of our GCB runners to use the newest Keras Core version -- doing that now.

ianstenbit:
CI failures are unrelated -- seems like DeepLab + YOLOV8 have some breakages with the latest Keras Core version. I'll open a separate PR for those.

DavidLandup0:
Need a hand with DLV3 or YOLO?

ianstenbit:
You're welcome to look if you'd like -- for DeepLab it's a deserialization issue. Haven't looked at YOLO yet.

You can repro by installing latest Keras Core version and running the large tests of those models with TF backend.

DavidLandup0:
Sure! Sign me up for YOLO if it's not *too* urgent then :)

PR #1949, opened 2023-07-13
Add Keras Core as an install dep

### LINKED ISSUE #1944: `No module named 'keras_core'`
After upgrading keras_cv to 0.6.x, importing the library fails with `No module named 'keras_core'`

Screenshot 2023-07-13 at 11 36 51

It works if I manually install `keras_core` but I didn't see that instruction in the doc.

ianstenbit:
Fixes #1944


jbischof (approved):
Thanks!

Issue #1727, opened 2023-04-24
Adding Research Paper In The Docstring of the Code 


NiharJani2002:
**Short Description**
Adding research paper in the docstring of the code 

**Benefits**
Mentioning research papers as docstrings in a Keras-CV repository is beneficial because it provides the following:
1)Credibility and proper attribution, 
2)Offers valuable background and context,
3)Supports reproducibility and validation, 
4)Educational resources for users.

**For Whom this Issue is concerned:**
1) @ianstenbit 
2) @LukeWood 
3) @tanzhenyu 
4) @qlzh727 
5) @fchollet 

**Maintainers, Can You assign me this issue?**

ianstenbit:
@NiharJani2002 could you point out some specific components that are missing references to research papers?

Thanks!
In general we do like having links to relevant research in docstrings.

NiharJani2002:
**@ianstenbit, thanks for showing interest**

What I mean is suppose, In the Simple case of the VGG16 model, 
Link: https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/vgg16.py

Mentioning the original research paper by keeping in mind the whole code was written, for example, since VGG16 is a widely recognized model and it also has a standard paper related to it, 
Link of standard paper: https://arxiv.org/abs/1409.1556

Along With that, we can also mention some reference papers related to VGG16, which can be helpful while studying the code of VGG16.

Since VGG16 is a standard model and easy to understand, there is one reference paper, but some code might be required to mention more than one research paper to be said. Research has to be done on this, and finally, after discussion, it can be mentioned in the docstring.

By this, if a newbie has come to the repository, they can first take a grasp by reading research papers and then proceed to understand the code. Reading the research paper will create a base before moving to understand the code, which can seem helpful for them.

**In brief, new sections of docstring should be specifically dedicated to the research paper**
Example:
'''
Reference Paper:

1. Paper1: Link
2. Paper2: Link
3. 
4. 
5. 
6. Papern: Link

'''

**I am interested in solving this issue and eager to contribute to community, feel free to assign this issue to me**

NiharJani2002:
@ianstenbit 

Any further updates ?

ianstenbit:
Sorry for the delay -- we were very caught up with Keras Core and some triage got left behind.

I don't think we need a separate reference paper section in docstring, but in places where there is not a reference to the original paper in a model docstring, feel free to open a PR to add it!

NiharJani2002:
[...]

PR #1954, opened 2023-07-15
U-Net Code, did not have a original research paper, attached along wi…

NiharJani2002:
…th code, added original research paper, followed the way keras-cv adds research paper along with code

# What does this PR do?
This PR Adds original research paper of U-Net In The Code (unet.py)

Fixes # (issue)
In Continuation with Pull Request #1727 

## Before submitting
Ans) YES
Improves The DOCSTRING Of U-Net (unet.py)
Followed the Standard Convention of Adding Research Paper as seen from previous examples  Keras-CV repo.

## Who can review?
@ianstenbit

### DIFF
--- a/keras_cv/models/__internal__/unet.py
+++ b/keras_cv/models/__internal__/unet.py
@@ -181,7 +181,7 @@ def UNet(
         - [U-Net: Convolutional Networks for Biomedical Image Segmentation](https://arxiv.org/abs/1505.04597)
 
             U-Net relies on strong use of Data Augmentation and architecture consists of a contracting path to capture 
-            context and a symmetric expanding path that enables precise localization.
+            
     
     Args:
         input_shape: the rank 3 shape of the input to the UNet
--- a/keras_cv/models/__internal__/unet.py
+++ b/keras_cv/models/__internal__/unet.py
@@ -179,10 +179,7 @@ def UNet(
 
     Reference:
         - [U-Net: Convolutional Networks for Biomedical Image Segmentation](https://arxiv.org/abs/1505.04597)
-
-       
-            
-    
+        
     Args:
         input_shape: the rank 3 shape of the input to the UNet
         down_block_configs: a list of (filter_count, num_blocks) tuples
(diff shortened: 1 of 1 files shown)


keras_cv/models/__internal__/unet.py
```diff
@@ -177,6 +177,12 @@ def UNet(
     All function parameters require curried functions as inputs which return a
     function that acts on tensors as inputs.
 
+    Reference:
+        - [U-Net: Convolutional Networks for Biomedical Image Segmentation](https://arxiv.org/abs/1505.04597)
+
+            U-Net relies on strong use of Data Augmentation and architecture consists of a contracting path to capture
```
ianstenbit, line 183: I think we can drop this second sentence, but the reference part LGTM -- thanks!

ianstenbit (approved):
Thanks for the PR! Looks good after one small change

keras_cv/models/__internal__/unet.py
ianstenbit: Sorry I think I wasn't clear -- let's only include the link (and drop the part that starts with "U-Net relies on strong use ..."

NiharJani2002, line 183:
```suggestion
       
```

NiharJani2002 (commented):
Per Your Latest Suggestion, Now Research Paper Is There In The Code.
"U-Net relies on strong...." is removed.

keras_cv/models/__internal__/unet.py
```diff
@@ -177,6 +177,12 @@ def UNet(
     All function parameters require curried functions as inputs which return a
     function that acts on tensors as inputs.
 
+    Reference:
+        - [U-Net: Convolutional Networks for Biomedical Image Segmentation](https://arxiv.org/abs/1505.04597)
+
```
ianstenbit, line 182: We ended up with a lot of newlines here, let's just have one newline between the reference link and the `Args` section. Then I can merge 👍

NiharJani2002: Removed The Extra Lines Between The Research Paper And The Code

ianstenbit:
@NiharJani2002 The linter is unhappy. You can run `sh shell/format.sh` to fix most issues locally and then run `sh shell/lint.sh` to verify that the code format is correct. In this case, we'll need to add `  # noqa: E501` to the end of the model docstring to make the linter happy.

NiharJani2002:
I will check This out (Linter), and generate a new Pull Request, Till Then I am closing this pull request.,
Thanks for the new tool Linter.

PR #1955, opened 2023-07-17
Readme patch

DavidLandup0:
# What does this PR do?
Just adds a `keras` prefix to the `ops` call in the README file, to avoid an error being thrown.

## Before submitting

## Who can review?

@ianstenbit


ianstenbit (approved):
Thanks for the fix!

PR #1958, opened 2023-07-18
Add segmentation mask support to Mosaic and JitteredResize

ianstenbit:
Also fixed the `keras` import on a few other layers.

### DIFF
--- a/examples/layers/preprocessing/segmentation/mosaic_demo.py
+++ b/examples/layers/preprocessing/segmentation/mosaic_demo.py
@@ -1,0 +1,34 @@
+# Copyright 2023 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


examples/layers/preprocessing/segmentation/mosaic_demo.py
```diff
@@ -0,0 +1,34 @@
...
+"""
+import demo_utils
+import tensorflow as tf
+
+from keras_cv.layers import preprocessing
+
+
+def main():
+    ds = demo_utils.load_oxford_iiit_pet_dataset()
+    jittered_resize = preprocessing.Mosaic()
```
jbischof, line 28: `s/jittered_resize/mosaic`?

keras_cv/layers/preprocessing/jittered_resize.py
```diff
@@ -205,6 +205,26 @@ def augment_images(self, images, transformations, **kwargs):
         )
         return tf.cast(scaled_images, self.compute_dtype)
 
+    def augment_segmentation_masks(
```
jbischof, line 208: Why can't we call `self.augment_images` here?

jbischof (approved):
Thanks!

examples/layers/preprocessing/segmentation/mosaic_demo.py
ianstenbit: done

PR #1961, opened 2023-07-19
Use p5 output instead of p4 for deeplab high-level features

ianstenbit:
This was a mistake that I made during the port to Keras Core, and it pretty significantly hurts performance of the model


jbischof:
Do you mean this to be a draft still?

ianstenbit:
> Do you mean this to be a draft still?

I was just waiting until CI looked good to mark it for review. Sorry I preemptively added you. Should be good to go now.

jbischof (approved):

PR #1968, opened 2023-07-23
RandomGaussianBlur crashes when factor=0

muxamilian:
```
>>> import keras_cv
Using TensorFlow backend
>>> blurrer = keras_cv.layers.RandomGaussianBlur(3, 0.)
>>> blurrer.get_random_transformation()
(<tf.Tensor: shape=(3, 1, 1, 1), dtype=float32, numpy=
array([[[[nan]]],


       [[[nan]]],


       [[[nan]]]], dtype=float32)>, <tf.Tensor: shape=(1, 3, 1, 1), dtype=float32, numpy=
array([[[[nan]],

        [[nan]],

        [[nan]]]], dtype=float32)>)
```

# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/keras_cv/layers/preprocessing/random_gaussian_blur.py
+++ b/keras_cv/layers/preprocessing/random_gaussian_blur.py
@@ -63,7 +63,7 @@ class RandomGaussianBlur(BaseImageAugmentationLayer):
     def get_random_transformation(self, **kwargs):
         # `factor` must not become too small otherwise numerical issues occur.
         # keras.backend.epsilon() behaves like 0 without causing `nan`s
-        factor = max(self.factor(), keras.backend.epsilon())
+        factor = tf.math.maximum(self.factor(), keras.backend.epsilon())
         blur_v = RandomGaussianBlur.get_kernel(factor, self.y)
         blur_h = RandomGaussianBlur.get_kernel(factor, self.x)
         blur_v = tf.reshape(blur_v, [self.y, 1, 1, 1])
--- a/keras_cv/layers/preprocessing/random_gaussian_blur.py
+++ b/keras_cv/layers/preprocessing/random_gaussian_blur.py
@@ -33,12 +33,10 @@ class RandomGaussianBlur(BaseImageAugmentationLayer):
         factor: A tuple of two floats, a single float or a
             `keras_cv.FactorSampler`. `factor` controls the extent to which the
             image is blurred. Mathematically, `factor` represents the `sigma`
-            value in a gaussian blur. The closer `factor` is to 0, the more 
-            this operation becomes a no-op operation, and high values make the 
-            blur stronger. However, factor MUST NOT BECOME 0 as then numerical 
-            instability occurs and `nan` results. In order to ensure the value 
-            is always the same, please pass a tuple with two identical floats: 
-            `(0.5, 0.5)`.
+            value in a gaussian blur. `factor=0.0` makes this layer perform a
+            no-op operation, and high values make the blur stronger. In order to
+            ensure the value is always the same, please pass a tuple with two
+            identical floats: `(0.5, 0.5)`.
     """
 
     def __init__(self, kernel_size, factor, **kwargs):
@@ -63,7 +61,9 @@ class RandomGaussianBlur(BaseImageAugmentationLayer):
                 )
 
     def get_random_transformation(self, **kwargs):
-        factor = self.factor()
+        # `factor` must not become too small otherwise numerical issues occur.
+        # 0.01 behaves like 0 without causing `nan`s
+        factor = max(self.factor(), 0.01)
         blur_v = RandomGaussianBlur.get_kernel(factor, self.y)
         blur_h = RandomGaussianBlur.get_kernel(factor, self.x)
         blur_v = tf.reshape(blur_v, [self.y, 1, 1, 1])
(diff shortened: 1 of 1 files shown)


keras_cv/layers/preprocessing/random_gaussian_blur.py
```diff
@@ -33,10 +33,12 @@ class RandomGaussianBlur(BaseImageAugmentationLayer):
         factor: A tuple of two floats, a single float or a
             `keras_cv.FactorSampler`. `factor` controls the extent to which the
             image is blurred. Mathematically, `factor` represents the `sigma`
-            value in a gaussian blur. `factor=0.0` makes this layer perform a
-            no-op operation, and high values make the blur stronger. In order to
-            ensure the value is always the same, please pass a tuple with two
-            identical floats: `(0.5, 0.5)`.
+            value in a gaussian blur. The closer `factor` is to 0, the more
```
ianstenbit, line 36: It looks like we can just fix this numerical issue by adding an epsilon to `factor`  on line 115 where we divide by `factor` -- this should resolve the issue.

ianstenbit (changes requested):
Thanks for the PR! I think we can fix the numerical issue pretty simply instead of just documenting it 😄

keras_cv/layers/preprocessing/random_gaussian_blur.py
muxamilian: In my code I now use 0.01 and it gives the intended behavior. 
```
>>> blurrer = keras_cv.layers.RandomGaussianBlur(3, 0.01)
>>> blurrer.get_random_transformation()
(<tf.Tensor: shape=(3, 1, 1, 1), dtype=float32, numpy=
array([[[[0.]]],


       [[[1.]]],


       [[[0.]]]], dtype=float32)>, <tf.Tensor: shape=(1, 3, 1, 1), dtype=float32, numpy=
array([[[[0.]],

        [[1.]],

        [[0.]]]], dtype=float32)>)
```

keras_cv/layers/preprocessing/random_gaussian_blur.py
```diff
@@ -61,7 +61,9 @@ def __init__(self, kernel_size, factor, **kwargs):
                 )
 
     def get_random_transformation(self, **kwargs):
-        factor = self.factor()
+        # `factor` must not become too small otherwise numerical issues occur.
+        # 0.01 behaves like 0 without causing `nan`s
+        factor = max(self.factor(), 0.01)
```
ianstenbit, line 66: Instead of a hard-coded `0.01`, let's use `keras.backend.epsilon()`

ianstenbit (changes requested):
Thanks -- this is close! Let's just use `keras.backend.epsilon` instead (see my comment in-line)

keras_cv/layers/preprocessing/random_gaussian_blur.py
```diff
@@ -61,7 +61,9 @@ def __init__(self, kernel_size, factor, **kwargs):
                 )
 
     def get_random_transformation(self, **kwargs):
-        factor = self.factor()
+        # `factor` must not become too small otherwise numerical issues occur.
+        # keras.backend.epsilon() behaves like 0 without causing `nan`s
+        factor = max(self.factor(), keras.backend.epsilon())
```
ianstenbit, line 66: Looks like this is breaking tests -- probably because `self.factor()` returns a tf Tensor and `epsilon` is a python float.

Perhaps better would be to say `factor = self.factor() + keras.backend.epsilon()`

You can run the test locally to verify that this isn't breaking

ianstenbit (changes requested):
Thanks! Looks like this now breaks some tests -- take a look at my inline comment

keras_cv/layers/preprocessing/random_gaussian_blur.py
muxamilian: It would be very convenient if the tests ran on each update of the pull request automatically. I think the issue could be fixed now. Can you rerun the tests?

muxamilian: Tested it locally, it seems to work now. I think the problem was that tensorflow has problems compiling the  `max`.

ianstenbit: They do run automatically for folks who have contributed before. However, you should always run tests locally instead of depending on CI to run tests

ianstenbit: Great! Hopefully CI will be happy here as well.

When using `max` in TF you need to use the tf-built in version, not the Python version, if you want to use TF tensors and have it compile into the TF graph. So in this case we could have used the TF version, but just adding the constant value also works just fine 😄

muxamilian: In the meantime I tested it locally and it worked. There is no documentation on how to run the tests but it is not so hard to figure out. Could be improved though.

muxamilian: CI tests look good to me except some unrelated formatting issue in some other code.

ianstenbit: Agreed! The formatting issue is unrelated and I have a fix open in #1993.

We have some (limited) guidance on running tests [here](https://github.com/keras-team/keras-cv/blob/master/.github/CONTRIBUTING.md#run-tests)

ianstenbit (approved):
Thank you!

PR #1977, opened 2023-07-25
Add the full Waymo 3D object detection example

ianstenbit:
This was originally going to be a keras.io guide, but given the addressable audience, and the fact that we'll be publishing a separate blog post with Waymo, I decided to have it live here instead.


ianstenbit:
> Why not keras.io? Sorry I didn't follow

@jbischof
Basically, this just serves as something we can reference from the Waymo blog post -- it's not meant to be independently discoverable on keras.io

A more generic, formalized "guide" would belong on KerasIO but this is more like an appendix to a pending publication from Waymo

examples/training/object_detection_3d/waymo/train_pillars.py
```diff
@@ -1,129 +1,320 @@
...
-        default_value=-1.0, shape=[global_batch, 1, 200000, 8]
+"""
+Now we can build our model!
+
+We'll define a function to create the model so that we can initialize it inside
+of a tf.distribute scope later on.
+"""
+
+
+def build_centerpillar_model():
```
jbischof, line 192: All I see is "caterpillar" model 😆 . Maybe `center_pillar`?

examples/training/object_detection_3d/waymo/train_pillars.py
```diff
@@ -1,129 +1,320 @@
...
+    """
+    Our first model component is a voxelization layer. This will be used to
+    dynamically map coordinates of a point to a voxel in 3D space.
+    """
+    voxelization_point_net = tf.keras.Sequential(
```
jbischof, line 197: Maybe this should be synced with your Keras Core port?

examples/training/object_detection_3d/waymo/train_pillars.py
```diff
@@ -1,129 +1,320 @@
...
-train_ds = train_ds.apply(
-    tf.data.experimental.dense_to_ragged_batch(
-        global_batch, drop_remainder=True
+
+"""
+Now we can build and compile our model!
+We use a one device strategy in this tutorial, but any strategy will work.
+"""
+
+strategy = tf.distribute.OneDeviceStrategy("/gpu:0")
```
jbischof, line 289: Why is a strategy needed for one device? Don't we do this all the time in our examples without this complication?

jbischof (approved):
Very nice, thank you!

This is a very advanced topic and more technically demanding guide than anything we've ever published. One day it could be cool to write an "intro to 3D OD" style guide that explains what a voxel is but for now this will be very helpful to those in the know.

wof
I'd like to buy a voxel for $250!

PR #1067, opened 2022-12-01
Update stable_diffusion `num_steps` to be 50

LukeWood:
fixes https://github.com/keras-team/keras-cv/issues/1043

PR #1980, opened 2023-07-26
Fix broken StableDiffusion tests

ianstenbit:
These tests were broken by #1067

### DIFF
--- a/keras_cv/models/stable_diffusion/stable_diffusion_test.py
+++ b/keras_cv/models/stable_diffusion/stable_diffusion_test.py
@@ -21,6 +21,7 @@ from keras_cv.tests.test_case import TestCase
 
 
 @pytest.mark.extra_large
+@pytest.mark.tf_keras_only
 class StableDiffusionTest(TestCase):
     def test_end_to_end_golden_value(self):
         prompt = "a caterpillar smoking a hookah while sitting on a mushroom"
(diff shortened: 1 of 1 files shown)


keras_cv/models/stable_diffusion/stable_diffusion_test.py
```diff
@@ -12,30 +12,34 @@
...
 
+import pytest
 import tensorflow as tf
 from tensorflow.keras import mixed_precision
 
 from keras_cv.models import StableDiffusion
 from keras_cv.tests.test_case import TestCase
 
 
+@pytest.mark.extra_large
```
jbischof, line 23: Should we add `@pytest.mark.tf_keras_only`?

keras_cv/models/stable_diffusion/stable_diffusion_test.py
```diff
@@ -12,30 +12,34 @@
...
 class StableDiffusionTest(TestCase):
-    def DISABLED_test_end_to_end_golden_value(self):
+    def test_end_to_end_golden_value(self):
         prompt = "a caterpillar smoking a hookah while sitting on a mushroom"
         stablediff = StableDiffusion(128, 128)
 
-        img = stablediff.text_to_image(prompt, seed=1337)
+        img = stablediff.text_to_image(prompt, seed=1337, num_steps=25)
```
jbischof, line 29: You could also just change the answer to `[32, 255, 238]`

jbischof (approved):
Thanks!

keras_cv/models/stable_diffusion/stable_diffusion_test.py
ianstenbit: Yeah, I just wanted to make sure that something wasn't broken and we did in fact still have a stable forward pass. At the very least, this makes the test much faster because it does fewer diffusion steps.

PR #1982, opened 2023-07-26
Migrate StableDiffusion to Keras Core

### LINKED ISSUE #1952: Port Stable Diffusion to Keras Core

jbischof:
Resolves #1952

Manually ran `extra_large` tests on local machine.


jbischof:
Currently getting error:
```
AttributeError: module 'keras.api._v2.keras' has no attribute 'random' 
```

keras_cv/models/stable_diffusion/noise_scheduler.py
```diff
@@ -198,14 +195,14 @@ def add_noise(
         noise,
         timesteps,
     ):
-        sqrt_alpha_prod = tf.gather(self.alphas_cumprod, timesteps) ** 0.5
+        sqrt_alpha_prod = ops.take(self.alphas_cumprod, timesteps) ** 0.5
```
jbischof, line 198: @mattdangerw are these equivalent?

mattdangerw: Looks like 1d inputs? should be!

keras_cv/models/stable_diffusion/stable_diffusion.py
```diff
@@ -325,19 +328,21 @@ def _get_initial_alphas(self, timesteps):
...
+            return random.normal(
+                (batch_size, self.img_height // 8, self.img_width // 8, 4),
+                seed=self.seed_generator,
             )
 
     @staticmethod
     def _get_pos_ids():
-        return tf.convert_to_tensor(
-            [list(range(MAX_PROMPT_LENGTH))], dtype=tf.int32
+        return ops.convert_to_tensor(
```
jbischof, line 343: Should probably just be `arange` or something more numpy

ianstenbit:
I screwed up this branch so I'm going to nuke this PR and start over

ianstenbit:
I think this is ready for another look.

The only outstanding question I have is:

How do we want to handle testing?

Golden values will be different across backends because the random seeding can't be made exactly identical for all 3. We could hypothetically store some floats in an NPZ file in the repo to address this, but that feels ugly to me.

Maybe we just test the numerics with only 1 backend, and for the others just assert that it is able to do a forward pass?

ianstenbit (approved):
This generally LGTM (although I added some commits so we should def. get an approval from Matt).

WRT testing -- I am happy to figure that out here or in a separate PR.

jbischof:
> I think this is ready for another look.
> 
> The only outstanding question I have is:
> 
> How do we want to handle testing?
> 
> Golden values will be different across backends because the random seeding can't be made exactly identical for all 3. We could hypothetically store some floats in an NPZ file in the repo to address this, but that feels ugly to me.
> 
> Maybe we just test the numerics with only 1 backend, and for the others just assert that it is able to do a forward pass?

I agree with this strategy @ianstenbit! Now that we're reduced the runtime of these tests, I propose adding the e2e golden value test to GCB and leaving the rest for manual debugging. WDYT?

keras_cv/models/stable_diffusion/stable_diffusion_test.py
```diff
@@ -13,24 +13,23 @@
...
+from keras_cv.backend import ops
+from keras_cv.backend import random
 from keras_cv.models import StableDiffusion
 from keras_cv.tests.test_case import TestCase
 
 
 @pytest.mark.extra_large
-@pytest.mark.tf_keras_only
 class StableDiffusionTest(TestCase):
+    @pytest.mark.large
```
jbischof, line 26: Actually this is already marked for GCB testing. Don't remember if I did this or you, but perfect!

jbischof:
@ianstenbit friendly ping on testing here

jbischof:
@mattdangerw friendly ping for review

ianstenbit (approved):
New test approach LGTM -- thanks Jonathan!
Feel free to merge.

PR #1983, opened 2023-07-26
Remove pinning of keras-core, set min TF vers to 2.13

sampathweb:
- Removes pinning of `keras-core` version so it works on latest.
- Sets min-TF version as 2.13 since that's the dependency in `keras-core`
- Updates minimum python version to 3.8, which is the minimum for TF 2.13

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -29,10 +29,10 @@ jobs:
           ${{ runner.os }}-pip-
     - name: Install dependencies
       run: |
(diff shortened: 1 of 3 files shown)


ianstenbit (approved):
Great, thank you!

ianstenbit:
Note that `version_check.py` is still only enforcing TF 2.11+, which is intentional because for `tf.keras` this is a reasonable version check, and our Waymo integration is pinned to TF 2.11 for now.

PR #1987, opened 2023-07-28
Add the Segment Anything Model to KerasCV

### LINKED ISSUE #1679: Add Segment Anything Model (SAM)
**Short Description**

Facebook research.

A recent new segmentation models, by facebook. The Segment Anything Model (SAM) produces high quality object masks from input prompts such as points or boxes, and it can be used to generate masks for all objects in an image. It has been trained on a [dataset](https://segment-anything.com/dataset/index.html) of 11 million images and 1.1 billion masks, and has strong zero-shot performance on a variety of segmentation tasks.

**Papers**

https://ai.facebook.com/blog/segment-anything-foundation-model-image-segmentation/

**Existing Implementations**

https://github.com/facebookresearch/segment-anything

**Other Information**

Paper: https://ai.facebook.com/research/publications/segment-anything/
Website: https://segment-anything.com/
Demo: https://segment-anything.com/demo
[Model Checkpoints](https://github.com/facebookresearch/segment-anything#model-checkpoints)

tirthasheshpatel:
# What does this PR do?

This PR implements the Segment Anything Model in multi-backend Keras.

Fixes #1679
See also #1933

## Before submitting

      Pull Request section?
      to it if that's the case.

## Who can review?

@ianstenbit @DavidLandup0


keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
```diff
@@ -0,0 +1,303 @@
...
+
+        self.built = True
+
+    def call(
+        self,
+        image_embeddings,
+        image_pe,
+        sparse_prompt_embeddings,
+        dense_prompt_embeddings,
+        multimask_output,
```
ianstenbit, line 199: should this have a default?

keras_cv/models/segmentation/segment_anything/sam.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2023 The KerasCV Authors
```
ianstenbit, line 1: I think we should probably have a top-level `SegmentAnything` model which takes an `ImageEncoder` as a backbone and subclasses `Task`. Then the high-level workflows can live on that model.

Then we can also include a preset which includes your ported weights!

Let's also include a reference to the paper and original implementation

ianstenbit (commented):
Thanks for the PR! 

This is super exciting 🎊 

Just a few comments as I took a quick look

keras_cv/models/segmentation/segment_anything/sam.py
DavidLandup0: The image encoder for SAM is a near 1:1 from Detectron2's ViTDet.
IMO it makes sense to have ViTDet as a standalone class/network rather than a SAM encoder only.
That way we can train it from scratch, have it as a standalone object detection model, a backbone for SAM and reuse the same code across all of that

keras_cv/models/segmentation/segment_anything/sam_layers.py
```diff
@@ -0,0 +1,106 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from keras_cv.backend import keras
+from keras_cv.backend import ops
+
+
+@keras.utils.register_keras_serializable(package="keras_cv")
+class MLPBlock(keras.layers.Layer):
```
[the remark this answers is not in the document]
tirthasheshpatel: I is used a few times in the mask decoder actually. So, just inline the dense layers would just duplicate a lot of code. Is there any side effect of having this? If not, I'd prefer keeping it.

keras_cv/models/segmentation/segment_anything/sam_layers.py
```diff
@@ -0,0 +1,106 @@
...
+            {
+                "embedding_dim": self.embedding_dim,
+                "mlp_dim": self.mlp_dim,
+                "activation": self.activation,
+            }
+        )
+
+
+@keras.utils.register_keras_serializable(package="keras_cv")
+class SAMLayerNormalization(keras.layers.Layer):
```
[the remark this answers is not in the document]
tirthasheshpatel: There is `keras.layers.LayerNormalization(epsilon=1e-6)`. I will push this in the next batch of commits.

keras_cv/models/segmentation/segment_anything/sam.py
tirthasheshpatel: I agree @DavidLandup0, I will move the layer to `keras_cv/layers`.

tirthasheshpatel: > I think we should probably have a top-level `SegmentAnything` model which takes an `ImageEncoder` as a backbone and subclasses `Task`. Then the high-level workflows can live on that model.

That's the plan. I will add a `Task` model in the `kera_cv/models/segmentation/segment_anything/sam.py` file. I am not yet sure how exactly the training step would be implemented with the `Task` API, we could just raise a `NotImplementedError` for now and write a train step as a follow-up.

keras_cv/models/segmentation/segment_anything/sam_layers.py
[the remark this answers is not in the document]
ianstenbit: If it's re-used in many places then I am alright with it -- it looked to me like it was only used once or twice but I probably missed some uses

keras_cv/models/segmentation/segment_anything/sam_layers.py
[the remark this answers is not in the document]
DavidLandup0: We should probably double check and don't take my word for it, but I'm not sure if the numerics are the same. `keras.layers.LayerNormalization` is:

```
        # Compute the batch normalization.
        inv = 1 / ops.sqrt(variance + self.epsilon)
        if scale is not None:
            scale = ops.cast(scale, inputs.dtype)
            inv = inv * scale
        x = -mean * inv
        if offset is not None:
            offset = ops.cast(offset, inputs.dtype)
            x = offset + x
        outputs = inputs * ops.cast(inv, inputs.dtype) + ops.cast(
            x, inputs.dtype
        )

        outputs = ops.cast(outputs, input_dtype)

        # If some components of the shape got lost due to adjustments, fix that.
        outputs = ops.reshape(outputs, ops.shape(inputs))
```

For SAM, they call it `LayerNorm2d()` in the official implementation, but the official impl is taken directly from Detectron2 which has BatchNorm2D and LayerNorm in turn taken from ConvNeXt: https://github.com/facebookresearch/ConvNeXt/blob/d1fa8f6fef0a165b27399986cc2bdacc92777e40/models/convnext.py#L119

Technically, the LayerNorm re-implementation in ConvNeXt, SAM and Detectron2 shouldn't be the same LayerNorm from PyTorch. 

More in these issues: 
- https://github.com/facebookresearch/ConvNeXt/issues/112
- https://github.com/facebookresearch/ConvNeXt/issues/136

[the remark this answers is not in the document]
tirthasheshpatel: Thanks for linking these issues @DavidLandup0, I too was wondering why this was reimplemented. After some testing, I can confirm that `LayerNormalization(epsilon=1e-6)` is numerically equivalent to `SAMLayerNormalization()` for segment anything. Here's the code I used to test:

```python
import numpy as np
import torch
import keras_core as keras
from keras_cv.models.segmentation.segment_anything import sam_layers

sam_ln = sam_layers.SAMLayerNormalization()
ln = keras.layers.LayerNormalization(epsilon=1e-6)

sam_ln.build((1, 512, 512, 3))
ln.build((1, 512, 512, 3))

sam_ln.set_weights(ln.weights)

x_np = np.random.randint(0, 256, size=(1, 512, 512, 3), dtype=np.uint8)
x_np = x_np.astype(np.float32)
x = torch.tensor(x_np, requires_grad=True)
x_sam = torch.tensor(x_np, requires_grad=True)

x_out_sam = sam_ln(x_sam)
x_out = ln(x)

x_out_sam.backward(torch.ones_like(x_out_sam))
x_out.backward(torch.ones_like(x_out))

np.testing.assert_allclose(
    x_out_sam.detach().numpy(),
    x_out.detach().numpy(),
    rtol=8e-5
)
np.testing.assert_allclose(
    ln.weights[0].value.grad.detach().numpy(),
    sam_ln.weights[2].value.grad.detach().numpy(), rtol=6e-7
)
np.testing.assert_allclose(
    ln.weights[1].value.grad.detach().numpy(),
    sam_ln.weights[3].value.grad.detach().numpy()
)
np.testing.assert_allclose(
    x_sam.grad.detach().numpy(),
    x.grad.detach().numpy(), atol=3e-7
)
```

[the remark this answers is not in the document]
DavidLandup0: Awesome, thanks for checking! That simplifies things a lot :D

keras_cv/models/segmentation/segment_anything/sam.py
tirthasheshpatel: Update: I have moved the image encoder to a standalone backbone. Let me know if that looks good to you. Thanks for the suggestion @DavidLandup0!

keras_cv/layers/detectron2_layers.py
```diff
@@ -0,0 +1,431 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from keras_cv.backend import keras
+from keras_cv.backend import ops
+from keras_cv.models.segmentation.segment_anything.sam_layers import MLPBlock
+
+
+def get_rel_pos(query_size, key_size, rel_pos):
```
ianstenbit, line 20: Let's move the helper functions to the bottom of the file

keras_cv/layers/detectron2_layers_test.py
```diff
@@ -0,0 +1,48 @@
...
+            num_heads=16,
+            key_dim=1280 // 16,
+            use_bias=True,
+            input_size=(64, 64),
+        )
+        x = np.ones(shape=(1, 64, 64, 1280))
+        x_out = ops.convert_to_numpy(attention_with_rel_pe(x))
+        self.assertEqual(x_out.shape, (1, 64, 64, 1280))
+
+    def test_windowed_transformer_encoder(self):
```
ianstenbit, line 35: Let's add a test for ViTDetPatchingAndEmbedding as well

keras_cv/models/backbones/detectron2/detectron2_backbone.py
```diff
@@ -0,0 +1,203 @@
...
+from keras_cv.layers.detectron2_layers import ViTDetPatchingAndEmbedding
+from keras_cv.layers.detectron2_layers import WindowedTransformerEncoder
+from keras_cv.models.backbones.backbone import Backbone
+from keras_cv.models.backbones.detectron2.detectron2_backbone_presets import (
+    backbone_presets,
+)
+from keras_cv.utils.python_utils import classproperty
+
+
+@keras.utils.register_keras_serializable(package="keras_cv.models")
```
ianstenbit, line 27: Sorry that this changed while this PR is in-flight, but if you sync to master this should now be

```
from keras_cv.api_export import keras_cv_export

@keras_cv_export("keras_cv.models.ViTDetBackbone")
```
(Same for all new public API symbols)

keras_cv/models/backbones/detectron2/detectron2_backbone.py
```diff
@@ -0,0 +1,203 @@
...
+            attention in the transformer encoder blocks. Defaults to `0`.
+        global_attention_indices (list, optional): Indexes for blocks using
+            global attention. Defaults to `[7, 15, 23, 31]`.
+        layer_norm_epsilon (int, optional): The epsilon to use in the layer
+            normalization blocks in transformer encoder. Defaults to `1e-6`.
+    """
+
+    def __init__(
+        self,
+        img_size=1024,
```
ianstenbit, line 68: We've standardized on `input_shape` for this in other backbones (accepting a tuple of `(height, width, channels)`)

keras_cv/models/backbones/detectron2/detectron2_backbone.py
```diff
@@ -0,0 +1,203 @@
...
+from keras_cv.layers.detectron2_layers import WindowedTransformerEncoder
+from keras_cv.models.backbones.backbone import Backbone
+from keras_cv.models.backbones.detectron2.detectron2_backbone_presets import (
+    backbone_presets,
+)
+from keras_cv.utils.python_utils import classproperty
+
+
+@keras.utils.register_keras_serializable(package="keras_cv.models")
+class ViTDetBackbone(Backbone):
```
ianstenbit, line 28: I assume that this backbone doesn't produce pyramid_level_outputs since it's a transformer architecture -- let's call this out in the docstring, and maybe even create an @property for self.pyramid_level_inputs which throws a nice NotImplementedError

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
```diff
@@ -0,0 +1,273 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from keras_cv.backend import keras
+from keras_cv.backend import ops
+
+
+@keras.utils.register_keras_serializable(package="keras_cv")
+class MLP(keras.layers.Layer):
```
ianstenbit, line 20: Is this different from `MLPBlock`?

keras_cv/models/segmentation/segment_anything/sam_test.py
```diff
@@ -0,0 +1,225 @@
...
+from keras_cv.models.segmentation.segment_anything.sam_transformer import (
+    TwoWayMultiHeadAttention,
+)
+from keras_cv.models.segmentation.segment_anything.sam_transformer import (
+    TwoWayTransformer,
+)
+from keras_cv.tests.test_case import TestCase
+
+
+class TestSAM(TestCase):
```
ianstenbit, line 36: nit: SAMTest

ianstenbit (commented):
Great progress!

ianstenbit:
@tirthasheshpatel LMK when you're ready for another review on this 😄

keras_cv/layers/detectron2_layers.py
tirthasheshpatel: Addressed in [ac7f30e](https://github.com/keras-team/keras-cv/pull/1987/commits/ac7f30ee2cb0ee7569244cf67dd18a50ef805752)

keras_cv/layers/detectron2_layers_test.py
tirthasheshpatel: Addressed in [ac7f30e](https://github.com/keras-team/keras-cv/pull/1987/commits/ac7f30ee2cb0ee7569244cf67dd18a50ef805752)

keras_cv/models/backbones/detectron2/detectron2_backbone.py
tirthasheshpatel: Addressed in [ac7f30e](https://github.com/keras-team/keras-cv/pull/1987/commits/ac7f30ee2cb0ee7569244cf67dd18a50ef805752)

keras_cv/models/backbones/detectron2/detectron2_backbone.py
tirthasheshpatel: Addressed in [ac7f30e](https://github.com/keras-team/keras-cv/pull/1987/commits/ac7f30ee2cb0ee7569244cf67dd18a50ef805752)

keras_cv/models/backbones/detectron2/detectron2_backbone.py
tirthasheshpatel: I tried to add a `input_shape` attribute but since I am not using the `Functional` syntax for building the model, it doesn't allow me to add that attribute.

In case you didn't notice, I am using the `call` method to specify the computations in the model instead of passing a symbolic input through each layer/operation in the `__init__` method. 

I noticed that it's just easier not to deal with symbolic inputs with Keras Core. One of the main reasons why Keras Core struggles with symbolic inputs is that it doesn't do shape inference. For example, in Keras, this works:

```python
from tensorflow import keras

x = keras.Input([2, 3])
tf.shape(x)[0] * 10  # Note that even though the shape at axis 0 is None,
                     # TensorFlow returns a symbolic tensor making computations
                     # like these valid instead of throwing an exception.
```

but Keras Core fails, since `x.shape[0]` is `None`.

I think it should not be difficult to convert the implementation to fully use the `Functional` syntax but we would have to manually check if the shapes we are getting are `None`s or not. So, for example, we could do this in Keras Core:

```python
import keras_core
from keras_core import ops

x = keras_core.Input([2, 3])
if x.shape[0] is not None:
    x = ops.reshape(x, (x.shape[0] * 2, 3))
else:
    x = ops.reshape(x, (None, 3))
```

or use some other operation.

Sorry for not highlighting these details properly beforehand! I will add a bunch of comments where I do something unintuitive so it's easier for future reviewers.

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
tirthasheshpatel: We could unite those. The only difference is that the `MLPBlock` has architecture `embedding_dim -> mlp_dim -> embedding_dim` while `MLP` has architecture `input_dim -> [hidden_dim] * (num_layers - 1) -> output_dim`. Looks like a low-hanging fruit, will address in the next commit.

tirthasheshpatel: Done.

keras_cv/models/segmentation/segment_anything/sam_test.py
tirthasheshpatel: Done.

keras_cv/models/backbones/detectron2/detectron2_backbone.py
```diff
@@ -0,0 +1,203 @@
...
+    """A ViT image encoder that uses a windowed transformer encoder and
+    relative positional encodings.
+
+    Args:
+        img_size (int, optional): The size of the input image. Defaults to
+            `1024`.
+        patch_size (int, optional): the patch size to be supplied to the
+            Patching layer to turn input images into a flattened sequence of
+            patches. Defaults to `16`.
+        in_chans (int, optional): The number of channels in the input image.
```
tirthasheshpatel, line 41: The docs are outdated here, will update.

tirthasheshpatel:
Hi @ianstenbit, thank you very much for your reviews so far! Very helpful!

> LMK when you're ready for another review on this

I have a `Task` API for the SAM model ready, I will just have to make some more changes before pushing it + add tests. I will mark the PR ready for review once I do that and help start reviews. But if you have time, feel free to review!

keras_cv/layers/__init__.py
```diff
@@ -17,6 +17,10 @@
 from tensorflow.keras.layers import RandomWidth
 
 from keras_cv.layers.augmenter import Augmenter
+from keras_cv.layers.detectron2_layers import AddPositionalEmbedding
```
DavidLandup0, line 20: Do we want to put these under a `detectron2` namespace?

DavidLandup0, line 20: Since this would be exported as part of the public API - we have `PatchingAndEmbedding` which does patching with a Conv2D and then adds embeddings in this same form. Do we want to update that layer to use `AddPositionalEmbedding` as well for conformity?

keras_cv/layers/detectron2_layers.py
```diff
@@ -0,0 +1,477 @@
...
+            axes=(2, 0, 3, 1, 4),
+        )
+        qkv = ops.reshape(qkv, (3, B * self.num_heads, H * W, self.key_dim))
+        queries, keys, values = ops.unstack(qkv, axis=0)
+        attention_map = (queries * self.scale) @ ops.transpose(
+            keys, axes=(0, 2, 1)
+        )
+
+        if self.use_rel_pos:
+            attention_map = add_decomposed_rel_pos(
```
DavidLandup0, line 106: This should probably be a private method as part of a layer

keras_cv/layers/detectron2_layers.py
```diff
@@ -0,0 +1,477 @@
...
+                "input_size": self.input_size,
+                "activation": self.activation,
+                "layer_norm_epsilon": self.layer_norm_epsilon,
+            }
+        )
+        return config
+
+
+@keras_cv_export("keras_cv.layers.ViTDetPatchingAndEmbedding")
+class ViTDetPatchingAndEmbedding(keras.layers.Layer):
```
DavidLandup0, line 259: This is the same as the ViT patching and embedding but without positional embedding. 
I'm torn between being able to turn off PE in the default layer and adding that as a flag and having a new layer for this...

keras_cv/layers/detectron2_layers.py
```diff
@@ -0,0 +1,477 @@
...
+            {
+                "img_size": self.img_size,
+                "patch_size": self.patch_size,
+                "embed_dim": self.embed_dim,
+            }
+        )
+        return config
+
+
+def get_rel_pos(query_size, key_size, rel_pos):
```
DavidLandup0, line 345: This should probably be a private method or turned into a public layer.
I.e.: https://github.com/DavidLandup0/deepvision/blob/main/deepvision/layers/decomposed_relative_positional_embedding.py

keras_cv/layers/detectron2_layers.py
```diff
@@ -0,0 +1,477 @@
...
+        ops.cast(ops.maximum(query_size // key_size, 1), dtype="float32")
+    )
+    relative_coordinates = (query_coordinates - key_coordinates) + ops.cast(
+        key_size - 1, dtype="float32"
+    ) * ops.cast(ops.maximum(query_size // key_size, 1), dtype="float32")
+    relative_coordinates = ops.cast(relative_coordinates, dtype="int32")
+    return ops.take(rel_pos_resized, relative_coordinates, 0)
+
+
+def add_decomposed_rel_pos(
```
DavidLandup0, line 384: Same comment as above

keras_cv/layers/serializable_sequential.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+# TODO(tirthasheshpatel): Use `Sequential` model once the bug is resolved.
+# Temporarily substitute the `Sequential` model with this because a
+# bug in Keras/Keras Core prevents the weights of a sequential model to
+# load in TensorFlow if they are saved in JAX/PyTorch and vice versa.
+# This only happens when the `build` method is called in the `__init__`
+# step.
+@keras_cv_export("keras_cv.layers.SerializableSequential")
+class SerializableSequential(keras.layers.Layer):
```
DavidLandup0, line 27: Is this still an issue in Keras Core?

keras_cv/models/__init__.py
```diff
@@ -43,6 +43,18 @@
 from keras_cv.models.backbones.densenet.densenet_backbone import (
     DenseNetBackbone,
 )
+from keras_cv.models.backbones.detectron2.detectron2_aliases import (
```
DavidLandup0, line 46: Afaik, the backbone is basically the same as the official ViTDet, so there may not be a need to call it a SAM{name}Backbone

keras_cv/models/__init__.py
```diff
@@ -166,5 +178,8 @@
     YOLOV8Detector,
 )
 from keras_cv.models.segmentation import DeepLabV3Plus
+from keras_cv.models.segmentation import MaskDecoder
```
DavidLandup0, line 181: May be better as SAMMaskDecoder for clarity

keras_cv/models/backbones/detectron2/data/sam_vitdet_b_out.npz
DavidLandup0, line 1: Probably left by accident?

keras_cv/models/backbones/detectron2/detectron2_backbone.py
DavidLandup0: The backbone outputs the same shapes itself, but they do use a feature pyramid output: https://arxiv.org/pdf/2203.16527.pdf

TL:DR for the paper, the simple feature pyramid on the right turned out to be the most performant for them. 
image

keras_cv/models/backbones/detectron2/detectron2_backbone.py
```diff
@@ -0,0 +1,203 @@
...
+            }
+        )
+        return config
+
+    @classproperty
+    def presets(cls):
+        """Dictionary of preset names and configurations."""
+        return copy.deepcopy(backbone_presets)
+
+    # @classproperty
```
DavidLandup0, line 199: stray comment?

keras_cv/models/segmentation/segment_anything/sam_layers.py
```diff
@@ -0,0 +1,66 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from keras_cv.api_export import keras_cv_export
+from keras_cv.backend import keras
+from keras_cv.layers.serializable_sequential import SerializableSequential
+
+
+@keras_cv_export("keras_cv.layers.MLP")
+class MLP(keras.layers.Layer):
```
DavidLandup0, line 21: This is a public class here - it should probably be a private subclass, especially since there was an MLP with the same name in a layer related to this, iirc

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
```diff
@@ -0,0 +1,230 @@
+# Copyright 2023 The KerasCV Authors
```
DavidLandup0, line 1: I'd probably put this layer under sam layers

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
```diff
@@ -0,0 +1,230 @@
...
+from keras_cv.api_export import keras_cv_export
+from keras_cv.backend import keras
+from keras_cv.backend import ops
+from keras_cv.layers.serializable_sequential import SerializableSequential
+from keras_cv.models.segmentation.segment_anything.sam_layers import MLP
+
+
+@keras_cv_export("keras_cv.models.MaskDecoder")
+class MaskDecoder(keras.models.Model):
+    """Mask decoder for the segment anything model.
```
DavidLandup0, line 24: Nit: "Segment Anything (SAM)"

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
```diff
@@ -0,0 +1,230 @@
...
+
+from keras_cv.api_export import keras_cv_export
+from keras_cv.backend import keras
+from keras_cv.backend import ops
+from keras_cv.layers.serializable_sequential import SerializableSequential
+from keras_cv.models.segmentation.segment_anything.sam_layers import MLP
+
+
+@keras_cv_export("keras_cv.models.MaskDecoder")
+class MaskDecoder(keras.models.Model):
```
DavidLandup0, line 23: As mentioned before, to avoid confusion, probably best if this is called SAMMaskDecoder or something along those lines

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
```diff
@@ -0,0 +1,230 @@
...
+            `True`.
+        iou_head_depth (int): The depth of the dense net used to predict the
+            IoU confidence score.
+        iou_head_hidden_dim (int): The number of units in the hidden layers
+            used in the dense net to predict the IoU confidence score.
+        activation (str, optional): Activation to use in the mask upscaler
+            network. Defaults to "gelu".
+
+    References:
+        - [Segment Anything](https://arxiv.org/abs/2304.02643)
```
DavidLandup0, line 58: We probably want a code reference as well

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
```diff
@@ -0,0 +1,230 @@
...
+                "iou_head_hidden_dim": self.iou_head_hidden_dim,
+                "activation": self.activation,
+            }
+        )
+        return config
+
+    @classmethod
+    def from_config(cls, config):
+        config.update(
+            {"transformer": keras.layers.deserialize(config["transformer"])}
```
DavidLandup0, line 228: Shouldn't we be able to pass the config here instead of the object if it's serializable?

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
```diff
@@ -0,0 +1,321 @@
...
+            scale (float): The standard deviation of the random frequencies.
+        """
+        super().__init__(**kwargs)
+        self.num_positional_features = num_positional_features
+        self.scale = scale
+        init_func = lambda *a, **kw: self.scale * ops.random.normal(
+            shape=(2, self.num_positional_features), dtype=self.dtype
+        )
+        self.positional_encoding_gaussian_matrix = self.add_weight(
+            name="positional_emcoding_gaussian_matrix",
```
DavidLandup0, line 43: emcoding -> encoding

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
```diff
@@ -0,0 +1,321 @@
...
+            shape=(2, self.num_positional_features),
+            dtype=self.dtype,
+            trainable=False,
+            initializer=init_func,
+        )
+
+        self.built = True
+
+    def __positional_encodings(self, coords):
```
DavidLandup0, line 52: Maybe a layer?

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
```diff
@@ -0,0 +1,321 @@
...
+        grid = ops.ones(shape=(H, W), dtype=self.dtype)
+        y_embed = ops.cumsum(grid, axis=0) - 0.5
+        x_embed = ops.cumsum(grid, axis=1) - 0.5
+        y_embed = y_embed / ops.cast(H, self.dtype)
+        x_embed = x_embed / ops.cast(W, self.dtype)
+        return self.__positional_encodings(
+            ops.stack([x_embed, y_embed], axis=-1)
+        )
+
+    def call_with_coords(self, coords_input, image_size):
```
DavidLandup0, line 78: Maybe refactoring to a cleaner name? "encode_coordinates" sounds a lot nicer than the official "call_with_coords" IMO

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
```diff
@@ -0,0 +1,321 @@
...
+            {
+                "num_positional_features": self.num_positional_features,
+                "scale": self.scale,
+            }
+        )
+        return config
+
+
+@keras_cv_export("keras_cv.models.PromptEncoder")
+class PromptEncoder(keras.models.Model):
```
DavidLandup0, line 110: Same comment as for the MaskDecoder

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
```diff
@@ -0,0 +1,321 @@
...
+            self.top_left_corner_embed,
+            self.bottom_right_corner_embed,
+            self.not_a_point_embed,
+            self.no_mask_embed,
+        ]:
+            layer.build(None)
+
+        self.built = True
+
+    def get_dense_pe(self):
```
DavidLandup0, line 210: I think this should be a private method, the same as the positional embeddings for the previous layer - or potentially turned into a layer themselves?

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
```diff
@@ -0,0 +1,321 @@
...
+        corner_embedding = ops.stack(
+            [top_left_embedding, bottom_right_embedding], axis=1
+        )
+        return corner_embedding
+
+    def __embed_mask(self, mask):
+        mask_embedding = self.mask_downscaler(mask)
+        return mask_embedding
+
+    def call(self, points=None, labels=None, box=None, mask=None):
```
DavidLandup0, line 270: Maybe batch together the statments under the `if` and `elif` statements? `if points is not None -> get bs -> do point embedding` for example

DavidLandup0 (commented):
Thank you for all of the work that's gone into this! Super nice to see it shaping up. 🎊 

Just sharing a couple of my thoughts:

There are a few methods that may be better off as layers (for example, window partitioning, window unpartitioning, etc.) which are usable for other architectures, such as MaxViT. For example, here are [WindowPartitioning](https://github.com/DavidLandup0/deepvision/blob/main/deepvision/layers/window_partitioning.py) and [WindowUnpartitioning](https://github.com/DavidLandup0/deepvision/blob/main/deepvision/layers/window_unpartitioning.py) from DeepVision

We have a great opportunity to clean up the official codebase with a few changes here. For instance, to stay consistent with `AddPositionalEmbedding`, we'll want to turn the [decomposed positional embeddings](https://github.com/DavidLandup0/deepvision/blob/main/deepvision/layers/decomposed_relative_positional_embedding.py), etc. into layers as well as opposed to the official repo's functions.

We also have the opportunity to standardize some existing layers - such as `PatchingAndEmbedding` being the same as `ViTDetPatchingAndEmbedding` + `AddPositionalEmbedding`. We could do a general layer that consists of patching and a general layer consisting of embedding, and re-implement the former with the two.

These are all primarily API decisions, which are always tricky... Ooooor, some of these can be put under the SAM folder without shipping them as part of the public API?

ianstenbit:
> Hi @ianstenbit, thank you very much for your reviews so far! Very helpful!
> 
> > LMK when you're ready for another review on this
> 
> I have a `Task` API for the SAM model ready, I will just have to make some more changes before pushing it + add tests. I will mark the PR ready for review once I do that and help start reviews. But if you have time, feel free to review!

Thanks! I just looked through David's review and I'd say that's plenty of stuff to look at -- thanks for the great progress on this, and thanks for the thorough review @DavidLandup0!

keras_cv/layers/__init__.py
tirthasheshpatel: Wouldn't adding a new layer in a pre-existing class invalidate the weights set for the ViT model? Also, since `PatchingAndEmbedding` is still a TensorFlow Keras layer, I think, for the time being, it'd be easier to keep the two separate.

Although, I'd add a comment about this as a TODO so we don't forget to do it in the future. What do you think?

keras_cv/layers/detectron2_layers.py
tirthasheshpatel: Are you referring to the `PathingAndEmbedding` class for the ViT model in KerasCV? I addressed that here: https://github.com/keras-team/keras-cv/pull/1987#discussion_r1304907301

keras_cv/models/__init__.py
tirthasheshpatel: Sounds good, thanks for looking into it!

keras_cv/models/backbones/detectron2/data/sam_vitdet_b_out.npz
tirthasheshpatel: I have added this intentionally. This is used in the tests to verify that the model weights are loaded correctly and that the forward pass in all backends yields the same result.

keras_cv/models/backbones/detectron2/detectron2_backbone.py
tirthasheshpatel: This method loads the presets with weights. I will uncomment it later once the model layers are finalized and the final weights are uploaded.

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
tirthasheshpatel: I don't think so. Keras errors out if I don't serialize and deserialize manually. RetinaNet and YOLOV8 also serialize and deserialize the layer arguments manually. Perhaps, @ianstenbit has a better idea about why this is required.

tirthasheshpatel:
> There are a few methods that may be better off as layers (for example, window partitioning, window unpartitioning, etc.) which are usable for other architectures, such as MaxViT. For example, here are [WindowPartitioning](https://github.com/DavidLandup0/deepvision/blob/main/deepvision/layers/window_partitioning.py) and [WindowUnpartitioning](https://github.com/DavidLandup0/deepvision/blob/main/deepvision/layers/window_unpartitioning.py) from DeepVision
> 
> We have a great opportunity to clean up the official codebase with a few changes here. For instance, to stay consistent with `AddPositionalEmbedding`, we'll want to turn the [decomposed positional embeddings](https://github.com/DavidLandup0/deepvision/blob/main/deepvision/layers/decomposed_relative_positional_embedding.py), etc. into layers as well as opposed to the official repo's functions.

I agree with both your comments!

> We also have the opportunity to standardize some existing layers - such as `PatchingAndEmbedding` being the same as `ViTDetPatchingAndEmbedding` + `AddPositionalEmbedding`. We could do a general layer that consists of patching and a general layer consisting of embedding, and re-implement the former with the two.

ViT layers and the backbone are still TensorFlow Keras. I agree with your comment but I think it should be addressed when porting the layer to Keras Core. I will add a TODO comment so we can track this. Does that work?

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
tirthasheshpatel: Done.

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
tirthasheshpatel: Done.

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
tirthasheshpatel: Done.

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
tirthasheshpatel: Done.

keras_cv/models/segmentation/segment_anything/sam_mask_decoder.py
tirthasheshpatel: Done.

keras_cv/models/__init__.py
tirthasheshpatel: Done.

keras_cv/models/__init__.py
tirthasheshpatel: Done.

keras_cv/layers/detectron2_layers.py
tirthasheshpatel: Done.

keras_cv/layers/detectron2_layers.py
tirthasheshpatel: Done.

keras_cv/layers/detectron2_layers.py
```diff
@@ -0,0 +1,477 @@
...
+        return input_shape
+
+    def call(self, x):
+        shortcut = x
+        x = self.layer_norm1(x)
+        # Window Partition
+        if self.window_size > 0:
+            H, W = x.shape[1], x.shape[2]
+
+            x, HW_padded = window_partition(x, self.window_size)
```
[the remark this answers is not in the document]
tirthasheshpatel: Done. Instead of creating two classes, one for partitioning and one for unpartitioning, I handled both in a single class. Let me know if that looks good.

keras_cv/layers/detectron2_layers.py
tirthasheshpatel: Done.

keras_cv/layers/detectron2_layers.py
```diff
@@ -0,0 +1,477 @@
...
+# limitations under the License.
+
+from keras_cv.api_export import keras_cv_export
+from keras_cv.backend import keras
+from keras_cv.backend import ops
+from keras_cv.models.segmentation.segment_anything.sam_layers import MLP
+
+
+@keras_cv_export("keras_cv.layers.MultiHeadAttentionWithRelativePE")
+class MultiHeadAttentionWithRelativePE(keras.layers.Layer):
```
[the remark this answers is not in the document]
tirthasheshpatel: Done.

keras_cv/layers/__init__.py
tirthasheshpatel: Done.

tirthasheshpatel: Done.

keras_cv/models/backbones/detectron2/detectron2_backbone.py
tirthasheshpatel: Hi @DavidLandup0, thanks for pointing out the paper! I didn't know the authors also proposed a FPN! I did look into the paper but don't know how the pyramid-level inputs would fit in the backbone here. Given this PR has already blown up a bit, I'd prefer to do this as a follow-up. Maybe you can take it up if you have time :)

keras_cv/models/backbones/detectron2/detectron2_backbone.py
tirthasheshpatel: Done.

keras_cv/models/segmentation/segment_anything/sam_layers.py
tirthasheshpatel: Done. I have removed the export.

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
tirthasheshpatel: Is this useful elsewhere? Otherwise, it'd be easier to leave it as a private method.

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
tirthasheshpatel: Removed it and made it a part of the call method itself to make the inputs and output well-defined for the Task model.

keras_cv/models/segmentation/segment_anything/sam_prompt_encoder.py
tirthasheshpatel: Removed, this code wasn't XLA compatible. I have rewritten this using `ops.cond` and added a vectorized test for it in `keras_cv/models/segmentation/segment_anything/sam_test.py`.

tirthasheshpatel:
> 1. I ported the weights to the Keras Core model and am able to load in any backend but loading weights is broken between Keras Core and TensorFlow Keras (xref

An update on this: the legacy weights `*.h5` load in both TF Keras and Keras Core (all backends). I think we can just use that until the saving fixes are available in TF Keras.

keras_cv/layers/serializable_sequential.py
tirthasheshpatel: I ended up removing it since the legacy weights load in all backends in Keras Core and also in TF Keras. I think until some saving issues are addressed, we should just use the legacy weights. Let me know what you both think!

keras_cv/models/backbones/detectron2/detectron2_backbone.py
```diff
@@ -0,0 +1,215 @@
...
+from keras_cv.models import utils
+from keras_cv.models.backbones.backbone import Backbone
+from keras_cv.models.backbones.detectron2.detectron2_backbone_presets import (
+    backbone_presets,
+)
+from keras_cv.utils.python_utils import classproperty
+
+
+@keras_cv_export("keras_cv.models.ViTDetBackbone")
+class ViTDetBackbone(Backbone):
```
ianstenbit, line 31: Sorry if we've already been over this -- why is the name of the backbone different from the name of the file that it's in?

Should we rename the files `vit_det_backbone`?

keras_cv/models/backbones/detectron2/detectron2_backbone.py
ianstenbit: Doing this as a follow-up sgtm

keras_cv/models/segmentation/segment_anything/sam.py
```diff
@@ -0,0 +1,250 @@
...
+    SAMMaskDecoder,
+)
+from keras_cv.models.segmentation.segment_anything.sam_prompt_encoder import (
+    SAMPromptEncoder,
+)
+from keras_cv.models.task import Task
+
+
+@keras_cv_export("keras_cv.models.SegmentAnythingModel")
+class SegmentAnythingModel(Task):
```
ianstenbit, line 30: Let's add a preset to load up the fully-pretrained version of SAM.

ianstenbit (commented):
The only thing left that I see is adding a preset for the pre-trained version of SAM.

Thanks for your great work!

keras_cv/models/backbones/detectron2/detectron2_backbone.py
tirthasheshpatel: > why is the name of the backbone different from the name of the file that it's in?

I think `ViTDetBackbone` is more descriptive than `Detectron2Backbone`. I called the namespace `detectron2` since that's where the ViT model comes from.

> Should we rename the files `vit_det_backbone`?

I don't mind doing that. Will push this in the next batch of commits.

tirthasheshpatel: Done.

keras_cv/models/segmentation/segment_anything/sam.py
tirthasheshpatel: Done.

keras_cv/models/segmentation/segment_anything/sam_test.py
```diff
@@ -0,0 +1,377 @@
...
+            {
+                "image_embeddings": features,
+                "image_pe": outputs_ex["dense_positional_embeddings"],
+                "sparse_prompt_embeddings": outputs_ex["sparse_embeddings"],
+                "dense_prompt_embeddings": outputs_ex["dense_embeddings"],
+            },
+        )
+        masks_ex, iou_pred_ex = outputs_ex["masks"], outputs_ex["iou_pred"]
+
+        self.assertAllClose(masks, masks_ex, atol=1e-5)
```
ianstenbit, line 275: Looks like for the TF + JAX backends this tolerance needs to be more lie 3e-5

ianstenbit (approved):
This is awesome -- thanks Tirth!

Just one little fix to make GCBRun happy

keras_cv/models/segmentation/segment_anything/sam_test.py
```diff
@@ -0,0 +1,377 @@
...
+        self.assertAllClose(iou_pred, iou_pred_ex, atol=1e-5)
```
tirthasheshpatel, line 276:
```suggestion
        self.assertAllClose(masks, masks_ex, atol=5e-5)
        self.assertAllClose(iou_pred, iou_pred_ex, atol=5e-5)
```

keras_cv/models/segmentation/segment_anything/sam_test.py
tirthasheshpatel: Done, thanks!

tirthasheshpatel:
Thanks, @DavidLandup0 @ianstenbit for your reviews! This was fun to work on. Excited to have this in KerasCV!

The next steps are to add some guides to use and train the model. It would also be nice to have some benchmarks. On it now! But I will also create a tracking issue in case the community wants to take over some of these tasks.

ianstenbit:
> Thanks, @DavidLandup0 @ianstenbit for your reviews! This was fun to work on. Excited to have this in KerasCV!
> 
> The next steps are to add some guides to use and train the model. It would also be nice to have some benchmarks. On it now! But I will also create a tracking issue in case the community wants to take over some of these tasks.

Thank you Tirth for your outstanding work on this -- we really appreciate it!

I think our long-term goal should be to add support for text prompts. There are some community projects out there which demonstrate the feasibility of this, and I think it would be a great step for us.

But I 100% agree that some guides and training are the right place to start!

PR #1988, opened 2023-07-28
Add support of segmentation mask in Augmix layer

cosmo3769:
# What does this PR do?

Related: #1992 

Add support of segmentation mask in Augmix layer. Here is the [colab link](https://colab.research.google.com/drive/1pHrD2bJqc-tGTdwVWkcHZKrIoppzDgA8?usp=sharing) for demo. 

Output generated by demo:

Screenshot 2023-07-31 at 8 08 50 PM

## Who can review?

@ianstenbit @jbischof

### DIFF
--- a/examples/layers/preprocessing/segmentation/demo_utils.py
+++ b/examples/layers/preprocessing/segmentation/demo_utils.py
@@ -28,6 +28,7 @@ std = tf.constant([0.229, 0.224, 0.225])
 def normalize(input_image, input_mask):
     input_image = tf.image.convert_image_dtype(input_image, tf.float32)
     input_image = (input_image - mean) / tf.maximum(std, backend.epsilon())
+    input_image = input_image / 255
     input_mask -= 1
     return input_image, input_mask
 
--- a/keras_cv/layers/preprocessing/aug_mix.py
+++ b/keras_cv/layers/preprocessing/aug_mix.py
@@ -306,6 +306,55 @@ class AugMix(BaseImageAugmentationLayer):
         )
         return augmented
 
+    def _apply_segmentation_masks_op(self, segmentation_masks, op_index):
+        augmented = segmentation_masks
+        augmented = tf.cond(
+            op_index == tf.constant([0], dtype=tf.int32),
+            lambda: self._auto_contrast(augmented),
+            lambda: augmented,
+        )
+        augmented = tf.cond(
+            op_index == tf.constant([1], dtype=tf.int32),
+            lambda: self._equalize(augmented),
+            lambda: augmented,
+        )
+        augmented = tf.cond(
+            op_index == tf.constant([2], dtype=tf.int32),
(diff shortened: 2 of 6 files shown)


cosmo3769:
I am getting this lint error here, though this error is not picked up in my local:

```
./keras_cv/datasets/waymo/load.py:80:8: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
```

Similar error I am getting in this #1991. 
How to resolve this? 🤔

ianstenbit:
Lint error is unrelated and will be fixed by #1993

ianstenbit (changes requested):
Thanks for the PR!

ariG23498:
Hey @cosmo3769 this PR should not auto-close #1992 

Could you please edit that part out?

ianstenbit:
> Hey @cosmo3769 this PR should not auto-close #1992
> 
> Could you please edit that part out?

I've updated the description

keras_cv/layers/preprocessing/aug_mix.py
```diff
@@ -328,6 +328,36 @@ def augment_image(self, image, transformation=None, **kwargs):
     def augment_label(self, label, transformation=None, **kwargs):
         return label
 
+    def augment_segmentation_mask(
+        self, segmentation_masks, transformation=None, **kwargs
+    ):
+        chain_mixing_weights = self._sample_from_dirichlet(
```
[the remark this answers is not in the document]
cosmo3769: Looking into it.

ianstenbit:
Awesome! This looks better now -- I tinkered with the demo and the behavior seems correct

examples/layers/preprocessing/segmentation/demo_utils.py
```diff
@@ -28,6 +28,7 @@
 def normalize(input_image, input_mask):
     input_image = tf.image.convert_image_dtype(input_image, tf.float32)
     input_image = (input_image - mean) / tf.maximum(std, backend.epsilon())
+    input_image = input_image / 255
```
ianstenbit, line 31: Does this break other demos?

ianstenbit: Actually this seems incorrect -- we're already scaling by mean and stddev so this shouldn't be done

cosmo3769: No, I checked other demos as well.

ianstenbit (approved):
Looks good except for the change with demo_utils. Thank you!

examples/layers/preprocessing/segmentation/demo_utils.py
cosmo3769: In this, the value range of image is greater than 255. As we are setting Augmix value range to be [0, 255], so the demo does not work without making this change. Any better alternative? 🤔

PR #1995, opened 2023-07-31
Use Namex for API generation

ianstenbit:
This is long overdue!

Some notes:
- I am currently _not_ exporting any legacy models
- I currently _am_ exporting the 3D OD model even though it's tf.keras only for now.

I didn't change import paths for anything with one exception:
`keras_cv.models.retinanet.RetinaNetLabelEncoder` -> `keras_cv.models.retinanet.LabelEncoder` (and same for YOLOV8), as this felt redundant. Perhaps instead we should export them as `keras_cv.layers.RetinaNetLabelEncoder`, but that's a bit inconsistent because we have some model-specific things like the DetectionHead for RetinaNet (which we could, of course, export as `keras_cv.layers.RetinaNetDetectionHead` if we wanted.

One thing that I wasn't certain about is whether we should add a `formats` prefix to box formats. Currently they're exported as
`keras_cv.bounding_box.XYXY` but maybe `keras_cv.bounding_box.formats.XYXY` makes more sense? Once nice thing about namex is that we can do both!


ianstenbit:
Lint issue unrelated and will be fixed by #1993

jbischof (approved):
Thanks!

mattdangerw (approved):
Looks good to me!

One thing I have found helpful for testing this out is to run `pip_build.py --install` and use the resulting package to rebuild the keras.io docs for CV. Helps catch silly typos.

PR #2003, opened 2023-08-01
Port CenterPillar to Keras Core

ianstenbit:
TODOs (all done now -- sending for review)

- [x] Avoid `_core` versions of voxel_utils
- [x] Figure out where to put `segment_max`
- [x] Use `keras_core.backend.name_scope`
- [x] Reorder components in `center_pillar.py`


keras_cv/models/object_detection_3d/center_pillar.py
```diff
@@ -11,22 +11,16 @@
...
-from keras_cv.models.object_detection_3d.center_pillar_backbone_presets import (
-    backbone_presets,
-)
-from keras_cv.models.task import Task
-from keras_cv.utils.python_utils import classproperty
 
 
-@keras_cv_export("keras_cv.models.MultiHeadCenterPillar")
-class MultiHeadCenterPillar(Task):
+class MultiHeadCenterPillar(keras.Model):
```
ianstenbit, line 23: This should still subclass `Task` -- this was a mistake

keras_cv/models/object_detection_3d/center_pillar.py
```diff
@@ -56,31 +51,31 @@ def __init__(
...
+        super().__init__(**kwargs)
         self._voxelization_layer = voxel_net
-        self._backbone = backbone
+        self._unet_layer = backbone
         self._multiclass_head = multiclass_head
         self._prediction_decoder = prediction_decoder
         self._head_names = self._multiclass_head._head_names
+        self.built = True
+
+    def call(self, input_dict, training=None):
```
ianstenbit, line 62: This should also still be a functional subclass model and we should drop the `call`

keras_cv/layers/object_detection_3d/centernet_label_encoder.py
jbischof, line 1: This file still has a LOT of `tensorflow` code, is that intentional?

keras_cv/layers/object_detection_3d/voxel_utils.py
```diff
@@ -118,20 +99,15 @@ def compute_voxel_origin(
...
     return voxel_origin
 
 
-@keras_cv_export(
-    "keras_cv.object_detection_3d.voxel_utils.point_to_voxel_coord"
-)
-def point_to_voxel_coord(
-    point_xyz: tf.Tensor, voxel_size: Sequence[float], dtype=tf.int32
-) -> tf.Tensor:
+def point_to_voxel_coord(point_xyz, voxel_size, dtype=tf.int32):
```
jbischof, line 110: This function also has a lot of TF

jbischof (commented):
@ianstenbit I thought this model wasn't possible to port due to custom tf ops and I obviously see a lot of TF code remaining. Is the idea that this is a TensorFlow-only model that works with Keras Core? If so, is this something we can flag for users?

keras_cv/layers/object_detection_3d/centernet_label_encoder.py
ianstenbit: Yep! This lives outside of the model and runs in `tf.data`

keras_cv/layers/object_detection_3d/voxel_utils.py
ianstenbit: (Same here -- this is for `tf.data`)

ianstenbit:
> @ianstenbit I thought this model wasn't possible to port due to custom tf ops and I obviously see a lot of TF code remaining. Is the idea that this is a TensorFlow-only model that works with Keras Core? If so, is this something we can flag for users?

This does work with other backends, but requires data encoding in a tf.data pipeline, so that code is all still written in TF.

keras_cv/backend/__init__.py
```diff
@@ -42,6 +42,8 @@
 
 if multi_backend():
     import keras_core as keras
+
+    keras.backend.name_scope = keras.src.backend.name_scope
```
mattdangerw, line 46: this looks like it is exported as `keras.name_scope`, can we just use it like that?

mattdangerw: https://github.com/keras-team/keras-core/blob/b35cdfb18d7b49ff40f294943be4fefb10f0c85e/keras_core/backend/exports.py#L27

keras_cv/layers/object_detection_3d/heatmap_decoder.py
```diff
@@ -138,65 +141,67 @@ def __init__(
...
-        _, top_index = tf.math.top_k(heatmap, k=self.max_num_box)
+        heatmap = ops.reshape(heatmap, [b, h * w])
+        _, top_index = ops.top_k(heatmap, k=self.max_num_box)
 
         # [B, H, W, ?]
         box_prediction = prediction[:, :, :, 2:]
-        f = box_prediction.get_shape().as_list()[-1]
-        box_prediction = tf.reshape(box_prediction, [b, h * w, f])
-        heatmap = tf.reshape(heatmap, [b, h * w])
+        f = list(box_prediction.shape)[-1]
```
mattdangerw, line 167: nit: I don't think that the list cast does anything, you can index into a shape on all backends.

keras_cv/layers/object_detection_3d/voxel_utils.py
```diff
@@ -52,30 +46,20 @@ def compute_feature_map_ref_xyz(
...
-    voxel_origin = tf.cast(
-        compute_voxel_origin(spatial_size, voxel_size),
-        dtype=global_xyz.dtype,
-    )
+    voxel_origin = (compute_voxel_origin(spatial_size, voxel_size),)
     # [H, W, Z, 3]
     voxel_coord = voxel_coord + voxel_origin
     # [H, W, Z, 3]
-    ref = voxel_coord_to_point(voxel_coord, voxel_size, dtype=global_xyz.dtype)
+    ref = ops.cast(voxel_coord * np.array(voxel_size), global_xyz.dtype)
```
mattdangerw, line 54: why not ops.array here?

keras_cv/layers/object_detection_3d/voxel_utils.py
```diff
@@ -118,20 +99,15 @@ def compute_voxel_origin(
...
     voxel_origin = spatial_size[::2]
-    voxel_origin = tf.constant(
-        [o / v for o, v in zip(voxel_origin, voxel_size)], dtype=tf.float32
+    voxel_origin = np.array(
+        [o / v for o, v in zip(voxel_origin, voxel_size)], "float32"
     )
-    voxel_origin = tf.math.round(voxel_origin)
-    voxel_origin = tf.cast(voxel_origin, dtype=tf.int32)
+    voxel_origin = np.round(voxel_origin)
+    # voxel_origin = np.cast(voxel_origin, dtype="int32")
```
mattdangerw, line 106: why did this get commented out? should we delete forever, or still a todo here?

keras_cv/layers/object_detection_3d/voxel_utils.py
```diff
@@ -258,8 +213,8 @@ def combined_static_and_dynamic_shape(
     Returns:
       A list of size tensor.shape.ndims containing integers or a scalar tensor.
     """
-    static_tensor_shape = tensor.shape.as_list()
-    dynamic_tensor_shape = shape_int_compatible(tensor)
+    static_tensor_shape = list(tensor.shape)
```
mattdangerw, line 216: potentially this is already what `ops.shape` doing for the tf backend now? not totally sure though

keras_cv/layers/object_detection_3d/voxelization.py
```diff
@@ -136,28 +128,28 @@ def call(
 
         # [B, N, dim]
         # convert point voxel to positive voxel index
-        point_voxel_xyz = (
-            point_voxel_xyz_int - voxel_origin[tf.newaxis, tf.newaxis, :]
+        point_voxel_xyz = point_voxel_xyz_int - ops.cast(
+            ops.expand_dims(ops.expand_dims(voxel_origin, axis=0), axis=0),
```
mattdangerw, line 132: nit, I think `expand_dims(x, axis=(0, 1))` should work and maybe be a bit more readable.

keras_cv/layers/object_detection_3d/voxelization_test.py
```diff
@@ -75,15 +65,17 @@ def test_voxelization_output_shape_with_z(self):
...
 
     def test_voxelization_numerical(self):
-        point_net = keras.layers.Lambda(lambda x: x)
         layer = DynamicVoxelization(
-            point_net=point_net,
             voxel_size=[1.0, 1.0, 10.0],
             spatial_size=[-5, 5, -5, 5, -2, 2],
         )
+        # Make the point net a no-op to allow us to verify the voxelization.
+        layer.point_net_dense = keras.layers.Lambda(lambda x: x)
```
mattdangerw, line 76: Isn't there a `keras.layers.Indentity` for this?

mattdangerw (approved):
Looks good! Feel free to merge whenever, just left some random low level comments for things that popped out.

keras_cv/backend/__init__.py
ianstenbit: Oh awesome! Yes we can -- this wasn't available when I wrote this PR.

keras_cv/layers/object_detection_3d/voxel_utils.py
ianstenbit: Ahh yeah we can get rid of this

keras_cv/layers/object_detection_3d/voxel_utils.py
ianstenbit: Ahh yes, since your change to Keras Core I think this method is really no longer useful. I'm removing it.

keras_cv/layers/object_detection_3d/voxelization.py
ianstenbit: Done

keras_cv/layers/object_detection_3d/voxelization_test.py
ianstenbit: TIL -- thanks!

Issue #1992, opened 2023-07-31
Master Issue to track Preprocessing Layer Support for BBox and Segmentation Masks


ariG23498:
Segmentation Masks:
- [x] AugMix: https://github.com/keras-team/keras-cv/pull/1988
- [x] FourierMix: https://github.com/keras-team/keras-cv/pull/1991
- [ ] RandAugment
- [x] RandomCrop: https://github.com/keras-team/keras-cv/pull/2027
- [x] RandomCutout: https://github.com/keras-team/keras-cv/pull/2004
- [x] RandomShear: https://github.com/keras-team/keras-cv/pull/2021
- [x] RandomTranslation: https://github.com/keras-team/keras-cv/pull/2024
- [x] RandomZoom - https://github.com/keras-team/keras-cv/pull/2010
- [ ] RandomAspectRatio

BBox:
- [ ] CutMix
- [ ] FourierMix
- [x] RandAugment: https://github.com/keras-team/keras-cv/issues/728
- [x] RandomCutout: https://github.com/keras-team/keras-cv/issues/730
- [ ] RandomZoom - @ariG23498 

The list directly comes from [this README](https://github.com/keras-team/keras-cv/tree/master/keras_cv/layers/preprocessing).

CC: @ianstenbit

innat:
Partially similar. 
https://github.com/keras-team/keras-cv/discussions/93

ariG23498:
Hi @cosmo3769 -- I have noticed you taking up on a lot of segmentation map layers 🤗 . Thanks for that. Could you first claim the layers in this issue? -- That would help us track the progress better.

cosmo3769:
@ariG23498 Surely. I would tag this issue first before opening a new PR.

cosmo3769:
Hi @ariG23498, I am working on RandomShear layer.

cosmo3769:
@ariG23498 Could you please add RandomAspectRatio in the 'Segmentation Mask' section in the description above. It's missing here.

IMvision12:
@ariG23498 Taking up RandomTranslation and RandomCrop

Zekrom-7780:
@ariG23498 can I take Cutmix and FourierMix in BBox?

ariG23498:
> @ariG23498 can I take CutMix and FourierMix in BBox?

Hey @Zekrom-7780 please feel free to take them up! As soon as you have made a PR link this issue and I will update the description accordingly. Thanks!

[...]

PR #2004, opened 2023-08-02
Add support of segmentation mask in RandomCutout

cosmo3769:
# What does this PR do?

Related: #1992 

Add support of segmentation mask in Augmix layer. Here is the [colab link](https://colab.research.google.com/drive/169KKaG4yLUT842Vzj9YDBYFb8rKYt0_b?usp=sharing) for demo.

Output generated by demo:

Screenshot 2023-08-02 at 7 22 48 AM

## Who can review?

@ianstenbit @jbischof

### DIFF
--- a/examples/layers/preprocessing/segmentation/random_cutout_demo.py
+++ b/examples/layers/preprocessing/segmentation/random_cutout_demo.py
@@ -1,0 +1,34 @@
+# Copyright 2023 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""random_cutout_demo.py shows how to use the RandomCutout preprocessing layer.
+
+Uses the oxford iiit pet_dataset.  In this script the pets
+are loaded, then are passed through the preprocessing layers.
+Finally, they are shown using matplotlib.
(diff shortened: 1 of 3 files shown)


ianstenbit:
Thanks for the PR!

I think we need to think a bit about how to handle segmentation masks in this case.
For example, in GridMask, we actually keep the full segmentation mask unedited as the ground-truth label. It may be that this is also the best approach here.

My concern with zeroing-out the mask is that we're effectively telling the model that this is neither foreground class nor background class. Maybe that's okay, but I don't have a great intuition for that.

It could be ideal to instead set the masked-out area to have the background class, but that introduces some API awkwardness as the caller needs to specify what the background class is.

My intuition is that the best approach is probably to leave the seg mask unmodified, but it's hard to say for sure.

I'd propose that we test both options with my (WIP) [KerasCV segmentation guide](https://colab.research.google.com/gist/ianstenbit/2f1939c0a4f4ed0c0cdb4909694f6b77/kerascv-semantic-segmentation-guide.ipynb) and see which one yields the best results

cosmo3769:
Hey @ianstenbit, I have tested it and I find out that it works best when we don't apply random cutout to mask. Here is the [colab link](https://colab.research.google.com/drive/1LWAJv87TOe2Pjzy618hgBjqADIeh0MnI?usp=sharing). 

* With RandomCutout applied to mask - 80.75% test accuracy.
* With RandomCutout **not** applied to mask - 83.89% test accuracy.

ianstenbit:
> Hey @ianstenbit, I have tested it and I find out that it works best when we don't apply random cutout to mask. Here is the [colab link](https://colab.research.google.com/drive/1LWAJv87TOe2Pjzy618hgBjqADIeh0MnI?usp=sharing).
> 
> * With RandomCutout applied to mask - 80.75% test accuracy.
> * With RandomCutout **not** applied to mask - 83.89% test accuracy.
> 
> I have updated the code to remove randomcutout from mask.

Awesome thank you for doing this test -- that's awesome!

ianstenbit (approved):
Thanks for the PR!

PR #2016, opened 2023-08-08
Add bug template

jbischof:
We currently only offer a "feature request" type. Users are generally creating bugs from the "other" category and only leaving a stack trace, which is very difficult to replicate.

Ideally almost all bugs would include a colab gist to speed replication time.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,38 @@
+---
+name: Bug
+about: File a bug/issue
+title: '[BUG] <title>'
+labels: bug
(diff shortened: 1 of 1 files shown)


sampathweb (approved):

mattdangerw (approved):
Looks good to me!

We probably have a separate class of bugs that are related to installing/using the library outside of colab. For these, python version, os version are helpful, but I'm not sure if it's even worth calling this out.

jbischof:
Sure @mattdangerw if that becomes an issue (ha!) we can use one of those fancy yaml templates. For now I will add a note about this to the colab request.

PR #2028, opened 2023-08-15
Use correct convention for static shape in point_cloud grouping

ianstenbit:
This also has the side effect of allowing for Numpy inputs instead of just TF tensor inputs to point_cloud `group_points_by_boxes`, which is more pleasant for tinkering.


sampathweb (approved):

sampathweb:
LGTM

PR #2037, opened 2023-08-23
Remove forward slashes from layer names for backbones

ianstenbit:
Evidently these result in "/" existing in Keras Variable names which is disallowed in Keras Core


mattdangerw (approved):
LGTM!

PR #2044, opened 2023-08-30
Re-write YOLOV8LabelEncoder from scratch

ianstenbit:
I have run a few training experiments and verified that numerics are not affected.

I am going to send an upstream fix for PyTorch so that I can get rid of the little indexing hack here with `ops.take_along_axis`, but for now this is working locally and I don't want to wait on a Keras Core release.

### DIFF
--- a/keras_cv/models/object_detection/yolo_v8/yolo_v8_label_encoder.py
+++ b/keras_cv/models/object_detection/yolo_v8/yolo_v8_label_encoder.py
@@ -79,6 +79,11 @@ class YOLOV8LabelEncoder(keras.layers.Layer):
     def assign(
         self, scores, decode_bboxes, anchors, gt_labels, gt_bboxes, gt_mask
     ):
+        """Assigns ground-truth boxes to anchors.
+
+        Uses the task-aligned assignment strategy for matching ground truth
+        and anchor boxes based on prediction scores and IoU.
+        """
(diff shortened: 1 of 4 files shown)


keras_cv/models/object_detection/yolo_v8/yolo_v8_label_encoder.py
```diff
@@ -133,24 +76,142 @@ def __init__(
         self.beta = beta
         self.epsilon = epsilon
 
+    def assign(
```
jbischof, line 79: Would it be possible to add a docstring or at least a more descriptive name to this function? It seems to be the workhorse.

jbischof: Also any thoughts on defining independently vs inline?

jbischof (approved):
Seems like a stylistic win for sure!

keras_cv/models/object_detection/yolo_v8/yolo_v8_label_encoder.py
ianstenbit: Yeah -- added a brief docstring. I think defining it independently is a bit nicer because it has to live in an ops.cond call anyway, and this way it's a bit separated from the guard-clauses in `call`

PR #2012, opened 2023-08-04
[WIP] Updating FasterRCNN to use Task API

ariG23498:
CC: @ianstenbit

PR #2048, opened 2023-09-01
Remove direct imports of tf.keras

ianstenbit:
When referencing tf.keras, we should going forward always import it via TF (because in the future this will reference Keras 3)

This leaves in one reference (in models/object_detection/predict_utils.py) because it's going to be removed in #2012 anyway, and it's a bit hairy to fix I think.


grasskin (approved):
LGTM! Thank you

jbischof (approved):
Awesome 🧹 !

PR #2051, opened 2023-09-05
adding deeplab_v3_plus_presets

divyashreepathihalli:


keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus_presets.py
```diff
@@ -0,0 +1,42 @@
...
+from keras_cv.models.backbones.resnet_v2 import resnet_v2_backbone_presets
+
+deeplab_v3_plus_presets = {
+    "deeplab_v3_plus_resnet50_pascalvoc": {
+        "metadata": {
+            "description": (
+                "DeeplabV3Plus with a ResNet50 v2 backbone. "
+                "Trained on PascalVOC 2012 Semantic segmentation task, which "
+                "consists of 20 classes and one background class. This model "
+                "achieves a final categorical accuracy of 89.34% on evaluation "
```
ianstenbit, line 24: What is the MIoU?

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus_test.py
```diff
@@ -74,6 +74,13 @@ def test_weights_change(self):
             self.assertNotAllEqual(w1, w2)
             self.assertFalse(ops.any(ops.isnan(w2)))
 
+    @pytest.mark.large
```
ianstenbit, line 77: Let's make this a forward-pass numerical verification as well (like [this test](https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/object_detection/yolo_v8/yolo_v8_detector_test.py#L238))

ianstenbit (changes requested):
Thank you Divya! Looks good, just a few little tweaks.

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus_presets.py
divyashreepathihalli: added.

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus_test.py
divyashreepathihalli: done

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus.py
```diff
@@ -229,16 +233,99 @@ def from_config(cls, config):
...
+        if preset in cls.backbone_presets:
+            backbone_cls = keras.saving.get_registered_object(
+                metadata["class_name"]
+            )
+            backbone = backbone_cls.from_preset(preset, load_weights)
+            return cls(backbone, **kwargs)
+
+        # Otherwise must be one of class presets
+        config = metadata["config"]
+        config["backbone"]["config"]["input_shape"] = input_shape
```
ianstenbit, line 313: Sorry if I'm behind on this conversation -- can't we just do this in `Task` instead of just for DeepLab?

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus_test.py
```diff
@@ -74,6 +74,18 @@ def test_weights_change(self):
...
+    def test_with_model_preset_forward_pass(self):
+        model = DeepLabV3Plus.from_preset(
+            "deeplab_v3_plus_resnet50_pascalvoc",
+            num_classes=21,
+            input_shape=[512, 512, 3],
+        )
+        image = np.ones((1, 512, 512, 3))
+        output = model(image)
+        expected_output = np.zeros((1, 512, 512, 3))
```
ianstenbit, line 86: Ahh I hadn't really considered that this would be the expected output, but of course this makes sense 😄 

Can you quickly verify that with randomly initialized weights this assertion would fail?

ianstenbit (commented):
Looks good! Just a quick question about the preset bit -- @jbischof curious about your take as well. (IMO we should make this `from_preset` change on the `Task` class instead of just for DeepLabV3Plus

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus_test.py
divyashreepathihalli: That would still be all zeros - which is background. I tried to test with a little block in the image like this - 
```
image = np.zeros((512, 512,3), dtype=np.int8)
image[256:300, 256:300] = 230
```
None of this would match the 20+1 classes (personal, animal, items, etc) the model was trained on.

ianstenbit: Interesting -- maybe in the long run we should have a test image we use here, but for now this is fine I think.

divyashreepathihalli: SG! Here is a test image I used for now to verify functionality
image

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus.py
divyashreepathihalli: we can! Just updated.

ianstenbit (approved):
LGTM -- @jbischof PTAL quickly to verify that you're cool with this change to `from_preset`

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus_test.py
```diff
@@ -74,6 +74,18 @@ def test_weights_change(self):
...
+        self.assert_all_close(output, expected_output)
```
ianstenbit, line 87: assertAllClose -- this failed on GCB

keras_cv/models/task.py
```diff
@@ -93,6 +94,8 @@ def from_preset(
             load_weights: Whether to load pre-trained weights into model.
                 Defaults to `None`, which follows whether the preset has
                 pretrained weights available.
+            input_shape : input shape that will be passed to backbone
+                initialization, Defaults to `None`.
```
jbischof, line 98: You might want to clarify that `None` will not be passed to the backbone.

jbischof (approved):
Thanks look amazing!

keras_cv/models/task.py
divyashreepathihalli: done!

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus_test.py
divyashreepathihalli: updated

keras_cv/models/task.py
jbischof: For maximum clarity, I might say

```python
"""
If `None`, the preset value will be used.
"""
```

keras_cv/models/task.py
```diff
@@ -93,6 +94,9 @@ def from_preset(
             load_weights: Whether to load pre-trained weights into model.
                 Defaults to `None`, which follows whether the preset has
                 pretrained weights available.
+            input_shape : input shape that will be passed to backbone
+                initialization, Defaults to `None`, which will not be passed to
```
jbischof, line 98: For maximum clarity, I might say

```python
"""
If `None`, the preset value will be used.
"""
```

keras_cv/models/segmentation/deeplab_v3_plus/deeplab_v3_plus_test.py
ianstenbit, line 86: Channels shape here should be 21 to match `num_classes` -- this test failed on GCB again. Once you make the change could you run the test locally to verify that it passes?

divyashreepathihalli: Thanks Ian, I just updated the test. Since my local setup is throwing random errors with loading the preset I haven't been able to run the test. Hopefully its all fixed now.

ianstenbit:
@divyashreepathihalli `test_with_model_preset_forward_pass` is failing due to the numerics not matching, which is more or less what I'd expect -- shouldn't it be close to all 1s for the background class and all 0s for the non-background class?

PR #2052, opened 2023-09-05
Update centernet_label_encoder.py

qlzh727:
# What does this PR do?
Remove the unnecessary tf.constant from the encorder.

This was identified and verified internally with a centerpillar model running with multibackend keras.

The tf.constant was raising an error when the input is already a tensor instance and without dtype casting.

## Before submitting
      Pull Request section?
      to it if that's the case.


ianstenbit (approved):
Thank you!

PR #2064, opened 2023-09-08
Fix some imports in KerasCV that are invalid with Namex

ianstenbit:
With namex, it doesn't work internally to `import keras_cv` and use that, instead we need to specify better paths. This is breaking the 0.6.2 release right now.


mattdangerw (approved):
lgtm!

PR #2079, opened 2023-09-18
Bump python version requirement to 3.9

mattdangerw:
Keras-core now requires at least 3.9, and tf 2.14 (out imminently) will require 3.9 as well.

We should follow suit for cv/nlp.

### DIFF
--- a/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
+++ b/keras_cv/layers/preprocessing/base_image_augmentation_layer.py
@@ -135,6 +135,7 @@ class BaseImageAugmentationLayer(base_class):
             seed=seed, force_generator=force_generator
         )
         super().__init__(**kwargs)
+        self.built = True
         self._convert_input_args = False
 
     @property
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,2 +1,2 @@
-FROM mcr.microsoft.com/vscode/devcontainers/python:3.8
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.9
 COPY setup.sh /setup.sh
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -15,10 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python 3.8
+    - name: Set up Python 3.9
       uses: actions/setup-python@v1
       with:
-        python-version: 3.8
+        python-version: 3.9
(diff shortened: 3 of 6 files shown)


jbischof (approved):

mattdangerw:
@ianstenbit we probably need to do some bumping of version in containers before things are all the way green?

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -135,6 +135,7 @@ def __init__(self, seed=None, **kwargs):
             seed=seed, force_generator=force_generator
         )
         super().__init__(**kwargs)
+        self.built = True
```
ianstenbit, line 138: Is this change related?

ianstenbit (commented):
Thanks! Just want to verify the relationship to the two smaller non-config changes

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
ianstenbit: Whoops nevermind I see there are build errors with the new Keras core version

mattdangerw: Yeah, we have CI failures without. Looks like keras-core has some stricter requirements now about unbuilt state.

https://github.com/keras-team/keras-cv/actions/runs/6227820288/job/16903274533

The CSPDarkNetBackboneTest failure was pretty straightforward, we were adding a layer that was never used (and so never built), probably a bad idea and easy to avoid.

This one is a little more interesting. The error was:

```
FAILED keras_cv/layers/preprocessing/random_augmentation_pipeline_test.py::RandomAugmentationPipelineTest::test_supports_empty_layers_argument - UserWarning: `build()` was called on layer 'random_augmentation_pipeline_12', however the layer does not have a `build()` method implemented and it looks like it has unbuilt state. This will cause the layer to be marked as built, despite not being actually built, which may cause failures down the line. Make sure to implement a proper `build()` method.
```

Currently, it seems safe to assume that no augmentation layer has a `build()`` method, and just set `self.built = True` on base class. If there ever was an augmentation method that wanted a `build()` (to set some sort of non-trainable variables that depend on input shape I guess?), then we might want each augmentation layer to either have a `build` method or set `self.built = True`.

ianstenbit:
I'm uploading the updated versions of the Docker containers to GCB now. They shouldn't be strictly required here since this code should work with the latest and second-latest versions of Keras Core, but it's good to verify

ianstenbit:
Docker images updated.

ianstenbit (approved):

PR #2083, opened 2023-09-22
[SAM] Fixed shape issue in example in docstring

cosmo3769:
# What does this PR do?

Fixed below given error:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-15-3e0d7f8c6706> in <cell line: 2>()
      1 no_box = np.empty((1, 0, 2, 2))
      2 padded_points = np.concatenate(
----> 3     [points, np.zeros(1, 1, 2)], axis=1
      4 )
      5 padded_labels = np.concatenate(

TypeError: Cannot interpret '1' as a data type
```

@ianstenbit @tirthasheshpatel


ianstenbit (approved):
Thanks for the fix!

PR #2084, opened 2023-09-22
Use "masks" and "boxes" as prompt keys for SAM

ianstenbit:
Even though these inputs only accept 1 input per image, we use plurals as our standard for input keys in KerasCV (take, for example, "images", which is plural even though there is only 1 input image per batch)

cc @tirthasheshpatel


mattdangerw (approved):
LGTM!

jbischof (approved):

PR #2085, opened 2023-09-22
Support partial prompting in SAM

ianstenbit:
This lets a user specify _only_ boxes, points, or masks (or some pair thereof) without having to pass sentinel values for the other inputs.

This PR also adds a few TODOs for the last cleanup tasks I have for SAM

### DIFF
--- a/keras_cv/models/segmentation/segment_anything/sam.py
+++ b/keras_cv/models/segmentation/segment_anything/sam.py
@@ -84,20 +84,14 @@ class SegmentAnythingModel(Task):
         decoder.
 
     These prompts can be mixed and matched but at least one of the prompts
-    must be present. To turn off a particular prompt, a zero size array of the
-    following shapes must be passed:
+    must be present. To turn off a particular prompt, simply exclude it from
+    the inputs to the model.
 
-    (1) For points prompts, the expected shape is `(batch, num_points, 2)`. If
-        no point prompt is desired, pass an input of shape `(batch, 0, 2)`.
-        The labels must have shape `(batch, 0)` in case of no point prompt.
-    (2) For box prompt, the expected shape is `(batch, 1, 2, 2)`. The second
-        dimension (`box.shape[1]`) represents whether a box prompt is present
-        or not. If no box prompt is present, an input of shape
-        `(batch, 0, 2, 2)` is expected.
-    (3) Similarly, mask prompts have shape `(batch, 1, H, W, 1)`. Here too,
-        the first dimension (`mask.shape[1]`) indicates the presence of a mask
-        prompt. To turn off mask prompts, an input of shape
-        `(batch, 0, H, W, 1)` must be passed.
+    # TODO(ianstenbit): Remove the need for the `1` axes, and fix the box shape.
+    (1) For points prompts, the expected shape is `(batch, num_points, 2)`.
+        The labels must have a corresponding shape of `(batch, num_points)`.
+    (2) For box prompt, the expected shape is `(batch, 1, 2, 2)`.
+    (3) Similarly, mask prompts have shape `(batch, 1, H, W, 1)`.
 
     For example, to pass in all the prompts, do:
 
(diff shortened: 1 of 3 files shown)


keras_cv/models/segmentation/segment_anything/sam.py
```diff
@@ -125,9 +119,8 @@ class SegmentAnythingModel(Task):
     (i.e. `masks[:, 1:, ...]`) are alternate predictions that can be used if
     they are desired over the first one.
 
-    Now, in case of only points and box prompts, do:
+    Now, in case of only points and box prompts, simply exclude the masks:
 
-    >>> no_input_mask = np.empty((1, 0, 256, 256, 1))
```
tirthasheshpatel, line 130: I think you forgot to remove the `no_mask_input` key from the input dictionary below.

ianstenbit: Good catch -- thanks!

tirthasheshpatel (commented):
The SAM Task model is written in a way that expects the inputs to have all the prompts. But, since this PR changes that, we would also need to change the task model to accept partial prompts.

The end-to-end model tests always pass all the prompts so this wasn't caught by the CI.

Maybe, just override the `call` and `predict` methods to add the required inputs and dispatch the call to the `super()` class's method. What do you think @ianstenbit? Do you think there are other methods that the users want to use?

ianstenbit:
> The SAM Task model is written in a way that expects the inputs to have all the prompts. But, since this PR changes that, we would also need to change the task model to accept partial prompts.
> 
> The end-to-end model tests always pass all the prompts so this wasn't caught by the CI.
> 
> Maybe, just override the `call` and `predict` methods to add the required inputs and dispatch the call to the `super()` class's method. What do you think @ianstenbit? Do you think there are other methods that the users want to use?

Thanks for calling this out -- I think this is a reasonable approach. I'm updating CI to cover this use case.

jbischof (approved):
Nice!

keras_cv/models/segmentation/segment_anything/sam.py
```diff
@@ -203,6 +195,15 @@ def __init__(self, *, backbone, prompt_encoder, mask_decoder, **kwargs):
         self.prompt_encoder = prompt_encoder
         self.mask_decoder = mask_decoder
 
+    # TODO(ianstenbit): Do something more elegant to handle empty prompts.
+    def predict_step(self, *args, **kwargs):
```
tirthasheshpatel, line 199: I don't think this needs to be done at each predict step. I was suggesting something like:

```suggestion
    def predict(self, inputs, ...):
        B = self._get_batch_size(inputs)
        points = inputs.get("points", ops.empty(B, 0, 2))
        ...
        inputs = {
            "points": points,
            ...
        }
        super().predict(inputs, ...)
```

Any reason why override the `predict_step` instead? I think it'd add some overhead at each step.

tirthasheshpatel: One more thing: the `call` method will also need to be updated since it also expects a full prompt dictionary.

ianstenbit: Your proposed solution won't work well when the inputs are not just a single dictionary (e.g. if they're a `tf.data.Dataset` with many batches), so I used `predict_step`

And yes, `call` still requires all the arguments. I think the only workaround for this is to no longer use a functional subclass model but instead a pure subclass model, which is probably what we should do in the long run. For now I was just trying to get the `predict` API working. Before 0.7 is cut we should resolve this.

tirthasheshpatel: > Your proposed solution won't work well when the inputs are not just a single dictionary (e.g. if they're a `tf.data.Dataset` with many batches), so I used `predict_step`

Ah, that makes sense. Thanks for clarifying!

> And yes, `call` still requires all the arguments. I think the only workaround for this is to no longer use a functional subclass model but instead a pure subclass model, which is probably what we should do in the long run. For now I was just trying to get the `predict` API working. Before 0.7 is cut we should resolve this.

Does Keras compile the model if the `call` method is overridden?

PR #2092, opened 2023-09-28
Doc updates to switch branding to Keras 3

ianstenbit:
Shamelessly copied from @mattdangerw's [PR](https://github.com/keras-team/keras-nlp/pull/1259) on KerasNLP

Keeping as a draft until that PR is merged.


README.md
```diff
@@ -48,25 +48,51 @@ pip to install directly from the master branch on github:
...
+
+> [!IMPORTANT]
+> Make sure to set the `KERAS_BACKEND` before import any Keras libraries, it
+> will be used to set up Keras when it is first imported.
+Until the Keras 3 release, KerasCV will use a preview of Keras 3 on PyPI named
+[keras-core](https://pypi.org/project/keras-core/).
+
+> [!IMPORTANT]
+> If you want to import Keras and guarantee that you're getting the same version
+> that KerasCV is using, just use `from keras_cv.backend import keras`. This
```
mattdangerw, line 83: I think we should not use this, as it is not available in our exported API, or would be only via `from keras_cv.src.backend import keras`.

mattdangerw (approved):
Looks good! Just the one comment we already discussed offline.

PR #2093, opened 2023-10-02
Export task models as both models.task.Model and models.Model

ianstenbit:
Right now we have a bit of inconsistency here, and this standardizes our task model exports.


jbischof:
Why is this needed? Should we include `ImageClassifier`?

ianstenbit:
> Why is this needed? Should we include `ImageClassifier`?

It's not strictly needed, but it's a nice quality of life change IMO which makes the namespacing a bit more accessible.

jbischof (approved):

PR #2101, opened 2023-10-16
Use keras nightly for CI

ianstenbit:
Keras nightly is now Keras3!


mattdangerw:
Here's the sibling PR for KerasNLP -> https://github.com/keras-team/keras-nlp/pull/1273/files

I am trying to consolidate our dependency specification into requirements files, which I think we can mostly do at this point. That way dev environment ~= test environment.

One silly thing we should probably do is update the names of the actions.
image

PR #2104, opened 2023-10-19
Update README.md author list

divyashreepathihalli:


README.md
```diff
@@ -260,7 +260,7 @@ Here is the BibTeX entry:
 ```bibtex
 @misc{wood2022kerascv,
```
ianstenbit, line 261: I think we should probably plan to change this to eventually point to the citation of the paper we're writing, and for now let's just make an authors/maintainers list and include you there instead.

That way, folks can also easily see who to reach out to with questions as well

divyashreepathihalli: That sounds good!

PR #2109, opened 2023-10-20
Use the proper title for example

Philmod:
Otherwise it won't be rendered correctly by the [`TFKerasDocumentationGenerator`](https://github.com/keras-team/keras-io/blob/fc340b9989cdf17fba44e66efa22758afad39b87/scripts/docstrings.py#L27-L28)

Screenshot 2023-10-20 at 10 13 12


ianstenbit (approved):
Thank you!

Issue #2102, opened 2023-10-17
 InvalidArgumentError occurred when I implemeted image generation using mixed float16


y-vectorfield:
### Current Behavior:
InvalidArgumentError occurred when I implemented image generation using mixed float16.
Environment: https://www.tensorflow.org/tutorials/generative/generate_images_with_stable_diffusion

```sh
---------------------------------------------------------------------------
InvalidArgumentError                      Traceback (most recent call last)
[<ipython-input-13-80d21f9296ec>](https://localhost:8080/#) in <cell line: 2>()
      1 # Warm up model to run graph tracing before benchmarking.
----> 2 model.text_to_image("warming up the model", batch_size=3)
      3 
      4 start = time.time()
      5 images = model.text_to_image(

3 frames
[/usr/local/lib/python3.10/dist-packages/keras_cv/src/models/stable_diffusion/stable_diffusion.py](https://localhost:8080/#) in text_to_image(self, prompt, negative_prompt, batch_size, num_steps, unconditional_guidance_scale, seed)
     81         encoded_text = self.encode_text(prompt)
     82 
---> 83         return self.generate_image(
     84             encoded_text,
     85             negative_prompt=negative_prompt,

[/usr/local/lib/python3.10/dist-packages/keras_cv/src/models/stable_diffusion/stable_diffusion.py](https://localhost:8080/#) in generate_image(self, encoded_text, negative_prompt, batch_size, num_steps, unconditional_guidance_scale, diffusion_noise, seed)
    236             )
    237             a_t, a_prev = alphas[index], alphas_prev[index]
--> 238             pred_x0 = (latent_prev - math.sqrt(1 - a_t) * latent) / math.sqrt(
    239                 a_t
    240             )

[/usr/local/lib/python3.10/dist-packages/tensorflow/python/util/traceback_utils.py](https://localhost:8080/#) in error_handler(*args, **kwargs)
    151     except Exception as e:
    152       filtered_tb = _process_traceback_frames(e.__traceback__)
--> 153       raise e.with_traceback(filtered_tb) from None
    154     finally:
    155       del filtered_tb

[/usr/local/lib/python3.10/dist-packages/tensorflow/python/framework/ops.py](https://localhost:8080/#) in raise_from_not_ok_status(e, name)
   5886 def raise_from_not_ok_status(e, name) -> NoReturn:
   5887   e.message += (" name: " + str(name if name is not None else ""))
[...]

PR #2124, opened 2023-11-01
Fix the issue (#2102): Add cast to float32 from float16 into Stable D…

y-vectorfield-fj:
# What does this PR do?

This PR fixed  Issue #2102 (InvalidArgumentError occurred when I implemeted image generation using mixed float16) .

Fixes # (issue)
#2102

# Who can review?
@ianstenbit 
@jbischof 
Someone who know this problem.

### DIFF
--- a/keras_cv/models/stable_diffusion/stable_diffusion.py
+++ b/keras_cv/models/stable_diffusion/stable_diffusion.py
@@ -235,6 +235,7 @@ class StableDiffusionBase:
                 + unconditional_guidance_scale * (latent - unconditional_latent)
             )
             a_t, a_prev = alphas[index], alphas_prev[index]
+            latent = ops.cast(latent, latent_prev.dtype)
             pred_x0 = (latent_prev - math.sqrt(1 - a_t) * latent) / math.sqrt(
                 a_t
             )


keras_cv/models/stable_diffusion/stable_diffusion.py
```diff
@@ -235,6 +235,8 @@ def generate_image(
                 + unconditional_guidance_scale * (latent - unconditional_latent)
             )
             a_t, a_prev = alphas[index], alphas_prev[index]
+            if latent.dtype == "float16":
```
ianstenbit, line 238: I think we should be able to unconditionally cast latent to the dtype of latent_prev, e.g. `ops.cast(latent, latent_prev.dtype)`.

Is there a reason that won't work?

ianstenbit (commented):
Thanks for the PR!

keras_cv/models/stable_diffusion/stable_diffusion.py
y-vectorfield: Thank you very much for your comment.
I think you are right!
I will try your advice and modify the code.

y-vectorfield-fj: Thank you very much for your comment.
I think you are right!
I will try your advice and modify the code.

y-vectorfield-fj: @ianstenbit 
I tried your suggestion and modified the code!
Thank you!!
If there is no problem, please approve my modification.

PR #2129, opened 2023-11-06
Test against keras 3

divyashreepathihalli:


requirements.txt
```diff
@@ -1,14 +1,14 @@
 absl-py
 packaging
 pandas
-tensorflow
+tf-nightly==2.16.0.dev20231026  # Pin a working nightly until rc0.
```
sampathweb, line 4: With this change are we breaking compatibility with older TF?  Also, TF 2.16 is not going to be released until Feb'24.  Can we be compatible with TF 2.15 which will work with Keras 3 except for pip compatibility check?

sampathweb: Also the GitHub actions all use TF 2.13, so I think we need to think through the min TF Version we want to be compatible with.

divyashreepathihalli: @sampathweb I am trying to mirror keras_nlp - https://github.com/keras-team/keras-nlp/pull/1273/files

sampathweb: If we are doing this, we need to update GitHub actions as well and also agree that we are dropping support for TF < 2.16.  I don't think we are ready to do that yet.  So wanted to discuss this change further.

divyashreepathihalli: As per our discussion, I have changed the requirements to use stable version of TF. This PR will only change the names of the tests.

mattdangerw: Oops sorry probably miscommunication here! I had thought we had discussed having a test run that test against a stable version of tf, but *also* have multiple tests run that checks Keras 3 on all backends. Let's hit both the compat check and the new nightly check in our testing setup.

Here's how we are doing this on KerasNLP:
https://github.com/keras-team/keras-nlp/blob/master/.github/workflows/actions.yml#L30-L34
and
https://github.com/keras-team/keras-nlp/blob/master/.github/workflows/actions.yml#L66-L69

If we want to test keras-cv against an earlier version of tensorflow like 2.13 we can do that by updating the version pin in the former link. I don't think we can spare the compute to test every tf version, but this way we will span checking the oldest version of tf we would like to support, and the newest (tf and keras 2.16).

(also note the keras-nlp has a slightly different setup for requirements files. it would be great to consolidate at some point, but probably a separate pr/discussion)

mattdangerw: @sampathweb @divyashreepathihalli does that make sense to both of you? If not we can sync up!

sampathweb: Agree.  We should keep similar setup for `keras-cv` where we test stable version of TensorFlow 2.13 or 2.14 with Keras 2 setup and use pinned `tf-nightly` with Keras 3 for multibackend similar to how its done in `keras-nlp`.

divyashreepathihalli: updated requuirements and actions.file to tests againts tf 2.14 for keras 2 and nightly version of tf for keras 3

.github/workflows/actions.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           ${{ runner.os }}-pip-
     - name: Install dependencies
       run: |
-        pip install tensorflow>=2.13.0
+        pip install tensorflow==2.14
```
mattdangerw, line 36: Do we want to keep this a `2.13` for now? Also nit, but I think `~=2.13` is slightly better practice as it will pull the latest 2.13 release always.

.github/workflows/actions.yml
```diff
@@ -117,7 +117,7 @@ jobs:
...
```
mattdangerw, line 120: Same here, use `~=`

.github/workflows/actions.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           ${{ runner.os }}-pip-
     - name: Install dependencies
       run: |
-        pip install tensorflow>=2.13.0
+        pip install tf-nightly-cpu==2.16.0.dev20231103  # Pin a working nightly until rc0.
         pip install "jax[cpu]"
         pip install torch>=2.0.1+cpu
         pip install torchvision>=0.15.1
```
mattdangerw, line 76: We should be able to remove the keras-core line below this. Keras core will never be used in Keras 3 is installed.

mattdangerw (commented):
Just a couple comments!

requirements.txt
```diff
@@ -1,14 +1,14 @@
...
 tensorflow-datasets
 flake8
 regex
 isort
 black
 pytest
 pycocotools
-keras-core
 build
+keras-core>=0.1.6
```
sampathweb, line 13: Wouldn't `tf-nightly` already pull in `keras 3` via `keras-nightly`.  Do we need to install `keras-core`?

requirements.txt
```diff
@@ -1,14 +1,14 @@
 absl-py
 packaging
 pandas
-tensorflow
+tf-nightly-cpu==2.16.0.dev20231103  # Pin a working nightly until rc0.
```
sampathweb, line 4: Also, we should install `tf-nightly` without the `-cpu`.  If the user's machine has GPU, this might lead to installing a CPU only version.  For GitHub workflow, `tf-nightly` would be suffice.

.github/workflows/actions.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           ${{ runner.os }}-pip-
     - name: Install dependencies
       run: |
-        pip install tensorflow>=2.13.0
+        pip install tf-nightly-cpu==2.16.0.dev20231103  # Pin a working nightly until rc0.
```
sampathweb, line 73: We should remove the `-cpu` suffix.  Just `tf-nightly`

requirements.txt
divyashreepathihalli: This is needed if users are using keras 2 and set a backend. we will remove this once keras 3 is released.

requirements.txt
divyashreepathihalli: done

.github/workflows/actions.yml
divyashreepathihalli: done!

.github/workflows/actions.yml
divyashreepathihalli: NLP was using 2.14 and I wanted to keep the two consistent.

PR #2132, opened 2023-11-06
Improve keras 3 detection

divyashreepathihalli:
To-dos added for after this PR is merged
convert these layers to keras 3. These layers are disabled for keras 3 for now.
dropblock_2d: https://github.com/keras-team/keras-cv/issues/2135
BaseAugmentationLayer3D: https://github.com/keras-team/keras-cv/issues/2136
fix legacy keras.engine import in predict_utils.py
https://github.com/keras-team/keras-cv/issues/2134

These changes have been tested here 
https://colab.sandbox.google.com/drive/1pGSji4tto6Fu88LXR0z4p7ExJZWNajGK#scrollTo=9s7Q0WbIa6vB


keras_cv/backend/__init__.py
```diff
@@ -11,91 +11,28 @@
...
 def assert_tf_keras(src):
-    if multi_backend():
+    if config.multi_backend():
         raise NotImplementedError(
             f"KerasCV component {src} does not yet support Keras Core, and can "
             "only be used in `tf.keras`."
         )
 
 
 def supports_ragged():
```
mattdangerw, line 29: These next three fns seem like they could just live in `config` rather then proxying them here.

```
from keras_cv.backend import config

if config.multi_backend():
    ...

if config.supports_ragged():
   ...
```

Is fairly clear usage throughout the library.

keras_cv/backend/ops.py
```diff
@@ -11,11 +11,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from keras_cv.backend.config import detect_if_tensorflow_uses_keras_3
 from keras_cv.backend.config import multi_backend
 
-if multi_backend():
+if detect_if_tensorflow_uses_keras_3():
```
mattdangerw, line 17: I don't get how this could work. If tensorflow is using keras 3 (and keras 2 is properly uninstalled), won't `keras.preprocessing` not exist here?

keras_cv/backend/tf_ops.py
```diff
@@ -11,14 +11,17 @@
...
-    convert_to_numpy,
-)
-from keras_core.src.backend.tensorflow.core import *  # noqa: F403, F401
-from keras_core.src.backend.tensorflow.math import *  # noqa: F403, F401
-from keras_core.src.backend.tensorflow.nn import *  # noqa: F403, F401
-from keras_core.src.backend.tensorflow.numpy import *  # noqa: F403, F401
+from keras_cv.backend import config
+
+if config.multi_backend():
+    from keras_core.src.backend.tensorflow import *  # noqa: F403, F401
```
mattdangerw, line 17: If we running on Keras 3, we should make sure to import all these from keras 3 right? And not keras core? We should probably assert that things can run if you...

```
install this branch from source
pip uninstall -y tensorflow keras keras-core
pip install tf-nightly
```

Also what happens on the else of this if statement? Seems like some ops might not exists that are being relied on elsewhere.

keras_cv/models/object_detection/predict_utils.py
```diff
@@ -15,12 +15,13 @@
 import tensorflow as tf
 
 try:
-    from keras.src.engine.training import _minimum_control_deps
-    from keras.src.engine.training import reduce_per_replica
+    # To-do: these imports need to fixed - Issue 2134
```
mattdangerw, line 18: I would just toss the whole link on a new line for easy browsing
`# https://github.com/keras-team/keras-cv/issues/2134`

mattdangerw (changes requested):
Left a few comments. I am a little suspicious of the tf_ops file changes, but I may not be fully understand that yet.

keras_cv/backend/__init__.py
divyashreepathihalli: updated

keras_cv/backend/tf_ops.py
divyashreepathihalli: Thank you for the repro colab. I have added the test for this here - https://colab.research.google.com/drive/1pGSji4tto6Fu88LXR0z4p7ExJZWNajGK#scrollTo=ftRC6cQ63iSp&line=3&uniqifier=1
it was erroring out before but now it is loading correctly

keras_cv/models/object_detection/predict_utils.py
divyashreepathihalli: done

keras_cv/backend/ops.py
divyashreepathihalli: Updated

keras_cv/backend/__init__.py
mattdangerw: nit: still think this might be slightly clearer to expose in the `config.py` module, as your are checking the status of feature support, but not a big deal! fine either way

keras_cv/backend/ops.py
```diff
@@ -25,5 +28,5 @@
         from keras_core.src.utils.image_utils import (  # noqa: F403, F401
             smart_resize,
         )
-else:
+if not multi_backend():
     from keras_cv.backend.tf_ops import *  # noqa: F403, F401
```
mattdangerw, line 32: Why are we importing `tf_ops` twice? Once here and once from the toplevel `__init__.py`? Do we need both?

keras_cv/backend/random.py
```diff
@@ -15,6 +15,9 @@
...
 
 if multi_backend():
-    from keras_core.random import *  # noqa: F403, F401
+    try:
+        from keras.random import *  # noqa: F403, F401
+    except ImportError:
+        from keras_core.random import *  # noqa: F403, F401
 else:
-    from keras_core.src.backend.tensorflow.random import *  # noqa: F403, F401
+    from tensorflow.random import *  # noqa: F403, F401
```
mattdangerw, line 23: These seems dangerous, and not what we had before. `tensorflow.random` is not the same API surface as `keras.random` or `keras_core.random` (which are the same).

We should probably do something like this
https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/backend/random.py

That is, keras_core is the "compat shim" for the random API on Keras 2. On Keras 3 we just use `keras.random` directly.

keras_cv/layers/regularization/dropblock_2d.py
```diff
@@ -145,20 +153,29 @@ def __init__(
         seed=None,
         **kwargs,
     ):
-        super().__init__(seed=seed, **kwargs)
-        if not 0.0 <= rate <= 1.0:
+        # To-do: remove this once th elayer is ported to keras 3
+        # https://github.com/keras-team/keras-cv/issues/2136
+        if detect_if_tensorflow_uses_keras_3():
```
mattdangerw, line 158: If we want to follow the KerasNLP approach, we could just save the result of `detect_if_tensorflow_uses_keras_3` as a global in the config file, and expose a fn to check here.

`if backend.config.keras_3(): raise ValueError`

keras_cv/models/object_detection/retinanet/retinanet_test.py
```diff
@@ -251,7 +251,10 @@ def test_call_with_custom_label_encoder(self):
         model(ops.ones(shape=(2, 224, 224, 3)))
 
     def test_tf_dataset_data_generator(self):
-        if backend.multi_backend() and keras.backend.backend() != "tensorflow":
+        if (
```
mattdangerw, line 254: You already handle this case in the `backend()` function you added to `config.py` above. Just import `config` and use `if config.backend() != "tensorflow"` here.

Example:
https://github.com/keras-team/keras-nlp/blob/e7bbf2b6c565f23a5be28557b60427d02223d091/keras_nlp/models/generative_task.py#L73

keras_cv/backend/tf_ops.py
```diff
@@ -11,24 +11,39 @@
...
+        )
+        from keras_core.src.backend.tensorflow.core import *  # noqa: F403, F401
+        from keras_core.src.backend.tensorflow.math import *  # noqa: F403, F401
+        from keras_core.src.backend.tensorflow.nn import *  # noqa: F403, F401
+        from keras_core.src.backend.tensorflow.numpy import *  # noqa: F403, F401, E501
+
 
 # Some TF APIs where the numpy API doesn't support raggeds that we need
 from tensorflow import broadcast_to  # noqa: F403, F401
+from tensorflow import repeat  # noqa: F403, F401
```
mattdangerw, line 40: Why changing the order here? No big deal, but is it important? Looks like these were alpha sorted before.

keras_cv/backend/tf_ops.py
```diff
@@ -11,24 +11,39 @@
...
-from keras_core.src.backend.tensorflow import (  # noqa: F403, F401
-    convert_to_numpy,
-)
-from keras_core.src.backend.tensorflow.core import *  # noqa: F403, F401
-from keras_core.src.backend.tensorflow.math import *  # noqa: F403, F401
-from keras_core.src.backend.tensorflow.nn import *  # noqa: F403, F401
-from keras_core.src.backend.tensorflow.numpy import *  # noqa: F403, F401
+from keras_cv.backend import config
+
+if config.multi_backend():
```
mattdangerw, line 16: Do we need all the branching and try except?

Can we just do?

```
if detect_if_tensorflow_uses_keras_3():
    from keras.src.backend.tensorflow import *
    ...
else:
    from keras_core.src.backend.tensorflow import *
    ...
```

keras_cv/backend/keras.py
```diff
@@ -0,0 +1,49 @@
...
+        from tensorflow.keras import saving
+    from tensorflow.keras import utils
+
+    if not hasattr(saving, "deserialize_keras_object"):
+        saving.deserialize_keras_object = utils.deserialize_keras_object
+    if not hasattr(saving, "serialize_keras_object"):
+        saving.serialize_keras_object = utils.serialize_keras_object
+    if not hasattr(saving, "register_keras_serializable"):
+        saving.register_keras_serializable = utils.register_keras_serializable
+    # TF Keras doesn't have this rename.
```
mattdangerw, line 48: Ian had a shim for `get_registered_object` above. Do we need it? Is it used anywhere?

mattdangerw (changes requested):
Looks good! Only big question is with `random`. I don't think we should import `tensorflow.random` and `keras.random` as if they were interchangeable.

keras_cv/backend/__init__.py
divyashreepathihalli: done

keras_cv/backend/ops.py
divyashreepathihalli: I think this was added to have tf_ops to be available through `keras_cv.backend.ops.*` instead if `keras_cv.backend.tf_ops.*`

keras_cv/backend/random.py
divyashreepathihalli: done

keras_cv/layers/regularization/dropblock_2d.py
divyashreepathihalli: done

keras_cv/backend/tf_ops.py
divyashreepathihalli: I am not sure why this happened :').  Changed it back.

keras_cv/backend/tf_ops.py
divyashreepathihalli: done

keras_cv/backend/keras.py
divyashreepathihalli: So I had removed Ian's original code on this in init. I added back his code back now.

mattdangerw (approved):
Very nice work! This looks great!

PR #2139, opened 2023-11-10
Adds Nightly package for `keras-cv`

sampathweb:
Adds Nightly package for `keras-cv`

### DIFF
--- a/pip_build.py
+++ b/pip_build.py
@@ -28,6 +28,7 @@ python3 pip_build.py --install
 ```
 """
 import argparse
+import datetime
 import glob
 import os
 import pathlib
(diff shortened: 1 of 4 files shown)


divyashreepathihalli:
@sampathweb Thank you for the PR! LGTM but I would wait till our tests are passing with the new keras 3 compat changes before releasing nightlies.

sampathweb:
> @sampathweb Thank you for the PR! LGTM but I would wait till our tests are passing with the new keras 3 compat changes before releasing nightlies.

Sure.  We can wait to merge.

pip_build.py
```diff
@@ -49,14 +50,34 @@ def ignore_files(_, filenames):
     return [f for f in filenames if "test" in f]
 
 
-def build():
+def export_version_string(version, is_nightly=False):
```
mattdangerw, line 53: It looks like our version string in `__init__.py` is the current release version. Maybe we should update that to `0.7.0` so it is pointing to the next major version we will release?

mattdangerw (approved):
looks good to me thanks! just one comment

pip_build.py
sampathweb: Done

sampathweb:
I will merge this PR now.  Plan to fix the tests before nightly runs tonight.

PR #2165, opened 2023-11-18
Keras 3 compatibility updates

divyashreepathihalli:
Uses tf.random.Generator instead of keras.backend.RandomGenerator
removed legacy_make_seed  

Tests disabled - https://github.com/keras-team/keras-cv/issues/2167
- all 3d preprocessing tests - has legacy baserandomlayer
- rand_augment_test - need to investigate
- random_shear_test - need to investigate
- repeated_augmentation_test - need to investigate
- resizing_test - has ragged tensor tests and is failing on jax and torch

### DIFF
--- a/benchmarks/vectorized_mosaic.py
+++ b/benchmarks/vectorized_mosaic.py
@@ -17,6 +17,8 @@ import time
 import matplotlib.pyplot as plt
 import numpy as np
 import tensorflow as tf
+from tensorflow import keras
+
 from keras_cv import bounding_box
 from keras_cv.layers import Mosaic
 from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
@@ -24,10 +26,11 @@ from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
 )
 from keras_cv.layers.preprocessing.vectorized_base_image_augmentation_layer import (  # noqa: E501
     IMAGES,
+)
+from keras_cv.layers.preprocessing.vectorized_base_image_augmentation_layer import (
     LABELS,
 )
 from keras_cv.utils import preprocessing as preprocessing_utils
-from tensorflow import keras
 
 
 class OldMosaic(BaseImageAugmentationLayer):
--- a/benchmarks/vectorized_mosaic.py
+++ b/benchmarks/vectorized_mosaic.py
@@ -27,7 +27,7 @@ from keras_cv.layers.preprocessing.base_image_augmentation_layer import (
 from keras_cv.layers.preprocessing.vectorized_base_image_augmentation_layer import (  # noqa: E501
     IMAGES,
 )
-from keras_cv.layers.preprocessing.vectorized_base_image_augmentation_layer import (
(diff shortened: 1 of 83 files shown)


benchmarks/vectorized_mosaic.py
```diff
@@ -101,7 +101,9 @@ def _batch_augment(self, inputs):
             minval=0,
             maxval=batch_size,
             dtype=tf.int32,
-            seed=self._random_generator.make_legacy_seed(),
+            seed=self._random_generator.uniform(
```
sampathweb, line 104: Are we changing the functionality here when we use random uniform to generate seed?

keras_cv/layers/preprocessing/fourier_mix.py
```diff
@@ -54,10 +54,12 @@ def __init__(self, alpha=0.5, decay_power=3, seed=None, **kwargs):
 
     def _sample_from_beta(self, alpha, beta, shape):
         sample_alpha = tf.random.gamma(
-            shape, alpha=alpha, seed=self._random_generator.make_legacy_seed()
+            shape,
+            alpha=alpha,
```
sampathweb, line 58: If we remove the seed, are we expecting same results like before?

keras_cv/layers/preprocessing/random_color_jitter_test.py
```diff
@@ -22,9 +21,9 @@
 class RandomColorJitterTest(TestCase):
     # Test 1: Check input and output shape. It should match.
     def test_return_shapes(self):
-        batch_input = np.ones((2, 512, 512, 3))
-        non_square_batch_input = np.ones((2, 1024, 512, 3))
-        unbatch_input = np.ones((512, 512, 3))
+        batch_input = tf.ones((2, 512, 512, 3))
```
sampathweb, line 24: why do we need to change them to `tf.`?  I think `np.` should also work since we allow for numpy args with the variable set in base layer

sampathweb: Same in many tests

keras_cv/layers/preprocessing/random_crop_test.py
```diff
@@ -31,7 +32,7 @@ class RandomCropTest(TestCase):
         ("random_crop_full_width", 3, 8),
     )
     def test_output_shape(self, expected_height, expected_width):
-        np.random.seed(1337)
+        tf.random.set_seed(1337)
```
sampathweb, line 35: Same here.  Why do we need to use `tf` instead of `np.`?

divyashreepathihalli: done

keras_cv/layers/preprocessing/random_color_jitter_test.py
divyashreepathihalli: done

sampathweb:
Replaced by `keras-3-updates` PR

Issue #2167, opened 2023-11-20
fix failing tests in keras 3


divyashreepathihalli:
- [x] all 3d preprocessing tests - has legacy baserandomlayer
- [ ] rand_augment_test - need to investigate
- [ ] random_shear_test - need to investigate
- [ ] repeated_augmentation_test - need to investigate
- [ ] resizing_test - has ragged tensor tests and is failing on jax and torch
- [ ] SqueezeAndExcite2D - serialization test

sampathweb:
@divyashreepathihalli - Is this still an open issue?

divyashreepathihalli:
@sampathweb yes, we still need to fix these

sachinprasadhs:
Closing this as we have migrated to Keras Hub.

PR #2170, opened 2023-11-20
Keras 3 updates

divyashreepathihalli:
Uses tf.random.Generator instead of keras.backend.RandomGenerator
removed legacy_make_seed

Tests disabled - https://github.com/keras-team/keras-cv/issues/2167

all 3d preprocessing tests - has legacy baserandomlayer
rand_augment_test - need to investigate
random_shear_test - need to investigate
repeated_augmentation_test - need to investigate
resizing_test - has ragged tensor tests and is failing on jax and torch

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,105 +8,38 @@ permissions:
   contents: read
 
 jobs:
-  deploy-with-custom-ops:
-    # This job is currently skipped until we cut a release with custom ops.
-    if: false
-    name: Build and deploy release wheels with custom ops
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        # To switch on windows-2022/latest, please verify the bazel version:
-        # https://github.com/bazelbuild/bazel/issues/14232#issuecomment-1011247429
-        os: ['macos-12', 'windows-2019', 'ubuntu-18.04']
-        py-version: ['3.9', '3.10', '3.11']
-        tf-version: ['2.13.0']
-        use-macos-arm: [false]
-        include:
-          - os: 'macos-12'
-            tf-version: '2.13.0'
-            py-version: '3.9'
-            use-macos-arm: true
-          - os: 'macos-12'
-            tf-version: '2.13.0'
-            py-version: '3.10'
-            use-macos-arm: true
-      fail-fast: false
-    steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python
-      uses: actions/setup-python@v1
-      with:
-        python-version: ${{ matrix.py-version }}
-    - name: Get pip cache dir
-      id: pip-cache
-      run: |
-        python -m pip install --upgrade pip setuptools wheel auditwheel twine
-        echo "::set-output name=dir::$(pip cache dir)"
-    - name: pip cache
-      uses: actions/cache@v2
-      with:
-        path: ${{ steps.pip-cache.outputs.dir }}
-        key: ${{ runner.os }}-pip-${{ hashFiles('setup.py') }}
-        restore-keys: |
-          ${{ runner.os }}-pip-
-    - name: Install dependencies
-      run: |
-        python -m pip install tensorflow-cpu==${{ matrix.tf-version }}
-        python -m pip install -e ".[tests]" --progress-bar off --upgrade
-    - name: Configure Build Environment
-      run: |
-        python build_deps/configure.py
-    - name: Reinstall TensorFlow (MacOS ARM)
-      if: ${{ matrix.os == 'macos-12' && matrix.use-macos-arm}}
-      run: |
-        python -m pip uninstall -y tensorflow-cpu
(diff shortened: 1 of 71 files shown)


.github/workflows/release.yml
```diff
@@ -8,105 +8,38 @@ permissions:
...
+            pip install -r requirements.txt --progress-bar off
+      - name: Build a binary wheel and a source tarball
+        run: |
+          export BUILD_WITH_CUSTOM_OPS=false
+          python pip_build.py
+      - name: Publish distribution to PyPI
+        if: startsWith(github.ref, 'refs/tags')
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          password: ${{ secrets.PYPI_API_TOKEN }}
```
mattdangerw, line 45: newline end of file

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
```diff
@@ -130,11 +130,13 @@ def augment_image(self, image, transformation):
     """
 
     def __init__(self, seed=None, **kwargs):
-        force_generator = kwargs.pop("force_generator", False)
-        self._random_generator = keras_backend.RandomGenerator(
-            seed=seed, force_generator=force_generator
-        )
+        if seed:
```
mattdangerw, line 133: maybe better to do the explicit `if seed is not None` check here, avoid unexpected cases of people passing strange seed input and getting the global generator behavior

keras_cv/layers/preprocessing_3d/input_format_test.py
```diff
@@ -15,98 +15,105 @@
...
 from absl.testing import parameterized
 
+from keras_cv.backend.config import keras_3
 from keras_cv.layers import preprocessing_3d
 from keras_cv.layers.preprocessing_3d import base_augmentation_layer_3d
 from keras_cv.tests.test_case import TestCase
 
-POINT_CLOUDS = base_augmentation_layer_3d.POINT_CLOUDS
-BOUNDING_BOXES = base_augmentation_layer_3d.BOUNDING_BOXES
+if not keras_3():
```
mattdangerw, line 23: is there a reason we can't just do this with the pytest annotation like elsewhere?

keras_cv/layers/serialization_test.py
```diff
@@ -378,11 +379,30 @@ class SerializationTest(TestCase):
...
+                cv_layers.GlobalRandomScaling,
+                cv_layers.GlobalRandomTranslation,
+                cv_layers.GroupPointsByBoundingBoxes,
+                cv_layers.RandomCopyPaste,
+                cv_layers.RandomDropBox,
+                cv_layers.SwapBackground,
+                cv_layers.SqueezeAndExcite2D,  # TODO: Fails in Keras 3
+            ]
+            if layer_cls in skip_layers:
+                return
```
mattdangerw, line 400: `self.skipTest("Not supported on Keras 3")` (this will show up in the test logs and make it obvious what is happening

requirements.txt
```diff
@@ -1,14 +1,13 @@
...
-isort
-black
-pytest
-pycocotools
-keras-core
-build
-namex
+# Tensorflow.
+tf-nightly-cpu==2.16.0.dev20231109  # Pin a working nightly until rc0.
+keras-nightly==3.0.0.dev2023111903  # TODO recent nightly fails keras_cv/models/object_detection/retinanet/retinanet_test.py
```
mattdangerw, line 3: typo with the path pasted here?

mattdangerw (approved):
Looks good! Don't have great visibility into what layers we need to disable for Keras 3 temporarily. But code overall looks clean!

requirements.txt
sampathweb: Thats how it's in PyPi - https://pypi.org/project/keras-nightly/3.0.0.dev2023111903/
The reason for it to look off is we may be saving down to Minute and TF only getting name down to the hour.

keras_cv/layers/serialization_test.py
sampathweb: That makes sense. Done

keras_cv/layers/preprocessing/base_image_augmentation_layer.py
sampathweb: That makes sense. Done

.github/workflows/release.yml
sampathweb: Done

requirements.txt
mattdangerw: oh sorry my bad, the line was wrapping on `keras_cv/models/object_detection/retinanet/retinanet_test.py` so i didn't realize it was part of the comment for the failure

keras_cv/layers/preprocessing_3d/input_format_test.py
sampathweb: The TEST_CONDITIONS list would also need to be moved inside the tests since those layers are also not compatible with Keras 3.  So left it as-is for now.  We can update it once we make them keras 3 compatible.

PR #2171, opened 2023-11-20
fix noise scheduler error in stable diffusion

divyashreepathihalli:
fixes noise scheduler error about incompatible dtypes for multiplication


sampathweb (approved):
LGTM!

PR #2180, opened 2023-11-22
Fixes Saving Format related test failures for Keras 3

sampathweb:


sampathweb (commented):
Merging the changes since discussed it offline in chat to test saving only on `.keras` format.

PR #2194, opened 2023-11-28
Fix bug when upranking passthrough inputs to RandAugment

mattdangerw:
- RandAugment sometimes will choose a "no augmentation" option and passthrough inputs unaltered.
- Preprocessing normalization routines were not making copies of inputs and sometimes mutating layer input directly (mutating the input dict to cast dtypes and uprank tensors).
- RandAugment under the passthrough option would return these inputs directly.

The net effect was sometimes attempting to uprank during a passthrough call, breaking `tf.map_fn`.


mattdangerw:
@sampathweb thanks! will fix that then pull this in

sampathweb (approved):
LGTM after the lint fix

PR #2197, opened 2023-11-28
Fix the issue(keras-team#2195): Improve readability and comprehensibility of Stable Diffusion source

y-vectorfield:
# What does this PR do?
I added the casting method to Stable Diffusion source as y-vectorfield-fj(my previous account name).
However, this method is implemented abruptly and unnaturally.
Hence, I modified the source and added comments to improve readability and comprehensibility.

# Who can review?
@ianstenbit
@sampathweb

### DIFF
--- a/keras_cv/models/stable_diffusion/stable_diffusion.py
+++ b/keras_cv/models/stable_diffusion/stable_diffusion.py
@@ -235,7 +235,9 @@ class StableDiffusionBase:
                 + unconditional_guidance_scale * (latent - unconditional_latent)
             )
             a_t, a_prev = alphas[index], alphas_prev[index]
-            latent = ops.cast(latent, latent_prev.dtype)
+            # Keras backend array need to cast explicitly
+            target_dtype = latent_prev.dtype
+            latent = ops.cast(latent, target_dtype)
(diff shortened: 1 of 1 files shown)


sampathweb:
@y-vectorfield - This PR only adds a comment - `# Keras backend array need to cast explicitly`.  It may not be required.  If it was adding any doc strings for the user, it might be useful.  So for now, we could close this PR?

sampathweb (approved):

y-vectorfield:
> @y-vectorfield - This PR only adds a comment - `# Keras backend array need to cast explicitly`. It may not be required. If it was adding any doc strings for the user, it might be useful. So for now, we could close this PR?

@sampathweb, thank you very much for your comments and review.
I think this modification is necessary for the improvement of Keras's CV.
If we add the state-of-the-art DL models, we will refer to the previous models.
Hence, I thought it was necessary to improve readability and comprehensibility.
If there is no problem, please merge my modification.

divyashreepathihalli:
Merged, Thank you!

y-vectorfield:
@divyashreepathihalli, thank you very much!

PR #2200, opened 2023-11-28
Adapt KerasCV for Kaggle integration with new preset format

nkovela1:
This PR changes the preset logic to a new format based on Keras V3 Saving, along with preset utilities to easily integrate with Kaggle.

### DIFF
--- a/keras_cv/models/task.py
+++ b/keras_cv/models/task.py
@@ -17,6 +17,7 @@ import os
 
 from keras_cv.api_export import keras_cv_export
 from keras_cv.backend import keras
+from keras_cv.models.backbones.backbone import Backbone
 from keras_cv.utils.preset_utils import check_preset_class
 from keras_cv.utils.preset_utils import load_from_preset
 from keras_cv.utils.python_utils import classproperty
@@ -167,7 +168,7 @@ class Task(keras.Model):
         preset_cls = check_preset_class(preset, (cls, cls.backbone_cls))
 
         # Backbone case.
-        if preset_cls == cls.backbone_cls:
+        if issubclass(preset_cls, Backbone):
             backbone = load_from_preset(
                 preset,
                 load_weights=load_weights,
--- a/keras_cv/models/task.py
+++ b/keras_cv/models/task.py
@@ -75,6 +75,10 @@ class Task(keras.Model):
         }
 
     @classproperty
+    def backbone_cls(cls):
+        return None
+
+    @classproperty
     def backbone_presets(cls):
         """Dictionary of preset names and configs for compatible backbones."""
         return {}
@@ -160,10 +164,10 @@ class Task(keras.Model):
                 preset, load_weights, input_shape, **kwargs
             )
 
-        check_preset_class(preset, cls)
+        preset_cls = check_preset_class(preset, (cls, cls.backbone_cls))
 
         # Backbone case.
-        if preset in cls.backbone_presets:
+        if preset_cls == cls.backbone_cls:
             backbone = load_from_preset(
                 preset,
                 load_weights=load_weights,
--- a/keras_cv/models/task.py
+++ b/keras_cv/models/task.py
@@ -160,7 +160,7 @@ class Task(keras.Model):
                 preset, load_weights, input_shape, **kwargs
             )
 
-        preset_cls = check_preset_class(preset, cls)
+        check_preset_class(preset, cls)
 
         # Backbone case.
         if preset in cls.backbone_presets:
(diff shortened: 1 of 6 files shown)


keras_cv/utils/preset_utils_test.py
```diff
@@ -0,0 +1,81 @@
...
+from keras_cv.models import ImageClassifier
+from keras_cv.models import RetinaNet
+from keras_cv.models import YOLOV8Detector
+from keras_cv.tests.test_case import TestCase
+from keras_cv.utils import preset_utils
+
+
+class PresetUtilsTest(TestCase):
+    @parameterized.parameters(
+        (ImageClassifier, "resnet50_v2_imagenet_classifier", "classification"),
```
mattdangerw, line 32: How'd you come up with this list? Are all these model testing something substantially different?

keras_cv/models/task.py
```diff
@@ -112,57 +154,29 @@ def from_preset(
...
-                f"""Pretrained weights not available for preset "{preset}". """
-                "Set `load_weights=False` to use this preset or choose one of "
-                "the following presets with weights:"
-                f""" "{'", "'.join(cls.presets_with_weights)}"."""
-            )
-
-        metadata = cls.presets[preset]
-        # Check if preset is backbone-only model
+        # Backbone case.
         if preset in cls.backbone_presets:
```
mattdangerw, line 166: This we need to reformulate. The goal is to be able to load a "preset directory" from disk, kaggle, etc, canonically, without needing metadata about it that preset in particular in the library itself (`cls.backbone_presets`). See https://github.com/keras-team/keras-nlp/blob/kaggle/keras_nlp/models/task.py#L219, the difference is we are just checking the preset config against a fixed class to determine if we are in the `backbone` or `task` case.

Here, since KerasCV allows mixing and matching backbones, we probably want to check if the preset class is a subclass of `Backbone`?

keras_cv/utils/preset_utils.py
```diff
@@ -0,0 +1,153 @@
...
+            org, model, variant = segments
+            kaggle_handle = f"{org}/{model}/keras/{variant}/1"
+        elif len(segments) == 4:
+            org, model, variant, version = segments
+            kaggle_handle = f"{org}/{model}/keras/{variant}/{version}"
+        else:
+            raise ValueError(
+                "Unexpected kaggle preset handle. Kaggle model handles should "
+                "have the form kaggle://{org}/{model}/{variant}[/{version}]. "
+                "For example, 'kaggle://keras/bert/bert_base_en'. "
```
mattdangerw, line 49: Update this to a more CV like example.

mattdangerw (changes requested):
Looks good! Just a couple comments.

keras_cv/utils/preset_utils_test.py
nkovela1: This list was chosen for simple models across each of the task types (classification, object detection, and segmentation).

keras_cv/models/task.py
nkovela1: For simplicity sake, I've reformulated it similar to the KerasNLP approach. I've added a property for `backbone_cls` in Task and subclassed it where applicable (only 2 instances). Thanks for catching this!

keras_cv/utils/preset_utils.py
nkovela1: Updated, thanks!

keras_cv/models/task.py
```diff
@@ -112,57 +158,29 @@ def from_preset(
...
-            )
+        preset_cls = check_preset_class(preset, (cls, cls.backbone_cls))
 
-        metadata = cls.presets[preset]
-        # Check if preset is backbone-only model
-        if preset in cls.backbone_presets:
-            backbone_cls = keras.saving.get_registered_object(
-                metadata["class_name"]
+        # Backbone case.
+        if preset_cls == cls.backbone_cls:
```
mattdangerw, line 170: I don't think this is what we want actually. I believe with CV (unlike NLP), the intent is to allow mixing and matching of backbones and tasks. So we should probably check the `preset_cls` is a subclass of `Backbone`, as I was mentioning earlier.

keras_cv/models/task.py
mattdangerw: Ah missed this comment and replied elsewhere. But I think that while a 1-1 mapping of backbone cls <> task cls is a valid assumption for KerasNLP, is it not for KerasCV, so we cannot do this here. We need to allow mixing and matching of backbones with a task.

keras_cv/models/task.py
nkovela1: Got it, I've updated to check if it's a subclass of Backbone

mattdangerw (approved):
LGTM! Thanks.

PR #2206, opened 2023-12-01
Switch all available presets to Kaggle format

nkovela1:
This PR converts all available presets in KerasCV to the Kaggle format, removing all configs from presets and replacing them with Kaggle links.


nkovela1:
Commented out the preset conversion script (which breaks install).

keras_cv/utils/preset_utils.py
```diff
@@ -138,11 +153,22 @@ def check_preset_class(
...
+            config = json.load(config_file)
+    except:
+        raise ValueError(
+            f"The specified preset  `{preset}` is unknown. "
+            "Please check documentation to ensure the correct preset "
+            "handle is being used."
+        )
     cls = keras.saving.get_registered_object(config["registered_name"])
-    if not isinstance(classes, (tuple, list)):
+    # Subclass checking
```
mattdangerw, line 166: This is where I was getting confused too. This first check isn't to actually check subclass, it is just making sure `classes` is an iterable.

keras_cv/utils/preset_utils.py
```diff
@@ -138,11 +153,22 @@ def check_preset_class(
...
+    if not issubclass(classes, (tuple, list)):
         classes = (classes,)
+    # Alias subclass checking
+    if any(issubclass(alias, cls) for alias in classes):
```
mattdangerw, line 170: I think we can just move this check down two lines. Always do subclass checking. And update KerasNLP to match.

keras_cv/utils/preset_utils.py
nkovela1: Yeah I was confused here too, reverted to isinstance here!

keras_cv/utils/preset_utils.py
nkovela1: Yeah, combined the two into one check, thanks!

cloudbuild/unit_test_jobs.jsonnet
```diff
@@ -34,7 +34,7 @@ local unittest = base.BaseTest {
     )
   ],
   command: [
-    'pytest --run_large --durations 0',
+    'pytest --run_extra_large --durations 0',
```
mattdangerw, line 37: let's revert this before submit. also, are there any large test that aren't getting picked up here?

should we try `pytest --run_extra_large --durations 0 keras_cv/`? or have we run that manually?

mattdangerw (approved):
Looks great!

keras_cv/models/backbones/backbone.py
```diff
@@ -121,9 +95,21 @@ def from_preset(
             load_weights=False,
         ```
         """
-        # TODO: delete me!
+        if preset in cls.presets_without_weights and load_weights is True:
```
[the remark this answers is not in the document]
mattdangerw: So I guess we should have `load_from_preset` default to `load_weights=None`.
- If None and there are weights in our `config.json`, load them.
- If True and there are not weights in our `config.json`, file this error.

keras_cv/utils/preset_utils.py
```diff
@@ -138,12 +153,23 @@ def check_preset_class(
...
+            "handle is being used."
+        )
     cls = keras.saving.get_registered_object(config["registered_name"])
+    # Subclass checking
     if not isinstance(classes, (tuple, list)):
         classes = (classes,)
-    if cls not in classes:
+    # Alias subclass checking
+    if cls not in classes and not any(
+        issubclass(alias, cls) for alias in classes
```
mattdangerw, line 171: Actually shouldn't we reverse this? `issubclass(cls, Backbone)` (for example)?

nkovela1: I've added both subclass and alias checking here, should be good now!

cloudbuild/unit_test_jobs.jsonnet
nkovela1: Let's keep this going since the runtime is 12 mins for JAX gcb testing, not bad!

keras_cv/utils/preset_utils.py
```diff
@@ -138,12 +153,23 @@ def check_preset_class(
...
+            "Please check documentation to ensure the correct preset "
+            "handle is being used."
+        )
     cls = keras.saving.get_registered_object(config["registered_name"])
+    # Subclass checking
     if not isinstance(classes, (tuple, list)):
         classes = (classes,)
-    if cls not in classes:
+    # Alias subclass checking
+    if cls not in classes and not any(
```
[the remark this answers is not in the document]
nkovela1: Modified to both subclass and alias checking.

keras_cv/utils/preset_utils.py
mattdangerw: maybe just delete the comment on this line, it doesn't really track (and duplicates below)

keras_cv/models/backbones/backbone.py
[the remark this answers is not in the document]
nkovela1: Fixed, though I'll need to regenerate the models using my script.

keras_cv/utils/preset_utils.py
nkovela1: Removed, thanks

PR #2213, opened 2023-12-06
Fixing conversion of box formats on the original boxes

### LINKED ISSUE #2201: Redundant return in `convert_format`
In the following lines I have noticed that the bounding boxes are converted in place. I am not sure why we return the dictionary again. This does not raise a bug, but breaks the mental model of inplace and out of place changes.
https://github.com/keras-team/keras-cv/blob/925d99d852accdea6fea47c7a778e7a0e82f68cc/keras_cv/bounding_box/converters.py#L372-L381

It would rather be useful to create a **new** dictionary and return the converted bbox where the original is kept as is.

If that is a good change, I would like to contribute to this (will be a quick fix)

ariG23498:
Fixes #2201 

CC: @divyashreepathihalli

### DIFF
--- a/keras_cv/bounding_box/converters.py
+++ b/keras_cv/bounding_box/converters.py
@@ -370,8 +370,8 @@ def convert_format(
             `"float32"`.
     """
     if isinstance(boxes, dict):
-        duplicate_boxes = boxes.copy()
-        duplicate_boxes["boxes"] = convert_format(
+        converted_boxes = boxes.copy()
+        converted_boxes["boxes"] = convert_format(
             boxes["boxes"],
             source=source,
             target=target,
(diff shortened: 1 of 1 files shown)


divyashreepathihalli (changes requested):

divyashreepathihalli:
@ariG23498 I think we should update all the code where we are converting Bboxes in place too.

ariG23498:
@divyashreepathihalli I think that is a good idea, do you think there is a good (efficient) way to find out such places in the code? If not, I could do that manually.

divyashreepathihalli:
Haha! we would have to do it manually. It is of a lower priority atm. I will go ahead and merge this PR for now and you can open a new PR for the rest of the code if you would like.

divyashreepathihalli (approved):

PR #2219, opened 2023-12-07
Change CenterPillar name

nkovela1:
The official name used to be "WaymoOpenDataset", but this isn't accurate.


mattdangerw (approved):
Thanks!

PR #2221, opened 2023-12-07
Update conversion script to fix model naming

nkovela1:
This PR gets rid of auto-naming in preset generation, ensuring that the name in the preset config corresponds to its class name.

### DIFF
--- a/keras_cv/tools/convert_presets.py
+++ b/keras_cv/tools/convert_presets.py
@@ -13,10 +13,18 @@
 # limitations under the License.
 
 import os
+import re
 
 import keras_cv  # noqa: E402
 
 BUCKET = "keras-cv-kaggle"
+
+
+def to_snake_case(name):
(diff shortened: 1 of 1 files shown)


keras_cv/tools/convert_presets.py
```diff
@@ -95,7 +105,7 @@ def convert_task_presets():
         )
         for preset in task_preset_keys:
             save_weights = preset in task_cls.presets_with_weights
-            kwargs = {}
+            kwargs = {"name": to_snake_case(task_cls.__name__)}
```
mattdangerw, line 108: Will this not hit the else branch below? pull `task = task_cls.from_preset(preset, **kwargs)` out of the if/else and leave on `kwargs.udpate` in the if.

nkovela1: Ah my mistake, fixed!

mattdangerw (approved):
lgtm!

PR #2222, opened 2023-12-07
Fix SegFormer Presets

nkovela1:
This PR enables proper functionality for SegFormer presets and adds basic tests.

The following behavior is created here:
SegFormer.from_preset("segformer_b0") -> missing required argument num_clases
SegFormer.from_preset("segformer_b0", num_classes=2) -> ok
SegFormer.from_preset("mit_b0", num_classes=2) -> the exact same

### DIFF
--- a/keras_cv/models/segmentation/segformer/segformer_test.py
+++ b/keras_cv/models/segmentation/segformer/segformer_test.py
@@ -48,7 +48,7 @@ class SegFormerTest(TestCase):
 
     def test_segformer_preset_error(self):
         with self.assertRaises(TypeError):
-            model = SegFormer.from_preset("segformer_b0")
+            _ = SegFormer.from_preset("segformer_b0")
 
     @pytest.mark.large
     def test_segformer_call(self):


divyashreepathihalli (approved):

keras_cv/models/segmentation/segformer/segformer.py
```diff
@@ -165,6 +166,29 @@ def get_config(self):
...
+            "segformer_b0": "mit_b0",
+            "segformer_b1": "mit_b1",
+            "segformer_b2": "mit_b2",
+            "segformer_b3": "mit_b3",
+            "segformer_b4": "mit_b4",
+            "segformer_b5": "mit_b5",
+        }
+        if preset in aliases:
+            preset = aliases[preset]
+        backbone = MiTBackbone.from_preset(
```
mattdangerw, line 187: Could we just chain to super here? And add the mit_b preset to `backbone_presets`?

To hit this branch
https://github.com/keras-team/keras-cv/blob/f64fba0decee9306874dc5f88aaa6d20538be0d7/keras_cv/models/task.py#L134-L141

That would have the advantage of being totally uniform with how other tasks to this, I think.

mattdangerw (approved):
Looks great!

PR #2224, opened 2023-12-08
Adds Kokoro GPU Tests

sampathweb:
Adds Kokoro GPU Tests similar to Keras-NLP and Keras


.kokoro/github/ubuntu/gpu/build.sh
```diff
@@ -0,0 +1,83 @@
...
+   echo "PyTorch backend detected."
+   pip install -r requirements-torch-cuda.txt --progress-bar off
+fi
+
+pip install --no-deps -e "." --progress-bar off
+
+# Run Extra Large Tests for Continuous builds
+if [ "${RUN_XLARGE:-0}" == "1" ]
+then
+   pytest --check_gpu --run_large --run_extra_large --durations 0 \
```
mattdangerw, line 54: General question, why are we doing so much enumeration of testing here. Seems hard to maintain. Why not just run it all?

(probably good to land this equivalent to before, just have been curious about this)

mattdangerw (approved):
Looks good!

.kokoro/github/ubuntu/gpu/build.sh
sampathweb: I kept it same as `cloudbuild` config.  We should open it all.  But that's how the GPU cloudbuild is setup so copied the command as-is

PR #2241, opened 2023-12-13
Change reference of SeedGen and Serialization

sampathweb:


keras_cv/backend/random.py
```diff
@@ -36,6 +38,13 @@ def next(self, ordered=True):
             self._current_seed[0] += 1
             return self._current_seed[:]
 
+    def get_config(self):
+        return {"seed": self._seed, **self._kwargs}
```
divyashreepathihalli, line 42: you would not need `**self._kwargs`

sampathweb: Why?  If the user passes kwargs we want to capture that in the config that's serializied.

divyashreepathihalli: adding reference from keras 3 here - https://github.com/keras-team/keras/pull/18933

divyashreepathihalli (changes requested):
LGTM! Hey Ramesh just one change requested.

keras_cv/backend/random.py
sampathweb: The reconstructed object will not be same as one originally created if we don't serialize the kwargs as well

keras_cv/layers/regularization/drop_path.py
```diff
@@ -50,7 +50,11 @@ class DropPath(keras.layers.Layer):
     def __init__(self, rate=0.5, seed=None, **kwargs):
         super().__init__(**kwargs)
         self.rate = rate
-        self.seed = keras.random.SeedGenerator(seed=seed)
+        self._seed_val = seed
+        if not isinstance(seed, random.SeedGenerator):
```
divyashreepathihalli, line 54: The seed value that is expected for drop_path is an `integer`. This check is unnecessary, we should directly initialize the `SeedGenerator` with the seed.

divyashreepathihalli (changes requested):

keras_cv/backend/random.py
divyashreepathihalli: Resolving based on offline discussion. Adding `kwargs` to get_config is not recommended because we have no way to guarantee the provided args would be serializable or not.

keras_cv/layers/regularization/drop_path.py
sampathweb: Done

PR #2245, opened 2023-12-14
Update Random to Stateless - Disables Stable Diffussion tests

sampathweb:
* Changes Random to `stateless` for Keras2 to match Keras3
* Disables Stable Diffusion (For Keras2 and Keras3) and DeepLabV3 for Keras2. - #2246


divyashreepathihalli (approved):
LGTM! thank you for the fix

Issue #2134, opened 2023-11-08
Fix keras.engine import in predict_utils.py


divyashreepathihalli:

sampathweb:
@divyashreepathihalli - Is this still an open issue?

divyashreepathihalli:
yes

divyashreepathihalli:
@sampathweb  - I have the fix here - https://github.com/keras-team/keras-cv/pull/2248

PR #2248, opened 2023-12-15
Fix keras.engine import in predict_utils.py

divyashreepathihalli:
fix https://github.com/keras-team/keras-cv/issues/2134


sampathweb (approved):
LGTM

Issue #2135, opened 2023-11-08
Implement dropblock_2d regularization layer in keras 3


divyashreepathihalli:
keras-cv/keras_cv/layers/regularization/dropblock_2d.py

divyashreepathihalli:
@sami-bg thank you for your interest. This layer would not be straightforward because it is subclassing an Keras 2 legacy internal baseRandomLayer.

divyashreepathihalli:
No worries! If you are interested we would love to have as many computer vision keras io examples converted to keras 3. This is a priority and could be a good one to start with. - https://github.com/keras-team/keras-io/tree/master/examples/vision

here is a list of them we are yet to convert 

- [ ] learnable_resizer
- [ ]  deit
- [ ]  mixup
- [ ]  perceiver_image_classification
- [ ]  shiftvit
- [ ]  bit
- [ ]  simsiam
- [ ]  generative/dcgan_overriding_train_ste
- [ ]  forwardforward
- [ ]  nnclr
- [ ]  handwriting_recognition
- [ ]  xray_classification_with_tpus
- [ ]  edsr
- [ ]  mobilevit
- [ ]  yolov8
- [ ]  video_classification
- [ ]  nl_image_search
- [ ]  supervised-contrastive-learning
- [ ]  depth_estimation
- [ ]  fully_convolutional_network
- [ ]  vivit
- [ ]  randaugment
- [ ]  consistency_training
- [ ]  basnet_segmentation
- [ ]  barlow_twins
- [ ]  adamatch
- [ ]  pointnet
- [ ]  vit_small_ds
- [ ]  retinanet
- [ ]  temporal_latent_bottleneck
- [ ]  near_dup_search
- [ ]  3D_image_classification
- [ ]  swin_transformers
- [ ]  focal_modulation_network

sami-bg:
I'll give it a try! Thank you for the suggestion :)

PR #2250, opened 2023-12-15
Fix drop block

divyashreepathihalli:
Fixes https://github.com/keras-team/keras-cv/issues/2135

### DIFF
--- a/keras_cv/layers/regularization/dropblock_2d.py
+++ b/keras_cv/layers/regularization/dropblock_2d.py
@@ -12,23 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import tensorflow as tf
-
-from keras_cv.backend import config
-
-if config.keras_3():
-    base_layer = tf.keras.layers.Layer
-else:
-    from tensorflow.keras.__internal__.layers import BaseRandomLayer
-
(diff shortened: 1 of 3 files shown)


keras_cv/layers/regularization/dropblock_2d.py
```diff
@@ -153,93 +145,87 @@ def __init__(
...
+            ops.size(block_pattern), "float32"
+        )
         return (
-            x / tf.cast(percent_ones, x.dtype) * tf.cast(block_pattern, x.dtype)
+            x
+            / ops.cast(percent_ones, x.dtype)
+            * ops.cast(block_pattern, x.dtype)
         )
 
     def get_config(self):
```
sampathweb, line 231: While you are uding the layer, can you also update `get_config`.  I think we could directly say - 
```
config = super().get_config()
config.update({
            "rate": self._rate,
            "block_size": (self._dropblock_height, self._dropblock_width),
            "seed": self.seed,
 })
return config
  ```
This is the pattern used in `keras` repo - https://github.com/keras-team/keras/blob/master/keras/optimizers/adam.py#L148

sampathweb (approved):
LGTM.  Lets run GPU tests to make sure they all pass

Issue #2136, opened 2023-11-08
Implement base_augmentation_layer_3d in keras 3


divyashreepathihalli:
keras-cv/keras_cv/layers/preprocessing_3d/base_augmentation_layer_3d.py

PR #2252, opened 2023-12-15
fix 3d preprocessing layer

divyashreepathihalli:
Fixes https://github.com/keras-team/keras-cv/issues/2136


sampathweb (approved):
LGTM.  Thanks for all the cleanup and using the `random.SeedGenerator`

PR #2255, opened 2023-12-19
Test presets with Keras saving fixes

nkovela1:


mattdangerw:
I think we no longer need these so close, but can re-open if we do!

PR #2272, opened 2024-01-04
Align keras requirments fo torch

sampathweb:
* Use Cuda 12.1 for Torch to align with TF and Jax. Also, aligns torch install with that of Colab that uses torch+cu121 packages.
* Remove extra-url for TensorFlow as its no longer required.

Align with Keras and Keras-NLP


divyashreepathihalli (approved):

mattdangerw (approved):
lgtm!

PR #2273, opened 2024-01-04
Backwards compatibility fixes for Keras <=2.15

nkovela1:
This PR adds backwards compatibility fixes for issues with weights reloading in Keras 2.15 and bumps version numbers of presets to pull from the new compatible weights (after Keras saving fixes).


divyashreepathihalli (approved):
Thanks for the updates!

PR #2275, opened 2024-01-05
Remove weight traversal output

nkovela1:
Accidentally left in a weight traversal output while debugging.


mattdangerw (approved):
thanks!

PR #2283, opened 2024-01-08
Fix YOLOv8Detector deserialization

nkovela1:


keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
```diff
@@ -645,6 +645,18 @@ def get_config(self):
...
+    def from_config(cls, config):
+        config["backbone"] = keras.saving.deserialize_keras_object(
+            config["backbone"]
+        )
+        label_encoder = config.get("label_encoder")
+        if label_encoder is not None:
+            config["label_encoder"] = keras.saving.deserialize_keras_object(
+                config["label_encoder"]
+            )
+        return cls(**config)
```
mattdangerw, line 658: also deserialize prediction decoder

keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
```diff
@@ -645,6 +645,18 @@ def get_config(self):
             "prediction_decoder": self._prediction_decoder,
```
mattdangerw, line 645: we should probably also call `keras.saving.serialize_keras_object` on `self.label_encoder` and `self._prediction_decoder`, seems not strictly necessary, but still more consistent

keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
nkovela1: Ah missed that, thanks!

keras_cv/models/object_detection/yolo_v8/yolo_v8_detector.py
nkovela1: Fixed, thanks!

mattdangerw (approved):
lgtm!

PR #2287, opened 2024-01-08
update dataset link

divyashreepathihalli:


mattdangerw (approved):
lgtm! do we need to make similar updates on keras-team/keras-io anywhere?

divyashreepathihalli:
> lgtm! do we need to make similar updates on keras-team/keras-io anywhere?

The Keras io guides use this like - 
`from keras_cv.datasets.pascal_voc.segmentation import load as load_voc`
So no changes is needed in Keras io

PR #2293, opened 2024-01-10
Fix stable diffusion

divyashreepathihalli:
Fix here - https://github.com/keras-team/keras/pull/19046
Updates to Noise Scheduler 
https://arxiv.org/pdf/2305.08891v2.pdf


sampathweb:
Still failing for Keras 2 GPU.

dugujiujian1999:
Looks like we have to fix this again for keras-core 0.17

Issue #2302, opened 2024-01-15
AnchorGenerator ValueError


tomrtk:
### Current Behavior:
 

Calling `AnchorGenerator` with an image fails with a ValueError:

```console
Traceback (most recent call last):
  File "/work/kerascv-bug/test.py", line 23, in <module>
    main()
  File "/work/kerascv-bug/test.py", line 16, in main
    anchors = ag(image=img)
  File "/work/kerascv-bug/venv/lib/python3.10/site-packages/keras_cv/src/layers/object_detection/anchor_generator.py", lin
e 179, in __call__
    results[key] = bounding_box.convert_format(
  File "/work/kerascv-bug/venv/lib/python3.10/site-packages/keras_cv/src/backend/scope.py", line 37, in wrapper
    return function(*args, **kwargs)
  File "/work/kerascv-bug/venv/lib/python3.10/site-packages/keras_cv/src/bounding_box/converters.py", line 395, in convert
_format
    _validate_image_shape(image_shape)
  File "/work/kerascv-bug/venv/lib/python3.10/site-packages/keras_cv/src/bounding_box/converters.py", line 501, in _valida
te_image_shape
    raise ValueError(
ValueError: Expected image_shape to be either a tuple, list, Tensor. Received image_shape=(416, 416, 3)
```

### Expected Behavior:

No error.

### Steps To Reproduce:

[gist](https://colab.research.google.com/gist/tomrtk/27b90895d8933a2443183c798796681a/anchorgenerator-bug.ipynb)

### Version:

- 0.8.1

### Possible fix

Changing this line to use `tf.shape` seem to fix the issue locally.

https://github.com/keras-team/keras-cv/blob/9fbf36b9bd57f2fb8f4c93192106fd60e8c4e605/keras_cv/layers/object_detection/anchor_generator.py#L175

PR #2305, opened 2024-01-18
Fix image.shape type

sachinprasadhs:
Fixed the image.shape type when the image is passed as an argument in the AnchorGenerator class using Keras ops for Tensorflow backend.

`type(img.shape)` for tensorflow backend is `tensorflow.python.framework.tensor_shape.TensorShape` which fails in the code [here](https://github.com/keras-team/keras-cv/blob/master/keras_cv/bounding_box/converters.py#L478) when checking the instance with tuple or list.
For `toch` type is  `torch.size` and for `jax` type is `tuple` which will pass the above condition.
Fixes: https://github.com/keras-team/keras-cv/issues/2302


divyashreepathihalli (approved):
Thank you for the fix!

Issue #2321, opened 2024-01-29
Some Tests fail for TF Backend with Seg Fault


sampathweb:
SegFormer Tests are failing with Segmentation Fault on TF Backend.  They work fine in Torch and Jax Backends.

Log: https://source.cloud.google.com/results/invocations/84cffec0-1890-479a-8e6f-0a732ad1c891/log

Actually, we have seg fault for other tests too for TensorFlow backend.  Here's one on Deep Lab V3 - https://source.cloud.google.com/results/invocations/d6798a84-3033-439d-99ad-a76e6fad5688/log

tirthasheshpatel:
The CI is now green on master so closing this issue. We can reopen if it comes up again.

PR #2322, opened 2024-01-29
Fix CI Test for Basnet OOM

sampathweb:
* Fixes CI Test for BasNet OOM
* Disables Segformer Tests for TF backend since it seg faults.  Created a Issue for it #2321


tirthasheshpatel (approved):
Looks good, let's tackle the tensorflow error in a follow-up. It was failing for `SegFormer` before but now fails with `DeepLabV3`. Pretty sure it's an OOM but PyTest's multi-threading is getting in the way of TensorFlow's exception handling.

Issue #2298, opened 2024-01-12
KerasCV augmentations throws error in `tf.data` with Keras V3


awsaf49:
### Current Behavior:
KerasCV augmentations in `tf.data` throw error. After digging deeper, I noticed this error occurs with Keras V3 but not with KerasCore.  

Here is the error message
```
NotImplementedError: in user code:

    File "/tmp/ipykernel_26/3817339408.py", line 22, in apply_augmentation  *
        return (augmenter(images), labels)
    File "/opt/conda/lib/python3.10/site-packages/keras_cv/src/layers/augmenter.py", line 44, in __call__  *
        inputs = layer(inputs)
    File "/opt/conda/lib/python3.10/site-packages/keras/src/utils/traceback_utils.py", line 123, in error_handler  **
        raise e.with_traceback(filtered_tb) from None
    File "/opt/conda/lib/python3.10/site-packages/keras_cv/src/layers/preprocessing/vectorized_base_image_augmentation_layer.py", line 447, in call
        outputs = tree.map_structure(
    File "/opt/conda/lib/python3.10/site-packages/tree/__init__.py", line 435, in map_structure
        [func(*args) for args in zip(*map(flatten, structures))])
    File "/opt/conda/lib/python3.10/site-packages/tree/__init__.py", line 435, in <listcomp>
        [func(*args) for args in zip(*map(flatten, structures))])
    File "/opt/conda/lib/python3.10/site-packages/keras_cv/src/layers/preprocessing/vectorized_base_image_augmentation_layer.py", line 450, in <lambda>
        lambda x: ops.convert_to_tensor(x) if x is not None else x,
    File "/opt/conda/lib/python3.10/site-packages/jax/_src/numpy/lax_numpy.py", line 2134, in array
        leaves = [_convert_to_array_if_dtype_fails(leaf) for leaf in leaves]
    File "/opt/conda/lib/python3.10/site-packages/jax/_src/numpy/lax_numpy.py", line 2134, in <listcomp>
        leaves = [_convert_to_array_if_dtype_fails(leaf) for leaf in leaves]
    File "/opt/conda/lib/python3.10/site-packages/jax/_src/numpy/lax_numpy.py", line 2177, in _convert_to_array_if_dtype_fails
        return np.asarray(x)

    NotImplementedError: Exception encountered when calling RandomFlip.call().
    
[...]

PR #2323, opened 2024-01-30
Fix format and Update Vectorized Base to fix tf.data with non-TF backends

sampathweb:
* Fixes Code Format with latest Black
* Fixes issue #2298 of using tf.data with non-TF Backend in Keras3

Only relevant changes are - 
* keras_cv/layers/preprocessing/vectorized_base_image_augmentation_layer.py
* keras_cv/layers/preprocessing/vectorized_base_image_augmentation_layer_test.py

Tests on GPU will still fail for other reasons.


tirthasheshpatel (approved):
Looks good, thanks @sampathweb! We can merge after the Kokoro run (just to make sure the fix doesn't add any new failures).

divyashreepathihalli (approved):
LGTM! thanks for the fix

PR #2325, opened 2024-01-30
Fix TensorFlow-GPU flaky failure

tirthasheshpatel:
Locally, BasNet tests consume around 12GB of memory on CPU. This PR reduces the memory consumption by scaling down the input sizes and batch size. Some GPU tests on Kokoro are failing with a `"Fatal Python error"` which might be due to excess memory usage.

@divyashreepathihalli @sampathweb Let's see if the GPU tests pass now. If not, I will try to go deeper into the failure and update the fix.


sampathweb (approved):
LGTM.  Thanks for the fix.

PR #2357, opened 2024-02-22
CLIP : Enable CLIP.from_preset() Kaggle tests

divyashreepathihalli:
- [x] update Kaggle handle
- [x] Change default CLIP config to base patch 32
- [x] update weights conversion file


sampathweb (approved):
LGTM

PR #2371, opened 2024-03-04
Implemented Coca architecture

VarunS1997:
# What does this PR do?

Implements the work done in the "CoCa": Contrastive Captioners are Image-Text Foundation Models" (https://arxiv.org/pdf/2205.01917.pdf). 

This PR requires: 
 [x] Model Implementation
 [ ] Testing

## Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/keras_cv/layers/attention_pooling.py
+++ b/keras_cv/layers/attention_pooling.py
@@ -1,0 +1,27 @@
+from keras import layers
+
+
+class AttentionPooling(layers.Layer):
+
+    # TODO: Add args
+    def __init__(self,
+                 proj_dim,
+                 num_heads,
+                 **kwargs):
+        super().__init__(self, **kwargs)
+
+        self.proj_dim = proj_dim
+        self.num_heads = num_heads
+
+
+    def build(self, input_shape):
+        self.multi_head_attn = layers.MultiHeadAttention(
+            self.num_heads,
+            self.proj_dim
+        )
+
+        self.layer_norm = layers.LayerNormalization()
+
+    def call(self, query, value, *args, **kwargs):
+        x = self.multi_head_attn(query, value)
+        return self.layer_norm(x)
--- a/keras_cv/models/feature_extractor/CoCa/coca_model.py
+++ b/keras_cv/models/feature_extractor/CoCa/coca_model.py
@@ -1,0 +1,133 @@
+# Copyright 2024 The KerasCV Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import numpy as np
+from keras import Sequential
+from keras_cv.api_export import keras_cv_export
+from keras_nlp.layers import RotaryEmbedding, TransformerDecoder
+from keras_cv.layers import TransformerEncoder as CVTransformerEncoder
+from keras_cv.models.task import Task
+from keras_cv.layers.attention_pooling import AttentionPooling
+from keras_cv.layers.vit_layers import PatchingAndEmbedding
+
+
+@keras_cv_export(["keras_cv.models.CoCa"])
+class CoCa(Task):
+    def __init__(self,
+                 img_query_dim,
+                 text_proj_dim,
+                 img_patch_size=18,
+                 encoder_depth=40,
+                 encoder_heads=16,
+                 encoder_intermediate_dim=6144,
+                 encoder_width=1408,
+                 decoder_intermediate_dim=5632,
+                 unimodal_decoder_heads=18,
+                 multimodal_decoder_heads=18,
+                 con_queries=1,
+                 cap_queries=256,
+                 con_heads=16,
+                 cap_heads=16,
+                 cap_loss_weight=0.5,
+                 con_loss_weight=0.5,
+                 **kwargs):
+        super().__init__(**kwargs)
+
+        self.img_patch_size = img_patch_size
+        self.img_query_dim = img_query_dim
+
+        self.encoder_depth = encoder_depth
+        self.encoder_heads = encoder_heads
+        self.encoder_width = encoder_width
+        self.encoder_intermediate_dim = encoder_intermediate_dim
+
+        self.text_proj_dim = text_proj_dim
+        self.decoder_intermediate_dim = decoder_intermediate_dim
+        self.unimodal_decoder_heads = unimodal_decoder_heads
+        self.multimodal_decoder_heads = multimodal_decoder_heads
+
+        self.con_queries = con_queries
+        self.con_heads = con_heads
+        self.con_loss_weight = con_loss_weight
+
+        self.cap_queries = cap_queries
+        self.cap_heads = cap_heads
+        self.cap_loss_weight = cap_loss_weight
+
+    def build(self, input_shape):
+        super().build(input_shape)
+
+        self.image_patching = PatchingAndEmbedding(self.encoder_width, self.img_patch_size)
+        self.image_encoder = Sequential([
+            CVTransformerEncoder(self.img_query_dim, self.encoder_heads, self.encoder_intermediate_dim)
+            for _ in range(self.encoder_depth)
+        ])
+
+        self.cls_token = self.add_weight(shape=[1, 1, self.text_proj_dim], name="cls_token", trainable=True)
+
+        self.text_embedding = RotaryEmbedding()
+        self.unimodal_text_decoder = Sequential([
+            TransformerDecoder(self.decoder_intermediate_dim, self.unimodal_decoder_heads)
+            for _ in range(self.con_queries)
+        ])
+        self.multimodal_text_decoder = TransformerDecoder(
+            self.decoder_intermediate_dim,
+            self.multimodal_decoder_heads
+        )
+
+        self.con_query = self.add_weight(shape=[1, 1, self.con_queries], trainable=True)
+        self.cap_query = self.add_weight(shape=[1, 1, self.cap_queries], trainable=True)
+
+        self.con_attn_pooling = AttentionPooling(self.img_query_dim, self.con_heads)
+        self.cap_attn_pooling = AttentionPooling(self.img_query_dim, self.cap_heads)
+
+    def call(self, images, texts):
+        """
+        Forward pass of the Coca Model
+
+        :param images: [batch_size, height, width, channels] representing images
+        :param texts: Tensor, typically represented as [batch_size, sequence_length, feature_length] or
+            [batch_size, sequence_length, num_heads, feature_length]. The sequence_length and/or feature_length
+            are required.
+        :return: output of the captioning Transformer Decoder with captioning cross-attention
+        """
+        img_encoding = self.image_patching(images)
+        img_encoding = self.image_encoder(img_encoding)  # [batch, patches_len+1, img_query_dim]
+
+        # This is only needed for loss calculations
+        # con_feature = self.con_attn_pooling(self.con_query, img_encoding)
+        cap_feature = self.cap_attn_pooling(self.cap_query, img_encoding)
+
+        text_tokens = np.concatenate(texts, self.cls_token)
+        mask = np.concatenate((np.ones_like(texts), np.zeros_like(self.cls_token)))
+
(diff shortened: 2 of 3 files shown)


keras_cv/models/feature_extractor/CoCa/coca_model.py
```diff
@@ -0,0 +1,150 @@
...
+        self.con_loss_weight = con_loss_weight
+
+        self.cap_queries = cap_queries
+        self.cap_heads = cap_heads
+        self.cap_loss_weight = cap_loss_weight
+
+    def build(self, input_shape):
+        super().build(input_shape)
+
+        self.image_patching = PatchingAndEmbedding(self.encoder_width, self.img_patch_size)
```
divyashreepathihalli, line 74: lets move all the layer definitions to `__init__()`

keras_cv/models/feature_extractor/CoCa/coca_model.py
```diff
@@ -0,0 +1,150 @@
...
+        self.image_encoder = Sequential([
+            CVTransformerEncoder(self.img_query_dim, self.encoder_heads, self.encoder_intermediate_dim)
+            for _ in range(self.encoder_depth)
```
divyashreepathihalli, line 77: And build all the layers here like `layer.build(<expected_input_shape>)`

divyashreepathihalli (changes requested):
This is great @VarunS1997 !! left a few comments. Also, please add a colab to verify the output.

keras_cv/models/feature_extractor/CoCa/coca_model.py
```diff
@@ -0,0 +1,150 @@
...
+                 encoder_depth=40,
+                 encoder_heads=16,
+                 encoder_intermediate_dim=6144,
+                 encoder_width=1408,
+                 unimodal_decoder_depth=18,
+                 multimodal_decoder_depth=18,
+                 decoder_intermediate_dim=5632,
+                 unimodal_decoder_heads=16,
+                 multimodal_decoder_heads=16,
+                 con_queries=1,
```
divyashreepathihalli, line 39: lets expand `con` -> `contrastive` here and everywhere.

divyashreepathihalli (changes requested):

divyashreepathihalli:
One additional overhead work is needed.

please add keras_cv/models/feature_extractor/coca  to this file
https://github.com/keras-team/keras-cv/blob/master/.kokoro/github/ubuntu/gpu/build.sh
to line 72 and 86

PS: we will fix this overhead soon, but in the mean time this is what we need to do to make sure the large GPU tests run.

keras_cv/layers/attention_pooling.py
```diff
@@ -0,0 +1,36 @@
...
+
+class AttentionPooling(layers.Layer):
+    """Implements the Pooled Attention Layer used in "coca": Contrastive Captioners are Image-Text Foundation Models"
+    (https://arxiv.org/pdf/2205.01917.pdf), consisting of a Multiheaded Attention followed by Layer Normalization.
+
+    Args:
+        proj_dim: The dimensions of the attention heads
+        num_heads: The number of attention heads in the multi-headed attention layer
+    """
+    def __init__(self,
```
mattdangerw, line 12: we should make sure to format with `black`, this looks like another formatting

keras_cv/models/feature_extractor/coca/coca_model.py
```diff
@@ -0,0 +1,244 @@
...
+        ])
+
+        self.contrastive_attn_pooling = AttentionPooling(self.encoder_width, self.contrastive_attn_heads)
+        self.captioning_attn_pooling = AttentionPooling(self.encoder_width, self.captioning_attn_heads)
+
+        # These are learnable weights defined in build as per Keras recommendations
+        self.cls_token = None
+        self.contrastive_query = None
+        self.captioning_query = None
+    """ Contrastive Captioner foundational model implementation.
```
mattdangerw, line 91: docstring should go above `__init__` after the class line

keras_cv/models/feature_extractor/coca/coca_model.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # This is only needed for loss calculations
+        # contrastive_feature = self.con_attn_pooling(self.contrastive_query, img_encoding)
+
+        # [batch_size, encoder_width, captioning_query_length]
+        captioning_feature = self.captioning_attn_pooling(self.captioning_query, img_encoding)
+
+        # [batch_size, sequence_length+1, text_dim]
+        text_tokens = np.concatenate(texts, self.cls_token)
+        mask = np.concatenate((np.ones_like(texts), np.zeros_like(self.cls_token)))
```
mattdangerw, line 209: I can't tell if this mask is the right shape or not. Usually you want something `(batch_size, seq_length, seq_length)` (or seq lenth + 1 if that is the effective sequence length. What is `text_dim` here?

mattdangerw (commented):
Thanks! Left a few comments.

keras_cv/models/feature_extractor/coca/coca_model.py
VarunS1997: text_dim is the dimensionality of the text embeddings

keras_cv/models/feature_extractor/coca/coca_model.py
```diff
@@ -0,0 +1,244 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import numpy as np
+from keras import Sequential
+from keras_cv.api_export import keras_cv_export
+from keras_nlp.layers import RotaryEmbedding, TransformerDecoder
```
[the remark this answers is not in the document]
divyashreepathihalli: I think as we support more multi modal models we should depend on KerasNLP. If the tf-text install issue is resolved, we should add it.

keras_cv/models/feature_extractor/coca/coca_model.py
```diff
@@ -0,0 +1,244 @@
...
+        img_encoding = self.image_encoder(img_encoding)  # [batch_size, img_patches_len+1, encoder_width]
+
+        # This is only needed for loss calculations
+        # contrastive_feature = self.con_attn_pooling(self.contrastive_query, img_encoding)
+
+        # [batch_size, encoder_width, captioning_query_length]
+        captioning_feature = self.captioning_attn_pooling(self.captioning_query, img_encoding)
+
+        # [batch_size, sequence_length+1, text_dim]
+        text_tokens = np.concatenate(texts, self.cls_token)
```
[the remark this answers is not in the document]
VarunS1997: Yeah, I considered that but I think because the token doesn't actually leave the model's intermediate representation, it's okay to leave here.

Otherwise, it would create a weird API where the input is of length N, but the output is length N-1.

keras_cv/layers/attention_pooling.py
```diff
@@ -0,0 +1,33 @@
...
+            self.num_heads, self.head_dim
+        )
+
+        self.layer_norm = layers.LayerNormalization()
+
+    def build(self, input_shape):
+        super().build(input_shape)
+
+        self.multi_head_attn.build(input_shape)
+        self.layer_norm.build(input_shape)
```
divyashreepathihalli, line 29: is the input shape for layer_norm correct?

keras_cv/models/feature_extractor/coca/coca_model.py
```diff
@@ -0,0 +1,311 @@
...
+
+        self.captioning_query_length = captioning_query_length
+        self.captioning_attn_heads = captioning_attn_heads
+        self.captioning_loss_weight = captioning_loss_weight
+
+        # Layer Definitions
+        self.image_patching = PatchingAndEmbedding(
+            self.encoder_width, self.img_patch_size
+        )
+        self.image_encoder = Sequential(
```
divyashreepathihalli, line 119: Sequential might not work, the model will not build properly.

keras_cv/models/feature_extractor/coca/coca_model.py
```diff
@@ -0,0 +1,311 @@
...
+                    self.encoder_width,
+                    self.encoder_heads,
+                    self.encoder_intermediate_dim,
+                )
+                for _ in range(self.encoder_depth)
+            ]
+        )
+
+        self.text_embedding = RotaryEmbedding()
+        self.unimodal_text_decoder = Sequential(
```
divyashreepathihalli, line 131: Again, sequential might not work well. Please double check.

keras_cv/models/feature_extractor/coca/coca_model.py
```diff
@@ -0,0 +1,311 @@
...
+            [
+                TransformerDecoder(
+                    self.decoder_intermediate_dim, self.unimodal_decoder_heads
+                )
+                for _ in range(self.unimodal_decoder_depth)
+            ]
+        )
+        self.multimodal_text_decoder = Sequential(
```
divyashreepathihalli, line 139: same comment about Sequential

keras_cv/models/feature_extractor/coca/coca_model.py
```diff
@@ -0,0 +1,311 @@
...
+            )
+
+        # Build Layers
+        self.image_patching.build(images_shape)
+
+        # Add 1 for CLs token appended by patching
+        num_patches = (images_shape[1] // self.img_patch_size) * (
+            images_shape[2] // self.img_patch_size
+        ) + 1
+        self.image_encoder.build((batch_size, self.encoder_width, num_patches))
```
divyashreepathihalli, line 198: you could keep batch_size as `None`
example 
`self.image_encoder.build((None, self.encoder_width, num_patches))`

divyashreepathihalli (changes requested):
Thanks! I left a few comments
Also, please do add a colab demo to show that the model is working as expected.

keras_cv/layers/attention_pooling.py
```diff
@@ -0,0 +1,33 @@
+from keras import layers
+
+
+class AttentionPooling(layers.Layer):
+    """Implements the Pooled Attention Layer used in "coca": Contrastive Captioners are Image-Text Foundation Models"
+    (https://arxiv.org/pdf/2205.01917.pdf), consisting of a Multiheaded Attention followed by Layer Normalization.
```
innat-asj, line 6: You probably need to rename this pooling layer for COCO. There are many ways to implement attention pooling layer and so this could be confusing later. Or, you can define a separate layers dedicated for coco models.

layers/coco_layers.py

define AttentionPooling in the coco_layers.py

keras_cv/models/feature_extractor/coca/coca_model.py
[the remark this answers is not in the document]
mattdangerw: Sgtm! Though if we switch to a hard dependency here, we should probably add `keras-nlp` as a dependency in `setup.py` (which comes with a transitive dependency on `tensorflow-text` and `tensorflow` just fyi).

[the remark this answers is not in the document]
VarunS1997: Do we want to include that in this PR? There's already some imports of Keras-NLP in other places of Keras CV.

If we make it a separate PR, it'll make it easier to rollback if we need to. Considering it's a new dependency, might be worth separating.

keras_cv/models/feature_extractor/coca/coca_model.py
VarunS1997: Just for my understanding, is there a specific reason to do that over setting the batch_size?

keras_cv/layers/attention_pooling.py
VarunS1997: Fixed it, let me know if it still doesn't look right!

keras_cv/models/feature_extractor/coca/coca_model.py
```diff
@@ -0,0 +1,279 @@
...
+    This model implements the "Contrastive Captioners are image-Text Foundational Models" by Yu, et al.
+    (https://arxiv.org/pdf/2205.01917.pdf). In short, the coca model combines the ideas of Contrastive techniques
+    such as CLIP, with Generative Captioning approaches such as SimVLM.
+
+    The architecture of clip can be described as an Image Visual Transformer Encoder in parallel to self-attention-only
+    Text Transformer Decoder, the outputs of both of which are passed into a multimodal Transformer Decoder. The
+    contrastive loss from the ViT and the uni-modal Text Decoder is combined with a captioning loss from the multi-modal
+    Decoder in order to produce the combined total loss.
+
+    Basic Usage:
```
sachinprasadhs, line 40: This has to be changed to `Example:` since we follow only `Example` or `Examples:` as a standard format.

divyashreepathihalli:
@VarunS1997 will you be completing this one?

PR #2378, opened 2024-03-11
Update CI to use TF 2.16

sampathweb:
Update CI to use TF 2.16 similar to Keras and Keras NLP.

### DIFF
--- a/requirements-tensorflow-cuda.txt
+++ b/requirements-tensorflow-cuda.txt
@@ -1,5 +1,5 @@
 # Tensorflow with cuda support.
-tf-nightly[and-cuda]~=2.16.1  # Pin to TF 2.16
+tensorflow[and-cuda]~=2.16.1  # Pin to TF 2.16
 
 # Torch cpu-only version.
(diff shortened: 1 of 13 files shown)


requirements-tensorflow-cuda.txt
```diff
@@ -1,5 +1,5 @@
 # Tensorflow with cuda support.
-tf-nightly[and-cuda]==2.16.0.dev20240209  # Pin a working nightly until rc0.
+tensorflow[and-cuda]~=2.16.1  # Pin to TF 2.16
```
mattdangerw, line 2: Should we converge on a standard way to pin jax/torch/tf? Either `~=` on patch version, `>=`, `==`? I can't think of a great reason why each library needs a different approach, and I see all three here.

divyashreepathihalli (approved):

requirements-tensorflow-cuda.txt
sampathweb: Agree Changed to `>=`

PR #2385, opened 2024-03-13
Fix the SAM end-to-end model predict test

tirthasheshpatel:
Fixes some of the failures seen in https://github.com/keras-team/keras-cv/actions/runs/8259807616/job/22594330565?pr=2383

One of the tests in SAM changed the global dtype policy. Until now, if the test failed, the policy was never reset due to which other unrelated tests failed. This PR should fix this issue and reduce the failure count on [this ci run](https://github.com/keras-team/keras-cv/actions/runs/8259807616/job/22594330565?pr=2383).

### DIFF
--- a/keras_cv/models/segmentation/segment_anything/sam_test.py
+++ b/keras_cv/models/segmentation/segment_anything/sam_test.py
@@ -41,6 +41,8 @@ from keras_cv.models.segmentation.segment_anything.sam_transformer import (
 )
 from keras_cv.tests.test_case import TestCase
 
+keras.config.disable_traceback_filtering()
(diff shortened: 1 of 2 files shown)


keras_cv/models/segmentation/segment_anything/sam_test.py
```diff
@@ -41,6 +41,8 @@
 )
 from keras_cv.tests.test_case import TestCase
 
+keras.config.disable_traceback_filtering()
+
```
tirthasheshpatel, line 45: Unrelated changes. Will remove

PR #2401, opened 2024-03-21
Fix Segformer test for num_classes 1 with softmax

sampathweb:
Keras has recently started to emit warnings when softmax has only one class.  This PR fixes it by having more than one class at the output.


innat:
@sampathweb 
You're compiling segformer model with `BinaryCrossentropy()` but the last layer activations is hard coded to [softmax](https://github.com/keras-team/keras-cv/blob/2028392f927a92366bfe0b85e718a7cd6190142f/keras_cv/models/segmentation/segformer/segformer.py#L139). An `activation` params should be introduced [here](https://github.com/keras-team/keras-cv/blob/2028392f927a92366bfe0b85e718a7cd6190142f/keras_cv/models/segmentation/segformer/segformer.py#L85).

divyashreepathihalli (approved):

PR #2432, opened 2024-04-30
Bug Fix for CenterNetBoxLoss

TillBeemelmanns:
I am currently doing some experiments with the Waymo 3D Object Detector **Centernet** and found out that there is a bug in the [CenterNetBoxLoss](https://github.com/keras-team/keras-cv/blob/7136ceb3c05335c90670651cc4f5a2ebc1a645ac/keras_cv/src/losses/centernet_box_loss.py#L37) function. 

I trained the model for 2 epochs with the **original loss** function which results in the following predictions on the val set:
waymo_open_dataset_0

Notably there is a constant yaw offset caused by the false `ops.floor()` operation in the loss function applied on the gt heading. 

This PR fixes this issue. The following image depicts the predictions after 2 epochs using the fixed loss function.
waymo_open_dataset_0
(Note, that the network has not converged and no NMS has been applied)

The problem probably raised during the conversion from the Tensorflow op [tf.math.floormod](https://www.tensorflow.org/api_docs/python/tf/math/floormod) to Keras ops. However, 
`tf.math.floormod(a, b)` != `ops.floor(ops.mod(a, b))`
but
`tf.math.floormod(a, b)` = `ops.mod(a, b)`

Hence `ops.floor()` simply removes the accuracy of the gt heading during loss computation.

Additionally, I implemented a test case for the heading classification (+regression) part of the `CenterNetBoxLoss`.

Frameworks & Versions:
- keras-cv==0.8.2
- keras==3.2.0
- tensorflow==2.16.1

@divyashreepathihalli  @sampathweb

### DIFF
--- a/keras_cv/src/losses/centernet_box_loss.py
+++ b/keras_cv/src/losses/centernet_box_loss.py
@@ -66,13 +66,11 @@ class CenterNetBoxLoss(keras.losses.Loss):
 
     def heading_regression_loss(self, heading_true, heading_pred):
         # Set the heading to within 0 -> 2pi
-        heading_true = ops.floor(ops.mod(heading_true, 2 * math.pi))
+        heading_true = ops.mod(heading_true, 2 * math.pi)
 
         # Divide 2pi into bins. shifted by 0.5 * angle_per_class.
         angle_per_class = (2 * math.pi) / self.num_heading_bins
-        shift_angle = ops.floor(
-            ops.mod(heading_true + angle_per_class / 2, 2 * math.pi)
-        )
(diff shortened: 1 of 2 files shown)


divyashreepathihalli (approved):

divyashreepathihalli:
@TillBeemelmanns Thank you for the update!

TillBeemelmanns:
Thanks for approving. I will resolve the remaining failing checks end of this week.

TillBeemelmanns:
@divyashreepathihalli I think I fixed the Pytorch pipeline. Could you rerun the workflow ?

TillBeemelmanns:
@divyashreepathihalli  All checks have passed. Feel free to merge the PR.

PR #2435, opened 2024-05-01
Improve error handling for missing TensorFlow dependency in KerasCV

SamanehSaadat:
This PR addresses [this issue](https://github.com/keras-team/keras/issues/19542): user doesn't have `tensorflow` installed and doesn't expect to get error as they're using `pytorch` backend.

### DIFF
--- a/keras_cv/src/backend/tf_ops.py
+++ b/keras_cv/src/backend/tf_ops.py
@@ -46,6 +46,7 @@ from tensorflow import reduce_max as max  # noqa: F403, F401
 from tensorflow import split  # noqa: F403, F401
 
 import numpy as np
+
 try:
     import tensorflow as tf
 except ImportError:
--- a/keras_cv/src/bounding_box/converters_test.py
+++ b/keras_cv/src/bounding_box/converters_test.py
@@ -16,6 +16,7 @@ import itertools
 
 import numpy as np
 import pytest
+
 try:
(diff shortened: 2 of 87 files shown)


keras_cv/src/backend/tf_ops.py
```diff
@@ -46,7 +46,13 @@
...
 
 import numpy as np
-import tensorflow as tf
+try:
+    import tensorflow as tf
+except ImportError:
+    raise ImportError(
+        "To use KerasCV, please install TensorFlow: `pip install tensorflow`. "
+        "The TensorFlow package is required for data preprocessing with any backend."
+    )
```
innat, line 55: Will keras-cv always depend on tensorflow for data processing layers?

keras_cv/src/backend/tf_ops.py
```diff
@@ -46,7 +46,14 @@
 from tensorflow import split  # noqa: F403, F401
```
smitlg, line 46: These imports will also fail in case tensorflow is not installed. Making this same change across all the files makes it more hard to maintain. Instead, this should be wrapped around in some module like `keras_cv/utils/conditional_imports.py` and import tensorflow from this wrapper.

divyashreepathihalli (approved):
Thanks Samaneh! looks like code reformatting is needed!

PR #2484, opened 2024-08-29
feat: Add `minimum_box_area_ratio` to filter out poor, partial boxes

thomas-coldwell:
# What does this PR do?

This PR adds logic to the `bounding_box.clip_to_image` to attenuate the boxes that fall below a certain threshold of their original area in an image. This is particularly useful when creating a training dataset and wanting to ensure partial elements of an object that could now belong to any class e.g. a foot are not detected as a particular class leading to a noisy output during inference.

This can be tested with the jittered resize demo by setting this value to 0.95 (most boxes will be filtered out as at least 5% has been cropped / clipped out) and then setting it to 0.0 which will make it behave as before.

@divyashreepathihalli and @sampathweb

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/examples/layers/preprocessing/bounding_box/jittered_resize_demo.py
+++ b/examples/layers/preprocessing/bounding_box/jittered_resize_demo.py
@@ -23,6 +23,7 @@ def main():
         target_size=(512, 512),
         scale_factor=(3 / 4, 4 / 3),
         bounding_box_format="xyxy",
+        minimum_box_area_ratio=0.5
     )
     result = dataset.map(jittered_resize, num_parallel_calls=tf.data.AUTOTUNE)
     demo_utils.visualize_data(result, bounding_box_format="xyxy")
--- a/keras_cv/src/bounding_box/utils.py
+++ b/keras_cv/src/bounding_box/utils.py
@@ -66,7 +66,7 @@ def _relative_area(boxes, bounding_box_format):
 
 @keras_cv_export("keras_cv.bounding_box.clip_to_image")
 def clip_to_image(
-    bounding_boxes, bounding_box_format, images=None, image_shape=None
+    bounding_boxes, bounding_box_format, images=None, image_shape=None, minimum_box_area_ratio=0.0
 ):
     """clips bounding boxes to image boundaries.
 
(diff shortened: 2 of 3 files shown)


kennyweigel (approved):

keras_cv/src/layers/preprocessing/jittered_resize.py
```diff
@@ -103,6 +103,7 @@ def __init__(
         crop_size=None,
         bounding_box_format=None,
         interpolation="bilinear",
+        minimum_box_area_ratio=0.0,
```
divyashreepathihalli, line 106: let us remove this from the augmentation layers. We need to keep this consistent with the other layers. Let us not touch this.

examples/layers/preprocessing/bounding_box/jittered_resize_demo.py
```diff
@@ -23,6 +23,7 @@ def main():
         target_size=(512, 512),
         scale_factor=(3 / 4, 4 / 3),
         bounding_box_format="xyxy",
+        minimum_box_area_ratio=0.5
```
divyashreepathihalli, line 26: remove this as well. Maybe add a doc string to `clip_to_image`, explain the args and add an example there

divyashreepathihalli (changes requested):
Thanks for the PR! Left a few comments!

keras_cv/src/bounding_box/utils.py
```diff
@@ -106,17 +107,19 @@ class ID set to -1, indicating that there is no object present in them.
...
     )
+    area_ratios = ops.divide(areas, original_areas)
     clipped_bounding_boxes = bounding_box.convert_format(
         clipped_bounding_boxes,
         source="rel_xyxy",
         target=bounding_box_format,
         images=images,
         image_shape=image_shape,
     )
+    passed = ops.logical_and(areas > 0.0, area_ratios > minimum_box_area_ratio)
```
divyashreepathihalli, line 118: can we rename passed to something more meaningful - ` valid_box` maybe?

PR #2489, opened 2024-09-13
Add activation function option for yolov8 detector.

digital-nomad-cheng:
# What does this PR do?
Currently, only YOLOV8Backbon accept `activation` as argument for changing to other activation function other than `swish`.  But for some hardware, `swish` activation function is not well optimized. It does not make sense to only change the backbone to other activation function while keep the detector(head) default to swish.
So this pull meant to give option to choose activation in detector as well.

### DIFF
--- a/keras_cv/src/models/object_detection/yolo_v8/yolo_v8_detector.py
+++ b/keras_cv/src/models/object_detection/yolo_v8/yolo_v8_detector.py
@@ -104,7 +104,7 @@ def get_anchors(
     return all_anchors, all_strides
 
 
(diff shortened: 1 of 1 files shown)


divyashreepathihalli:
Thanks for the PR. There was recent changes to YOLV8. Can you please rebase and also runn the `shell/format.sh` at root to clear formatting errors

digital-nomad-cheng:
> Thanks for the PR. There was recent changes to YOLV8. Can you please rebase and also runn the `shell/format.sh` at root to clear formatting errors

Done

